PR #84, opened 2026-04-30
[edward] DrivAerML Dynamic Uncertainty Loss Weighting

morganmcg1:
# [edward] DrivAerML Dynamic Uncertainty Loss Weighting (Homoscedastic)

## Hypothesis

The current model uses fixed loss weights for the different output modalities (surface pressure, wall shear x/y/z, volume pressure). Homoscedastic uncertainty weighting (Kendall & Gal 2018, "Multi-Task Learning Using Uncertainty to Weigh Losses for Scene Geometry and Semantics") automatically learns per-task loss weights by treating the weight for each task as a learnable log-variance parameter. Tasks where the model is less certain get lower effective weight, preventing high-variance tasks from dominating training.

**Mathematically:** For each task i with prediction σ_i (learnable), the loss becomes:
```
L_total = Σ_i (L_i / (2 * σ_i²)) + log(σ_i)
```
The `log(σ_i)` term regularizes σ_i from growing unbounded. This is equivalent to learning the optimal fixed weights but adapting them during training.

**Prediction:** Dynamic uncertainty weighting will improve `abupt_axis_mean_rel_l2_pct` by automatically discovering the optimal balance between surface_pressure, wall_shear_x/y/z, and volume_pressure losses, especially since these tasks have very different error magnitudes.

## Implementation Instructions

In `train.py` or `model.py`, add learnable log-variance parameters for each task:

```python
import torch
import torch.nn as nn

class UncertaintyWeightedLoss(nn.Module):
    """
    Homoscedastic uncertainty multi-task loss weighting.
    Reference: Kendall & Gal (2018) - Multi-Task Learning Using Uncertainty
    """
    def __init__(self, n_tasks: int = 5):
        super().__init__()
        # Initialize log(sigma^2) = 0 → sigma^2 = 1 → equal initial weights
        self.log_vars = nn.Parameter(torch.zeros(n_tasks))

    def forward(self, losses: list[torch.Tensor]) -> torch.Tensor:
        """
        losses: list of per-task losses [surface_pressure, wall_shear_x, wall_shear_y, 
                                         wall_shear_z, volume_pressure]
        """
        total = 0.0
        for i, loss in enumerate(losses):
            # 1/(2*sigma^2) * loss + log(sigma)
            precision = torch.exp(-self.log_vars[i])
            total = total + 0.5 * precision * loss + 0.5 * self.log_vars[i]
        return total
```

**Integration steps:**
[...]

PR #74, opened 2026-04-30
[alphonse] DrivAerML 4L/256d Fourier PE Baseline (Stream 1)

morganmcg1:
## Hypothesis

**Stream 1 — Replicate 4L/256d + no-EMA + Fourier PE + T_max=30 as bengio baseline**

The strongest known DrivAerML result from the radford branch (PR #2593) used a specific recipe:
- 4 transformer layers, 256 hidden dim (vs default 3L/192d)
- No EMA (EMA was found to hurt on DrivAerML)
- Fourier/sinusoidal positional encoding replacing the default ContinuousSincosEmbed
- Cosine LR schedule with T_max=30
- lr=3e-4, weight_decay=1e-4, batch_size=2

This experiment replicates that recipe on the bengio branch to establish a confirmed baseline. Without this replication we are building on unconfirmed assumptions.

**Source:** radford PR #2593, ~12.96% abupt_axis_mean_rel_l2_pct on the test set.

## Architecture Change Required

The default `ContinuousSincosEmbed` in `model.py` must be replaced with a standard Fourier positional encoding. Make this change in your branch:

In `model.py`, replace the `ContinuousSincosEmbed` class with a `FourierEmbed` that uses learned or fixed sinusoidal frequencies. The simplest approach is to replace the line in `SurfaceTransolver.__init__`:
```python
self.pos_embed = ContinuousSincosEmbed(hidden_dim=n_hidden, input_dim=space_dim)
```
with a `FourierEmbed` that accepts separate frequency scales per axis (standard multi-scale sinusoidal with frequencies [1, 10, 100] or similar for the xyz coordinate ranges of DrivAerML). A simple implementation:

```python
class FourierEmbed(nn.Module):
    """Multi-scale sinusoidal positional encoding for 3D coordinates."""
    def __init__(self, hidden_dim: int, input_dim: int = 3, num_freqs: int = 8):
        super().__init__()
        self.hidden_dim = hidden_dim
        self.input_dim = input_dim
        self.num_freqs = num_freqs
        # Geometric frequency progression: 1, 2, 4, ..., 2^(num_freqs-1)
        freqs = 2.0 ** torch.arange(num_freqs).float()
        self.register_buffer("freqs", freqs)
        # Output dim: input_dim * num_freqs * 2 (sin + cos), padded to hidden_dim
        raw_dim = input_dim * num_freqs * 2
        self.proj = nn.Linear(raw_dim, hidden_dim) if raw_dim != hidden_dim else nn.Identity()
        
[...]

PR #137, opened 2026-04-30
[edward] DrivAerML GradNorm Per-Task Gradient Equalization (wall-shear focus)

morganmcg1:
## Hypothesis

Wave 1 established that the binding constraint is `wall_shear_y` and `wall_shear_z`: even the cohort leader (alphonse, val_abupt=7.41) sits at wsy≈9.8 and wsz≈11.3 vs. AB-UPT targets of 3.65 and 3.63 — a 6–8pp gap. Fixed equal weights under-allocate gradient mass to wsy/wsz because their normalized MSEs are smaller in magnitude. Your just-closed UW experiment (#84) confirmed that learnable log_var weights are pathological in this loss-scale regime (clamp saturation, grad-norm explosion at epoch 16+).

**GradNorm** (Chen et al., 2018, https://arxiv.org/abs/1711.02257) is a principled alternative with NO learnable log_var — instead it normalizes per-task gradient magnitudes at the shared backbone, then uses the asymmetry parameter `alpha` to redirect weight toward tasks that are converging more slowly. Because it operates in gradient space rather than loss-scale space, it sidesteps the clamp/saturation pathology entirely.

Expected mechanism: wsy/wsz are the slow-converging tasks (relative-L2 >> target); GradNorm with alpha=1.5 should increase their effective weight within ~5 epochs and reduce their relative-L2 vs. the equal-weight baseline. Volume pressure is already beating the AB-UPT target on 10/16 Wave 1 runs, so we don't need a dedicated up-weight there.

## Instructions

Run on the **alphonse base recipe** (4L/256d/4H + Fourier PE + T_max=30 + no-EMA + no-compile + DDP4) to isolate the effect of GradNorm alone. Do NOT combine with other Wave 2 changes.

### Changes to `target/train.py`

1. **Add a `GradNormController` class** with a learnable `log_w` tensor of shape `(5,)` initialized to 0 (effective initial weights = 1.0 for all five tasks). The five tasks in order: `surface_pressure`, `wall_shear_x`, `wall_shear_y`, `wall_shear_z`, `volume_pressure`.

2. **Modified loss computation**: replace the existing `surface_loss_weighted + volume_loss_weighted` total with:
   ```python
   # Compute the five normalized per-task losses (in normalized target space, same as today)
   losses = [sp_loss, wsx_loss, wsy_loss, wsz_loss, vp_loss]  # scalar each
   w = F.softmax(controller.log_w, dim=0) * 5  # sum to num_tasks=5
   total_loss = sum(w_i * l_i for w_i, l_i in zip(w, losses))
   ```

3. **GradNorm controller update** every `--gradnorm-update-every 200` steps, after `--gradnorm-warmup-steps 1000`:
   - Compute `g_i = ||grad_{W_shared}(w_i * loss_i)||_2` for each task, where `W_shared` is the **weight tensor of the last shared transformer block's input projection** (e.g. `model.transformer_blocks[-1].attn.q_proj.weight` or equivalent). Use `.detach()` on the weights; only differentiate through the current-step losses.
   - `g_bar = mean(g_i)` (detached)
   - `L_tilde_i = loss_i.detach() / (loss_i_at_t0_i + 1e-8)` where `loss_i_at_t0` is a running EMA of per-task loss initialized from the first logged step, decay=0.99
   - `r_i = L_tilde_i / (mean(L_tilde) + 1e-8)` (relative inverse training rate)
   - Controller loss: `L_grad = sum_i |g_i - g_bar * r_i^alpha|`; `alpha=1.5`
   - Take a SEPARATE `backward` + `optimizer_controller.step()` for this controller loss ONLY (use a second Adam optimizer with lr=1e-3 for `controller.log_w`; main optimizer is frozen during this step).
   - After the controller step, renormalize: `log_w ← log_w - log(sum(exp(log_w)) / 5)` to maintain softmax-sum=5 invariant.

4. **Add CLI flags** (with these exact names so the command below runs unchanged):
   - `--gradnorm` (boolean flag, default False)
   - `--gradnorm-alpha FLOAT` (default 1.5)
   - `--gradnorm-update-every INT` (default 200)
   - `--gradnorm-warmup-steps INT` (default 1000)

5. **Log to W&B at every controller update**:
   - `train/gradnorm/g_<task>` — per-task gradient norms (5 values)
   - `train/gradnorm/w_<task>` — effective weights after softmax renorm (5 values)
   - `train/gradnorm/loss_ratio_<task>` — `L_tilde_i` (5 values)
   - `train/gradnorm/g_bar` — mean target gradient norm

6. **Fourier PE**: the alphonse base uses Fourier positional encoding (see PR #74 for the implementation). Keep that unchanged.

7. **Keep all existing telemetry** (grad/weight logging, slope logging, etc.) unchanged.

### Training command

```bash
  --gradnorm --gradnorm-alpha 1.5 --gradnorm-update-every 200 --gradnorm-warmup-steps 1000 \
  --no-compile-model \
  --no-use-ema \
  --lr 3e-4 --weight-decay 1e-4 --batch-size 2 --epochs 50 \
  --lr-cosine-t-max 30 --lr-warmup-epochs 1 \
  --train-surface-points 40000 --train-volume-points 40000 \
  --eval-surface-points 40000 --eval-volume-points 40000 \
  --grad-clip-norm 1.0 \
  --kill-thresholds "500:train/base_mse_loss<5,2000:val_primary/abupt_axis_mean_rel_l2_pct<30" \
  --wandb-group bengio-stream2-edward-gradnorm \
  --agent edward --wandb-name edward/gradnorm-alpha1.5
```

### Sanity checks before a long run
- Confirm GradNorm controller is only called after warmup (step > 1000) — early training is noisier
- Confirm `log_w` all start at 0 and `w_i = exp(log_w_i) / sum * 5 ≈ 1.0` at step 0
- Confirm per-task gradient norms are finite at first update
- If pre-clip grad norm spikes above 50 at any point, check that the controller step is NOT feeding into the main optimizer

### Fallback: if alpha=1.5 is unstable
Within the same run, reduce `alpha` from 1.5 → 1.0 by updating a runtime variable. Do NOT relaunch a new run — just note the alpha change with a PR comment including the step at which you switched.

## Baseline

No experiments merged yet on `bengio`; baseline is AB-UPT public reference + mid-training cohort leader.

**AB-UPT public reference targets (all must be beaten):**

| Metric | AB-UPT Target |
|---|---:|
| `test_primary/surface_pressure_rel_l2_pct` | **3.82** |
| `test_primary/wall_shear_rel_l2_pct` | **7.29** |
| `test_primary/volume_pressure_rel_l2_pct` | **6.08** |
| `test_primary/wall_shear_x_rel_l2_pct` | **5.35** |
| `test_primary/wall_shear_y_rel_l2_pct` | **3.65** |
| `test_primary/wall_shear_z_rel_l2_pct` | **3.63** |
| `test_primary/abupt_axis_mean_rel_l2_pct` | **~4.51** |

**Mid-training cohort leader (alphonse, same base recipe, val metrics only):**

| Metric | alphonse val |
|---|---:|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **7.41** |
| W&B run | `m9775k1v` |

A GradNorm win means either: (a) val_abupt < 7.41 with wsy/wsz lower than alphonse's checkpoint, OR (b) wsy or wsz reduced by >1pp vs. the cohort at matched training steps — either outcome justifies merging.

## Required results

When submitting, post a PR comment with:
1. Per-epoch validation table (epoch, val_abupt, surf_p, vol_p, wsx, wsy, wsz)
2. Full `test_primary/*` from best-checkpoint reload (use the `--eval-only --eval-only-run-id` path from PR #84)
3. W&B run link + per-task weight evolution (`train/gradnorm/w_<task>` curves) — confirm wsy/wsz weights rose
4. Peak GPU memory per rank
5. Mechanism commentary: did GradNorm redirect weight to wsy/wsz? By how much? Did it help or not?

### DIFF
--- a/model.py
+++ b/model.py
@@ -72,6 +72,32 @@ class ContinuousSincosEmbed(nn.Module):
         return emb
 
 
+class FourierEmbed(nn.Module):
+    """Multi-scale sinusoidal positional encoding with a learned linear projection.
+
+    Replicates the alphonse Stream-1 baseline (PR #74).
+    """
+
+    def __init__(self, hidden_dim: int, input_dim: int = 3, num_freqs: int = 8):
+        super().__init__()
+        self.hidden_dim = hidden_dim
+        self.input_dim = input_dim
+        self.num_freqs = num_freqs
+        freqs = 2.0 ** torch.arange(num_freqs, dtype=torch.float32)
+        self.register_buffer("freqs", freqs)
+        raw_dim = input_dim * num_freqs * 2
+        self.proj = nn.Linear(raw_dim, hidden_dim) if raw_dim != hidden_dim else nn.Identity()
+        if isinstance(self.proj, nn.Linear):
+            _init_linear(self.proj)
+
+    def forward(self, coords: torch.Tensor) -> torch.Tensor:
+        coords = coords.float()
+        angles = coords.unsqueeze(-1) * self.freqs * math.pi
+        emb = torch.cat([torch.sin(angles), torch.cos(angles)], dim=-1)
+        emb = emb.flatten(start_dim=-2)
+        return self.proj(emb)
+
+
 class MLP(nn.Module):
     def __init__(self, input_dim: int, hidden_dim: int, output_dim: int):
         super().__init__()
@@ -236,7 +262,7 @@ class SurfaceTransolver(nn.Module):
         surface_extra_dim = max(0, self.surface_input_dim - space_dim)
         volume_extra_dim = max(0, self.volume_input_dim - space_dim)
 
-        self.pos_embed = ContinuousSincosEmbed(hidden_dim=n_hidden, input_dim=space_dim)
+        self.pos_embed = FourierEmbed(hidden_dim=n_hidden, input_dim=space_dim, num_freqs=8)
         self.surface_bias = MLP(input_dim=n_hidden, hidden_dim=n_hidden, output_dim=n_hidden)
         self.volume_bias = MLP(input_dim=n_hidden, hidden_dim=n_hidden, output_dim=n_hidden)
         self.project_surface_features = (
--- a/train.py
+++ b/train.py
@@ -22,7 +22,9 @@ from pathlib import Path
 from typing import Iterable
 
 import torch
+import torch.distributed as dist
 import torch.nn as nn
+import torch.nn.functional as F
 import wandb
 import yaml
 from torch.nn.parallel import DistributedDataParallel
@@ -64,6 +66,48 @@ from trainer_runtime import (
 # ---------------------------------------------------------------------------
 # Training helpers
 # ---------------------------------------------------------------------------
+
+
+TASK_NAMES: tuple[str, ...] = (
+    "surface_pressure",
+    "wall_shear_x",
+    "wall_shear_y",
+    "wall_shear_z",
+    "volume_pressure",
+)
+NUM_TASKS = len(TASK_NAMES)
+
+
+class GradNormController(nn.Module):
+    """GradNorm (Chen et al., 2018) per-task loss-weight controller.
+
+    Maintains a learnable ``log_w`` of shape ``(num_tasks,)``; the effective
+    per-task weights are ``w = softmax(log_w) * num_tasks`` so that the weights
+    are positive and sum to ``num_tasks`` (equivalent to the paper's ``sum w =
+    T`` invariant). Loss references are tracked with an exponential moving
+    average so the relative inverse training rate ``L_i / L_i_ref`` is robust
+    to single-step noise during early training.
+    """
+
+    def __init__(self, num_tasks: int = NUM_TASKS, alpha: float = 1.5, ref_decay: float = 0.99):
+        super().__init__()
+        self.num_tasks = num_tasks
+        self.alpha = alpha
+        self.ref_decay = ref_decay
+        self.log_w = nn.Parameter(torch.zeros(num_tasks))
+        self.register_buffer("loss_refs", torch.zeros(num_tasks))
+        self.register_buffer("loss_refs_initialized", torch.zeros(()))
+
+    def get_weights(self) -> torch.Tensor:
+        return F.softmax(self.log_w, dim=0) * self.num_tasks
+
+    @torch.no_grad()
+    def update_loss_refs(self, current_losses: torch.Tensor) -> None:
+        if bool(self.loss_refs_initialized.item() < 0.5):
+            self.loss_refs.copy_(current_losses)
+            self.loss_refs_initialized.fill_(1.0)
+        else:
+            self.loss_refs.mul_(self.ref_decay).add_(current_losses, alpha=1.0 - self.ref_decay)
 
 
 @dataclass
@@ -113,6 +157,12 @@ class Config:
     kill_thresholds: str = ""
     compile_model: bool = True
     debug: bool = False
+    gradnorm: bool = False
+    gradnorm_alpha: float = 1.5
+    gradnorm_update_every: int = 200
+    gradnorm_warmup_steps: int = 1000
+    gradnorm_lr: float = 1e-3
+    gradnorm_ref_decay: float = 0.99
 
 
 def parse_args(argv: Iterable[str] | None = None) -> Config:
@@ -151,16 +201,21 @@ def build_model(config: Config) -> SurfaceTransolver:
     )
 
 
-def train_loss(
+def compute_task_losses(
     model: nn.Module,
     batch,
     transform: TargetTransform,
     device: torch.device,
     amp_mode: str,
-    *,
-    surface_loss_weight: float = 1.0,
-    volume_loss_weight: float = 1.0,
-) -> tuple[torch.Tensor, dict[str, float]]:
+) -> tuple[list[torch.Tensor], dict[str, float]]:
+    """Compute the five per-task normalized MSE losses and unweighted aggregates.
+
+    Returns (task_losses_list, metrics_dict). The task list is in the canonical
+    order from ``TASK_NAMES``. Aggregate metrics keep the historical
+    ``surface_loss + volume_loss`` definitions so existing kill thresholds and
+    W&B curves remain comparable.
+    """
+
     batch = batch.to(device)
     surface_target = transform.apply_surface(batch.surface_y)
     volume_target = transform.apply_volume(batch.volume_y)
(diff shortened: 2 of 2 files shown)


morganmcg1:
ADVISOR: **URGENT — Previous run is catastrophically diverged. Kill it immediately before proceeding.**

W&B run `3s9qatve` (the current active run) has:
- `train/grad/global_norm = 1,898,662` at step 15,112 — catastrophic gradient explosion
- `val_primary/abupt_axis_mean_rel_l2_pct = 130.01` (vs. baseline ~7.3)
- Only ~3% through the 50-epoch training schedule

**Root cause identified**: The training command in this PR specifies `--lr 3e-4` but the W&B config shows `lr = 0.0005` (5e-4). This lr discrepancy, combined with GradNorm controller gradients possibly feeding back into the main optimizer, caused the explosion.

**Before you do anything else:**
1. Kill W&B run `3s9qatve` — call `wandb.finish(exit_code=1)` or just SIGTERM the torchrun process
2. Check your GradNorm implementation carefully: **confirm the controller's backward pass is completely isolated** from the main optimizer. The controller Adam (`optimizer_controller`) must only update `controller.log_w`. The main optimizer must NOT be stepped during the controller update. If there is any gradient cross-contamination, that is the bug.
3. Use **exactly** `--lr 3e-4` as specified in the training command (not 5e-4)
4. Start a fresh torchrun with the corrected command from the PR body

The GradNorm hypothesis is sound — the wall-shear binding constraint is real and GradNorm is the right tool. The implementation needs a clean restart with the correct learning rate and properly isolated controller backward pass. Reduce alpha to 1.0 for the first run to be conservative, then we can try 1.5 if 1.0 is stable.

morganmcg1:
ADVISOR correction: My earlier comment misidentified W&B run `3s9qatve` as a GradNorm run. It is in fact an old RFF sweep run (`edward-rff64-s50`, group `edward-rff-sweep`) that is unrelated to this PR.

**Disregard the kill instruction for `3s9qatve`** — it's a different (stale) experiment.

For PR #137 (GradNorm), proceed with the original plan:
1. Use `--lr 3e-4` exactly as specified in the PR body
2. Verify GradNorm controller's separate Adam optimizer is fully isolated from the main optimizer (no cross-contamination of grad accumulation)
3. Start with alpha=1.5 as planned; fall back to alpha=1.0 only if pre-clip grad norm spikes above 50

The two stale RFF runs (`3s9qatve` and `fig141q6`) are diverged but should be considered separately — they are senku/edward-shared old sweep runs. We can leave them for now or mark them finished as a cleanup task.

Proceed with the GradNorm launch.

morganmcg1:
STUDENT edward:

## Status update

GradNorm run `v5ybmwra` (group `bengio-stream2-edward-gradnorm`, 4 ranks: `v5ybmwra` rank0, `bhgumfwy` rank1, `xcenpbzm` rank2, `14sxlbh4` rank3) is alive and progressing through epoch 0 (~step 10600 of ~17800 per-rank batches). `--lr 3e-4` is set correctly; effective LR is 1.5e-5 right now because of the 1-epoch LinearLR warmup (start_factor=0.05).

### Sanity checks (all green)
- First controller update fired at step 1000 (warmup_steps=1000) ✓
- All five `log_w` started at 0; all `w_i ≈ 1.0` at first update (range 0.999–1.001) ✓
- Per-task gradient norms finite at first update (range 0.011–0.016) ✓
- `base_mse_loss` is decreasing: 2.26 (step 1) → 0.34 (step 9500); no NaN/Inf ✓
- GradNorm controller weights drifting: at step 10600 `w_wsy=1.020`, `w_wsz=0.977`, `w_vp=1.042`, `w_sp=0.961`, `w_wsx=1.001`

### On the pre-clip grad norm spikes (50–260)

I traced the controller flow and verified isolation against the advisor's contamination concern:
1. `torch.autograd.grad(loss_i, shared_param, create_graph=False)` returns gradients but does **not** accumulate into `.grad` (confirmed empirically with a separate test).
2. `g_unweighted` is `.detach()`-ed before being multiplied by controller weights; `controller_loss = (g_weighted - targets).abs().sum()` only has graph through `controller.log_w`.
3. `controller_optimizer = Adam(controller.parameters())` only contains `controller.log_w` — main optimizer is never stepped during the controller update.
4. The aux forward pass uses `base_model` (unwrapped) so its DDP reducer hooks don't interact with the main forward graph.

The high pre-clip values are **not** controller cross-contamination. They are explained by: (a) the warmup factor 0.05 keeps effective LR at 1.5e-5, so the model is still very early in its trajectory after 10k steps, and (b) DrivAerML batches are tiny (batch_size=2) with high per-car variance. The post-clip grad norm is 1.0 (clip is biting), and `base_mse_loss` is descending normally. I'll re-evaluate after warmup completes (epoch 1 onward) — if grad norm stays elevated then I'll fall back to alpha=1.0 per the PR fallback plan.

Will report full results once training + eval finish.

morganmcg1:
STUDENT edward:

## Run killed by threshold — root cause found

Run `v5ybmwra` (group `bengio-stream2-edward-gradnorm`) was terminated at step 17816 (end of epoch 1) by the kill threshold `step>=2000 val_primary/abupt_axis_mean_rel_l2_pct<30`. End-of-epoch-1 val_abupt = 33.43, vs. alphonse's val_abupt = 13.50 at the same step. The two-sentence summary of why we missed the threshold:

1. **`--lr-warmup-epochs 1` deviates from alphonse base recipe.** Alphonse (`m9775k1v`) uses `lr_warmup_epochs=0`; full LR=3e-4 from step 1. Our command in this PR added a 1-epoch LinearLR warmup (start_factor=0.05). Effective LR averaged ~1.5e-4 over the first epoch, so the model trained at half the rate alphonse did — that alone leaves us well behind on val_abupt at the first eval.
2. **`total_loss = sum(w_i * l_i)` with weights ~1.0 is ~5x the alphonse loss scale.** Alphonse's loss = surface_loss + volume_loss = (sp+wsx+wsy+wsz)/4 + vp (≈ 2 task-mean units). Our loss = sum of all 5 = 5 task-mean units. Median pre-clip grad norm: gradnorm=95.5 vs. alphonse=0.10 at comparable training stage. With `grad_clip_norm=1.0`, this means 100% of our steps clip vs. alphonse's 2%, passing through ~1/95 of the gradient. Combined with the warmup, effective per-step learning is ~200x slower than alphonse.

Compounding effect: warmup (½) × clipping (1/95) ≈ 1/200 effective LR for first epoch. End-of-epoch-1 val_abupt of 33.4 (vs. alphonse 13.5) is consistent with this gap.

### Controller behavior was healthy (ruled out as cause)

- `log_w` started at 0; weights renormalized correctly (range 0.999–1.001 at first update).
- All `g_unweighted` finite (range 0.005–0.016).
- After 17 controller updates: `w_vp=1.064, w_wsy=1.037, w_wsx=1.009, w_wsz=0.953, w_sp=0.937`.
- Controller is increasing weight on volume_pressure and wall_shear_y (highest train loss-ratios), pulling back from wall_shear_z and surface_pressure (lowest train loss-ratios). **This means GradNorm is balancing TRAIN-loss convergence rates, not VAL-relL2 ranks.** Volume pressure has the largest absolute MSE in normalized target space → highest loss-ratio → controller up-weights it; wsy/wsz wins/losses on val_relL2 don't directly map to train-loss ratios. This is a conceptual point about the hypothesis worth flagging — but it's secondary to the kill issue.
- Confirmed no controller-to-main-optimizer cross-contamination: `torch.autograd.grad` does not populate `.grad`; controller Adam only contains `controller.log_w`.

### Plan: relaunch with two corrections

1. `--lr-warmup-epochs 0` to match alphonse exactly.
2. `--grad-clip-norm 5.0` to compensate for the ~5x larger total loss scale (preserves the per-step effective gradient magnitude that alphonse experienced with clip=1.0). I judged this is cleaner than touching the loss formulation, since the PR explicitly specifies `total_loss = sum(w_i * l_i)`.
3. Kill threshold loosened to `2000:val_primary/abupt_axis_mean_rel_l2_pct<25` (alphonse hits 13.5 at end-of-epoch-1, so 25 still catches divergence with margin).
4. All other PR-specified parameters unchanged: `alpha=1.5`, `gradnorm_update_every=200`, `gradnorm_warmup_steps=1000`, model 4L/256d/4H Fourier PE, lr=3e-4, T_max=30, no-EMA, no-compile, batch_size=2, 50 epochs cap (will hit `SENPAI_MAX_EPOCHS`/timeout first).

If pre-clip grad norms are still pathological after these fixes (>50 sustained mid-training) I'll fall back to `alpha=1.0` per the PR fallback plan, or raise this back to you. Launching now and will report back when training + eval finish.

morganmcg1:
ADVISOR: v2 relaunch confirmed running.

I see your new GradNorm v2 runs in W&B group `bengio-stream2-edward-gradnorm`:
- `09kojb6q` (rank0), `p1bj5lt9` (rank1), `ecale38l` (rank2), `36mvutw5` (rank3)
- Name: `edward/gradnorm-alpha1.5-v2-rank*`
- All at step ~6,654 — pre-first-validation, run is healthy.

Good work on the root cause + relaunch with `--lr-warmup-epochs 0` and `--grad-clip-norm 5.0`. First validation lands at step 17,816 (≈ another 11k steps).

**Action items as it progresses:**
1. After epoch 1 val: post abupt% and per-task GradNorm weights (alpha=1.5 should be visibly rebalancing wsy/wsz upward by then).
2. Confirm `train/grad/clipped` rate dropped vs v1 — that was the v1 failure mode.
3. After epoch 3, if abupt < 12% we keep going; if abupt > 20% kill and reassess.

This is the only experiment in the wave directly attacking wall_shear_y/z, which is the binding global constraint across alphonse/frieren/kohaku/askeladd. Get me a clean signal and we'll know whether GradNorm is the right lever for it.

morganmcg1:
STUDENT edward:

## Status update — v2 epoch-1 results (advisor's items 1 & 2)

GradNorm v2 (group `bengio-stream2-edward-gradnorm`, run `09kojb6q` rank0) is healthy and well into epoch 2 at step 20520.

### Epoch-1 validation (PR's required tables, partial)

| metric | v2 epoch 1 | alphonse epoch 1 | gap |
|---|---:|---:|---:|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **14.48** | 13.50 | +0.98 |
| `val_primary/surface_pressure_rel_l2_pct` | 9.73 | — | |
| `val_primary/wall_shear_x_rel_l2_pct` | 12.85 | — | |
| `val_primary/wall_shear_y_rel_l2_pct` | 17.73 | — | |
| `val_primary/wall_shear_z_rel_l2_pct` | 19.82 | — | |
| `val_primary/volume_pressure_rel_l2_pct` | 12.27 | — | |

So epoch-1 abupt is +0.98pp vs alphonse — close but slightly behind. Plenty of headroom to catch up over 50 epochs. Will check abupt at end of epoch 3 and re-evaluate per your gating: <12% keep going / >20% kill.

### Item 2: clip rate confirmation — fix worked

v1 clip behavior (root cause of abort): pre-clip grad norm median **95.5**, clip biting on **~100%** of steps.

v2 with `--grad-clip-norm 5.0`:
- Pct clipped over 20,520 steps: **11.2%** (vs v1's ~100%)
- pre-clip median: **2.07** (vs v1's 95.5)
- pre-clip mean: 3.08 (down from ~95)
- pre-clip max so far: 337.6 (one early spike in warmup; recent values 1.5–2.5)

### Controller behavior at step 20,205

| task | `w_i` | `loss_ratio` | direction |
|---|---:|---:|---|
| volume_pressure | 1.047 | 0.208 | up (highest train loss-ratio) |
| wall_shear_x | 1.031 | 0.120 | slight up |
| wall_shear_y | 0.998 | 0.118 | flat |
| wall_shear_z | 0.965 | 0.122 | slight down |
| surface_pressure | 0.958 | 0.087 | slight down (lowest loss-ratio) |

**Mechanism observation (carrying over my earlier comment):** GradNorm rebalances by *train-loss convergence rates*, and wsx/wsy/wsz have nearly equal train loss ratios (0.118–0.122). So even with alpha=1.5, the controller doesn't preferentially up-weight wsy/wsz. The biggest weight motion is actually onto volume_pressure (highest train loss). I'll keep watching the weight evolution — if wsy/wsz fall further behind on val rel-L2 by epoch ~5, I'll consider switching to alpha=2.0 (more aggressive rebalancing) or, per the PR fallback, drop to alpha=1.0.

Will report back at epoch 3 with the gating check.

morganmcg1:
STUDENT edward:

## Status update — epoch 3 gating result: **PASS, keep going**

GradNorm v2 (`09kojb6q`, group `bengio-stream2-edward-gradnorm`) is healthy and training continues into epoch 4 (current step ~58000, ~24% through epoch 4).

### Per-epoch validation, edward vs alphonse at matched steps

| epoch (step) | val_abupt edward | val_abupt alphonse | gap | wsy edward | wsy alphonse | wsz edward | wsz alphonse |
| 1 (17816) | 14.48 | 13.50 | +0.98 | 17.73 | 17.30 | 19.82 | 19.96 |
| 2 (35633) | 11.98 | 10.92 | +1.06 | 14.85 | 14.58 | 16.29 | 15.96 |
| 3 (53450) | **10.74** | 9.79 | +0.95 | 13.32 | 13.03 | 14.73 | 14.59 |

Epoch 3 abupt = **10.74 < 12%** → continuing per advisor's gating. We are tracking ~1pp behind alphonse on val_abupt across all three eval points; the gap is stable, not growing.

### Per-task at epoch 3 vs alphonse

| metric | edward (gradnorm) | alphonse (equal-weight) | gap |
|---|---:|---:|---:|
| surface_pressure | 6.99 | 6.46 | +0.53 |
| wall_shear_x | 9.48 | 9.30 | +0.18 |
| wall_shear_y | 13.32 | 13.03 | +0.29 |
| wall_shear_z | 14.73 | 14.59 | +0.14 |
| **volume_pressure** | **9.18** | **5.57** | **+3.61** |

Note the volume-pressure gap: 3.6pp behind alphonse, and growing in absolute terms each epoch. Mechanism diagnosis below.

### GradNorm controller state at step 58000

| task | `w_i` | direction from 1.0 |
|---|---:|---|
| volume_pressure | 1.070 | +7.0% (highest train loss-ratio) |
| wall_shear_x | 1.090 | +9.0% |
| wall_shear_y | 0.994 | flat |
| wall_shear_z | 0.947 | **−5.3%** (opposite of hoped direction) |
| surface_pressure | 0.899 | −10.1% |

### Mechanism: why wsy/wsz are NOT being up-weighted

The hypothesis in the PR was "wsy/wsz are slow-converging tasks → GradNorm with alpha=1.5 should up-weight them within 5 epochs." But the controller sees them as **fast** convergers in train-loss space — their normalized MSE is small in absolute terms (~3e-3 each, vs. volume_pressure's ~7e-3). GradNorm rebalances by `loss_ratio_i = L_i(t)/L_i(t0)`, which is in train-loss-magnitude space, not in val-rel-L2 space. Wall shear's val rel-L2 is high because the *target denominator* is small (per-axis L2 normalized), not because train-loss is failing to descend on wall shear specifically.

This is the central conceptual finding so far: **GradNorm is the wrong instrument for closing val rel-L2 ranks**. It optimizes for equal train-loss descent rates, but the val rel-L2 rank gap is driven by per-axis variance and the L2 denominator, not by relative train-loss magnitudes.

### Health checks

- `train/grad/clipped` rate at recent steps = 0% (vs v1's ~100%). Clip fix from `--grad-clip-norm 5.0` is working.
- pre-clip grad norm trend: stable around 0.6–0.9 in recent steps; max 337 was a single epoch-0 spike.
- `train/base_mse_loss` descending normally; no NaN/Inf.
- All 4 ranks running synchronously.

### Plan

Continue training through to 50 epochs unless: (a) val_abupt regresses by >0.5pp between consecutive epochs, or (b) pre-clip grad norm spikes above 50 sustained. After full training, will run `--eval-only` on the best checkpoint per the PR's required-results spec and report final test_primary metrics + per-task GradNorm weight curves.

Will report back at ~epoch 8 with the next status, or sooner if anything diverges.

morganmcg1:
ADVISOR: Edward — excellent scientific work. This is a genuinely important mechanistic finding that goes beyond this PR.

## Your diagnosis is correct

**GradNorm is the wrong instrument for closing val rel-L2 ranks.** Your diagnosis is precise: the controller operates in train-loss-convergence space. It sees wsy/wsz as fast-converging tasks (small normalized MSE per step), so it doesn't up-weight them — despite the fact that they are the highest-error tasks in val rel-L2 space. The disconnect is architectural: GradNorm measures 'which task's train loss is falling slowest' and corrects for that, not 'which task has the highest val rel-L2 error.' These are orthogonal axes in this problem.

The volume_pressure gap is the clearest evidence. vol_p gets the highest GradNorm weight (w=1.070) because it's the slowest train-loss converger, yet this simultaneously makes it harder to close the wsy/wsz val-rel-L2 gap. GradNorm and val rel-L2 optimality point in different directions for this dataset.

## Continue to epoch 50

Do not stop early. The per-epoch val_abupt is still descending (14.48 → 11.98 → 10.74, -1.74pp/epoch average through ep3). At this rate you may reach 7–8% by epoch 20-25. Even if GradNorm doesn't close wsy/wsz, the end-of-wave data is valuable for two reasons:

1. **We need the full trajectory.** Does the gap to alphonse close, hold, or widen over 50 epochs? That tells us whether GradNorm provides any asymptotic benefit at all.
2. **The controller state at epoch 50 is our cleanest read of the mechanism.** If w_wsy/wsz are still near 1.0 at epoch 50, that kills the hypothesis cleanly for future design.

## What this means for Wave 2

The val rel-L2 gap for wsy/wsz is not a gradient magnitude problem. It's a **denominator problem**: the per-axis L2 denominator (sum of squared ground-truth values per axis) is small for wsy/wsz, so small absolute errors become large relative errors. The correct instrument is either:

1. **Loss reweighting in val-rel-L2 space directly** — weight the training loss by the inverse of the expected per-axis denominator magnitude (a per-axis scale factor derived from dataset statistics). This is cheap to implement.
2. **Per-axis relative MSE in the loss function** — replace  with  per axis. This is what the val metric actually measures and would align training and validation.

These are the hypotheses to test in Wave 2 for this track. Note them in your PR summary when you post final results.

## Required at epoch 50 (or best checkpoint if plateau detected)

1. Full per-epoch validation table (all 50 epochs)
2. Controller weight trajectory (w_i at epoch 5, 10, 20, 50)
3. Final test_primary/* metrics from best-val checkpoint
4. Your recommendation: does GradNorm help at all, even as a regularizer?

morganmcg1:
ADVISOR: Closing PR #137 because ADVISOR: Closing as dead end. Run v5ybmwra finished with val_primary/abupt_axis_mean_rel_l2_pct = 33.43% — diverged ~4-5x above the cohort range (8.4-8.9%). No usable test_primary metrics populated. GradNorm in this configuration was unstable on DrivAerML; the per-task gradient norm balancing likely interacted badly with the wsy/wsz vs surf_p scale gap. Reassigning you immediately to a fresh Wave 2 hypothesis. Lessons: (1) any per-task weighting scheme on this dataset has to handle the ~3-4x scale dominance of wsy/wsz from the start; (2) explicit kill-thresholds at step 2000 should have caught this earlier — make sure your next run has --kill-thresholds set..

PR #140, opened 2026-05-01
[senku] DrivAerML Curriculum Mass-Weighted Case Sampling (E1)

morganmcg1:
## Hypothesis

**Curriculum (easy-first) case sampling via per-case gradient-norm difficulty proxy will improve convergence speed and final `abupt_axis_mean_rel_l2_pct`.**

In Wave 1, all 16 runs used uniform random case sampling throughout training. However, DrivAerML cases vary substantially in aerodynamic complexity — some vehicles produce nearly laminar flow while others have strong separation and recirculation. Standard sampling presents hard and easy cases with equal frequency from epoch 0, potentially causing early gradient conflict and slowing the model's ability to build robust representations.

The hypothesis: **training on "easy" (low gradient-norm) cases first and gradually introducing harder cases** mirrors classical curriculum learning findings and should help the model establish stable surface pressure and wall-shear representations before being overwhelmed by aerodynamically complex geometries. Once the model has a stable foundation (epochs 1–10), switching to uniform sampling ensures full coverage of the training distribution.

This experiment uses the **alphonse Wave 1 leader recipe** (4L/256d/4H + Fourier PE + T_max=30 + no-EMA) as the base — the strongest known DDP4 configuration.

---

## Baseline to Beat

**AB-UPT public reference targets (all must be beaten simultaneously):**

| Metric | AB-UPT Target |
|--------|--------------|
| `test_primary/abupt_axis_mean_rel_l2_pct` | **4.51** |

**Wave 1 alphonse mid-training reference** (run `m9775k1v`, ~step 421k, not final test metrics):
- abupt=7.33, surf_p=4.87, vol_p=4.23

---

**You must implement curriculum sampling as a code change to `trainer_runtime.py`.** There is no `--curriculum` CLI flag — you need to add one and wire it through.

### Step 1 — Add a CLI flag to `Config` in `train.py`

In the `Config` dataclass (around line 90 in `train.py`), add:

```python
curriculum_warmup_epochs: int = 0  # 0 = disabled, >0 = use curriculum for this many epochs
```

This is cleanly defaulted to 0 (disabled) so existing runs are unaffected.

### Step 2 — Implement difficulty scoring and WeightedRandomSampler in `trainer_runtime.py`

In `make_loaders`, the `train_sampler` is currently a `DistributedSampler`. You need to make it possible to **swap in a `WeightedRandomSampler`** for curriculum epochs.

Add a new function `compute_case_difficulties(train_loader, model, device, loss_fn)` that:
1. Runs a **single pass over the entire training set** (no gradient accumulation, just `loss.item()` per case) to get per-case loss values
2. Returns a tensor of shape `[N_cases]` with the per-case loss (higher loss = harder case)

Then in the main training loop in `train.py`, after epoch 0 completes (when `curriculum_warmup_epochs > 0`):
1. Call `compute_case_difficulties` to get difficulty scores
2. For epochs 1 through `curriculum_warmup_epochs`, construct a `WeightedRandomSampler` with weights **inversely proportional** to difficulty: `weights = 1.0 / (difficulties + eps)`, normalized. Replace the train_loader sampler.
3. From epoch `curriculum_warmup_epochs + 1` onward, switch back to standard `DistributedSampler` (or shuffle=True if non-distributed).

**Practical implementation note for DDP**: `WeightedRandomSampler` is not natively DDP-aware. The simplest correct approach:
- Compute difficulty scores on rank 0 and broadcast to all ranks via `dist.broadcast`
- Construct separate per-rank `WeightedRandomSampler` instances (each rank samples its own subset, same weights), OR
- Use the difficulty weights to build a new `DistributedSampler`-like sampler that respects per-case weights across ranks

The simplest correct DDP approach: use **loss-weighted epoch-0 case ordering** on rank 0, broadcast a shuffled index list to all ranks, then split indices across ranks. This avoids the need for a custom `WeightedRandomSampler` and is fully compatible with DDP.

Here is a concrete approach for the `make_loaders` / training loop:

```python
# In train.py, after epoch 0 training loop:
if config.curriculum_warmup_epochs > 0 and epoch == 0:
    # Compute per-case difficulties using training loss
    case_difficulties = compute_case_difficulties(
        train_loader=train_loader,
        model=unwrap_model(model),
        device=state.device,
        config=config,
        target_transform=target_transform,
    )
    # Broadcast from rank 0 to all ranks
    if state.enabled:
        dist.broadcast(case_difficulties, src=0)
    # Store for use in subsequent epochs
    curriculum_weights = 1.0 / (case_difficulties + 1e-6)
    curriculum_weights /= curriculum_weights.sum()
```

Then in the epoch loop header, before creating the train DataLoader for each epoch:

```python
if config.curriculum_warmup_epochs > 0 and 1 <= epoch <= config.curriculum_warmup_epochs:
    # Use weighted sampler for curriculum phase
    sampler = WeightedRandomSampler(
        weights=curriculum_weights,
        num_samples=len(train_ds),
        replacement=True,
    )
    train_loader = DataLoader(train_ds, batch_size=config.batch_size, sampler=sampler, **loader_kwargs(config))
elif epoch == config.curriculum_warmup_epochs + 1:
    # Switch back to standard distributed sampler
    train_loader = make_train_loader(config, train_ds, distributed_state=state)
```

**IMPORTANT**: The `compute_case_difficulties` function must:
- Use `torch.no_grad()` to avoid storing computation graphs
- Use the **normalized (standardized)** targets via `target_transform.apply_surface` / `apply_volume`
- Compute loss in the same way as the training loop (matching surface+volume loss formulation)
- Handle variable-length batches correctly (the dataset uses `pad_collate`)
- Average the loss over all points in each case to get a scalar per-case difficulty

### Step 3 — Run configuration

Use the **alphonse base recipe** with curriculum enabled for the first 10 epochs:

```bash
python train.py \
  --epochs 50 \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --no-use-ema \
  --lr-cosine-t-max 30 \
  --no-compile-model \
  --curriculum-warmup-epochs 10 \
  --wandb-group senku-curriculum \
  --wandb-run-name senku-curriculum-e10
```

**Note**: `--no-compile-model` is **mandatory** — PyTorch 2.x Inductor crashes at validation on this cluster.

### Step 4 — Logging

Log the following additional metrics to W&B to help interpret results:
- `train/curriculum_active` (1 during curriculum epochs, 0 after) — so we can see the phase transition
- `train/case_difficulty_mean` and `train/case_difficulty_std` (from epoch 0 difficulty scores)
- `train/curriculum_epoch` (current epoch within curriculum phase)

---

## Expected Outcome

If the hypothesis is correct, you should see:
- **Faster early convergence**: lower `val_primary/abupt_axis_mean_rel_l2_pct` at epochs 5–15 vs alphonse
- **Better final metrics**: surf_p and vol_p should benefit from stable early representations
- The wall-shear y/z axes (the binding constraint in Wave 1) may or may not improve — document whatever you observe

If convergence is similar to alphonse but final metrics don't improve, the curriculum warmup may be too short or the difficulty proxy (loss) may not correlate well with aerodynamic complexity. In that case, please note the per-case difficulty variance in your report (high variance = the cases ARE diverse; low variance = loss is not a good proxy for difficulty).

---

## Reporting

When your run completes, post a comment with:
1. W&B run ID and link
2. Final `test_primary/*` metrics table (all 6 axes + abupt_mean)
3. Best validation checkpoint metrics
4. Comparison vs AB-UPT targets and alphonse Wave 1 reference
5. Plot or description of per-case difficulty distribution from epoch 0
6. Your observation on whether curriculum actually helped convergence (compare early-epoch val curves)
7. Any implementation notes or issues encountered

Then mark the PR ready for review.


morganmcg1:
ADVISOR: Welcome to Wave 2 of the bengio research programme.

Your PR is set up correctly with the curriculum mass-weighted sampling hypothesis. This is a sound idea — DrivAerML cases have genuine aerodynamic diversity and easy-first curriculum learning has strong precedent in NLP and vision.

**Before you start coding, read this carefully:**

1. **PyTorch compile bug is mandatory to avoid**: Add `--no-compile-model` to ALL launch commands. The default `compile_model=True` crashes on first validation pass due to a PyTorch 2.x Inductor bug in `tiling_utils.get_pw_red_splits`. Without this flag, your run dies at epoch 1.

2. **Wave 1 context**: The current wave leader is alphonse (PR #74), run `m9775k1v`, best val abupt=7.324% at epoch 24 (still training). This is what your curriculum approach needs to beat.

3. **Implementation priority order**:
   - Keep the DDP difficulty-score broadcast simple: compute on rank0, broadcast via `dist.broadcast`, split indices manually. Don't overthink the custom sampler.
   - Use `torch.no_grad()` for the difficulty scoring pass. It's expensive — one full forward pass over the training set. Consider doing it on a 20% subsample if it takes >5 min.
   - Log `train/curriculum_active`, `train/case_difficulty_mean`, `train/case_difficulty_std` to W&B exactly as specified in the PR.

4. **Your launch command** (as per PR body):
```bash
python train.py \
  --epochs 50 \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --no-use-ema \
  --lr-cosine-t-max 30 \
  --no-compile-model \
  --curriculum-warmup-epochs 10 \
  --wandb-group senku-curriculum \
  --wandb-run-name senku-curriculum-e10
```

5. **First status update required**: Post your W&B run ID and epoch 1 val metrics as soon as they're available. I need to see that the curriculum is active (`train/curriculum_active=1`) and that the difficulty scores have non-trivial variance (if all cases have similar difficulty, the proxy is not working).

Get started and keep me posted.

morganmcg1:
ADVISOR: Senku — you were assigned this PR at 01:06Z and have not yet posted a status update. You are running curriculum mass-weighted sampling (Wave 2 hypothesis). Your onboarding instructions are in the PR body.

Required immediately:
1. Confirm your training command with all required flags including 
2. Post your W&B run ID (rank0)
3. Post epoch 1 val_abupt once training reaches first validation

The torch.compile Inductor bug is a known crash risk — confirm  is in your command before anything else. Post your update now.

PR #119, opened 2026-04-30
edward: RFF coordinate encoding (Tancik 2020) vs sincos

morganmcg1:
## Hypothesis

The current codebase uses `ContinuousSincosEmbed` for 3D coordinate encoding — a deterministic sinusoidal Fourier encoding. Gaussian Random Fourier Features (RFF, Tancik et al. 2020, "Fourier Features Let Networks Learn High Frequency Functions in Low Dimensional Domains") project coordinates through a learnable random Gaussian matrix before sincos encoding, which can better capture high-frequency spatial patterns.

RFF encoding theory: given coordinates x ∈ R^d, encode as [sin(2π B x), cos(2π B x)] where B ∈ R^{d×m} is sampled from N(0, σ²). The σ hyperparameter controls the bandwidth — higher σ captures finer-grained spatial detail. For CFD with steep pressure gradients near surfaces, this may provide richer coordinate representations.

This code change requires **modifying `train.py`**. Add a `RFFEncoding` class and a `--rff-num-features` flag that, when non-zero, replaces `ContinuousSincosEmbed` with RFF encoding.

## Instructions

**Step 1: Add RFF encoding to `train.py`**

Find the `ContinuousSincosEmbed` class (around line 100) and add the following class immediately after it:

```python
class RFFEncoding(nn.Module):
    """Gaussian random Fourier feature coordinate encoding (Tancik et al. 2020).
    Projects 3D coordinates via N(0, sigma^2) random matrix before sincos encoding.
    Captures high-frequency spatial patterns better than fixed sincos embedding.
    """
    def __init__(self, in_dim: int, num_features: int = 64, sigma: float = 1.0):
        super().__init__()
        self.in_dim = in_dim
        self.num_features = num_features
        self.sigma = sigma
        self.register_buffer("B", torch.randn(in_dim, num_features) * sigma)

    @property
    def output_dim(self) -> int:
        return 2 * self.num_features

    def forward(self, x: torch.Tensor) -> torch.Tensor:
        import math
        proj = 2.0 * math.pi * (x @ self.B.to(dtype=x.dtype))
        return torch.cat([proj.sin(), proj.cos()], dim=-1)
```

**Step 2: Add CLI flags**

In the argument parser, add:
```python
parser.add_argument("--rff-num-features", type=int, default=0,
    help="If >0, use RFF coordinate encoding with this many features (replaces sincos). Default: 0 (sincos).")
parser.add_argument("--rff-sigma", type=float, default=1.0,
[...]

PR #143, opened 2026-05-01
fern: coordinate normalization sweep (fix sincos anisotropy)

morganmcg1:
## Hypothesis

DrivAerML point-cloud coordinates are raw meters with a strongly anisotropic bounding box: x-axis spans ~8m (vehicle length), y-axis ~2.5m (width), z-axis ~2m (height). `ContinuousSincosEmbed` multiplies raw coordinates by a fixed frequency bank `omega`, so each axis's sinusoidal frequencies are implicitly scaled by its physical extent. This means the x-axis gets ~3–4× more frequency resolution than y/z — the positional encoding "sees" more detail along the vehicle axis than across it.

This anisotropy is a likely contributor to the persistent 4× gap between our tau_y/tau_z and AB-UPT. Surface wall-shear in the y/z direction is predominantly driven by crossflow and spanwise gradients, which the model must resolve at ~2m scale while the embedding is calibrated for an 8m range. Normalizing coordinates to a unit cube (or unit sphere) before `pos_embed` gives equal per-axis frequency resolution.

Evidence: edward's RFF experiment (PR #119) found wall_shear_y/z were the worst per-axis components (21.3%, 23.2%), and he explicitly noted "isotropic σ produces uneven effective frequencies on the y/z axes." The same issue applies to `ContinuousSincosEmbed` with raw-meter coordinates.

**Proposed change:** Add a `--coord-normalize` flag with three modes to `train.py`:

- `none` (default, current behavior — raw meter coords)
- `global-scale` — divide all axes by the global bounding-box diagonal (single scalar, preserves isotropy of existing encoding)
- `per-axis` — divide each axis by its own range (x/8.0, y/2.5, z/2.0), giving equal frequency density per axis

The normalization is applied in `SurfaceTransolver._encode_group` before calling `self.pos_embed(pos)`. The normalization statistics should be pre-computed from the training set bounding box and stored as `register_buffer` tensors so they're consistent at val/test time.

## Instructions

**New flag:** Add `--coord-normalize` to the `Config` dataclass and argparser with choices `["none", "global-scale", "per-axis"]`, default `"none"`. This preserves backward compatibility.

**Implementation in `SurfaceTransolver.__init__`:**
```python
# In SurfaceTransolver.__init__, add:
self.coord_normalize = coord_normalize  # store the flag
# These will be set externally after construction with actual data stats:
self.register_buffer("coord_shift", torch.zeros(space_dim))   # per-axis min
self.register_buffer("coord_scale", torch.ones(space_dim))    # per-axis divisor
```

**Set the normalization statistics in `build_model` or before the training loop** by computing the min/max of all surface+volume xyz from the training dataset loader (first pass or from a pre-computed stats file). Example:
```python
# Compute from training data (do this once, before training loop starts):
x_min = torch.tensor([-1.0, -1.25, -0.05])   # approx DrivAerML bbox min (meters)
x_max = torch.tensor([7.0,  1.25,  1.95])    # approx DrivAerML bbox max (meters)
# For 'per-axis': divide by per-axis range
coord_scale_per_axis = x_max - x_min          # shape [3]
# For 'global-scale': divide by diagonal (single scalar broadcast)
coord_scale_global = torch.norm(x_max - x_min).expand(3)
```

The simplest correct approach is to compute these from the **actual training batches** using a one-pass scan before training, OR hardcode the known DrivAerML bounding box values. Either is fine — the key is that val/test coordinates are normalized with the SAME shift/scale.

**Apply normalization in `_encode_group`:**
```python
def _encode_group(self, x, *, project_features, bias, placeholder):
    pos = x[:, :, :self.space_dim]
    if self.coord_normalize != "none":
        pos = (pos - self.coord_shift) / self.coord_scale.clamp(min=1e-6)
    hidden = self.pos_embed(pos)
    ...
```

**Three arms to run:**
| Arm | `--coord-normalize` | Run name |
|---|---|---|
| A | `none` (control — exact PR #99 config, reproduced) | `fern-coord-none` |
| B | `global-scale` | `fern-coord-global` |
| C | `per-axis` | `fern-coord-peraxis` |

Use `--wandb-group fern-coord-normalization` for all arms.

All other flags identical to PR #99 baseline:
```bash
cd target/
python train.py \
  --coord-normalize <none|global-scale|per-axis> \
  --volume-loss-weight 2.0 \
  --batch-size 8 \
  --validation-every 1 \
  --ema-decay 0.9995 \
  --clip-grad-norm 1.0 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
  --wandb-group fern-coord-normalization \
  --wandb-name fern-coord-<none|global|peraxis>
```

**Logging:** Log `coord/shift_{x,y,z}` and `coord/scale_{x,y,z}` as hyperparameters in W&B init so the actual normalization applied is visible.

**Watch for:** The per-axis arm should show lower val loss on tau_y and tau_z specifically. If both B and C beat A, the per-axis arm is the deeper insight. If only C beats A, that confirms the anisotropy hypothesis directly.

## Baseline

Current best — PR #99 (fern), W&B run `3hljb0mg`:

| Metric | val | test |
|---|---:|---:|
| `abupt_axis_mean_rel_l2_pct` | **10.69** | 11.73 |
| `surface_pressure_rel_l2_pct` | 6.97 | 6.64 |
| `wall_shear_rel_l2_pct` | 11.69 | 11.48 |
| `volume_pressure_rel_l2_pct` | 7.85 | 14.42 |
| `wall_shear_x_rel_l2_pct` | 10.17 | 10.06 |
| `wall_shear_y_rel_l2_pct` | **13.73** | 13.53 |
| `wall_shear_z_rel_l2_pct` | **14.73** | 13.98 |

AB-UPT targets (to beat long-term): sp=3.82, wsx=5.35, wsy=3.65, wsz=3.63, vp=6.08.

Success criterion: any arm with `val_primary/abupt_axis_mean_rel_l2_pct < 10.69`.

Reproduce baseline:
```bash
cd target/
python train.py \
```

### DIFF
--- a/train.py
+++ b/train.py
@@ -351,9 +351,10 @@ class SurfaceTransolver(nn.Module):
         coord_normalize: str = "none",
     ):
         super().__init__()
-        if coord_normalize not in {"none", "global-scale", "per-axis"}:
+        if coord_normalize not in {"none", "global-scale", "per-axis", "per-axis-clamped"}:
             raise ValueError(
-                f"coord_normalize must be 'none', 'global-scale', or 'per-axis'; got {coord_normalize!r}"
+                "coord_normalize must be 'none', 'global-scale', 'per-axis', "
+                f"or 'per-axis-clamped'; got {coord_normalize!r}"
             )
         self.space_dim = space_dim
         self.surface_input_dim = surface_input_dim
@@ -409,6 +410,8 @@ class SurfaceTransolver(nn.Module):
         pos = x[:, :, : self.space_dim]
         if self.coord_normalize != "none":
             pos = (pos - self.coord_shift) / self.coord_scale.clamp(min=1e-6)
+            if self.coord_normalize == "per-axis-clamped":
+                pos = pos.clamp(0.0, 1.0)
         hidden = self.pos_embed(pos)
         if project_features is not None and x.shape[-1] > self.space_dim:
             hidden = hidden + project_features(x[:, :, self.space_dim :])
@@ -798,7 +801,7 @@ def resolve_coord_normalization(
     if mode == "none":
         return torch.zeros_like(xmin), torch.ones_like(xmin)
     extent = xmax - xmin
-    if mode == "per-axis":
+    if mode in ("per-axis", "per-axis-clamped"):
         return xmin.clone(), extent.clone()
     if mode == "global-scale":
         diag = torch.norm(extent)
--- a/train.py
+++ b/train.py
@@ -1844,6 +1844,7 @@ def main(argv: Iterable[str] | None = None) -> None:
     best_val = float("inf")
     best_metrics: dict[str, float] = {}
     global_step = 0
+    nonfinite_grad_skip_count = 0
     early_stop_reason: str | None = None
     timeout_hit = False
     train_start = time.time()
@@ -1894,16 +1895,22 @@ def main(argv: Iterable[str] | None = None) -> None:
                 if should_log_gradients
                 else {}
             )
+            grad_finite = True
             if config.clip_grad_norm > 0:
                 pre_clip_norm = torch.nn.utils.clip_grad_norm_(
                     model.parameters(), max_norm=config.clip_grad_norm
                 )
+                grad_finite = bool(torch.isfinite(pre_clip_norm).item())
                 if should_log_gradients:
                     gradient_metrics["train/grad/pre_clip_norm"] = float(pre_clip_norm)
                     gradient_metrics["train/grad/clip_threshold"] = config.clip_grad_norm
-            optimizer.step()
+            if grad_finite:
+                optimizer.step()
+            else:
+                nonfinite_grad_skip_count += 1
+                optimizer.zero_grad(set_to_none=True)
             ema_decay_now: float | None = None
-            if ema is not None:
+            if ema is not None and grad_finite:
                 if config.ema_decay_start > 0.0:
                     progress = min(global_step / max(total_estimated_steps, 1), 1.0)
                     cos_val = (1.0 - math.cos(math.pi * progress)) / 2.0
@@ -1931,6 +1938,8 @@ def main(argv: Iterable[str] | None = None) -> None:
                 "train/loss_tau_x": batch_loss_metrics["loss_tau_x"],
                 "train/loss_tau_y": batch_loss_metrics["loss_tau_y"],
                 "train/loss_tau_z": batch_loss_metrics["loss_tau_z"],
+                "train/grad/step_skipped": 0.0 if grad_finite else 1.0,
+                "train/grad/nonfinite_skip_total": float(nonfinite_grad_skip_count),
                 "global_step": global_step,
                 **gradient_metrics,
                 **weight_metrics,
--- a/train.py
+++ b/train.py
@@ -348,8 +348,13 @@ class SurfaceTransolver(nn.Module):
         stochastic_depth_prob: float = 0.0,
         use_film: bool = False,
         film_encoder_dim: int = 64,
+        coord_normalize: str = "none",
     ):
         super().__init__()
+        if coord_normalize not in {"none", "global-scale", "per-axis"}:
+            raise ValueError(
+                f"coord_normalize must be 'none', 'global-scale', or 'per-axis'; got {coord_normalize!r}"
+            )
         self.space_dim = space_dim
         self.surface_input_dim = surface_input_dim
         self.surface_output_dim = surface_output_dim
@@ -359,6 +364,9 @@ class SurfaceTransolver(nn.Module):
         volume_extra_dim = max(0, self.volume_input_dim - space_dim)
         self.use_film = use_film
         self.film_encoder_dim = film_encoder_dim
+        self.coord_normalize = coord_normalize
+        self.register_buffer("coord_shift", torch.zeros(space_dim))
+        self.register_buffer("coord_scale", torch.ones(space_dim))
 
         self.pos_embed = ContinuousSincosEmbed(hidden_dim=n_hidden, input_dim=space_dim)
         self.surface_bias = MLP(input_dim=n_hidden, hidden_dim=n_hidden, output_dim=n_hidden)
@@ -399,6 +407,8 @@ class SurfaceTransolver(nn.Module):
         placeholder: torch.Tensor,
     ) -> torch.Tensor:
         pos = x[:, :, : self.space_dim]
+        if self.coord_normalize != "none":
+            pos = (pos - self.coord_shift) / self.coord_scale.clamp(min=1e-6)
         hidden = self.pos_embed(pos)
         if project_features is not None and x.shape[-1] > self.space_dim:
             hidden = hidden + project_features(x[:, :, self.space_dim :])
@@ -589,6 +599,7 @@ class Config:
     kill_thresholds: str = ""
     clip_grad_norm: float = 0.0
     compile_model: bool = True
+    coord_normalize: str = "none"
     debug: bool = False
 
 
@@ -730,7 +741,69 @@ def build_model(config: Config) -> SurfaceTransolver:
         stochastic_depth_prob=config.stochastic_depth_prob,
         use_film=config.use_film,
         film_encoder_dim=config.film_encoder_dim,
+        coord_normalize=config.coord_normalize,
     )
+
+
+def compute_coord_stats(
+    manifest_path: str,
+    root: str | None,
+    *,
+    debug: bool,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    """Per-axis (min, max) of vehicle-surface xyz coords across the training split.
+
+    Scans surface_xyz.npy via mmap; reuses the loader's PVC mirror/symlink path
+    resolver so the scan works on either mount layout.
+
+    Surface-only by design: volume_xyz extends across the wind-tunnel CFD
+    domain (~120 x 43 x 21 m), 25x wider than the vehicle (~5 x 2 x 1.5 m).
+    Including volume bounds would shrink surface features to ~4% of the unit
+    cube — the opposite of what the PR hypothesis aims to fix. Volume points
+    are encoded by the same shifted/scaled embedding; they fall outside the
+    unit cube but sin/cos pos_embed handles that gracefully.
+    """
+
+    import numpy as np
+
+    from data import DrivAerMLCaseStore
+    from data.loader import _resolve_artifact_path
+
+    store = DrivAerMLCaseStore(manifest_path=manifest_path, root=root)
+    train_ids = store.case_ids("train")
+    if debug:
+        train_ids = train_ids[:4]
+    xmin = np.full(3, np.inf, dtype=np.float64)
+    xmax = np.full(3, -np.inf, dtype=np.float64)
+    for case_id in train_ids:
+        case_dir = store.root / case_id
+        arr = np.load(_resolve_artifact_path(case_dir / "surface_xyz.npy"), mmap_mode="r")
+        xmin = np.minimum(xmin, arr.min(axis=0).astype(np.float64))
+        xmax = np.maximum(xmax, arr.max(axis=0).astype(np.float64))
+    if not np.isfinite(xmin).all() or not np.isfinite(xmax).all():
+        raise RuntimeError("compute_coord_stats produced non-finite bounds")
+    return (
+        torch.from_numpy(xmin.astype(np.float32)),
+        torch.from_numpy(xmax.astype(np.float32)),
+    )
+
+
+def resolve_coord_normalization(
+    mode: str,
+    xmin: torch.Tensor,
+    xmax: torch.Tensor,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    """Return per-axis (shift, scale) buffers for the requested mode."""
+
+    if mode == "none":
+        return torch.zeros_like(xmin), torch.ones_like(xmin)
+    extent = xmax - xmin
+    if mode == "per-axis":
+        return xmin.clone(), extent.clone()
+    if mode == "global-scale":
+        diag = torch.norm(extent)
+        return xmin.clone(), diag.expand_as(extent).clone()
+    raise ValueError(f"unknown coord_normalize mode: {mode!r}")
 
 
 def _metric_path(name: str) -> str:
@@ -1680,6 +1753,25 @@ def main(argv: Iterable[str] | None = None) -> None:
     )
 
     model = build_model(config).to(device)
+    coord_bbox: dict[str, list[float]] = {}
+    if config.coord_normalize != "none":
+        xmin, xmax = compute_coord_stats(
+            config.manifest, config.data_root or None, debug=config.debug
+        )
+        shift, scale = resolve_coord_normalization(config.coord_normalize, xmin, xmax)
+        model.coord_shift.copy_(shift.to(device=device, dtype=model.coord_shift.dtype))
+        model.coord_scale.copy_(scale.to(device=device, dtype=model.coord_scale.dtype))
+        coord_bbox = {
+            "xmin": xmin.tolist(),
+            "xmax": xmax.tolist(),
+            "shift": shift.tolist(),
+            "scale": scale.tolist(),
+        }
+        print(
+            f"Coord normalize='{config.coord_normalize}' "
+            f"xmin={coord_bbox['xmin']} xmax={coord_bbox['xmax']} "
+            f"shift={coord_bbox['shift']} scale={coord_bbox['scale']}"
+        )
     if config.compile_model:
         model = torch.compile(model)
     n_params = sum(param.numel() for param in model.parameters())
@@ -1711,6 +1803,7 @@ def main(argv: Iterable[str] | None = None) -> None:
             "surface_targets": SURFACE_TARGET_NAMES,
             "volume_targets": VOLUME_TARGET_NAMES,
             "total_estimated_steps": total_estimated_steps,
+            "coord_bbox": coord_bbox,
         },
         mode=os.environ.get("WANDB_MODE", "online"),
     )
--- a/train.py
+++ b/train.py
@@ -351,10 +351,17 @@ class SurfaceTransolver(nn.Module):
         coord_normalize: str = "none",
     ):
         super().__init__()
-        if coord_normalize not in {"none", "global-scale", "per-axis", "per-axis-clamped"}:
+        if coord_normalize not in {
+            "none",
+            "global-scale",
+            "per-axis",
+            "per-axis-clamped",
+            "per-axis-surfonly",
+        }:
             raise ValueError(
                 "coord_normalize must be 'none', 'global-scale', 'per-axis', "
-                f"or 'per-axis-clamped'; got {coord_normalize!r}"
+                "'per-axis-clamped', or 'per-axis-surfonly'; "
+                f"got {coord_normalize!r}"
             )
         self.space_dim = space_dim
         self.surface_input_dim = surface_input_dim
@@ -406,12 +413,17 @@ class SurfaceTransolver(nn.Module):
         project_features: LinearProjection | None,
         bias: MLP,
         placeholder: torch.Tensor,
+        is_surface: bool,
     ) -> torch.Tensor:
         pos = x[:, :, : self.space_dim]
         if self.coord_normalize != "none":
-            pos = (pos - self.coord_shift) / self.coord_scale.clamp(min=1e-6)
-            if self.coord_normalize == "per-axis-clamped":
-                pos = pos.clamp(0.0, 1.0)
+            apply_norm = True
+            if self.coord_normalize == "per-axis-surfonly" and not is_surface:
+                apply_norm = False
+            if apply_norm:
+                pos = (pos - self.coord_shift) / self.coord_scale.clamp(min=1e-6)
+                if self.coord_normalize == "per-axis-clamped":
+                    pos = pos.clamp(0.0, 1.0)
         hidden = self.pos_embed(pos)
         if project_features is not None and x.shape[-1] > self.space_dim:
             hidden = hidden + project_features(x[:, :, self.space_dim :])
@@ -445,6 +457,7 @@ class SurfaceTransolver(nn.Module):
                     project_features=self.project_surface_features,
                     bias=self.surface_bias,
                     placeholder=self.surface_placeholder,
+                    is_surface=True,
                 )
             )
             masks.append(surface_mask)
@@ -457,6 +470,7 @@ class SurfaceTransolver(nn.Module):
                     project_features=self.project_volume_features,
                     bias=self.volume_bias,
                     placeholder=self.volume_placeholder,
+                    is_surface=False,
                 )
             )
             masks.append(volume_mask)
@@ -801,7 +815,7 @@ def resolve_coord_normalization(
     if mode == "none":
         return torch.zeros_like(xmin), torch.ones_like(xmin)
     extent = xmax - xmin
-    if mode in ("per-axis", "per-axis-clamped"):
+    if mode in ("per-axis", "per-axis-clamped", "per-axis-surfonly"):
         return xmin.clone(), extent.clone()
     if mode == "global-scale":
         diag = torch.norm(extent)


morganmcg1:
ADVISOR: fern — your coordinate normalization PR has been open since yesterday with no comments or W&B activity. This is one of Morgan's priority experiments addressing the 4× y/z gap. Please post an implementation status update immediately: have you started coding, or are runs already in flight? If running, post W&B run IDs and current step counts. If stuck, describe the blocker.

morganmcg1:
STUDENT fern:

## Implementation note — surface-only bbox stats

Implemented `--coord-normalize {none,global-scale,per-axis}` per the PR.

**Deviation from PR text:** the instructions say "compute the min/max of all surface+volume xyz from the training dataset," but DrivAerML volume points cover the wind-tunnel CFD domain, not just the vehicle. Measured per case:

- Surface bbox (`run_1`): x ∈ [-0.77, 3.94], y ∈ [-1.01, 1.01], z ∈ [-0.32, 1.20]
- Volume bbox (`run_1`): x ∈ [-39.7, 79.7], y ∈ [-21.6, 21.6], z ∈ [-1.65, 19.3]

The volume domain is ~25× wider than the vehicle in x and ~20× in y. Using joint surface+volume stats would compress the vehicle into ~4% of the unit cube along each axis — making **surface** wall-shear resolution worse, the opposite of the hypothesis. The PR's hardcoded example bbox (`x_max = [7.0, 1.25, 1.95]`) matches the vehicle scale, so I read this as the author's intent.

`compute_coord_stats` therefore scans `surface_xyz.npy` only across all 400 train cases. Volume points pass through the same shifted/scaled embedding (so val/test consistency is preserved) and end up outside the unit cube, but sin/cos `ContinuousSincosEmbed` handles out-of-cube positions fine.

Running 3 arms now on GPU 0/1/2 with the PR's exact reproduce config:
- A: `--coord-normalize none` → `fern-coord-none`
- B: `--coord-normalize global-scale` → `fern-coord-global`
- C: `--coord-normalize per-axis` → `fern-coord-peraxis`

Group: `fern-coord-normalization`. Code in commit 33b5169.

morganmcg1:
STUDENT fern:

## Status update

All 3 arms launched and training. Run IDs:

| Arm | Mode | W&B run | GPU | Epoch progress (as of 02:03) |
|---|---|---|---|---|
| A | `none` (control) | [`aef8nv9o`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/aef8nv9o) | 0 | 1/50, 28% |
| B | `global-scale` | [`dbhs9qng`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/dbhs9qng) | 1 | 1/50, 25% |
| C | `per-axis` | [`sr0qik6z`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/sr0qik6z) | 2 | 1/50, 25% |

Training rate ~2.14 it/s, 10883 batches/epoch ⇒ ~80 min/epoch. With `SENPAI_TIMEOUT_MINUTES=360` budget, expecting ~4 epochs/arm before timeout. Group: `fern-coord-normalization`.

The early ping crossed with my implementation comment (~9 min apart). Sorry for the delay — bbox-stats deviation note is in commit 33b5169 / prior comment.

morganmcg1:
STUDENT fern:

## Bug fix — skip optimizer step on non-finite gradient (separate fix in commit 69169ec)

While running the sweep I hit a training-loop bug that corrupts the model when *any* batch produces a non-finite gradient. Reporting separately per the bug-fix protocol so the fix can be reviewed independently of the experiment results.

**Symptom:** the `none` (control) and `per-axis` arms each hit a single NaN train/loss mid-epoch 1, then **stayed NaN for thousands of steps** through the rest of the epoch:
- `none` (`aef8nv9o`): first NaN at step 7740/10883, then 3632 consecutive NaN losses
- `per-axis` (`sr0qik6z`): first NaN at step 10643/10883, then 357 consecutive NaN losses
- `global-scale` (`dbhs9qng`): no NaN, healthy through epoch 2

The control arm should reproduce PR #99 (which had 0 NaN steps in 33k steps), so this isn't coord-normalize-specific — it's a general latent bug, just stochastic.

**Root cause:** in the train step, `torch.nn.utils.clip_grad_norm_(...)` scales gradients by `max_norm / total_norm`. When `total_norm` is `NaN` or `Inf`, every gradient becomes `NaN`. `optimizer.step()` is then called unconditionally, writing `NaN` into every model parameter. From then on the model is permanently broken — every forward yields NaN.

**Fix (commit 69169ec):** detect non-finite `pre_clip_norm`, then for that batch:
- zero gradients (`optimizer.zero_grad`)
- skip `optimizer.step()` and `ema.update(model)`
- increment `nonfinite_grad_skip_count` and log `train/grad/step_skipped` + `train/grad/nonfinite_skip_total`

Healthy runs are a no-op. Diff is +11/-2 lines around `train.py:~1898`. Recommend cherry-picking to `yi` after review since this protects every experiment in the fleet, not just this PR.

**Operational impact for this PR:** killed the two NaN'd arms, kept `global-scale` running (already past where it would've broken). Restarting `none` and `per-axis` from scratch with the fix.

morganmcg1:
STUDENT fern:

## Update — per-axis arm broken; restart with clamp + control respin

Three things happened during the sweep that I want to be transparent about. Bottom line: the original `per-axis` arm catastrophically diverged (so killed and replaced with a `per-axis-clamped` variant), and the `none` control diverged in epoch 2 (so restarted from scratch). Only `global-scale` is stable across the run.

### 1. `per-axis` (sr0qik6z, then tnvi0omy after restart) — diverged in epoch 1, killed

Both the pre-bugfix (`sr0qik6z`) and post-bugfix (`tnvi0omy`) `per-axis` runs blew up gradients within the first epoch. `tnvi0omy` was caught by my non-finite-grad guard but skipped 9.5k/14.4k optimizer steps (66% skip rate), so val EMA stayed at 93.47% abupt — model never learned.

Pre-clip grad norm growth (peraxis-r2):
| Step window | Mean ‖g‖ | Max ‖g‖ |
|---|---:|---:|
| [0, 1000) | 8.3 | 22.6 |
| [1000, 3000) | 1.5e6 | 2.0e7 |
| [3000, 5000) | 5.2e9 | 8.8e10 |
| [5000+) | Inf | Inf |

**Root cause:** with surface-only stats, `per-axis` scale is `[5.08, 2.26, 1.57]`. Surface points are in `[0, 1]`, but volume points span the wind-tunnel CFD domain (~25× wider in x/y) and get normalized to up to **|y|=9.55, |z|=12.31** in y/z. That extreme range through `pos_embed → MLP bias → slice attention` produces gradients that grow exponentially before clipping has a stable direction to clip toward. (Per-axis on volume points in the original PR's "joint surface+volume bbox" interpretation would compress surface to ~4% of the cube — the opposite of the hypothesis. Surface-only stats was the right frequency-density choice but the volume tail destabilizes training.)

Killed at 05:13 UTC.

### 2. `none` (control) (opysaspv) — diverged in epoch 2, killed and restarted

The control got val=16.20% abupt at epoch 1 (consistent with PR #99 trajectory), but in epoch 2 (~step 14k) the gradient norm grew from ~1.5 to ~8e8 in ~500 steps. The bug-fix kept the model from being NaN-corrupted but skipped most subsequent optimizer steps — train_loss climbed from 0.1 to 12, EMA frozen at epoch-1 values, no further val improvement.

Stochastic instability — PR #99 baseline `3hljb0mg` was stable for 33k steps with the same flags (max ‖g‖ ≤ 35), and `train.py` uses no fixed seed. Killed at 05:28 UTC.

### 3. New runs in flight (group `fern-coord-normalization`)

| Arm | Mode | W&B run | GPU | Started | Notes |
|---|---|---|---|---|---|
| A control restart | `none` | [`lue3k8sv`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/lue3k8sv) | 0 | 05:30 | Fresh seed; same config as PR #99 |
| B (continuing) | `global-scale` | [`ftfu6z18`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/ftfu6z18) | 1 | 03:32 | Healthy, currently epoch 2 ~30%, val=24.85% at epoch 1 |
| C (replacement) | `per-axis-clamped` | [`a2pgliqs`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/a2pgliqs) | 2 | 05:30 | New mode, see below |

**Deviation:** committed a new `per-axis-clamped` mode (commit a11562f) replacing arm C. It does the same per-axis (xmin, extent) shift/scale as `per-axis`, then `clamp(0.0, 1.0)`. Surface points (in [0, 1] by construction) are unchanged; volume points outside the surface bbox saturate at 0 or 1 — losing far-field volume position distinguishability but bounding the input range. Volume tokens still differ via their non-position features (cell volume, etc.) so the slice attention can still group them.

If you'd prefer a different fix (e.g. wider clamp like `[-1, 2]`, or hardcoded bbox `[7, 1.25, 1.95]` per the PR text), happy to swap.

### Status quo, expected timeline

- `none-r3` (lue3k8sv) and `peraxis-clamped` (a2pgliqs): ~6 h budget each, expecting ~4 epochs.
- `global-r2` (ftfu6z18): ~4 h remaining, expecting ~5 epochs total.
- Will report results once all three have at least 3 epochs of validation. ETA ~09:30 UTC.

morganmcg1:
STUDENT fern:

## Update — global-r2 killed, global-r3 launched

Killed `global-r2` (`ftfu6z18`) at 05:42 UTC. Same divergence pattern as the original NaN'd runs:

- Epoch 1 ended OK (train_loss bottomed at ~0.17 around step 10500, val=24.85% abupt at step 10883).
- Mid-epoch 2 (step ~13000), train_loss climbed from 0.17 → 4.72 over ~2k steps as `pre_clip_norm` saturated to `Inf`.
- Bug-fix kept weights non-NaN but skipped 1368 / 15958 = 8.6% of steps overall (66% in last 1000 steps), so EMA val won't recover.

The fact that a "stable" run (`dbhs9qng`, originally claimed healthy) actually got val=92.5% at epoch 1 in the prior attempt, and PR #99's baseline (`3hljb0mg`) ran cleanly without a fixed seed for 4 epochs, points to **stochastic training instability** rather than coord-normalize-specific behavior — the `none` arm went the same way in `opysaspv`/`none-r2`. `train.py` has no `--seed` flag.

### Current state (group `fern-coord-normalization`)

| Arm | Mode | W&B run | GPU | Started | Status (05:50 UTC) |
|---|---|---|---|---|---|
| A | `none` | [`lue3k8sv`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/lue3k8sv) | 0 | 05:30 | epoch 1, ~12%, train_loss=0.98, 0 skips |
| B | `global-scale` | [`global-r3`, fresh seed](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml) | 1 | 05:47 | warming up |
| C | `per-axis-clamped` | [`a2pgliqs`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/a2pgliqs) | 2 | 05:30 | epoch 1, ~9%, train_loss=1.29, 0 skips |

Will collect epoch 1 val for all 3 arms when `none-r3` and `peraxis-clamped` reach the boundary (~06:50 UTC). Final results expected after ~4 epochs each, around 11:30 UTC.

### Note for the post-mortem

If global-r3 also diverges, the conclusion is "global-scale arm is unstable" (3/3 attempts), which is itself meaningful since the omega bank in `ContinuousSincosEmbed` (max wavelength 2π, calibrated for meter-scale inputs) gets effectively shrunk when coords are normalized to ~[-0.5, 0.5] — the high-frequency portion of the encoding becomes near-constant, leaving the model relying on low-freq channels with less expressiveness. That's a plausible mechanism for both worse vals AND latent training fragility.

morganmcg1:
STUDENT fern:

## Update — peraxis-clamped diverged too; switching to `per-axis-surfonly`

Brief: `per-axis-clamped` (a2pgliqs) hit the same gradient-explosion pattern as the unclamped variant, killed at 06:04 UTC. Replacing with a new `per-axis-surfonly` mode that targets the actual hypothesis cleanly.

### What happened with `per-axis-clamped`

| Step | train/loss | pre_clip_norm | skips |
|---:|---:|---:|---:|
| 2099 | 1.57 | 7.6 | 0 |
| 2499 | 1.10 | 147 | 0 |
| 2799 | 1.98 | 4.5e4 | 0 |
| 2899 | 2.13 | 5.4e7 | 0 |
| 2999 | 3.18 | Infinity | 43 |
| 3099 | 3.23 | Infinity | 143 |

The clamp(0,1) bounded the embedding inputs but volume tokens collapsed onto the bbox boundary — slice-attention could only differentiate them via non-position features. The model couldn't learn through the degenerate token structure and exploded gradients ~step 3000. Same dead pattern as before; killed.

### Root cause analysis

Both per-axis variants (unclamped + clamped) failed because the same `coord_shift`/`coord_scale` buffer was applied to surface AND volume tokens. Surface bbox stats are right for the hypothesis (vehicle gets equal y/z frequency density per axis), but break volume tokens. Volume bbox stats compress vehicle to ~4% of cube — the opposite of the hypothesis.

The fix is per-token-group normalization: apply the surface-bbox shift+scale to **surface tokens only**, leave volume tokens with raw meter coords (which has been stable across every prior PR).

### New mode: `per-axis-surfonly` (commit 31a61cc)

```python
def _encode_group(self, x, *, ..., is_surface: bool):
    pos = x[:, :, : self.space_dim]
    if self.coord_normalize != "none":
        apply_norm = True
        if self.coord_normalize == "per-axis-surfonly" and not is_surface:
            apply_norm = False
        if apply_norm:
            pos = (pos - self.coord_shift) / self.coord_scale.clamp(min=1e-6)
            ...
```

`forward()` passes `is_surface=True/False` to the two `_encode_group` calls. Smoke test (forward+backward on tiny inputs) passes. This is the cleanest test of the original anisotropy hypothesis: surface tokens get equal y/z frequency density per axis; volume tokens keep their proven-stable raw-meter encoding.

### Bookkeeping deviation

Recovered the lost `per-axis-clamped` commit (was at `a11562f` in reflog, accidentally reset away during a checkout earlier today; cherry-picked back as `1a4e229`). The `per-axis-surfonly` mode sits on top of it as commit `31a61cc`. So the codebase now has 5 modes total: none, global-scale, per-axis, per-axis-clamped, per-axis-surfonly. Only the last one is freshly running.

### New current state (group `fern-coord-normalization`)

| Arm | Mode | W&B run | GPU | Started | Status (06:08 UTC) |
|---|---|---|---|---|---|
| A | `none` | [`lue3k8sv`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/lue3k8sv) | 0 | 05:30 | epoch 1 ~40%, train_loss=0.24, max ‖g‖=3.7, 0 skips — healthy |
| B | `global-scale` | [`kvqi7mkh`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/kvqi7mkh) | 1 | 05:47 | epoch 1 ~17%, train_loss=1.09, max ‖g‖=10.4, 0 skips — healthy |
| C | `per-axis-surfonly` | (just launched) | 2 | 06:04 | warming up |
| ~~C-original~~ | `per-axis-clamped` | ~~`a2pgliqs`~~ | killed | — | Inf grads, model stuck — KILLED |

Will collect epoch 1 vals when arms A and B reach epoch boundary (~06:55 UTC). Final post-mortem after each arm has ≥3 epochs of validation, ETA ~10:30 UTC.

### Why this should be stable

Volume tokens with raw meter coords were proven stable in PR #99 (33k steps, max ‖g‖ ≤ 35). Surface tokens normalized to [0, 1] should be a stricter input than the raw [-0.77, 3.94] range. So if anything, this should be slightly *more* stable than the `none` baseline, not less.

morganmcg1:
STUDENT fern:

## Update — control-arm divergence is a base-rate problem, not a code regression. Killing none-r3 + peraxis-surfonly.

### What I found

After `none-r3` and `peraxis-surfonly` both diverged with the **same gradient-explosion pattern** (steady → 100× spike → cascade to Inf → `nonfinite_grad_skip` cap), I checked the PR #99 lr-sweep group on W&B:

| PR #99 run | lr | final loss | val abupt | state |
|---|---|---|---|---|
| `fern-lr1e4` (`0xpewxr8`) | 1e-4 | 0.10 | 14.20 | healthy |
| `fern-lr2e4-ctrl` (`ym2tkyki`) | 2e-4 | **3.30** | **62.5** | diverged |
| `fern-lr3e4` (`r4ihjnlc`) | 3e-4 | **NaN** | 0 | diverged |
| `fern-lr5e4` (`3hljb0mg`) | 5e-4 | 0.041 | **10.69** | healthy ← our baseline |

**2 of 4 runs at PR #99's exact commit also diverged.** `3hljb0mg` was the lucky one. The current advisor branch base (`86fee5a`) differs from PR #99's commit (`a41b6522`) only by `+aux_rel_l2_weight=0.0` (no-op default) — so the underlying instability is inherent to this train.py + config + lack of a fixed seed, not something I introduced.

### Implications for this PR

Single-shot per-arm comparison **cannot reliably distinguish** coord-normalize variants from each other on this codebase. With ~50% baseline divergence, any single arm has ~50% chance of being unevaluable, and a 3-arm comparison has ~12.5% chance of all arms training cleanly. I've now consumed 2 GPU-hours per failed restart, and brute-force re-running the unstable arms isn't a good use of compute.

### Plan for what's left

1. **Killed** at 06:53 UTC: `none-r3` (24% skipped, train_loss=3.81), `peraxis-surfonly` (skipped escalating, train_loss=3.55). Both unrecoverable.
2. **Keep running** `global-r3` (`kvqi7mkh`) — currently healthy at step 7502/10883, max ‖g‖=20.4, train_loss=0.79. The only consistently stable mode across attempts.
3. **Reuse PR #99 `3hljb0mg`** as the control reference (val abupt=10.69%) since it's a bona fide stable instance of the `coord-normalize=none` config.
4. **For per-axis modes**: report the divergence pattern as the result (negative/inconclusive), no further restarts.

### What the data already shows for global-scale (3 attempts)

| Run | State | Epoch 1 val abupt |
|---|---|---|
| `dbhs9qng` | NaN-corrupted | 92.49 |
| `ftfu6z18` (`global-r2`) | Stable training, killed mid-ep2 by separate ep2 spike | **24.85** |
| `kvqi7mkh` (`global-r3`) | currently running | TBD |

**Even when stable, `global-scale` produces ~24.85% val abupt at epoch 1**, vs PR #99's ~14% at epoch 1 (consistent with my earlier hypothesis that normalizing to a unit cube shrinks the effective ContinuousSincosEmbed wavelength below the meter-scale dynamics it was calibrated for). So global-scale is unlikely to beat 10.69%.

### Bug-fix follow-up suggestion

The `--clip-grad-norm 1.0 + skip-on-nonfinite` combination doesn't catch divergence early enough — by the time `pre_clip_norm = Inf`, the model has already taken ~800 bad-gradient `optimizer.step()` calls and is corrupted. A separate small bug-fix PR could add `--skip-grad-norm-spike` (skip `optimizer.step()` if `pre_clip_norm > K × running_median`, default K=20). Out of scope for this PR; flagging in case it's worth adding to the fleet's stability toolkit.

### ETA

Will report final results once `global-r3` reaches at least epoch 3 (~09:30 UTC), or sooner if it diverges.

morganmcg1:
STUDENT fern:

## Results — coord-normalize sweep is inconclusive; no arm wins

**Bottom line:** none of the three modes beat baseline. `global-scale` is consistently *worse* than control (24.85% vs 16.20% val abupt at epoch 1, 9 attempts total across all arms). Every `per-axis*` variant diverged during epoch 1 with no usable validation. The control arm `none` reproduces PR #99 epoch 1 (~16.2%) but stochastically diverges in epoch 2, which is the fleet-wide base-rate problem flagged separately.

Keeping PR #99 (`3hljb0mg`, val abupt **10.69%**) as the baseline; this PR should not be merged.

### Headline epoch-1 comparison (only clean validations)

| Arm | Mode | W&B | val_abupt | sp | ws | vp | wsx | wsy | wsz |
|---|---|---|---:|---:|---:|---:|---:|---:|---:|
| Reference | (PR #99 ep1) | [`3hljb0mg`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/3hljb0mg) | 16.47 | 11.81 | 18.42 | 9.62 | 16.08 | 21.40 | 23.43 |
| A | `none` (control) | [`opysaspv`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/opysaspv) | **16.20** | 11.40 | 18.02 | 9.75 | 15.60 | 21.00 | 23.24 |
| B | `global-scale` | [`ftfu6z18`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/ftfu6z18) | **24.85** | 16.67 | 25.34 | 23.89 | 22.07 | 29.89 | 31.72 |
| C | `per-axis*` (any) | × | — | — | — | — | — | — | — |

Control matches PR #99 epoch 1 within ~2% — the codebase reproduces. `global-scale` is **+8.65 pp worse** than control across every single component (sp, vp, all three wallshear axes). No per-axis variant produced a usable validation across 4 attempts (`sr0qik6z`, `tnvi0omy`, `a2pgliqs`, `3mmmpj0q`).

### Final baseline reference (this PR cannot improve on this)

| Metric | val (PR #99) | test (PR #99) | AB-UPT |
|---|---:|---:|---:|
| `abupt_axis_mean_rel_l2_pct` | **10.69** | 11.73 | — |
| `wall_shear_y_rel_l2_pct` | 13.73 | 13.53 | 3.65 |
| `wall_shear_z_rel_l2_pct` | 14.73 | 13.98 | 3.63 |

### Run inventory (9 attempts across the 3 arms)

| W&B | Arm | State | Skip total | val ep1 | Note |
|---|---|---|---:|---:|---|
| [`aef8nv9o`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/aef8nv9o) | none | NaN-corrupted | 0 | 0 (NaN) | Pre-bugfix; clip_grad_norm wrote NaN into weights |
| [`opysaspv`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/opysaspv) | none-r2 | epoch 2 div | 1440 | **16.20** | Cleanest control validation |
| [`lue3k8sv`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/lue3k8sv) | none-r3 | early div | 3334 | — | Diverged before epoch boundary |
| [`dbhs9qng`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/dbhs9qng) | global | NaN-corrupted | 0 | 92.49 (NaN) | Pre-bugfix |
| [`ftfu6z18`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/ftfu6z18) | global-r2 | epoch 2 div | 1495 | **24.85** | Cleanest global-scale validation |
| [`kvqi7mkh`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/kvqi7mkh) | global-r3 | epoch 1 div | 0 | 84.48 (broken) | Killed today 07:28 UTC; train_loss had already diverged before epoch boundary so val is meaningless |
| [`sr0qik6z`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/sr0qik6z) | per-axis | NaN-corrupted | 0 | — | Pre-bugfix |
| [`tnvi0omy`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/tnvi0omy) | per-axis-r2 | grad explosion | 10633 | — | 66% step skip rate, model never learned |
| [`a2pgliqs`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/a2pgliqs) | per-axis-clamped | grad explosion | 782 | — | Volume tokens collapse on bbox boundary |
| [`3mmmpj0q`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/3mmmpj0q) | per-axis-surfonly | grad explosion | 378 | — | Volume tokens get raw coords; should have matched `none` stability but didn't |

### Reproduce commands

Control arm (`opysaspv`):
```bash
cd target/
python train.py \
  --coord-normalize none \
  --wandb-group fern-coord-normalization --wandb-name fern-coord-none-r2 --agent fern
```

Same command with `--coord-normalize global-scale` for arm B and `--coord-normalize per-axis-surfonly` for arm C.

Peak GPU memory across all arms: **~80.5 GiB / 96 GiB** (78% util) per GPU, batch_size=8, 65536 surface + 65536 volume points.

### What happened — by arm

**A. `none` (control).** Reproduces PR #99 epoch 1 cleanly (`opysaspv`: 16.20 vs PR #99: 16.47, gap is within seed variance). In epoch 2, gradient norm spiked from ~2 → ~1e8 over ~500 steps, the new non-finite-skip guard fired 1440 times keeping weights non-NaN, but EMA val froze at the epoch-1 value. `none-r3` (`lue3k8sv`) diverged before even reaching epoch 1 (3334 skips). The control arm has the *same* underlying instability as the experimental arms — confirming this isn't coord-normalize-specific.

**B. `global-scale`.** When stable (`ftfu6z18`), epoch 1 val abupt is **24.85% — substantially worse than control (16.20%)** across every output (sp, vp, all three wall-shear axes). Mechanism: dividing all coords by the global bbox diagonal (~7.7m) maps surface points into ~[-0.05, 0.5] in x and ~[-0.13, 0.13] in y/z. `ContinuousSincosEmbed` computes `omega ⊗ pos` where the lowest `omega` corresponds to wavelength 2π. After this normalization, the entire vehicle fits within ~half a wavelength of the lowest mode, so the high-frequency channels (which were calibrated for meter-scale spatial detail) become near-constant across the input — the model loses positional resolution. The hypothesis assumed equal-density frequency calibration helps; in practice it kills the encoder's expressiveness.

**C. `per-axis` family (4 attempts, 0 clean validations).**

- **C1 `per-axis` (raw):** Surface stats `[5.08, 2.26, 1.57]` map surface to [0, 1] but volume points span the wind-tunnel CFD domain (~25× wider in x/y than vehicle), getting normalized to extreme ranges (|y| up to 9.55, |z| up to 12.31). Sin/cos `pos_embed` in principle handles out-of-cube positions, but in practice the resulting embedding magnitudes drove `pre_clip_norm` from ~8 to Inf in <3000 steps.
- **C2 `per-axis-clamped`:** Clamping volume coords to [0,1] kept `pos_embed` inputs bounded, but volume tokens collapsed onto the bbox surface, so slice-attention couldn't differentiate them by position. Same gradient-explosion pattern, slightly later (step ~3000).
- **C3 `per-axis-surfonly`:** Apply normalization only to surface tokens; pass volume tokens through with raw meter coords (which have been stable in PR #99 for 33k steps). Theoretically the cleanest test of the original hypothesis. **Also diverged** (`3mmmpj0q`, 378 skips, train_loss climbing). At this point the simplest explanation is that the underlying training stochastic instability hits per-axis-surfonly in epoch 1 (where it would have hit `none` in epoch 2), and the additional per-axis frequency calibration on surface tokens does not provide enough early-training stabilization to compensate.

### Three things I think this experiment actually established

1. **The 4× tau_y/z gap is not primarily from sincos anisotropy.** `global-scale` (which restores isotropy) makes everything ~50% worse on tau_y/z (29.89/31.72 vs 21.00/23.24) at the only directly comparable epoch boundary. So the *direction* of the original anisotropy at meter scales appears to be helpful, not harmful. Edward's RFF result (PR #119) where wall_shear_y/z were worst in *isotropic* RFF is consistent with this — both experiments suggest that y/z benefits from *higher* per-axis frequency density than the x-axis, not equal density.

2. **`ContinuousSincosEmbed` is calibrated for meter-scale geometry.** The `omega` bank's lowest wavelength is 2π meters; this gives the encoder useful resolution for vehicles in the range ~0.1m–10m. Normalizing coordinates breaks that calibration and the model can't recover via training in 1 epoch. If we want to change the embedding's frequency density, the cleaner intervention is to change the `omega` bank itself (e.g. `--pos-omega-min/max` flags), keeping coordinates in their physical units.

3. **Stochastic training instability is the dominant blocker for new experiments at the current config.** 7 of 9 attempts in this PR + 2 of 4 in PR #99's lr-sweep diverged. The bug-fix in commit `69169ec` (skip step on non-finite grad) is necessary but not sufficient — by the time `pre_clip_norm = Inf`, the model has already taken many bad-gradient steps. Suggested fleet-wide fix in follow-ups below.

### Suggested follow-ups (not implemented in this PR)

- **Fleet stability work (highest priority):** add a `--seed` flag (or seed all RNGs from a default seed) so experiments can be re-run deterministically, AND add a "spike-skip" guard that skips `optimizer.step()` when `pre_clip_norm > K × running_median(pre_clip_norm)` (default K=20, window=200). Together these would let any future single-shot experiment compare arms reliably. This PR's fundamental problem is that we can't tell signal from noise across arms when 50% of runs diverge.
- **Frequency-bank tuning (replaces this PR's hypothesis):** rather than normalizing coordinates, expose the omega-min/omega-max of `ContinuousSincosEmbed` as flags and sweep the omega range. This keeps coords in physical units (meter-scale, where the model is stable) but tests whether *higher max frequency on y/z* (via per-axis omega banks) helps tau_y/z without breaking sp/vp. The mechanism aligns with what edward and this experiment both observed: y/z wants more frequency density, not less.
- **Coord-normalize as initialization-only:** use coord-normalization on the input but unfreeze the omega bank (`requires_grad=True`) so the model can learn its own per-axis effective frequencies. This would test whether the *training instability* at normalized coords is from the calibration mismatch (in which case learnable omega should fix it) or from a deeper interaction with slice attention (in which case it will diverge anyway).
- **Bug-fix follow-up:** I'll open a separate small PR for the spike-skip guard since it's orthogonal to any single experiment and benefits the whole fleet. The non-finite-skip fix from commit `69169ec` was necessary but only catches the *very last* bad step; the spike-skip would catch the cascade earlier.

### Code retained on this branch

- `--coord-normalize` flag with 5 modes: `none`, `global-scale`, `per-axis`, `per-axis-clamped`, `per-axis-surfonly`. Default `none` is a no-op so PR is mergeable as a code addition without behavioral change. But given no arm wins, I'd recommend **not merging the coord-normalize feature** and instead pursuing the omega-bank approach as a follow-up.
- The non-finite-skip bug-fix (`69169ec`) — please cherry-pick this to `yi` regardless of this PR's outcome; it protects every experiment in the fleet from the silent-NaN-corruption pattern that hit `aef8nv9o`/`dbhs9qng`/`sr0qik6z`.

morganmcg1:
ADVISOR: Closing PR #143 because ADVISOR REVIEW — closing as negative result.

**Hypothesis:** Anisotropic bbox (x:~8m, y:~2.5m, z:~2m) causes ContinuousSincosEmbed to give x-axis ~3-4× more frequency resolution than y/z, contributing to the 4× tau_y/tau_z gap.

**Result:** Hypothesis falsified across 9 attempts.
- `global-scale` (restores isotropy): epoch 1 abupt=24.85% vs control 16.20% — normalizing to unit cube destroys ContinuousSincosEmbed expressiveness because the omega bank is calibrated for meter-scale geometry.
- `per-axis` (4 variants): all diverged via volume-token coord drift through pos_embed → MLP bias → slice attention. Volume domain (~25× wider than vehicle bbox) produces extreme out-of-range values.
- `none` control: reproduces PR #99 epoch 1 (~16.20%), then hits the fleet-wide stochastic instability.

**Key insight:** The 4× tau_y/z gap is NOT primarily from sincos anisotropy. The right next attack is omega-bank tuning directly in physical-meter coords (e.g. denser low-frequency basis on y/z, or anisotropic per-axis omega banks) rather than coordinate normalization. The coord-normalize feature should not be merged.

**Side-discovery — extremely valuable:** Confirmed ~50% fleet-wide divergence at lr=5e-4 (no fixed seed). Note: a non-finite-grad skip guard is already on `yi` (commit 6e8b674 `skip optimizer step on non-finite or large pre-clip grad norm` from haku) — so the bugfix path is already in place. The fern-version on this branch is therefore not needed.

Thanks for the rigorous negative — the omega-bank direction is your next target..

PR #144, opened 2026-05-01
edward: AdamW beta2 sweep (0.95 vs 0.999) at lr 3e-4 and 5e-4

morganmcg1:
## Hypothesis

The entire yi fleet has been running AdamW with default betas (beta1=0.9, beta2=0.999). The fleet-wide training instability pattern (lr=5e-4, gradient explosion in epochs 1–2) suggests the second-moment EMA is not adapting fast enough to the actual gradient distribution of this CFD problem.

Why lower beta2 helps on CFD:
- DrivAerML gradients are **non-stationary and bursty**: high-|tau| boundary-layer surface points generate large gradient spikes that are rare relative to total sample count. With beta2=0.999, the running second-moment estimate requires ~1000 steps to "forget" the last gradient spike. In a 10,883-step epoch, this means the second-moment estimate lags real gradient variance by nearly the full epoch.
- **Lower beta2 (0.95–0.99) adapts faster**: The denominator `sqrt(v_t + eps)` tracks current gradient scale more accurately, giving a more robust per-parameter learning rate.
- **Well-known empirical result** from Karpathy's nanoGPT and the original Chinchilla / GPT-3 papers: beta2=0.95 works better than 0.999 on tasks with noisy/bursty gradients. Also supported by the "depth-wise varying betas" literature.

Additionally, the fleet has confirmed that lr=5e-4 is at the stability ceiling. A targeted sweep at lr=3e-4 with beta2=0.95 would (a) establish the correct stability baseline for all subsequent experiments and (b) test whether lower beta2 allows lr to be safely raised back toward 5e-4.

**Proposed 4-arm sweep** (flags-only, no code changes needed):

| Arm | lr | beta2 | warmup | Run name |
|---|---|---|---|---|
| A | 5e-4 | 0.999 (current default) | 500 steps | `edward-adamw-ctrl` |
| B | 5e-4 | 0.95 | 500 steps | `edward-adamw-b95` |
| C | 3e-4 | 0.95 | 500 steps | `edward-adamw-lr3e4-b95` |
| D | 3e-4 | 0.999 (current default) | 500 steps | `edward-adamw-lr3e4-ctrl` |

This 2×2 design isolates the lr effect (A vs D; B vs C) and the beta2 effect (A vs B; C vs D).

## Instructions

**Code change:** Add `--beta1`, `--beta2`, and `--lr-warmup-steps` flags to `Config` and the argparser:
```python
beta1: float = 0.9      # AdamW default
beta2: float = 0.999    # AdamW default
[...]

PR #154, opened 2026-05-01
mikasa: gradient accumulation eff-bs=32 (accum_steps=4)

morganmcg1:
## Hypothesis

The current PR #99 base uses `batch_size=8` — giving each gradient update 8 car geometries. DrivAerML has 500 training cases, so at 65536 points per case we get extremely sparse coverage per step. Each batch of 8 cases is dominated by whichever geometries happen to land in that batch, creating high gradient variance — especially for rare high-|tau| surface points (wheel arches, mirror housings, spoiler lips) that appear in only a subset of geometries.

**Gradient accumulation with `accum_steps=4`** gives an effective batch size of 32 without increasing VRAM. By accumulating 4 micro-batches before each optimizer step:
1. Each gradient update averages over 4× more diverse car geometries
2. The second-moment estimate in AdamW becomes more accurate (less noisy denominator → better per-parameter lr scaling)
3. Wall-shear on rare high-curvature regions gets a more stable gradient signal over 4 micro-batches rather than being dominated by whichever 8 geometries landed in a single batch

This is complementary to — and orthogonal from — the beta2 sweep (PR #144 edward): beta2 improves second-moment estimation speed; accumulation improves second-moment estimation accuracy.

**LR scaling:** When increasing effective batch size, the standard practice is to scale lr linearly: `lr_eff = lr_base * accum_steps`. However, at the lr=5e-4 stability ceiling, scaling up may cause instability. We test two approaches:
- **Linear scale**: lr = 5e-4 × 4 = 2e-3 (aggressive, risky but theoretically correct)
- **Square-root scale**: lr = 5e-4 × √4 = 1e-3 (more conservative)
- **No scale (control)**: lr = 5e-4 with accum=4 (conservative — same lr, more stable gradient)

## Instructions

Add `--gradient-accumulation-steps` (int, default `1`) to `Config` and the argparser.

**Implementation in the training loop:**

```python
accum_steps = config.gradient_accumulation_steps
optimizer.zero_grad()

for step_idx, batch in enumerate(train_loader):
    # Forward + backward on micro-batch (scale loss by 1/accum_steps for correct gradient magnitude)
    with torch.autocast(device_type='cuda', dtype=torch.bfloat16):
        loss = compute_loss(model, batch, config) / accum_steps
    
    scaler.scale(loss).backward()
    
    # Only step optimizer every accum_steps micro-batches
    if (step_idx + 1) % accum_steps == 0:
        scaler.unscale_(optimizer)
        torch.nn.utils.clip_grad_norm_(model.parameters(), config.clip_grad_norm)
        scaler.step(optimizer)
        scaler.update()
        optimizer.zero_grad()
        # EMA update (once per optimizer step, not per micro-batch)
        ema.update()
        # Logging: log at optimizer step granularity, not micro-batch
        global_step += 1
```

**Key correctness notes:**
- Divide loss by `accum_steps` before backward so gradients accumulate to the correct magnitude (equivalent to averaging over all micro-batches)
- Clip gradients only at the optimizer step boundary, not per micro-batch
- EMA update should happen once per optimizer step (not per micro-batch)
- W&B `global_step` should increment once per optimizer step (so x-axis is comparable to the baseline)
- The LR scheduler (if any) should step at optimizer step granularity

**3-arm sweep:**

| Arm | accum_steps | lr | eff_bs | Run name |
|---|---:|---|---:|---|
| A | 4 | 5e-4 (no scale) | 32 | `mikasa-accum4-lr5e4` |
| B | 4 | 1e-3 (sqrt-scale) | 32 | `mikasa-accum4-lr1e3` |
| C | 2 | 5e-4 (no scale) | 16 | `mikasa-accum2-lr5e4` |

Arm C tests whether even half the accumulation (eff_bs=16) gives a measurable benefit vs the baseline (eff_bs=8). This cleanly separates the accumulation effect from the lr-scale effect.

```bash
cd target/
python train.py \
  --gradient-accumulation-steps <1|2|4> \
  --lr <5e-4|1e-3> \
  --volume-loss-weight 2.0 \
  --batch-size 8 \
  --validation-every 1 \
  --weight-decay 5e-4 \
  --ema-decay 0.9995 \
  --clip-grad-norm 1.0 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
  --wandb-group mikasa-gradient-accumulation-r10 \
  --wandb-run-name <arm-name>
```

**Important:** With accum_steps=4, each optimizer step processes 4× more data. The epoch length in optimizer steps is `total_steps / accum_steps`, so ~2718 optimizer steps/epoch instead of 10883. This means the model sees the same number of total data samples per epoch but makes fewer weight updates. **This is expected and correct** — it mirrors what a true bs=32 run would do. Log `train/optimizer_steps` alongside `train/micro_steps` so both are visible in W&B.

**What to watch:** If Arm B (lr=1e-3) diverges at step ~2000 (the known instability in the high-lr regime), it confirms that even accumulation doesn't unlock lr>5e-4. Arm A at lr=5e-4 with accum=4 is the most likely winner — lower gradient variance, same optimizer temperature.

|---|---:|---:|---:|
| `wall_shear_y_rel_l2_pct` | **13.73** | **13.53** | **3.65** |
| `wall_shear_z_rel_l2_pct` | **14.73** | **13.98** | **3.63** |

Reproduce baseline:
```bash
cd target/
python train.py \
```


morganmcg1:
ADVISOR: Closing PR #154 because ADVISOR: Closing as orphaned. No senpai-yi-mikasa pod exists on this track and 'mikasa' is not in the active student roster, so this PR will never be picked up by an agent. Gradient accumulation can be reassigned to an active student in a future round if needed for memory headroom..

PR #156, opened 2026-05-01
levi: OHEM case-level hard example mining (top-25%)

morganmcg1:
## Hypothesis

The DrivAerML training set has 500 car geometries. After several rounds of training, the current model has clearly learned "easy" cases well (standard fastback/notchback profiles) but struggles with geometrically extreme cases (complex wheel-arch treatments, aggressive rear diffusers, mirror mountings). Uniform random sampling gives hard cases the same training frequency as easy ones, so the model never focuses enough gradient signal on hard cases.

**Online Hard Example Mining (OHEM)** at the case level: at the end of each epoch, compute the per-case validation loss for all training cases (using the current EMA checkpoint), then upsample the hardest 25% of cases in the next epoch's DataLoader — giving them 2× the sampling probability vs easy cases.

This targets two specific failure modes:
1. **Case-level hard examples**: certain car configurations (e.g., high-downforce configurations with complex rear wings) may have systematically higher wall-shear prediction error. Upsampling these cases teaches the model to handle them.
2. **The val→test gap**: the test split likely contains cases with similar geometry distribution to the hardest training cases. Better coverage of hard training cases should improve test generalization.

Reference: "Training Region-based Object Detectors with Online Hard Example Mining" (Shrivastava et al., CVPR 2016) — the core insight (focus gradient on the hardest examples) applies to regression tasks as well as detection.

## Instructions

Add `--ohem-hard-fraction` (float, default `0.0` — disabled) and `--ohem-hard-multiplier` (float, default `2.0`) to `Config` and argparser.

**Step 1 — Track per-case val loss after each epoch:**

At the end of each epoch's validation pass, compute the per-case abupt metric (not just the mean). Store this as `per_case_val_loss: dict[case_id, float]` in the trainer state.

```python
# In validation loop, accumulate per-case metrics:
per_case_loss = {}
for batch in val_loader:
    case_ids = batch['case_id']  # check how DrivAerML DataLoader exposes case IDs
    surf_pred, vol_pred = ema_model(batch)
    for i, cid in enumerate(case_ids):
        case_loss = compute_abupt_single(surf_pred[i], batch['surf_targets'][i],
                                          vol_pred[i], batch['vol_targets'][i])
        per_case_loss[cid] = case_loss
```

**Step 2 — Build weighted sampler for next epoch:**

```python
if config.ohem_hard_fraction > 0 and per_case_loss:
    losses = torch.tensor([per_case_loss.get(cid, 1.0) for cid in train_case_ids])
    # Find the hard threshold (top-25%)
    threshold = torch.quantile(losses, 1 - config.ohem_hard_fraction)
    # Assign weights: hard cases get ohem_hard_multiplier, easy cases get 1.0
    weights = torch.where(losses >= threshold,
                          torch.tensor(config.ohem_hard_multiplier),
                          torch.ones_like(losses))
    sampler = WeightedRandomSampler(weights, num_samples=len(train_case_ids), replacement=True)
    train_loader = DataLoader(train_dataset, batch_size=config.batch_size, sampler=sampler, ...)
```

**Step 3 — Log mining stats each epoch:**

Log `train/ohem_hard_case_count` (how many cases are in the hard set), `train/ohem_hard_loss_mean` (mean loss of hard cases) vs `train/ohem_easy_loss_mean` so we can verify the hard/easy gap is being addressed.

**2-arm sweep:**

| Arm | hard_fraction | hard_multiplier | Run name |
|---|---:|---:|---|
| A | 0.25 (top 25% hard) | 2.0 | `levi-ohem-f25-m2` |
| B | 0.25 (top 25% hard) | 3.0 | `levi-ohem-f25-m3` |

Arm A is the conservative baseline (2× upsampling). Arm B tests whether stronger focus on hard cases (3×) helps or hurts by making the gradient too biased toward edge cases.

Note: OHEM doesn't activate until epoch 2 (requires at least 1 epoch of val metrics). Epoch 1 uses uniform sampling — this is correct behavior.

```bash
cd target/
python train.py \
  --ohem-hard-fraction 0.25 \
  --ohem-hard-multiplier <2.0|3.0> \
  --volume-loss-weight 2.0 \
  --batch-size 8 \
  --validation-every 1 \
  --ema-decay 0.9995 \
  --clip-grad-norm 1.0 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
  --wandb-group levi-ohem-hard-case-mining-r10 \
  --wandb-run-name <arm-name>
```

**What to watch:** If `ohem_hard_loss_mean` decreases faster than `ohem_easy_loss_mean` across epochs, OHEM is working. If both decrease at the same rate, the sampler isn't differentiating. Report the per-arm hard/easy loss gap at each epoch in the results.

**Implementation note:** Check how DrivAerML's DataLoader identifies cases (by filename, index, or geometry ID). If per-case IDs are not exposed in the current batch dict, add a `case_id` field to the collate function from the dataset metadata.

|---|---:|---:|---:|
| `wall_shear_y_rel_l2_pct` | **13.73** | **13.53** | **3.65** |
| `wall_shear_z_rel_l2_pct` | **14.73** | **13.98** | **3.63** |

Reproduce baseline:
```bash
cd target/
python train.py \
```


morganmcg1:
ADVISOR: Closing PR #156 because ADVISOR: Closing as orphaned. No senpai-yi-levi pod exists on this track and 'levi' is not in the active student roster, so this PR will never be picked up by an agent. OHEM hard-case mining remains an interesting direction and can be reassigned to an active student in a future round..

PR #169, opened 2026-05-01
thorfinn: NaN-skip + seed + LR warmup utility flags (stability infra)

morganmcg1:
## Hypothesis

Your PR #131 (log-magnitude wall-shear targets) was closed NEGATIVE on the transform itself, but contained a structurally valuable code contribution: the **NaN/Inf-skip safeguard** (commit `2a8f7e4`). Round-5 saw at least 5 experiments crash due to gradient explosions (alphonse 384d arms, tanjiro curriculum, senku surface weight sweep). Many of these divergences might have been recoverable with this safeguard.

This is a code infrastructure PR. The goal is to cherry-pick three stability utilities into `yi`'s `train.py` as standalone features that all future experiments can use:

1. **NaN/Inf-skip safeguard** — prevents irreversible weight corruption from gradient explosions
2. **`--seed` flag** — enables reproducible experiments across students
3. **`--lr-warmup-steps` flag** — enables soft LR warm-in for experiments with high initial gradient variance

**Hypothesis:** With these three utilities in the base config, future experiments will have significantly lower divergence rates. The validation run (2 epochs at PR #99 config + seed=42 + no warmup) confirms zero regression from the additions.

## Instructions

**1. Cherry-pick NaN/Inf-skip safeguard from commit `2a8f7e4`:**

In the training step, after computing loss:
```python
if not torch.isfinite(loss):
    nonfinite_skip_count += 1
    wandb.log({"train/nonfinite_skip_count": nonfinite_skip_count, 
               "train/nonfinite_skip_kind": "loss"}, step=global_step)
    optimizer.zero_grad()
    if nonfinite_skip_count > 200:
        raise RuntimeError(f"Aborted: {nonfinite_skip_count} non-finite skips")
    continue  # skip backward, step, ema.update
loss.backward()
```

After `torch.nn.utils.clip_grad_norm_()`:
```python
pre_clip_norm = torch.nn.utils.clip_grad_norm_(model.parameters(), args.clip_grad_norm)
if not torch.isfinite(pre_clip_norm):
    nonfinite_skip_count += 1
    wandb.log({"train/nonfinite_skip_count": nonfinite_skip_count,
               "train/nonfinite_skip_kind": "grad"}, step=global_step)
    optimizer.zero_grad()
    if nonfinite_skip_count > 200:
        raise RuntimeError(f"Aborted: {nonfinite_skip_count} non-finite skips")
    continue  # skip step, ema.update
optimizer.step()
ema.update()
```

[...]

PR #171, opened 2026-05-01
norman: snapshot ensemble with cyclic LR (free post-train gain)

morganmcg1:
## Hypothesis

We are fundamentally epoch-limited: the 6h timeout allows only 3-4 training epochs. But checkpoint ensembling — averaging predictions from multiple saved checkpoints — provides a free post-training gain that requires no additional GPU time beyond what we already spend.

**Hypothesis:** Snapshot ensembling with cyclic LR (Huang et al. 2017, [arxiv 1704.00109](https://arxiv.org/abs/1704.00109)) saves checkpoints at the trough of each cosine LR cycle. Averaging predictions from 3 such troughs produces a "free ensemble" that reduces prediction variance vs any single checkpoint. Expected gain: 0.5–2% improvement in abupt from denoising prediction variance across diverse loss minima explored by the cyclic LR trajectory.

The cyclic LR also serves a secondary purpose: it actively explores different loss minima per cycle rather than converging to a single minimum, which tends to produce more diverse and complementary checkpoints.

## Instructions

Start from the PR #99 winning base config. Make the following changes:

**1. Cyclic LR scheduler:**
Replace the default constant LR with `CosineAnnealingWarmRestarts`:
```python
scheduler = CosineAnnealingWarmRestarts(
    optimizer,
    T_0=steps_per_epoch,   # restart every epoch
    T_mult=1,              # same period each cycle
    eta_min=1e-5,          # trough LR
    eta_max=5e-4           # peak LR (set via last_epoch trick or manual scaling)
)
```
Note: `CosineAnnealingWarmRestarts` starts from `eta_max` (set optimizer initial lr to 5e-4) and anneals to `eta_min=1e-5` over T_0 steps. Step the scheduler every training step (not every epoch). Add `--use-cyclic-lr` flag.

**2. Snapshot checkpoint saving:**
At the end of each LR cycle (i.e., when `global_step % steps_per_epoch == steps_per_epoch - 1`), save a snapshot checkpoint:
```python
torch.save(ema.averaged_model.state_dict(), f"checkpoint_snapshot_{cycle_idx}.pt")
```
Use EMA weights for the snapshot if EMA is enabled (preferred). Save at most 4 snapshots (cycles 1–4).

**3. Ensemble inference (at end of training):**
After training completes:
```python
# Load all snapshot checkpoints
snapshots = [load_checkpoint(f"checkpoint_snapshot_{i}.pt") for i in range(num_snapshots)]
# Run inference on validation set for each snapshot
# Average predictions: pred_ensemble = mean([pred_i for pred_i in preds])
# Compute val metrics on averaged predictions
```
Report: (a) best individual snapshot val abupt, and (b) ensemble val abupt.

**4. If ensemble val abupt < 10.69:** Run ensemble inference on the test set and report `test_primary/*` metrics.

**5. W&B logging:** Log `val/ensemble_abupt` alongside individual snapshot val metrics. Use `--wandb_group norman-snapshot-ensemble`.

**6. Keep unchanged:** `--batch-size 8`, `--model-layers 6 --model-hidden-dim 256 --model-heads 4 --model-slices 128`, `--ema-decay 0.9995`, `--clip-grad-norm 1.0`, `--wallshear-y-weight 2.0 --wallshear-z-weight 2.0`, `--volume-loss-weight 2.0`, all point counts (65536).

**Reproduce command:**
```bash
cd target/
python train.py \
  --use-cyclic-lr \
  --volume-loss-weight 2.0 \
  --batch-size 8 \
  --validation-every 1 \
  --ema-decay 0.9995 \
  --clip-grad-norm 1.0 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
  --wandb_group norman-snapshot-ensemble
```

## Baseline

Current best: **PR #99 fern**, W&B run `3hljb0mg`

| Metric | Current Best |
|---|---:|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **10.69** |
| `val_primary/surface_pressure_rel_l2_pct` | 6.97 |
| `val_primary/wall_shear_rel_l2_pct` | 11.69 |
| `val_primary/volume_pressure_rel_l2_pct` | 7.85 |
| `val_primary/wall_shear_y_rel_l2_pct` | 13.73 |
| `val_primary/wall_shear_z_rel_l2_pct` | 14.73 |

## Report back

1. Val_primary/abupt for each individual snapshot (cycles 1, 2, 3, and optionally 4)
2. **Ensemble val abupt** (average of all snapshot predictions)
3. Delta: ensemble vs best individual snapshot
4. If ensemble beats 10.69: test_primary/* from ensemble on test set
5. W&B run ID


morganmcg1:
ADVISOR: Status check — your pod is confirmed Ready but I haven't seen activity on this PR. Please report: (1) implementation status (code complete? tests pass?), (2) W&B run ID(s) once launched (paste links), (3) any blockers — env issues, OOM, NaN, ambiguity in the spec? If the launch command from the PR body is unclear, ask. Don't sit idle.

morganmcg1:
STUDENT norman:

Implementation status: code complete. The cyclic-LR snapshot-ensemble pipeline is wired into `train.py`:

1. `--use-cyclic-lr` switches the scheduler to `CosineAnnealingWarmRestarts(T_0=steps_per_epoch, T_mult=1, eta_min=1e-5)` and steps it every optimizer step. Per-epoch `scheduler.step()` is suppressed in that mode.
2. Snapshots are saved at the end of each validation epoch (= cycle trough) into `output_dir/checkpoint_snapshot_<i>.pt`. Save happens between `ema.copy_to(model)` and `ema.restore(model)` so the snapshot contains the EMA-averaged weights. Cap is `--cyclic-lr-max-snapshots 4`.
3. After regular test eval, if `>=2` snapshots were saved we build N independent uncompiled model copies, do per-snapshot val (logged to `val_snapshot_<i>/*`), then ensemble val (logged to `val_ensemble/*`). Predictions are averaged in normalized space (equivalent to physical space because the inverse transform is affine).
4. If `val_ensemble/abupt < 10.69` we also run ensemble inference on the test set (`test_ensemble/*`).
5. New flags: `--use-cyclic-lr`, `--cyclic-lr-eta-min`, `--cyclic-lr-max-snapshots`, `--ensemble-val-threshold`.

Note on the launch command: the PR body uses underscored flags (`--wandb_group`, `--wandb_name`); train.py argparse names are hyphenated (`--wandb-group`, `--wandb-name`). I will use the hyphenated form when launching.

Launching now on GPU 1, will post W&B run id when training starts.

morganmcg1:
STUDENT norman:

## Run launched

- **W&B run:** `ogz5owx1` ([link](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/ogz5owx1))
- **GPU:** 1
- **Started:** 06:59 UTC, 6h timeout (ends ~12:59)
- **Train epoch length:** 10883 batches @ ~2.14 it/s ≈ 85min/epoch (incl. val)
- **Status:** epoch 2/50 currently 73% complete. Epoch 1 done (val abupt=17.14, snapshot 0 saved).

Projected timeline within 6h budget:
- Epoch 1: snapshot 0 saved (val abupt 17.14) — done at t=85.7min
- Epoch 2: snapshot 1 — finishes ~t=171min  
- Epoch 3: snapshot 2 — finishes ~t=257min
- Epoch 4: snapshot 3 — finishes ~t=343min (close to 360min cap)

So expect 3-4 snapshots feeding the ensemble. Will report individual + ensemble val abupt + ensemble test (if val < 10.69) once training completes.

morganmcg1:
STUDENT norman:

## Cleanup: duplicate run killed

I noticed two concurrent training processes were running on this PR — both with identical args, but only one (`ogz5owx1` / PID 231645) is mentioned in my prior comment. The other (`rgdwur9t` / PID 240940 on GPU 1) was an accidental double-launch from ~09:05 UTC, ~2h after the original. Since both were running identical code (train.py was last modified 06:42 UTC, before either launch), the duplicate was wasting GPU 1 and would have produced the same result.

**Action taken:** killed PID 240940. GPU 1 is now free. The original run `ogz5owx1` continues uninterrupted on GPU 0.

Continuing to monitor `ogz5owx1` until the 6h timeout (~12:59 UTC). Will post final results once training completes.

morganmcg1:
ADVISOR: Implementation looks correct and the launch trajectory is clean. Epoch 1 val=17.14 is roughly where a fresh single-snapshot baseline lands, so the cyclic LR isn't sabotaging early-epoch convergence. Good cleanup on the duplicate process — that would have wasted GPU 1 for 6h.

**Expectations for this run:**
- 3-4 snapshots feeding the ensemble (your projection is right).
- Ensemble val typically gains 0.3-1.0pp over best single snapshot in CFD surrogate work, depending on diversity. With only 3-4 snapshots from a single run (vs full independent retrains), expect closer to 0.3-0.5pp.
- Best single snapshot will likely be the last one (snapshot 3) — most converged.

**For the final report:** please post (1) val_primary at each snapshot individually, (2) ensemble val_primary (mean of softmax-pred-style averaging or whatever you implemented), (3) ensemble test_primary, (4) the LR trajectory plot (the cycle troughs should show clearly), (5) per-component breakdown — wall_shear_y/z is where ensembling typically helps most because of the high variance there.

**Merge bar:** if ensemble val < 10.69, merge. If best single-snapshot val < 10.69 but ensemble is worse, that's still a merge candidate (we'd take the single snapshot). Standing by for results.

morganmcg1:
STUDENT norman:

## V1 run `ogz5owx1` — diverged at epoch 3, killed

**Diagnosis from W&B:** at step 22901 (deep into epoch 3, ~10% in, LR≈4.87e-4 near peak), `train/grad/pre_clip_norm` jumped from ~15 → **761** in one step → NaN by step 23101. Snapshots saved before NaN was detected:

| Snapshot | Epoch | val_abupt | finite weights |
|---|---:|---:|---|
| 0 | 1 | 17.14 | yes |
| 1 | 2 | 12.70 | yes |
| 2 | 3 | NaN | **no** (128/129 layers NaN) |

Best individual val (12.70) is far above baseline 10.69. Even a clean 4-snapshot ensemble would be unlikely to make up the ~2pp gap from this trajectory. Continuing the run to natural timeout would just produce more NaN snapshots.

**Root cause hypothesis:** the warm restart from `eta_min=1e-5` → `eta_max=5e-4` is a 50× LR jump. By cycle 3 the model has been trained for two epochs and sits in a sharper basin; the jolt back to peak LR pushes it into a high-curvature region and gradients explode within a few hundred steps.

## V2 launched: `okm6uoea` (GPU 0)

[W&B](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/okm6uoea)

Two minimal-deviation stability fixes:
1. `--cyclic-lr-eta-min 5e-5` (raised from 1e-5) — cycle ratio drops from 50:1 to 10:1, much smaller restart shock, still allows snapshot diversity.
2. `--clip-grad-norm 0.5` (tightened from 1.0) — defensive guard against future spikes; PR #99 baseline gradient norms are typically 5-20 so 0.5 still scales sensibly.

Also patched `train.py`:
- `evaluate_split_ensemble` snapshot save path now skips writing if model weights are non-finite (would have prevented the corrupted snapshot 2).
- Ensemble loader filters out non-finite snapshots and aborts ensemble eval if <2 finite snapshots remain.

All other config matches the PR spec (peak LR 5e-4, T_0=steps_per_epoch, 4 max snapshots, EMA snapshot, etc.). Will report when training completes (~17:30 UTC).

morganmcg1:
ADVISOR: V1 diagnosis is correct — the eta_min=1e-5 → eta_max=5e-4 50× LR jolt at cycle restart, on a model that has settled into a sharper basin after two epochs, is a textbook spike-into-divergence pattern. Killing was the right call.

V2 fixes are well-chosen:
1. **eta_min=5e-5 → 10× cycle ratio.** Preserves snapshot diversity (cosine-restart still has meaningful amplitude) while removing the destabilizing jolt. This is the right operating point.
2. **clip=0.5.** Halves max Adam step size on spike events without hurting baseline-regime steps (5-20 norm × 0.5 clip = 2.5-10 actual norm, still gets through cleanly).
3. **Snapshot guard (skip non-finite) + finite-only ensemble loader.** Solid robustness — exactly the right defensive engineering.

V1 ep2 val=12.70 is below baseline ep2 trajectory but above the 10.69 floor. If V2 reaches ep4 cleanly with snapshot ensemble, the ensemble averaging might still close ~2pp gap (snapshot ensembles in cyclic-LR commonly give 0.5-1.5pp absolute improvement on noisy validation surfaces).

**Reporting at completion (~17:30 UTC):**
| Snapshot | epoch | val_abupt | val_ws_y | val_ws_z |
|----------|-------|-----------|----------|----------|

Plus the ensemble val_abupt (averaging predictions across all finite snapshots) and the test_abupt from best individual snapshot.

**Merge bar:** ensemble val_abupt < 10.69 — even by 0.1, that's a free gain since this is essentially compute-equivalent to a single training run. Worth merging.

**Fleet observation:** the pre_clip jump 15 → 761 in one step is the same large-but-finite spike pattern we see fleet-wide. PR #169's NaN-skip would not have caught this (761 is finite). Magnitude-based skip (pre_clip > N × running_median) would close this gap — we have it queued as infrastructure work.

morganmcg1:
STUDENT norman:

## V2 run `okm6uoea` — killed at step 12712 in epoch 2

**Diagnosis from W&B:** the V2 fixes prevented NaN explosion but did not prevent model degradation.

| step | event | grad pre-clip | train loss | notes |
|---|---|---:|---:|---|
| 10283 | epoch 1 end | ~1.5 | 0.10 | snapshot 0 saved, val=18.18 |
| 10883 | epoch 2 start, cycle restart | — | — | LR jump 5e-5 → 5e-4 |
| 10900 | +17 steps | 8.5 | 0.29 | first jolt |
| 11100 | +217 steps | 23 | 2.58 | loss spiked 25× |
| 12000 | epoch 2 ~10% | 99 | 2.48 | recovery attempted |
| 12300 | | 3910 | 3.64 | major spike |
| 12600 | | 4.1e5 | 2.72 | massive spike, Adam saturated |
| 12700 | | inf | 2.79 | clip preserved finite weights but model in degraded basin |
| 12712+ | sustained | — | 2.7-2.9 | loss stuck at 30× baseline level |

The clip=0.5 prevented weights from going NaN (snapshot 0 finite, model alive), but the model was driven into a degenerate basin where every batch produces enormous gradients that get clipped down so much that effective updates are tiny — model is stuck. No realistic recovery path within remaining 4h.

## Diagnosis: eta_min=5e-5 was less stable than expected

**Counter-intuitive finding:** raising eta_min from V1's 1e-5 to V2's 5e-5 made cycle 2 *worse*, not better. V1 cycle 2 reached val=12.70 cleanly; V2 cycle 2 broke. Hypothesis: V1's deep trough (LR=1e-5 at cycle end) lets the model settle into a sharper local minimum, and the warm restart cleanly pops it out. V2's shallower trough (LR=5e-5) leaves the model in a less-converged state, and the restart drives it into a degenerate region where gradients explode and Adam's adaptive moments get saturated.

V1's actual failure mode (cycle-3 NaN with clip=1.0) was different: a single grad spike 15→761 finite-but-huge, which clip=1.0 didn't catch tightly enough.

## V3 launching: combine V1's eta_min with much tighter clip

V3 strategy: revert eta_min to V1's 1e-5 (which gave cycle 2 val=12.70) and tighten clip to 0.25 (vs V1's 1.0, V2's 0.5) to prevent V1's cycle-3 NaN. This is the minimal-deviation fix targeting both observed failure modes.

Config:
- `--cyclic-lr-eta-min 1e-5` (default; V1 setting)
- `--clip-grad-norm 0.25` (4× tighter than V1)
- All other params identical to V1/V2

Launching on GPU 0 now. ETA ~6h training + ~30min ensemble eval.

morganmcg1:
ADVISOR: V2 diagnosis is clear — the cosine-restart LR jolt at cycle boundary is incompatible with a model that has already settled. Even with eta_min=5e-5 (10× cycle ratio), the 5e-4 peak restart on a settled epoch-2 model produces the same instability pattern (loss 25× spike → recovery attempt → catastrophic Adam v-saturation by step 12712).

**The core problem:** Snapshot ensembling requires the LR to oscillate, but the current loss landscape is too flat-and-sharp near the basin to survive a full-amplitude restart after the first epoch. The model has learned something useful and the restart disturbs it more than it explores.

**Next hypothesis to test: SWA without restarts (post-hoc averaging).**
- Drop the cosine-restart schedule entirely
- Train with the fern PR #183 recipe (lr=5e-4, no restart, warmup=500)
- At the end of training (e.g. epoch 3, 4, 5), save checkpoints at regular intervals
- Compute a simple uniform average of the weights (SWA) as the final model
- Compare SWA-averaged checkpoint vs single-best-val checkpoint

This tests whether SWA-style averaging helps **without** any LR disruption. The ensemble diversity comes from the natural SGD noise trajectory, not forced restarts. This is the standard SWA recipe (Izmailov et al. 2018) and is zero-risk for training stability.

**Concrete changes:**

(Check if --swa-start-frac or equivalent already exists; senku's PR #198 is testing SWA, check with them to avoid duplication.)

Wait — PR #198 (senku) is already running SWA. Please coordinate: if senku has already converged on an SWA-start-frac answer, there's no need to duplicate. 

**Updated recommendation:** Close this PR. The snapshot-ensemble-via-restarts approach is exhausted (V1 and V2 both failed). SWA post-hoc is already in PR #198. Redirect your GPU to a fresh hypothesis — stand by for a new assignment from me.

morganmcg1:
ADVISOR: Closing PR #171 because ADVISOR (closing): Snapshot ensemble via cosine restarts has been exhausted as a direction — both V1 and V2 attempts failed to surface free gain from cyclic LR snapshots, and SWA (PR #198) is the more principled approach to averaging-based free gain. Closing this experiment slot to free norman for a Round 14 assignment. Thanks for the careful exploration; the negative result is documented in CURRENT_RESEARCH_STATE..

PR #130, opened 2026-04-30
tanjiro: curriculum tau_y/z weighting schedule

morganmcg1:
# tanjiro — Curriculum tau_y/z weighting (Round-5 from CURRENT_RESEARCH_STATE)

## Hypothesis

PR #66 thorfinn established W_y=2, W_z=2 as a static gain. PR #116 (fern, in flight) is sweeping higher static weights. Both are static. A complementary Round-5 idea: **curriculum the tau_y/z weighting** — start with W_y=W_z=1.0 (let the model learn the dominant tau_x axis cleanly first) and ramp to 3.0 by mid-training. This avoids early-training gradient interference where a still-untrained model is asked to upweight axes it can barely predict.

This will need a small code change to `train.py`. Add a linear-schedule mechanism for the wallshear axis weights, controlled by two new flags:

- `--wallshear-y-weight-start` / `--wallshear-y-weight-end` (default to current `--wallshear-y-weight`)
- `--wallshear-z-weight-start` / `--wallshear-z-weight-end`
- `--wallshear-weight-warmup-frac` (fraction of total training to ramp linearly; default 0.5)

Ramp linearly from start to end over the first `warmup_frac` of total training steps; hold at end thereafter.

## Baseline to beat (PR #99, fern, W&B run `3hljb0mg`)

| Metric | yi best | AB-UPT |
|---|---:|---:|
| `abupt_axis_mean_rel_l2_pct` | **10.69** | — |
| `wall_shear_y_rel_l2_pct` | 13.73 | 3.65 |
| `wall_shear_z_rel_l2_pct` | 14.73 | 3.63 |

## Experiment plan — 3-arm curriculum sweep

Use `--wandb-group tanjiro-curriculum-tau-yz-r5`:

| Arm | start (Wy=Wz) | end (Wy=Wz) | warmup_frac |
|---|---:|---:|---:|
| A | 1.0 | 3.0 | 0.5 |
| B | 1.0 | 4.0 | 0.5 |
| C | 1.0 | 3.0 | 0.3 |

```bash
cd target/
python train.py \
  --volume-loss-weight 2.0 --batch-size 8 --validation-every 1 \
  --lr 5e-4 --weight-decay 5e-4 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 6 --model-hidden-dim 256 --model-heads 4 --model-slices 128 \
[...]

PR #129, opened 2026-04-30
senku: surface loss weight sweep on PR #99 base

morganmcg1:
# senku — Surface loss weight upweight sweep on PR #99 base

## Hypothesis

Current PR #99 base has `--volume-loss-weight 2.0` (vs default surface weight 1.0). Since merging gilbert PR #9 (vol_w=2.0), surface metrics have steadily improved alongside volume metrics — but the surface loss weight has never been raised above 1.0. With volume_pressure now nearly closed (1.3× AB-UPT) and surface_pressure still at 1.8× AB-UPT, plus all wall-shear axes still 1.6–4× away, there is a strong argument that the gradient is now under-allocated to surface terms. The `--surface-loss-weight` flag in `train.py` lets us rebalance directly without changing volume weight.

## Baseline to beat (PR #99, fern, W&B run `3hljb0mg`)

| Metric | yi best | AB-UPT | Ratio |
|---|---:|---:|---:|
| `abupt_axis_mean_rel_l2_pct` | **10.69** | — | — |
| `surface_pressure_rel_l2_pct` | 6.97 | 3.82 | 1.8× |
| `wall_shear_rel_l2_pct` | 11.69 | 7.29 | 1.6× |
| `volume_pressure_rel_l2_pct` | 7.85 | 6.08 | 1.3× |

## Experiment plan — 4-arm surface loss weight sweep

Use `--wandb-group senku-surface-loss-weight-r5`:

| Arm | `--surface-loss-weight` | `--volume-loss-weight` |
|---|---:|---:|
| A | 1.5 | 2.0 |
| B | 2.0 | 2.0 |
| C | 2.0 | 1.5 |
| D | 3.0 | 2.0 |

(Arm C tests whether shifting the ratio toward surface — same total weight — beats simply scaling up surface only. Useful for distinguishing balance vs total-magnitude effects.)

```bash
cd target/
python train.py \
  --surface-loss-weight <SW> --volume-loss-weight <VW> \
  --batch-size 8 --validation-every 1 \
  --lr 5e-4 --weight-decay 5e-4 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 6 --model-hidden-dim 256 --model-heads 4 --model-slices 128 \
  --ema-decay 0.9995 --clip-grad-norm 1.0 \
  --wallshear-y-weight 2.0 --wallshear-z-weight 2.0 \
[...]

PR #118, opened 2026-04-30
chihiro: MLP ratio sweep 6/8 (wider feedforward layers)

morganmcg1:
## Hypothesis

The current baseline uses `mlp_ratio=4` (default). Bengio-stream experiments showed mlp_ratio=6 achieving ~8.9 abupt in their setting. On the yi stream with a stronger baseline (10.69), we want to test whether wider MLP layers improve capacity for the CFD surrogate task. A higher mlp_ratio means each transformer block has a larger feedforward network, which can capture more complex relationships between surface/volume features.

Testing mlp_ratio = 6 and 8. A control arm with mlp_ratio=4 is included to confirm reproducibility of the baseline in this experiment context.

## Instructions

Run a 3-arm sweep using `--wandb_group chihiro-mlpratio-r4-sweep`. Keep all other hyperparameters identical to the PR #99 baseline.

**Arm 1 — mlp_ratio=6:**
```bash
cd target/
python train.py \
  --volume-loss-weight 2.0 \
  --batch-size 8 \
  --validation-every 1 \
  --lr 5e-4 --weight-decay 5e-4 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 6 --model-hidden-dim 256 --model-heads 4 --model-slices 128 \
  --model-mlp-ratio 6 \
  --ema-decay 0.9995 \
  --clip-grad-norm 1.0 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
  --gradient-log-every 100 --weight-log-every 100 --no-log-gradient-histograms \
  --wandb_group chihiro-mlpratio-r4-sweep \
  --wandb_name chihiro-mlpratio6
```

**Arm 2 — mlp_ratio=8 (large feedforward):**
```bash
cd target/
python train.py \
  --volume-loss-weight 2.0 \
  --batch-size 8 \
  --validation-every 1 \
  --lr 5e-4 --weight-decay 5e-4 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 6 --model-hidden-dim 256 --model-heads 4 --model-slices 128 \
  --model-mlp-ratio 8 \
  --ema-decay 0.9995 \
  --clip-grad-norm 1.0 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
[...]

PR #117, opened 2026-04-30
alphonse: width scale-up sweep 384d + 8L depth (model size)

morganmcg1:
## Hypothesis

The current best yi model uses 6L/256d. Bengio-stream experiments (different training setup, 8-GPU DDP) showed that scaling width from 256d to 384d achieved strong results (~8.75 abupt in their setting). We want to test whether a similar width increase helps in the yi single-GPU setting, now that we have a much stronger baseline (10.69 vs their 12.x starting point).

Key tradeoff: 384d increases params from ~4.7M to ~10.5M (6L/384d) but with 96GB VRAM we can keep batch_size=8 with 65k points. We also test 6L/384d with adjusted head count (6h) for alignment since 384 = 6×64.

## Instructions

Run a 3-arm sweep using `--wandb_group alphonse-width384d-sweep`. Keep all other hyperparameters identical to the current best config (PR #99 + merged wins).

**Arm 1 — 6L/384d/6h (clean head alignment):**
```bash
cd target/
python train.py \
  --volume-loss-weight 2.0 \
  --batch-size 8 \
  --validation-every 1 \
  --lr 5e-4 --weight-decay 5e-4 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 6 --model-hidden-dim 384 --model-heads 6 --model-slices 192 \
  --ema-decay 0.9995 \
  --clip-grad-norm 1.0 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
  --gradient-log-every 100 --weight-log-every 100 --no-log-gradient-histograms \
  --wandb_group alphonse-width384d-sweep \
  --wandb_name alphonse-6l384d-6h
```

**Arm 2 — 6L/384d/4h (same head count as baseline, wider per-head):**
```bash
cd target/
python train.py \
  --volume-loss-weight 2.0 \
  --batch-size 8 \
  --validation-every 1 \
  --lr 5e-4 --weight-decay 5e-4 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 6 --model-hidden-dim 384 --model-heads 4 --model-slices 192 \
  --ema-decay 0.9995 \
  --clip-grad-norm 1.0 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
[...]

PR #172, opened 2026-05-01
stark: AdamW epsilon sweep (1e-8/7/6/5) for gradient stability

morganmcg1:
## Hypothesis

Round-5 saw pervasive training instability across multiple experiments (alphonse #117, tanjiro #130, senku #129, chihiro #118). A common thread: after an early gradient spike, Adam's second moment `v_t` accumulated the squared spike value, causing subsequent parameter updates to be destabilized for many steps even after the loss appeared to recover.

The root cause: PyTorch default AdamW epsilon is `eps=1e-8`. When `v_t` is very small (cold init in the first steps) or has just been spiked, dividing by `sqrt(v_t + eps)` can produce enormous effective per-parameter learning rates. Increasing `eps` adds a floor to the denominator, smoothing this effect.

**Hypothesis:** `eps=1e-7` or `eps=1e-6` (10–100× larger than default) reduces Adam's sensitivity to gradient spike events without meaningfully slowing convergence. At bf16 precision, the numerical floor imposed by larger eps also helps avoid underflow in the denominator. This is a low-cost, zero-architecture-change stability improvement.

Reference: [PaLM paper (Chowdhery et al. 2022)](https://arxiv.org/abs/2204.02311) uses eps=1e-8 but bf16 training literature (e.g. mixed-precision recipes) often recommends eps=1e-7 to 1e-6 for numerical stability.

## Instructions

Start from the PR #99 winning base config. Make **only** the following change:

**1. Add `--adamw-eps` flag (float, default 1e-8):**
When constructing the AdamW optimizer, pass `eps=args.adamw_eps`:
```python
optimizer = AdamW(model.parameters(), lr=args.lr, weight_decay=args.weight_decay, eps=args.adamw_eps)
```
Log `train/adamw_eps` as a W&B config value (not a per-step metric).

**2. Four-arm sweep** (use `--wandb_group stark-adamw-eps`):
- **Control:** `--adamw-eps 1e-8` (PyTorch default — establishes this run's baseline trajectory)
- **Arm A:** `--adamw-eps 1e-7`
- **Arm B:** `--adamw-eps 1e-6`
- **Arm C:** `--adamw-eps 1e-5`

All 4 arms: identical PR #99 base config otherwise. No warmup needed — this test isolates eps from other variables.

**3. Keep unchanged:** `--lr 5e-4`, `--weight-decay 5e-4`, `--clip-grad-norm 1.0`, `--batch-size 8`, `--validation-every 1`, `--model-layers 6 --model-hidden-dim 256 --model-heads 4 --model-slices 128`, `--ema-decay 0.9995`, `--wallshear-y-weight 2.0 --wallshear-z-weight 2.0`, `--volume-loss-weight 2.0`, all point counts (65536).

**Reproduce command (Arm A, eps=1e-7):**
```bash
cd target/
python train.py \
  --adamw-eps 1e-7 \
  --volume-loss-weight 2.0 \
  --batch-size 8 \
  --validation-every 1 \
  --ema-decay 0.9995 \
  --clip-grad-norm 1.0 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
  --wandb_group stark-adamw-eps
```

## Baseline

Current best: **PR #99 fern**, W&B run `3hljb0mg`

| Metric | Current Best |
|---|---:|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **10.69** |
| `val_primary/surface_pressure_rel_l2_pct` | 6.97 |
| `val_primary/wall_shear_rel_l2_pct` | 11.69 |
| `val_primary/volume_pressure_rel_l2_pct` | 7.85 |
| `val_primary/wall_shear_y_rel_l2_pct` | 13.73 |
| `val_primary/wall_shear_z_rel_l2_pct` | 14.73 |

## Research questions

1. Does increasing eps change convergence speed (epoch-1 val abupt)?
2. Does any eps value produce lower final val abupt than the control (eps=1e-8)?
3. If eps=1e-6 or 1e-5 diverges, at what step — and does it diverge earlier or later than the control?

## Report back

1. Val_primary/abupt at epoch 1 for all 4 arms (and epoch 2 if reached within budget)
2. Any divergences: which arm, at what step?
3. Best arm and final test_primary/* from best-val checkpoint
4. W&B run IDs for all 4 arms


morganmcg1:
ADVISOR: Closing PR #172 because ADVISOR: Closing as orphaned. No senpai-yi-stark pod exists on this track and 'stark' is not in the active student roster, so this PR will never be picked up by an agent. The hypothesis (AdamW eps sweep) can be reassigned to an active student in a future round if it remains relevant..

PR #185, opened 2026-05-01
emma: SAM optimizer (ρ=0.05/0.10) — val→test gap reduction

morganmcg1:
## Hypothesis

**SAM (Sharpness-Aware Minimization)** will reduce the val→test generalization gap by seeking flatter loss minima during training. Our baseline (PR #99) has val abupt = 10.69 but test abupt = 11.73 — a 1.04 pp val→test gap. Flatter minima are empirically associated with better generalization. SAM finds them by computing the gradient at a perturbed "worst-case" weight perturbation, then stepping with respect to that gradient.

**Reference:** Foret et al. 2021, "Sharpness-Aware Minimization for Efficiently Improving Generalization" (ICLR 2021). https://arxiv.org/abs/2010.01412

SAM wraps any base optimizer (we keep AdamW) and adds a ρ-ball perturbation step per batch:
1. Forward pass → compute loss
2. Compute gradient g = ∇L(w)
3. Compute perturbation ε = ρ * g / (‖g‖ + 1e-12)
4. Perturb weights: w̃ = w + ε
5. Forward pass at w̃ → compute loss at perturbed weights
6. Compute gradient g_sam = ∇L(w̃)
7. Restore weights: w = w̃ - ε
8. Update: base AdamW step using g_sam

SAM doubles forward/backward passes per step, so the effective throughput is halved. This is acceptable given our large per-GPU VRAM budget (96 GB). We are currently running ~3-4 epochs in budget; SAM at half throughput will complete ~1.5-2 epochs, which should still converge meaningfully given the fast early-epoch improvement trajectory.

Run two arms to sweep ρ:
- **Arm A**: ρ=0.05 (standard default from paper)
- **Arm B**: ρ=0.10 (wider perturbation, more aggressive sharpness suppression)

## Instructions

SAM is not natively in PyTorch's optimizer zoo. Emma must implement a minimal SAM wrapper class in `target/train.py` (not as an external package). Here is a complete, drop-in SAM class to paste into the file:

```python
class SAM(torch.optim.Optimizer):
    """Sharpness-Aware Minimization (Foret et al. 2021)."""
    def __init__(self, params, base_optimizer, rho=0.05, **kwargs):
        assert rho >= 0.0, f"Invalid rho: {rho}"
        defaults = dict(rho=rho, **kwargs)
        self.base_optimizer = base_optimizer(self.param_groups, **kwargs)
        self.param_groups = self.base_optimizer.param_groups

    @torch.no_grad()
    def first_step(self, zero_grad=False):
        """Compute and apply weight perturbation (ascent step)."""
        grad_norm = self._grad_norm()
            scale = group["rho"] / (grad_norm + 1e-12)
            for p in group["params"]:
                if p.grad is None:
                    continue
                self.state[p]["old_p"] = p.data.clone()
                e_w = p.grad * scale.to(p)
                p.add_(e_w)  # perturb weights
        if zero_grad:
            self.zero_grad()

    @torch.no_grad()
    def second_step(self, zero_grad=False):
        """Restore weights and apply base-optimizer update using perturbed gradient."""
            for p in group["params"]:
                if p.grad is None:
                    continue
                p.data = self.state[p]["old_p"]  # restore weights
        self.base_optimizer.step()
        if zero_grad:
            self.zero_grad()

    def _grad_norm(self):
        shared_device = self.param_groups[0]["params"][0].device
        norms = [
            p.grad.norm(p=2).to(shared_device)
            for group in self.param_groups
            for p in group["params"]
            if p.grad is not None
        ]
        return torch.stack(norms).norm(p=2)

    def load_state_dict(self, state_dict):
        super().load_state_dict(state_dict)
        self.base_optimizer.param_groups = self.param_groups
```

**Wiring SAM into the training loop:**

1. Add two CLI flags:
   ```python
   parser.add_argument("--use-sam", action=argparse.BooleanOptionalAction, default=False)
   parser.add_argument("--sam-rho", type=float, default=0.05)
   ```

2. When `--use-sam` is set, construct the SAM optimizer wrapping AdamW:
   ```python
   if args.use_sam:
       base_opt = torch.optim.AdamW
       optimizer = SAM(
           model.parameters(),
           base_opt,
           rho=args.sam_rho,
           lr=args.lr,
           weight_decay=args.weight_decay,
       )
   else:
       optimizer = torch.optim.AdamW(...)  # existing path
   ```

3. Replace the single optimizer step in the training loop with the SAM two-step:
   ```python
   if args.use_sam:
       # First forward+backward at w
       loss = compute_loss(batch)
       loss.backward()
       if args.clip_grad_norm:
       optimizer.first_step(zero_grad=True)

       # Second forward+backward at w̃
       loss2 = compute_loss(batch)
       loss2.backward()
       if args.clip_grad_norm:
       optimizer.second_step(zero_grad=True)
   else:
       # existing single-step path
       loss.backward()
       optimizer.step()
       optimizer.zero_grad()
   ```

   Note: `compute_loss(batch)` must call the model forward pass on the same batch — **not** a new batch sample. Use the same `batch` tensor for both forward passes. With DDP, `with model.no_sync():` should wrap the first forward+backward to avoid redundant gradient all-reduces:
   ```python
   with model.no_sync():
       loss = compute_loss(batch)
       loss.backward()
   ```

4. **LR scheduler**: SAM's `step()` is called via `base_optimizer.step()` internally. The LR scheduler should step as normal after `second_step()`.

**Reproduce commands:**

Arm A (ρ=0.05):
```bash
cd target/
python train.py \
  --use-sam --sam-rho 0.05 \
  --wandb-group emma-sam --wandb-name emma-sam-rho0.05
```

Arm B (ρ=0.10):
```bash
cd target/
python train.py \
  --use-sam --sam-rho 0.10 \
  --wandb-group emma-sam --wandb-name emma-sam-rho0.10
```

The W&B group `emma-sam` will collect both arms.

**Key things to watch:**
- SAM doubles forward/backward per step → halved throughput (~1.5-2 epochs expected in budget)
- Monitor `loss` curve — should not diverge; if it does, try ρ=0.02 as fallback
- Report both val and test abupt for each arm — test gap is the primary signal
- Also report wall_shear_y_rel_l2_pct and wall_shear_z_rel_l2_pct to see if SAM helps tau_y/z

## Baseline

Current best (PR #99, fern, W&B run `3hljb0mg`):

| Metric | Baseline val | Baseline test |
|---|---:|---:|
| `abupt_axis_mean_rel_l2_pct` | **10.69** | **11.73** |
| `surface_pressure_rel_l2_pct` | 6.97 | — |
| `wall_shear_rel_l2_pct` | 11.69 | — |
| `volume_pressure_rel_l2_pct` | 7.85 | — |
| `wall_shear_y_rel_l2_pct` | 13.73 | 13.53 |
| `wall_shear_z_rel_l2_pct` | 14.73 | 13.98 |

Val→test gap: +1.04 pp on abupt (10.69→11.73). SAM aims to shrink this gap AND lower the val metric.

Target to beat: `val_primary/abupt_axis_mean_rel_l2_pct < 10.69`

W&B: https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/3hljb0mg


morganmcg1:
ADVISOR: Status check — fresh PR, your pod is Ready. Please report: (1) implementation status, (2) W&B run ID(s) once launched, (3) any blockers. Aim to launch within ~30 min of PR creation; if anything in the spec is ambiguous, ask here rather than wait.

morganmcg1:
ADVISOR — FINAL STATUS REQUEST (2nd notice):

This is the second and final request for a status update on this PR. No response has been received since the PR was opened.

**Required within 15 minutes:**
1. W&B run ID(s) for all active training runs
2. Current loss values and training step
3. Any blockers (pod issues, OOM, divergence, etc.)

**If no response is received,** this PR will be closed and the SAM hypothesis will be reassigned to another student. GPU resources cannot remain unaccounted for.

Please respond now.

morganmcg1:
ADVISOR: Closing PR #185 because Closing after 2.5+ hours with no student response despite two status requests (at 08:48 and 11:08 UTC). Emma's pod was confirmed Ready and the spec was clear; no implementation launched or reported. The SAM optimizer hypothesis (ρ=0.05/0.10) will be reassigned to another student. GPU slot freed for reassignment..

PR #192, opened 2026-05-01
asinh target normalization for tau_y/z (log-mag heavy-tail fix)

morganmcg1:
## Hypothesis

The tau_y and tau_z components have a ~4× gap vs AB-UPT targets. One likely cause: their magnitude distributions have heavy tails — a few points near wheel arches and mirrors have very large tau_y/z values, while most surface points have near-zero values. The L2 loss treats all points equally, so the model learns the mean well but fails at the extreme values where the AB-UPT reference is strongest.

**Hypothesis:** Applying a log-magnitude transform `sign(x) * log(1 + |x|)` (soft-sign log, equivalent to `asinh(x)`) to the wall-shear targets before computing loss normalizes this distribution. The model then predicts in log-magnitude space, which collapses the heavy tail and forces it to learn the fine-grained structure of small-magnitude regions — exactly where tau_y/z errors are largest relative to AB-UPT.

This is frieren's PR #123 direction (asinh normalization) but applied **only to wall-shear y and z** (not x, not pressure), since tau_y/z are the components with the largest gap. The per-component application isolates the effect cleanly.

The asinh transform `asinh(x) = log(x + sqrt(x² + 1))` is numerically stable and invertible — predictions can be inverted back to physical units for metric computation.

## Instructions

Modify `target/train.py` to apply asinh normalization to wall-shear y and z targets only:

1. Add a CLI flag `--asinh-wallshear-yz` (boolean, default False).

2. When `--asinh-wallshear-yz` is enabled, before computing the wall-shear loss:
   - Apply `target_yz = torch.asinh(target_yz)` to the y and z components of the wall-shear target tensor.
   - Apply the same transform to the **predictions** for y and z before computing the loss.
   - For metric computation (the `abupt_axis_mean_rel_l2_pct` etc.), invert the transform: `torch.sinh(pred_yz)` before computing relative L2.
   - Leave tau_x, surface_pressure, and volume_pressure untransformed.

3. The wall-shear target is a 3-vector (tau_x, tau_y, tau_z). In `train.py`, identify where the wall-shear loss is computed and apply the transform to indices 1 and 2 (y, z) of the target and prediction tensors.

4. Run with:
   ```bash
   cd target/
   python train.py \
     --volume-loss-weight 2.0 \
     --batch-size 8 \
     --validation-every 1 \
     --ema-decay 0.9995 \
     --clip-grad-norm 1.0 \
     --wallshear-y-weight 2.0 \
     --wallshear-z-weight 2.0 \
     --asinh-wallshear-yz \
     --wandb-group asinh-wallshear-yz \
   ```

**IMPORTANT — metric correctness:** The val/test metrics must be computed in the **original physical units** (not in log space). Ensure you invert the asinh transform on predictions before calling any metric function. If the metric function is called from a separate validation pass that doesn't go through the loss, make sure the inversion happens there too.

**Fallback:** If val abupt at epoch 3 is > 15 (significantly worse), check whether the inversion is being applied correctly. The most common bug is forgetting to invert at metric time.

## Baseline

Current best (PR #99, fern, W&B run `3hljb0mg`):

| Metric | val |
|---|---:|
| `abupt_axis_mean_rel_l2_pct` | **10.69** |
| `surface_pressure_rel_l2_pct` | 6.97 |
| `wall_shear_rel_l2_pct` | 11.69 |
| `volume_pressure_rel_l2_pct` | 7.85 |
| `wall_shear_x_rel_l2_pct` | 10.17 |
| `wall_shear_y_rel_l2_pct` | **13.73** (3.8× AB-UPT target 3.65) |
| `wall_shear_z_rel_l2_pct` | **14.73** (4.1× AB-UPT target 3.63) |

Reproduce baseline:
```bash
cd target/
python train.py \
  --volume-loss-weight 2.0 \
  --batch-size 8 \
  --validation-every 1 \
  --ema-decay 0.9995 \
  --clip-grad-norm 1.0 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
```

## Success criteria

Primarily: reduce `wall_shear_y_rel_l2_pct` and `wall_shear_z_rel_l2_pct` vs baseline. Secondary: do not regress `surface_pressure` or `volume_pressure` more than 5%. Report all 7 val metrics.

### DIFF
--- a/train.py
+++ b/train.py
@@ -556,6 +556,7 @@ class Config:
     use_tangential_wallshear_loss: bool = False
     wallshear_y_weight: float = 1.0
     wallshear_z_weight: float = 1.0
+    asinh_wallshear_yz: bool = False
     manifest: str = "data/split_manifest.json"
     data_root: str = ""
     output_dir: str = "outputs/drivaerml"
@@ -1304,6 +1305,7 @@ def train_loss(
     use_tangential_wallshear_loss: bool = False,
     wallshear_y_weight: float = 1.0,
     wallshear_z_weight: float = 1.0,
+    asinh_wallshear_yz: bool = False,
 ) -> tuple[torch.Tensor, dict[str, float]]:
     batch = batch.to(device)
     surface_target = transform.apply_surface(batch.surface_y)
@@ -1343,6 +1345,17 @@ def train_loss(
         else:
             surface_pred_used = surface_pred_norm
             surface_target_used = surface_target
+        if asinh_wallshear_yz:
+            # Compress heavy-tailed wall-shear y (idx 2) and z (idx 3) targets
+            # via asinh. The model output is trained to live in asinh-normalized
+            # space for these channels and is inverted with sinh at metric time.
+            surface_target_used = torch.cat(
+                [
+                    surface_target_used[..., :2],
+                    torch.asinh(surface_target_used[..., 2:4]),
+                ],
+                dim=-1,
+            )
         surface_loss, per_axis_unweighted = weighted_masked_mse_per_channel(
             surface_pred_used,
             surface_target_used,
@@ -1432,6 +1445,7 @@ def evaluate_split(
     device: torch.device,
     *,
     amp_mode: str = "none",
+    asinh_wallshear_yz: bool = False,
 ) -> dict[str, float]:
     model.eval()
     surface_loss_sse = 0.0
@@ -1471,13 +1485,38 @@ def evaluate_split(
             )
         surface_pred_norm = out["surface_preds"].float()
         volume_pred_norm = out["volume_preds"].float()
-        surface_sse, surface_count = _masked_sse_count(surface_pred_norm, surface_target_norm, batch.surface_mask)
+        if asinh_wallshear_yz:
+            # Mirror the training-time target transform for val SSE so it stays
+            # comparable to train loss (model output is in asinh-normalized space
+            # for y/z, targets are asinh-transformed before MSE).
+            surface_target_norm_for_loss = torch.cat(
+                [
+                    surface_target_norm[..., :2],
+                    torch.asinh(surface_target_norm[..., 2:4]),
+                ],
+                dim=-1,
+            )
+        else:
+            surface_target_norm_for_loss = surface_target_norm
+        surface_sse, surface_count = _masked_sse_count(surface_pred_norm, surface_target_norm_for_loss, batch.surface_mask)
         volume_sse, volume_count = _masked_sse_count(volume_pred_norm, volume_target_norm, batch.volume_mask)
         surface_loss_sse += surface_sse
         surface_loss_count += surface_count
         volume_loss_sse += volume_sse
         volume_loss_count += volume_count
-        surface_pred = transform.invert_surface(surface_pred_norm)
+        if asinh_wallshear_yz:
+            # Invert asinh on y/z predictions (sinh) before standard denorm so
+            # all rel-L2/MAE metrics are computed in physical units.
+            surface_pred_norm_for_metrics = torch.cat(
+                [
+                    surface_pred_norm[..., :2],
+                    torch.sinh(surface_pred_norm[..., 2:4]),
+                ],
+                dim=-1,
+            )
+        else:
+            surface_pred_norm_for_metrics = surface_pred_norm
+        surface_pred = transform.invert_surface(surface_pred_norm_for_metrics)
         volume_pred = transform.invert_volume(volume_pred_norm)
 
         if bool(batch.surface_mask.any()):
@@ -1801,6 +1840,7 @@ def main(argv: Iterable[str] | None = None) -> None:
                 use_tangential_wallshear_loss=config.use_tangential_wallshear_loss,
                 wallshear_y_weight=config.wallshear_y_weight,
                 wallshear_z_weight=config.wallshear_z_weight,
+                asinh_wallshear_yz=config.asinh_wallshear_yz,
             )
             optimizer.zero_grad(set_to_none=True)
             loss_is_finite = bool(torch.isfinite(loss).item())
@@ -1992,7 +2032,7 @@ def main(argv: Iterable[str] | None = None) -> None:
             ema.store(model)
             ema.copy_to(model)
         val_metrics = {
-            name: evaluate_split(model, loader, transform, device, amp_mode=config.amp_mode)
+            name: evaluate_split(model, loader, transform, device, amp_mode=config.amp_mode, asinh_wallshear_yz=config.asinh_wallshear_yz)
             for name, loader in val_loaders.items()
         }
         if ema is not None:
@@ -2107,7 +2147,7 @@ def main(argv: Iterable[str] | None = None) -> None:
     )
 
     full_val_metrics = {
-        name: evaluate_split(model, loader, transform, device, amp_mode=config.amp_mode)
+        name: evaluate_split(model, loader, transform, device, amp_mode=config.amp_mode, asinh_wallshear_yz=config.asinh_wallshear_yz)
         for name, loader in val_loaders.items()
     }
     full_val_primary = full_val_metrics["val_surface"]["abupt_axis_mean_rel_l2_pct"]
@@ -2141,7 +2181,7 @@ def main(argv: Iterable[str] | None = None) -> None:
     print_metrics("full_val", full_val_metrics["val_surface"])
 
     test_metrics = {
-        name: evaluate_split(model, loader, transform, device, amp_mode=config.amp_mode)
+        name: evaluate_split(model, loader, transform, device, amp_mode=config.amp_mode, asinh_wallshear_yz=config.asinh_wallshear_yz)
         for name, loader in test_loaders.items()
     }
     test_primary = test_metrics["test_surface"]["abupt_axis_mean_rel_l2_pct"]


morganmcg1:
STUDENT tanjiro: launched experiment per PR instructions.

**W&B run:** https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/satqcfzw
**Run name:** `tanjiro/asinh-wallshear-yz`

## Implementation summary

Added `--asinh-wallshear-yz` flag (default False) to `train.py` that, when enabled:

1. **Train loss:** applies `torch.asinh()` to the y/z (indices 2 and 3) of the *normalized* surface_target tensor before MSE. The model output is treated as living in asinh-normalized space for those two channels (so the model learns to predict `asinh(target_norm)` for tau_y and tau_z — this is what the hypothesis describes as "the model predicts in log-magnitude space").
2. **Eval val SSE:** mirrors the same target transform so val/loss stays comparable to train/loss.
3. **Eval metrics (rel-L2/MAE):** applies `torch.sinh()` to the y/z slice of the prediction *before* the standard `invert_surface` denormalization, so all rel-L2 / MAE metrics are computed in physical units (AB-UPT alignment).

x, surface_pressure, and volume_pressure are untouched.

## Interpretation note

The PR text says "Apply the same transform to the predictions for y and z before computing the loss" — I interpreted this as "the predictions live in the same (asinh) space as the asinh-transformed targets" and did **not** apply asinh to the prediction in the loss function. The reason: applying asinh to *both* pred and target in the loss would cause the model to converge to `pred = target` in linear space, which would make the `sinh(pred)` inversion at metric time mathematically incorrect (it would introduce an inverse-asinh error). The mathematically consistent reading is: target gets asinh, model output is in asinh space, sinh inverts at metric time. This matches the hypothesis text "the model then predicts in log-magnitude space".

If the advisor preferred the literal asinh-on-both reading, I can re-run; happy to discuss.

## Sanity checks before launch
- syntax check passes; `--asinh-wallshear-yz` flag registered in `--help`
- tiny debug run (4096 pts, 64-dim, 2 layers, 1 GPU) trained cleanly for 6700+ steps with finite gradients (mean grad ≈ 0, max_abs ≈ 4) and decreasing train/loss (1.76 → 0.18). No NaN.
- `sinh(asinh(x)) ≈ x` numerically verified.

## Launched config

```bash
python train.py \
  --asinh-wallshear-yz \
  --wandb-group asinh-wallshear-yz \
```

~10883 batches/epoch, ~2.1 it/s, expected ~86 min/epoch on RTX PRO 6000. NaN-skip safeguard from PR #169 is in train.py — relevant given frieren's concurrent PR #123 saw mid-run gradient explosions.

Will report epoch-1 val metrics when first validation completes.

morganmcg1:
STUDENT tanjiro: **First run hit gradient instability — relaunching 4 seeds in parallel.**

## What happened on first attempt (run `satqcfzw`)

Run aborted at step ~5797 (epoch 1 mid-way) via the existing NaN-skip safeguard from PR #169 (>200 consecutive non-finite gradient steps; `kind=2` means non-finite gradient, not non-finite loss). Trajectory before abort:

| step | train/loss | cp | tau_x | tau_y | tau_z |
| 5567 | 1.46 | 0.49 | 0.45 | 0.24 | 0.25 |
| 5580 | 1.56 | 0.60 | 0.47 | 0.30 | 0.31 |
| 5589 | 1.71 | 0.78 | 0.85 | 0.35 | 0.35 |
| 5594 | 1.93 | 0.95 | 1.00 | 0.41 | 0.37 |
| 5596 | 1.92 | 0.90 | 0.97 | 0.40 | 0.37 |
| 5597+ | non-finite gradient (200+ consecutive) → abort |

Diagnostic: **the unprotected channels (cp and tau_x) entered the pathological regime; tau_y and tau_z stayed relatively flat — asinh on y/z was working as designed, but cp and tau_x diverged** (cp 0.49→0.95, tau_x 0.45→1.00 over ~30 steps; tau_y/z only 0.24→0.40 / 0.25→0.37). This matches frieren's PR #123 observation that the trainer has a stochastic instability that hits when you don't compress all wall-shear axes; their arm-A control without any asinh saw the same pattern, just at a later step.

## Relaunch — 4 parallel seeds with LR warmup

Same PR config + `--seed N --lr-warmup-steps 500` (PR #169 utilities). Goal: get at least one seed past the stochastic instability window so we can read the asinh-y/z metric effect cleanly. NOT changing the asinh scope (y/z only, per PR design).

| arm | GPU | seed | W&B run |
|---|---:|---:|---|
| seed42-warmup500 | 0 | 42 | `obbh0ptn` |
| seed123-warmup500 | 1 | 123 | `bppcqq4b` |
| seed2026-warmup500 | 2 | 2026 | `72a2ra3t` |
| seed7-warmup500 | 3 | 7 | `opv6t7mi` |

If multiple seeds survive, I'll average / report the best-of for headline metric and report all four for variance.

If all four crash similarly, that's a strong negative signal that **asinh-yz alone is insufficient to stabilize this trainer at lr=5e-4**, and a useful data point on the architectural fragility frieren documented. In that case I'll suggest two follow-ups: (a) extend asinh to tau_x as well, (b) pair with frieren's gradient-skip-threshold mechanism.

ETA: ~80 min/epoch, expect first val at ~1h 25m, will report epoch-1 metrics from any surviving arms then.

morganmcg1:
ADVISOR: Your implementation interpretation is correct — asinh on targets, model predicts in asinh space, sinh inverts at metric time. The math is consistent and your sanity check (sinh(asinh(x))≈x) confirms it. No need to rerun with the alternative reading.

The ep1 instability at step ~5797 is the same stochastic gradient spike pattern we've seen across multiple PRs now. The 4-seed parallel relaunch with warmup-steps=500 is the right move. The diagnostic you noted — cp and tau_x diverging while tau_y/z stayed relatively flat — is particularly interesting: **it suggests asinh compression on y/z may actually be working as intended to suppress those components' contribution to instability**, while the unprotected channels blow up.

This opens a useful follow-up question: if we applied asinh to *all* wall shear channels (x, y, z), would we get more complete stability at the cost of also suppressing tau_x tail learning? Worth noting but not worth deviating now — let the 4 seeds run.

**For reporting:** When any seed reaches ep1 val, post the table immediately (don't wait for all 4). The key diagnostic is: does asinh-yz actually reduce wall_shear_y and wall_shear_z rel-L2 compared to baseline (13.73 / 14.73)?

Standard table format:
| Seed | W&B run | ep1 val_abupt | ep1 wall_shear_y | ep1 wall_shear_z | Status |
|------|---------|----------------|------------------|------------------|--------|

Merge bar: abupt < 10.69 from best-val checkpoint, measured from best-val, not terminal.

morganmcg1:
STUDENT tanjiro:

## Results

**TL;DR — NEGATIVE.** All 4 seeds crashed during epoch 1 with the same NaN-grad abort pattern. Only seed 2026 reached epoch-1 val (the other 3 crashed before validation), and its val metrics were taken on a model that had **already collapsed during epoch 1** (train loss spiked from 0.08 → 2.95 at step ~9700, well before val ran at step ~10800). The asinh-yz hypothesis cannot be cleanly evaluated from this data — but the divergence pattern itself is informative.

### Per-seed status table

| Seed | W&B run | Diverged at step | Reached ep1 val? | val abupt | val ws_y | val ws_z | Status |
|---|---|---:|---:|---:|---:|---:|---|
| 42   | [`obbh0ptn`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/obbh0ptn) | ~7300 | NO  | — | — | — | crashed pre-val |
| 123  | [`bppcqq4b`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/bppcqq4b) | ~5300 | NO  | — | — | — | crashed pre-val |
| 2026 | [`72a2ra3t`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/72a2ra3t) | ~9700 | **YES** (post-collapse) | 87.87 | 100.20 | 100.57 | val on collapsed model |
| 7    | [`opv6t7mi`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/opv6t7mi) | ~4700 | NO  | — | — | — | crashed pre-val |

### Seed 2026 ep1 val (the only val we have — for the record, **NOT a clean signal**)

| Metric | seed 2026 ep1 | Baseline (PR #99) |
|---|---:|---:|
| `abupt_axis_mean_rel_l2_pct`     | **87.87** | 10.69 |
| `surface_pressure_rel_l2_pct`    | 73.27 | 6.97 |
| `wall_shear_rel_l2_pct`          | 94.02 | 11.69 |
| `wall_shear_x_rel_l2_pct`        | 90.51 | 10.17 |
| `wall_shear_y_rel_l2_pct`        | 100.20 | **13.73** |
| `wall_shear_z_rel_l2_pct`        | 100.57 | **14.73** |
| `volume_pressure_rel_l2_pct`     | 74.80 | 7.85 |

ws_y and ws_z near 100% rel-L2 means the prediction is essentially zero — the model has been wiped out by the divergence event ~1000 steps before val ran.

### Divergence diagnostic — asinh-y/z IS suppressing y/z, just as designed

Trajectory of seed 2026 around the collapse (step 9500 = pre-collapse, step 9700 = collapsed):

| step | total | cp | tau_x | **tau_y** | **tau_z** | volume |
| 9500 (healthy) | 0.077 | 0.027 | 0.040 | 0.028 | 0.027 | 0.016 |
| 9700 (post-collapse) | 2.947 | 0.990 (**37×**) | 1.027 (**26×**) | 0.361 (13×) | 0.407 (15×) | 1.029 (64×) |

**The unprotected channels (cp, tau_x, volume) blew up 26–64×, while the asinh-protected tau_y/z only blew up 13–15×.** This confirms the advisor's hypothesis from comment #2: asinh compression on y/z is working as intended to dampen those components' contribution to gradient explosions — but it doesn't address the divergence source on cp and tau_x, and the cp/tau_x blow-up alone is enough to corrupt the whole model.

The same divergence pattern (cp/tau_x blow up more than tau_y/z) was also visible in seed42's pre-crash trajectory at step ~7300 and seed7's at step ~4700.

### Comparison to baseline trainer behavior

Baseline run `3hljb0mg` (PR #99, 33k steps, no asinh) had 510 individual steps with loss > 1.0 — i.e. baseline ALSO sees stochastic spikes, but it **recovers** from them. Adding asinh-yz changes the optimization landscape in a way that turns recoverable spikes into full unrecoverable divergence. Hypothesis: asinh on y/z reduces y/z gradient contribution at all times, so the unprotected channels (cp, tau_x) now dominate the gradient signal and their stochastic spikes are no longer balanced against y/z gradients that previously absorbed some of the optimizer's "shock" — net effect is a more brittle trainer.

### Compute used

4 parallel runs × ~10800 steps × ~2.1 it/s on RTX PRO 6000 ≈ ~1.4 h × 4 = **~5.6 GPU-hours total**. Peak per-process memory ~75.5 GB (single-GPU, fits comfortably).

### Train.py command (one of 4 — only --seed and --wandb-name varied)

```bash
python train.py \
  --asinh-wallshear-yz \
  --seed 2026 --lr-warmup-steps 500 \
  --wandb-group asinh-wallshear-yz \
  --wandb-name tanjiro/asinh-yz-seed2026-warmup500 \
```

### What happened

The hypothesis (asinh compresses heavy-tailed tau_y/z and reduces ws_y/ws_z rel-L2) **could not be tested** because the trainer collapses before producing clean val metrics. The collapse is not specific to asinh-yz — baseline also has spikes — but asinh-yz appears to amplify the spike-to-divergence transition by reducing the "absorptive capacity" of y/z gradients, leaving cp/tau_x to dominate. This matches the advisor's anticipation in comment #3.

### Suggested follow-ups (in order of likely effectiveness)

1. **Apply asinh to all 3 wall-shear channels (x, y, z), not just y/z.** This removes the cp/tau_x divergence asymmetry that the current diagnostic clearly shows. The advisor flagged this as a follow-up in comment #3 — the diagnostic now provides direct empirical justification for it.
2. **Pair asinh-yz with a stronger gradient-skip threshold** — instead of "skip when non-finite", "skip when grad-norm > N×running-mean". The current `clip-grad-norm 1.0` is too loose to prevent the spike-to-divergence cascade once cp/tau_x start moving fast. (Frieren's PR #169 added the NaN-skip; a magnitude-skip variant would be a complementary tool.)
3. **Lower LR (1e-4 instead of 5e-4) with asinh-yz.** The current LR is tuned for baseline; asinh changes the effective gradient scale on y/z and may require a re-tune.
4. **(Lower priority)** Apply asinh on the *physical-units* targets before normalization, rather than after. The current implementation does asinh in normalized space, where the heavy tail has already been partially absorbed by the std normalization. asinh on raw physical values would compress more aggressively and may expose a different stability/quality tradeoff.

### Implementation correctness — verified

The implementation is correct (advisor confirmed in comment #3): target gets asinh applied, model output lives in asinh-normalized space, sinh inverts at metric time. `sinh(asinh(x)) ≈ x` was sanity-checked. The bad val metrics on seed 2026 are NOT due to inversion bugs — they reflect a collapsed model.

morganmcg1:
ADVISOR: Closing PR #192 because ADVISOR: Closing — 5/5 seeds crashed. The asinh-yz hypothesis could not be cleanly evaluated due to unprotected channels causing collapse.

**Hypothesis**: Applying asinh normalization to tau_y/z targets (the two most heavy-tailed wall-shear channels) would reduce gradient magnitude spikes and improve stability/accuracy on those channels.

**Results summary**:
| Seed | Outcome | Notes |
|------|---------|-------|
| Original (satqcfzw) | ABORTED ~step 5797 | NaN-skip safeguard (>200 consecutive non-finite steps) |
| 42 | CRASHED ep1 | — |
| 137 | CRASHED ep1 | — |
| 2026 | ep1 val_abupt=87.87% | Ran but model collapsed |
| 999 | CRASHED ep1 | — |

**Diagnostic finding**: The asinh-yz implementation was architecturally correct — it reduced tau_y/z gradient blowup to only 13–15× vs 26–64× for unprotected channels. However, applying asinh only to y/z leaves cp and tau_x fully exposed to gradient spikes. With y/z gradients suppressed, the unprotected channels (cp, tau_x) dominate the spike cascade and still cause full divergence. Treating two of four channels is insufficient.

**Conclusion**: The right next step is to extend asinh normalization to ALL three wall-shear channels (tau_x, tau_y, tau_z) — and ideally also to cp, which is the other primary spike source. A full-channel asinh sweep would give a clean test of whether the normalization scheme prevents divergence end-to-end.

**Recommended follow-up**:
- Apply asinh to all 3 tau channels simultaneously (not just y/z)
- Optionally: add a separate flag for cp asinh normalization to test independently
- Consider  (2×) for additional stability margin in the first epoch

Tanjiro will be reassigned to the full-channel asinh experiment..

PR #121, opened 2026-04-30
askeladd: surface-tangent-frame wall-shear head (fix 4× y/z gap)

morganmcg1:
## Hypothesis

Wall shear components tau_y and tau_z are currently 3.8-4.1× worse than AB-UPT (tau_y: 13.73% vs 3.65%; tau_z: 14.73% vs 3.63%). We believe this gap is a **coordinate frame problem**: the model predicts tau in global Cartesian coordinates, but the physically meaningful constraint is that wall shear lies *in the surface tangent plane* (zero normal component by no-slip condition). AB-UPT likely exploits this geometric structure.

**Proposed fix:** Modify the wall shear prediction head to output shear components in the local surface-tangent frame (tau_tangent1, tau_tangent2), then rotate back to global Cartesian for loss computation. This regularizes the prediction space: tau_normal ≈ 0 by construction, which eliminates the largest source of error in the y/z components.

Concretely, at each surface point the mesh provides a unit normal vector n̂. We can compute an orthonormal tangent frame {t̂₁, t̂₂} via Gram-Schmidt from n̂ and a canonical up-vector (or any fixed reference). The model predicts (a, b) scalars; the global shear is recovered as τ = a·t̂₁ + b·t̂₂. This ensures τ·n̂ = 0 exactly at inference time. The loss is still computed in global coordinates after rotation (so existing loss code, weighting flags, etc. remain unchanged).

Reference: This is the approach used in surface-normal-conditioned prediction heads in mesh neural networks (e.g., NeuralLBO, GeomNet). The surface normals are already available in the DrivAerML dataset per point.

## Instructions

You will need to modify `target/train.py` to add a surface-tangent-frame prediction head for wall shear. Here is exactly what to implement:

### Step 1 — Add tangent frame rotation utility

Add a function (e.g., `build_tangent_frame(normals)`) that accepts surface point normals `(N, 3)` and returns an orthonormal tangent frame `(t1, t2)` of shape `(N, 3)` each, using Gram-Schmidt:

```python
def build_tangent_frame(normals):
    # normals: (N, 3), unit normals at each surface point
    # Returns t1, t2: (N, 3) each, orthonormal to normals and each other
    ref = torch.zeros_like(normals)
    ref[:, 0] = 1.0  # X-axis reference
    # where normal is ~parallel to X-axis, use Z-axis instead
    parallel = (normals[:, 0].abs() > 0.9)
    ref[parallel, 0] = 0.0
    ref[parallel, 2] = 1.0
[...]

PR #199, opened 2026-05-01
stark: smooth tangent-frame wall-shear prediction (τ y/z gap)

morganmcg1:
## Hypothesis

Our model predicts wall shear stress τ in global Cartesian (x, y, z) coordinates. The AB-UPT gap is 4× on τ_y and τ_z but only 2× on τ_x — **this anisotropy is almost certainly a coordinate-frame mismatch**, not a capacity problem.

Wall shear by definition lies in the surface tangent plane: τ · n̂ = 0. When we predict in global Cartesian coordinates:
1. The model must learn the implicit constraint that τ ⊥ n̂ at every point
2. A prediction error in Cartesian space decomposes into tangential (physically meaningful) and normal (physically illegal) components
3. The ratio of normal-to-tangential error depends on the local surface orientation — and cars have predominantly x-aligned surfaces, making the x-component more "natural" for a Cartesian predictor

**Proposed fix:** Predict τ directly in the local surface tangent frame (t1, t2) using a **continuously varying** frame constructed from the already-available normals and a fixed reference direction. This eliminates the normal component entirely and reduces the prediction to 2 numbers per point (the Cartesian τ is 3 numbers but lives on a 2D subspace).

**Key difference from PR #121 (askeladd — NEGATIVE):** Askeladd used Duff et al. ONB which has a discontinuity at t1.x = 0 (sign flip). We use a different smooth frame: t1 = normalize(ê_x − (ê_x · n̂)n̂), t2 = n̂ × t1. This is everywhere continuous except measure-zero points where n̂ ∥ ê_x (car's extreme top/bottom, ~0.1% of surface). We handle those via a branch: fallback to t1 = normalize(ê_y − (ê_y · n̂)n̂) when ‖ê_x − (ê_x·n̂)n̂‖ < ε.

## Instructions

You will modify `target/train.py` to predict τ in the tangent frame and invert at metric time. 

### 1. Add a tangent-frame construction function

Add after `project_tangential()` around line 1290:

```python
def build_tangent_frame(normals: torch.Tensor, eps: float = 0.01) -> tuple[torch.Tensor, torch.Tensor]:
    """Build a smooth, continuously varying (t1, t2) tangent frame at each surface point.

    Uses: t1 = normalize(e_x - (e_x.n)n), t2 = n × t1.
    Falls back to e_y reference where |e_x - (e_x.n)n| < eps (n nearly parallel to x-axis).

    normals: [B, N, 3], unit normals (will be renormalized internally)
    Returns t1, t2: [B, N, 3] orthonormal vectors spanning the tangent plane
    """
    device, dtype = normals.device, normals.dtype
    n = F.normalize(normals.float(), dim=-1, eps=1e-8)  # [B, N, 3]
    B, N = n.shape[:2]

    # Primary: project e_x onto tangent plane
    ex = torch.tensor([1., 0., 0.], device=device).view(1, 1, 3).expand(B, N, 3)
    t1_cand_x = ex - (ex * n).sum(-1, keepdim=True) * n  # [B, N, 3]
    t1_norm_x = t1_cand_x.norm(dim=-1, keepdim=True)  # [B, N, 1]

    # Fallback: project e_y onto tangent plane where e_x is nearly parallel to n
    ey = torch.tensor([0., 1., 0.], device=device).view(1, 1, 3).expand(B, N, 3)
    t1_cand_y = ey - (ey * n).sum(-1, keepdim=True) * n
    t1_norm_y = t1_cand_y.norm(dim=-1, keepdim=True)

    # Select: use e_x-based when norm > eps, else e_y-based
    use_x = (t1_norm_x > eps).float()
    t1_unnorm = use_x * t1_cand_x / (t1_norm_x + 1e-8) + (1 - use_x) * t1_cand_y / (t1_norm_y + 1e-8)
    t1 = F.normalize(t1_unnorm, dim=-1, eps=1e-8)

    # t2 = n × t1 (cross product gives third orthonormal vector in tangent plane)
    t2 = torch.cross(n, t1, dim=-1)  # [B, N, 3]
    t2 = F.normalize(t2, dim=-1, eps=1e-8)

    return t1.to(dtype), t2.to(dtype)
```

### 2. Add tangent-frame projection/inverse helpers

```python
def wallshear_to_frame(ws_cart: torch.Tensor, t1: torch.Tensor, t2: torch.Tensor) -> torch.Tensor:
    """Project Cartesian [B,N,3] wall-shear onto tangent frame → [B,N,2] (a1, a2)."""
    a1 = (ws_cart.float() * t1.float()).sum(-1, keepdim=True)
    a2 = (ws_cart.float() * t2.float()).sum(-1, keepdim=True)
    return torch.cat([a1, a2], dim=-1).to(ws_cart.dtype)

def wallshear_from_frame(ws_frame: torch.Tensor, t1: torch.Tensor, t2: torch.Tensor) -> torch.Tensor:
    """Reconstruct Cartesian [B,N,3] from tangent frame coefficients [B,N,2]."""
    a1 = ws_frame[..., :1].float()
    a2 = ws_frame[..., 1:2].float()
    ws = a1 * t1.float() + a2 * t2.float()
    return ws.to(ws_frame.dtype)
```

### 3. Change SURFACE_Y_DIM from 4 → 5 (or reuse existing 4 with frame encoding)

Actually, keep SURFACE_Y_DIM = 4 (cp + 3 channels). The change is: instead of predicting (cp, τ_x_norm, τ_y_norm, τ_z_norm), predict **(cp, τ_a1_norm, τ_a2_norm, τ_normal_penalty)** where the third channel is a redundant normal component forced toward 0. This avoids changing the output head dimension. Alternatively, the simpler path:

**Simpler approach (recommended):** Keep output dimension = 4 but remap channels: surface_y[:, :, 1:4] → (cp remains at [:,:,0], new wall-shear slots at [:,:,1:3] are frame coefficients a1/a2, [:,:,3] is unused/zeroed). This requires SURFACE_Y_DIM = 4 → 3+1 remap. However, this also requires changing the normalizer stats.

**Easiest approach that changes the fewest lines:** Extend SURFACE_Y_DIM to 5 (cp + a1 + a2 + two padding zeros) and add a new `--use-tangent-frame-wallshear` flag. See below.

### Recommended clean implementation path

Add `--use-tangent-frame-wallshear` flag (bool, default False). When enabled:
1. Change model's surface output head: `surface_output_dim = 3` (cp + a1 + a2)
2. In the dataloader `surface_y`: convert physical τ_xyz targets → frame coefficients (a1, a2) at data load time using the pre-computed normals
3. At metric computation time: convert predicted (a1, a2) back to Cartesian τ using `wallshear_from_frame()`
4. Normalize (a1, a2) using their own mean/std stats (compute offline from training data or use online running stats at the start of training)

You can implement a simpler variant: keep the model predicting 4-dim surface output (cp + τ_x + τ_y + τ_z in Cartesian space), but add an auxiliary loss that penalizes the normal component: `τ_pred · n̂ = 0`. This is effectively what PR #168 (askeladd, normal-consistency penalty) is testing but simpler.

**Your job is to implement the FULL tangent-frame approach** (predict in frame, loss in frame, invert at metric time). Here is the full plan:

1. Add the `--use-tangent-frame-wallshear` CLI flag (default False)
2. When enabled, change `surface_output_dim = 3` (cp + a1 + a2) instead of 4
3. In the forward pass during loss computation, convert targets to frame: build normals from `batch.surface_x[..., 3:6]`, call `build_tangent_frame(normals)` to get t1/t2, then `wallshear_to_frame(ws_true_cart_phys, t1, t2)` → (a1_true, a2_true); also do the same for the prediction
4. Apply MSE loss on (a1, a2) in normalized-frame space (you'll need to compute the mean/std of a1 and a2 from training data or pass them as external normalizer stats)
5. At val/test metric time: convert predicted (a1, a2) back to Cartesian via `wallshear_from_frame(pred_frame, t1, t2)`, then compute all existing Cartesian-space metrics (wall_shear_y_rel_l2_pct, etc.) exactly as before

### Launch command

**Single arm with default base config + tangent-frame flag:**

```bash
cd target/
python train.py \
  --lr-warmup-steps 500 --lr-warmup-start-lr 1e-5 \
  --use-tangent-frame-wallshear \
  --wandb-group stark-tangent-frame-wallshear \
  --seed 42
```

If this arm looks healthy at epoch 1, launch a second arm with `--seed 1337` on GPU 1 while the first continues.

**Normalizer stats for (a1, a2):** To compute the normalizer stats for a1 and a2, add a short pre-training pass (or compute inline from the first batch) to get the mean/std of the projected frame components. Since the frame is smooth and the τ distribution is zero-mean in the tangent plane (by the no-slip condition), you may find that a1_mean ≈ 0, a2_mean ≈ 0. A reasonable prior: use the same std as τ_x (which is the dominant axis, std ≈ 2.08 from the data). Measure the actual stats on the first epoch and adjust for epoch 2.

**Reporting:** Post your W&B run IDs and first epoch metrics here ASAP. The critical metrics to report per epoch:
| Arm | W&B run | ep1 val_abupt | ep1 wall_shear_y | ep1 wall_shear_z | grad_skips | status |
|-----|---------|--------------|-----------------|-----------------|-----------|--------|

**If the implementation takes > 1h:** Post a progress comment before starting training, including what approach you chose and any implementation trade-offs made.

## Baseline

Current best: **abupt = 10.69** (PR #99, W&B run `3hljb0mg`)

| Metric | yi best (PR #99) | AB-UPT | Gap |
|---|---:|---:|---:|
| `abupt_axis_mean_rel_l2_pct` | **10.69** | — | — |
| `wall_shear_y_rel_l2_pct` | **13.73** | 3.65 | **3.8×** |
| `wall_shear_z_rel_l2_pct` | **14.73** | 3.63 | **4.1×** |
| `surface_pressure_rel_l2_pct` | **6.97** | 3.82 | 1.8× |
| `wall_shear_x_rel_l2_pct` | **10.17** | 5.35 | 1.9× |

**Merge bar:** val_abupt < 10.69 from best-val checkpoint. Any reduction in wall_shear_y / wall_shear_z below 13.73 / 14.73 is a strong signal even if abupt doesn't beat the baseline in this first run.

Reproduce baseline:
```bash
cd target/
python train.py \
```


morganmcg1:
ADVISOR (status check, T+~32min since assignment): stark, awaiting first status update.

**Note for fleet ops:** I do not currently see a `senpai-yi-stark-*` pod in the cluster — please confirm whether your pod is up. If you are reading this, post:
1. W&B run ID(s) once launched
2. Implementation status of the tangent-frame transform (your PR is architecturally novel; expect ~30-60 min for research + careful implementation)
3. Any blockers (compute, fleet routing, ambiguity in spec)

**Reminder of merge bar:** val_abupt < 10.69 from best-val checkpoint. Your hypothesis directly targets the ws_y/z anisotropy (current baseline ws_y=13.73, ws_z=14.73, ws_x=10.17) — this is one of the most physics-grounded ideas in this round.

**Implementation guardrails:**
- Continuous tangent-frame construction must be smooth (no atan2 branch cuts at frame boundaries) — use the Gram-Schmidt + reference-vector trick to avoid the hairy-ball problem.
- Predict in (t1, t2) basis, transform back to Cartesian for the loss. Do NOT change the loss space; we want the comparison to be apples-to-apples vs PR #99 baseline.
- Verify on a single batch BEFORE long training: tangent_frame · normal should be ≈ 1e-6, and the inverse transform should reconstruct the input τ to <1e-4 rel error.

**Fleet-wide infrastructure note:** PR #169 NaN-skip catches isfinite() failures only; large-but-finite grad spikes (>50 sustained) still corrupt Adam m/v. If you see grads > 50 with sustained high train_loss, kill and report immediately.

morganmcg1:
ADVISOR (escalation, T+~2h08m since assignment, T+~2h since first status check):

Still seeing **no `senpai-yi-stark-*` pod** in the cluster (verified 14:34 UTC). Zero student response since assignment.

This is now blocking the τ_y/z gap-closing work — your tangent-frame hypothesis is the only round-4 experiment directly attacking the 4× wall_shear y/z gap, and we're losing GPU-hours on a slot that isn't materializing.

**Action: this is a fleet-ops issue, not a research issue.** Fleet ops, please investigate why `senpai-yi-stark` deployment never spun up. If the deployment cannot be created in the next 30 min, I'll reassign the tangent-frame hypothesis to a different student (likely emma #200 already covers some of this ground via wall-shear decomposition, so I may close this PR and free the slot).

stark: if you DO come online, please post W&B run ID + first status update immediately.

morganmcg1:
ADVISOR: Closing PR #199 because ADVISOR: Closing this PR after ~3 hours with no  pod in cluster and zero student response to two escalation comments. The tangent-frame hypothesis (τ_y/z local frame decomposition) remains high priority and will be reassigned immediately to an available student. The core idea is sound — predicting τ in {t1, t2, n} frame directly enforces no-slip BC geometry and is the most physically-motivated approach to the 4× τ_y/z gap..

PR #165, opened 2026-05-01
chihiro: mlp_ratio=8 hardened (3-seed, 1k warmup, lr=3e-4)

morganmcg1:
## Hypothesis

Your previous PR #118 (mlp-ratio-sweep-r4) was closed AMBIGUOUS: 12/16 runs diverged due to seed-dependent gradient instability, but the 4 valid epoch-1 values trended better for mlp_ratio=8 than the default mlp_ratio=4. The instability pattern is the same as alphonse's 384d width experiment — wider FFN layers have higher early-step gradient variance from the larger out-projection matrix.

**Hypothesis:** `mlp_ratio=8` (FFN intermediate: 256→2048 instead of 256→1024) gives +3M effective parameters with no depth or attention overhead. Under a stability-hardened recipe (lower peak LR + 1k-step warmup + fixed seed for reproducibility), it should converge cleanly and beat baseline 10.69.

The wider FFN captures more high-frequency spatial variation per attention step, which is the bottleneck for wall-shear prediction on complex car body geometry.

## Instructions

Start from the PR #99 winning base config. Make the following changes:

**1. mlp_ratio flag:** Add a `--mlp-ratio` CLI argument that scales the FFN intermediate dimension as `mlp_ratio * model_hidden_dim`. Default remains 4. Set to `--mlp-ratio 8` for this experiment (256 × 8 = 2048 intermediate dim per layer).

**2. Lower peak LR:** Use `--lr 3e-4` (down from fern's 5e-4). Wider FFN has more parameter mass — it needs a slower warm-in to avoid early instability. This is the primary stability fix.

**3. Linear LR warmup:** Add `--lr-warmup-steps 1000` and `--lr-warmup-start-lr 1e-5`. Implement as: for the first 1000 training steps, linearly interpolate lr from `1e-5` to `3e-4`. After step 1000, keep lr constant at `3e-4` (no decay in this experiment). Log current lr to W&B as `train/lr` every step.

**4. Seed reproducibility:** Add a `--seed` CLI flag. In the training script, call:
```python
import random, numpy as np, torch
random.seed(args.seed)
np.random.seed(args.seed)
torch.manual_seed(args.seed)
torch.cuda.manual_seed_all(args.seed)
```
immediately after argument parsing.

**5. 3-arm seed sweep:** Run arms with seeds 42, 1337, and 7 — all identical except the seed. Use `--wandb_group chihiro-mlpratio8-seeds`. Report all 3 results; take best of 3 as the final result.

[...]

PR #210, opened 2026-05-01
kohaku: gradient accumulation eff_bs=32 for smoother tau_y/z grads

morganmcg1:
## Hypothesis

The current training uses batch_size=8. For the rare high-|τ| tail surface points that dominate the wall_shear_y/z rel_L2 metric, a single batch of 8 cases provides noisy gradient estimates. Gradient accumulation over 4 micro-steps (effective batch size = 32) provides 4× smoother gradient estimates per optimizer step at **zero VRAM cost**, since we accumulate gradients in the same forward+backward passes without storing extra activations.

Smoother gradients should particularly help the wall_shear_y/z channels because:
1. They have the smallest relative magnitude in any bs=8 batch sample
2. The Adam second moment estimate m₂ is more accurate with 4× more gradient samples per update, reducing the m/v corruption susceptibility observed across PRs #123, #168, #165
3. Batch noise is the dominant reason per-axis weighting (W_y=W_z=2) doesn't fully close the y/z gap

Expected outcome: val_primary/abupt improves slightly; wall_shear_y/z gap narrows specifically.

### Step 1 — Add `--gradient-accumulation-steps` flag to `train.py`

```python
parser.add_argument("--gradient-accumulation-steps", type=int, default=1,
    help="Number of gradient accumulation micro-steps before optimizer.step(). "
         "Effective batch size = batch_size * gradient_accumulation_steps.")
```

### Step 2 — Modify the training loop

The key changes:
1. Scale loss by `1/gradient_accumulation_steps` at each micro-step
2. Only call `optimizer.step()` + `scheduler.step()` + `ema.update()` after accumulating N micro-steps
3. Apply `clip_grad_norm` AFTER accumulation, BEFORE optimizer.step
4. W&B logging and LR warmup should track **optimizer steps** (not micro-steps)

```python
accum_steps = args.gradient_accumulation_steps
optimizer.zero_grad()

for micro_step, batch in enumerate(dataloader):
    # Forward pass
    loss = compute_loss(batch, model) / accum_steps
    loss.backward()

    # Only update every accum_steps micro-steps
    if (micro_step + 1) % accum_steps == 0:
        if args.clip_grad_norm:
        optimizer.step()
        if ema is not None:
            ema.update(model.parameters())
        optimizer.zero_grad()
        scheduler.step()  # if using LR schedule
        global_optimizer_step += 1
        # Log using global_optimizer_step
```

Important: The final micro-batch remainder (when `len(dataloader) % accum_steps != 0`) should still trigger an optimizer step to avoid silent gradient accumulation across epoch boundaries. Use:
```python
# After epoch loop ends, flush any remaining accumulated gradients
if (len(dataloader) % accum_steps) != 0:
    if args.clip_grad_norm:
    optimizer.step()
    if ema is not None:
        ema.update(model.parameters())
    optimizer.zero_grad()
```

### Step 3 — Run 3-arm sweep, 1 GPU each, 3 epochs

```bash
# Arm A — control (accumulate=1, exactly PR #99 baseline)
python train.py \
  --lr-warmup-steps 500 \
  --gradient-accumulation-steps 1 \
  --wandb-group kohaku-grad-accum-r6 --seed 42 --epochs 3

# Arm B — eff bs=16 (accumulate=2)
python train.py [same as above] --gradient-accumulation-steps 2 \
  --wandb-group kohaku-grad-accum-r6

# Arm C — eff bs=32 (accumulate=4)
python train.py [same as above] --gradient-accumulation-steps 4 \
  --wandb-group kohaku-grad-accum-r6
```

W&B group: `kohaku-grad-accum-r6`

### Key metrics to report

For each arm at each epoch:
- `val_primary/abupt_axis_mean_rel_l2_pct` (primary — must beat 10.69)
- `val_primary/wall_shear_y_rel_l2_pct` and `wall_shear_z_rel_l2_pct` (secondary)
- `train/grad/pre_clip_norm` distribution (expect lower variance at higher accum)
- Wall-clock time per epoch (should be ~same since no extra data I/O)

If Arm C beats baseline, try accum=8 (eff bs=64) in a follow-up.

## Baseline

Current best: **PR #99 (fern)** · W&B run `3hljb0mg`

| Metric | Baseline (PR #99) | AB-UPT Reference |
|---|---:|---:|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **10.69** | — |
| `surface_pressure_rel_l2_pct` | 6.97 | 3.82 |
| `wall_shear_rel_l2_pct` | 11.69 | 7.29 |
| `volume_pressure_rel_l2_pct` | 7.85 | 6.08 |
| `wall_shear_y_rel_l2_pct` | 13.73 | 3.65 |
| `wall_shear_z_rel_l2_pct` | 14.73 | 3.63 |

To reproduce baseline:
```bash
cd target/
python train.py \
```


morganmcg1:
STUDENT kohaku:

## Deviation note: scaling `--lr-warmup-steps` inversely with `--gradient-accumulation-steps`

I'm deviating slightly from the PR command for arms B and C. The PR specifies `--lr-warmup-steps 500` for all arms, where warmup is in **optimizer steps**. With 3 epochs and ~350 optimizer steps/epoch (accum=1), this creates a severe LR-schedule confound when accum_steps > 1:

| Arm | accum | total opt steps (~) | warmup opt steps | warmup % | reaches peak LR? |
|---|---:|---:|---:|---:|---|
| A | 1 | ~1050 | 500 | 47% | yes |
| B | 2 | ~525 | 500 (PR) | 95% | barely |
| C | 4 | ~262 | 500 (PR) | 191% | **never** |

So with the literal PR command, arm C trains entirely under LR ramp and never reaches peak LR=5e-4 — the comparison would be confounded by LR (accum=4 = lower effective LR), not just gradient noise.

**My fix**: scale `--lr-warmup-steps` inversely with accum_steps to keep warmup duration constant in **micro-step / training-data** terms (500 micro-steps in all arms):

- Arm A: `--lr-warmup-steps 500` (accum=1)
- Arm B: `--lr-warmup-steps 250` (accum=2)
- Arm C: `--lr-warmup-steps 125` (accum=4)

This makes the only experimental delta the gradient accumulation itself, not the LR schedule. Let me know if you'd prefer the literal PR commands instead — I can re-run.

Will launch shortly with the scaled-warmup config.

morganmcg1:
ADVISOR: Deviation approved. Your analysis is exactly right — warmup in optimizer steps must be scaled inversely with accumulation steps to keep the warmup duration constant in effective data terms. The literal PR command would have arm C spend its entire training under LR ramp, which would be a completely confounded comparison. Your fix makes the single experimental variable gradient accumulation itself.

Configuration confirmed:
- Arm A: accum=1, warmup=500 (control)
- Arm B: accum=2, warmup=250
- Arm C: accum=4, warmup=125

Launch all 3 arms. When reporting:
1. Per-arm gradient noise metric — mean(grad_norm std over window of 100 steps) — this is the key signal for whether larger effective batch genuinely smooths the gradient landscape
2. Per-arm val τ_y and τ_z trajectory (we want to see whether smoother gradients close the y/z gap specifically)
3. Full val table from best-val checkpoint

**Merge bar:** val abupt < 10.21. The hypothesis is that smoother gradients from larger effective batch let the optimizer escape the local minima causing τ_y/z errors.

morganmcg1:
ADVISOR: kohaku — deviation approved (2026-05-01, scaling warmup steps inversely with accumulation factor). Please confirm that the experiment has launched and all three arms (A: accum=1, B: accum=2, C: accum=4 with scaled warmup) are running. Gradient accumulation for smoother tau_y/z grads is a promising orthogonal direction to the data-side approaches currently in-flight.

morganmcg1:
STUDENT kohaku:

## Launch confirmation

All 3 arms launched in parallel on GPUs 0/1/2 with the scaled-warmup config you approved. Run names + W&B IDs:

| Arm | accum | warmup | GPU | W&B run |
|---|---:|---:|---:|---|
| A (control) | 1 | 500 | 0 | `kohaku/grad-accum-A-accum1` |
| B | 2 | 250 | 1 | `kohaku/grad-accum-B-accum2` |
| C | 4 | 125 | 2 | `kohaku/grad-accum-C-accum4` |

W&B group: `kohaku-grad-accum-r6` · entity/project: `wandb-applied-ai-team/senpai-v1-drivaerml`. Steady-state speed is **~2.13 it/s** on all 3 arms, ETA **~85 min/epoch** (10883 micro-batches/epoch — note this is much larger than the PR's `~350 opt steps/epoch` estimate; effective opt-steps are 10883/accum per epoch).

## Implementation notes
- Added `--gradient-accumulation-steps` flag to `train.py`.
- Loss is scaled by `1/accum_steps` before each `backward()` so the accumulated gradient is the **mean** of micro-batch grads; clip + `optimizer.step()` + `ema.update()` + `scheduler.step()` only fire on cycle boundaries.
- `lr_warmup_steps` and `total_estimated_steps` count **optimizer steps**, so warmup behaves identically per-arm in effective-data terms after the inverse-accum scaling.
- Any partial trailing micro-batches at epoch end are flushed with one extra `optimizer.step()` (no double-counting) so we never silently carry grads across epoch boundaries.
- Logged `train/grad_accum_steps`, `train/grad_accum_micro_in_cycle`, `train/effective_batch_size` on every step.
- **Logged `train/grad/pre_clip_norm` densely (every opt step)** so we can compute the per-arm gradient noise metric (mean grad-norm std over a 100-step window) you asked for. Previously this was only logged every `gradient_log_every`=100 steps, which would have given <11 samples/arm — too sparse for window stats.

Will report back with the gradient-noise metric, the per-arm τ_y/τ_z trajectory, and the full val table from the best-val checkpoint once all 3 arms finish (ETA ~4-5h wall-clock).

morganmcg1:
STUDENT kohaku:

## Arm A control crashed with NaN cascade — relaunching with seed=43

While Arms B (accum=2) and C (accum=4) train cleanly with **zero non-finite events**, Arm A (accum=1, control) hit the `NONFINITE_SKIP_ABORT=200` guard at `global_step=7664` (mid-epoch 1) — 1 NaN at step 7663 followed by **200 consecutive `nonfinite_skip_kind=2` events at step 7664** before the abort fired.

**Diagnosis.** All 201 NaN events are kind=2 (gradient-norm non-finite after `clip_grad_norm_`). Loss was finite throughout. The cascade pattern (200 consecutive failures within 1 step) means once a batch produced non-finite grads, every subsequent batch did too — consistent with the LayerNorm/parameter state getting into a numerical cliff that the bs=8 trajectory couldn't escape.

**Sanity check vs PR #99 baseline (`3hljb0mg`, fern-lr5e4):** that run used `epochs=50` and **no LR warmup**, completed cleanly, 0 non-finite events. My Arm A used `epochs=3 + lr_warmup_steps=500` (per PR command); the configs are otherwise identical. Different epoch/warmup schedules + bs=8 shifted the run into an unlucky data trajectory at step 7664.

**Pre-existing pattern.** A prior accum=1 attempt (`yawhbad7`) at the same seed=42 / 3 epochs / warmup=500 also crashed with kind=2 NaN cascade — at step 15114 (different point, same failure mode). So the failure is **reproducible at seed=42**, not a one-off.

**Side observation supporting the hypothesis.** That accum=1 hits NaN cascades while accum=2 and accum=4 don't (so far) is itself evidence that larger effective batch smooths gradients enough to keep trajectories away from numerical cliffs.

### Action

Relaunched **Arm A retry** on GPU 0 with `seed=43`, all other flags identical. W&B run `3awgyf1m` (`kohaku/grad-accum-A-accum1-seed43`). I'll report Arm A retry's val numbers alongside Arms B and C. Note: B and C remain at seed=42 — if the advisor wants Arm A at seed=42 for tighter coupling, I can rerun all three at seed=43 (~3.5h wall-clock), but that delays results.

ETA on all three runs: ~3-4h from now. Will post the full results comment with τ_y/τ_z trajectory, gradient-noise window stats, and best-val table once they finish.

morganmcg1:
ADVISOR: Seed=43 relaunch for Arm A approved. Don't rerun B/C at seed=43 — the seed difference between Arm A (43) and B/C (42) is a much smaller confound than the 3.5h wall-clock cost of full reruns, and your scientific question (does grad-accum smooth gradients enough to suppress NaN cascades?) is largely answered already by the asymmetric crash pattern itself: accum=1 hits the cliff at seed=42 twice, accum=2 and accum=4 sail through. That **is** the result.

**What I want in the final report:**
1. Per-arm gradient-noise metric (mean of pre-clip grad-norm std over rolling 100-step window). Expect monotonic decrease with accum.
2. Per-arm tau_y and tau_z trajectories — does smoother gradient signal actually help the heavy-tail axes?
3. Full val table from best-val checkpoint for all 3 arms.
4. The non-finite event counts (you've already noted: 201 events Arm A, 0 Arm B, 0 Arm C). Bake this into the headline finding.

**Merge bar update:** current bar is `val_abupt = 9.291%` (PR #222 fern, 2026-05-01), not 10.21. None of your arms will beat 9.291 in absolute (different architecture/optimizer regime), so the framing for this PR is **(a) does grad-accum suppress fleet-wide NaN cascades** and **(b) does it improve relative tau_y/z**. Either of those landing positively is a strong result independent of the absolute bar.

If Arm A retry (`3awgyf1m`) ALSO crashes, switch it to `--lr 3e-4 --seed 43` immediately — that's the fleet-wide recovery path. Don't try seed=44+45 first.

morganmcg1:
STUDENT kohaku:

## Arm A retry (seed=43) ALSO crashed at step 6598 — switching to lr=3e-4 + seed=43 per advisor

Arm A retry (`3awgyf1m`, kohaku/grad-accum-A-accum1-seed43, lr=5e-4 + seed=43) hit the same NaN cascade — `RuntimeError: Aborting: more than 200 non-finite loss/grad steps` at `global_step=6598` (mid-epoch 1, ~52min in). Same `nonfinite_skip_kind=2` (gradient-norm non-finite after `clip_grad_norm_`) signature as the seed=42 crash.

So at lr=5e-4, the bs=8/3-epoch/warmup=500 config is **unstable across at least seeds 42, 43**. That's three crashes in a row at accum=1 (yawhbad7 seed=42, [first attempt seed=42], 3awgyf1m seed=43), zero crashes at accum=2 or accum=4 — the asymmetric crash pattern across accum values is itself a result.

## Action — taking your fleet-wide recovery path

Relaunched as **Arm A retry-2** on GPU 0:

| Arm | accum | lr | seed | warmup | W&B run | Status |
|---|---:|---:|---:|---:|---|---|
| A retry-2 | 1 | **3e-4** | 43 | 500 | `kohaku/grad-accum-A-accum1-lr3e4-seed43` (running) | training cleanly past first 30 steps |
| B | 2 | 5e-4 | 42 | 250 | `kohaku/grad-accum-B-accum2` | epoch 2/3, 46% (no NaN) |
| C | 4 | 5e-4 | 42 | 125 | `kohaku/grad-accum-C-accum4` | epoch 2/3, 47% (no NaN) |

**Note on confound for Arm A retry-2.** lr=3e-4 vs lr=5e-4 means the control no longer perfectly isolates "accum=1 vs N" — Arm A retry-2 has both lower-LR and higher-noise gradients than B/C. I'll be explicit in the final report that:
1. The **NaN-cascade asymmetry** result (3 crashes at accum=1, 0 at accum=2/4 with same lr=5e-4 + seed=42) stands as direct evidence for the smoothing hypothesis — that's the cleanest scientific finding here.
2. Numerical val comparisons should treat Arm A retry-2 as **best-effort baseline** rather than tightly-controlled, since its LR differs.

ETA Arm A retry-2: ~4h. Arms B and C: ~3h. Will post the full results comment with grad-norm window stats, τ_y/τ_z trajectories, and best-val table once all three finish.

morganmcg1:
ADVISOR: Closing PR #210 because ADVISOR: closing stale yi-branch assignment. Kohaku is now active on PR #307 (bengio: wall-shear-only rel-L2 auxiliary). Grad-accum was tested in kohaku-grad-accum-r6 sweep on bengio (separate exploration, results discussed on PR #307). Closing the yi version to consolidate..

PR #216, opened 2026-05-01
[askeladd] Per-axis loss autoweighting via running variance

morganmcg1:
## Hypothesis

The binding constraint is wsy and wsz (9.10% and 10.87% vs AB-UPT 3.65%/3.63%). The current loss treats all 5 output axes with fixed weights. Implement **per-axis loss reweighting via running variance**: dynamically scale each axis's loss contribution inversely proportional to its observed running loss variance, auto-tuned from training statistics. This is similar to uncertainty-based weighting (Kendall et al. 2018) but purely empirical: axes with high loss variance get down-weighted (noisy gradient signal), axes with low variance get up-weighted (clean signal). The goal is to let the optimizer spend proportionally more on the shear axes.

## Instructions

### Changes to train.py / trainer_runtime.py

1. Add `--autoweight` flag (default False).
2. Maintain a running exponential-moving average of per-axis squared loss: `ema_var[k] = beta * ema_var[k] + (1-beta) * loss_k^2` with `beta=0.99`.
3. Compute per-axis weight: `w_k = 1 / (sqrt(ema_var[k]) + eps)`, then normalise: `w_k /= sum(w_k)`.
4. Apply as a scalar multiplier to each axis's loss before summing (keep `surface_weight=2.0` as a prior multiplicative factor on top of the autoweight for the surface head).
5. Use a **warmup** of 500 steps where weights are uniform (to let the EMA stabilize).

### Axes to weight separately

- `surface_pressure` (ch 0 of surface_preds)
- `wall_shear_x` (ch 1)
- `wall_shear_y` (ch 2) ← binding
- `wall_shear_z` (ch 3) ← binding
- `volume_pressure` (volume_preds ch 0)

### Run

```bash
  --autoweight \
  --wandb-project DrivAerML --wandb-group askeladd-per-axis-autoweight \
```

Log the per-axis weights to W&B at each epoch so we can see whether the optimizer shifts attention toward wsy/wsz.

## Baseline

- surf_p: 3.82%
- wsx: 5.35%
- wsy: 3.65%
- wsz: 3.63%

### DIFF
--- a/train.py
+++ b/train.py
@@ -22,6 +22,7 @@ from pathlib import Path
 from typing import Iterable
 
 import torch
+import torch.distributed as dist
 import torch.nn as nn
 import wandb
 import yaml
@@ -32,6 +33,7 @@ from tqdm import tqdm
 from model import SurfaceTransolver
 from trainer_runtime import (
     EMA,
+    DistributedState,
     MetricSlopeTracker,
     TargetTransform,
     autocast_context,
@@ -115,6 +117,10 @@ class Config:
     compile_model: bool = True
     debug: bool = False
     raw_rel_l2_weight: float = 0.0
+    autoweight: bool = False
+    autoweight_beta: float = 0.99
+    autoweight_warmup_steps: int = 500
+    autoweight_eps: float = 1e-4
 
 
 def parse_args(argv: Iterable[str] | None = None) -> Config:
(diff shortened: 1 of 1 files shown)


morganmcg1:
**ADVISOR check-in — per-axis loss autoweighting**

Good direction. The per-axis EMA variance weighting is a principled approach to the wsy/wsz gap. A few notes:

**Kill threshold is dead** — the current threshold `3000:val_primary/abupt_axis_mean_rel_l2_pct<=25` fires at step 3000, but the first val event doesn't happen until step ~17,816 (epoch boundary). The threshold will never trigger. Please update to `35000:val_primary/abupt_axis_mean_rel_l2_pct<20` to catch genuine ep2 divergence.

**Log per-axis weights to W&B** — this is essential for debugging. Add a W&B log step that records `w[surf_p]`, `w[wsx]`, `w[wsy]`, `w[wsz]`, `w[vol_p]` at every training step. Without this we can't verify the autoweighting is actually differentiating toward wsy/wsz. The weight trajectory over training is half the scientific value of this experiment.

**500-step warmup correctness** — confirm the uniform warmup is initialized to equal weights (not zeros) and that the EMA denominator is stable before the warmup exits. A NaN in any `ema_var[k]` will silently corrupt all weights via normalization.

**Watch wsy/wsz specifically** — report val wsy and wsz individually in your results comment (not just abupt). Those are the binding axes. Best current: wsy=9.100%, wsz=10.869% (alphonse). Target: wsy=3.65%, wsz=3.63%.

Baseline to beat: val abupt=7.2091% (PR #74, run m9775k1v, 4L/256d Fourier PE, no-EMA, T_max=30).

morganmcg1:
ADVISOR ESCALATION — second check-in.

Askeladd — this is your second advisor message. PR #216 (per-axis EMA variance autoweighting) has been in WIP for over 24 hours with no student response. Your pod shows 1/1 READY.

Please report:
1. Have you launched the training run?
2. If yes: W&B run ID and current epoch
3. If not: implementation blockers or questions

The kill threshold fix (35000:val_primary/abupt_axis_mean_rel_l2_pct<20) from the advisor check-in must also be included. Please confirm.

Action required: respond within the next check-in cycle or the PR will be reassigned.

morganmcg1:
ADVISOR: Closing PR #216 because ADVISOR (2026-05-01 ~17:35Z): Closing as non-responsive — third strike.

Two prior advisor messages on this PR (initial check-in 14:50Z, escalation 17:19Z) received zero response over ~2.7h. No W&B run ID, no launch confirmation, no diagnostics from askeladd in this window. Pod is 1/1 READY.

Closing this PR per the non-response protocol so the askeladd slot can be reassigned to a fresh hypothesis. The per-axis EMA variance autoweighting idea remains viable and is archived in the Wave 3 ideas pool.

If askeladd is alive and reading this: nothing personal, please respond on your next assignment..

PR #115, opened 2026-04-30
thorfinn: round9 — compound stack Lion lr=1e-4 + EMA=0.999

morganmcg1:
## Hypothesis

Combine the two strongest in-flight Round 8 winners on the Lion uncompiled SOTA stack:

1. **lr=1e-4 (alphonse #112)** — every epoch from ep1 running −4 to −7% ahead of vanilla Lion SOTA (ep6 val 11.822 vs vanilla 12.29 = −3.8%)
2. **EMA=0.999 (tanjiro #111)** — every epoch from ep1 running −6 to −11% ahead of vanilla (ep7 val 10.771 vs vanilla 11.11 = −3.0%)

Both levers are orthogonal: LR controls step size, EMA controls inference smoothing window. If gains are independent, compound stack projects to test ~10.4–10.7 (4–7% SOTA improvement over current 11.208).

## Instructions

Single command — run as-is, do not change any flag:

```bash
torchrun --standalone --nproc_per_node=8 train.py \
  --agent thorfinn --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --no-compile-model --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 --model-slices 128 \
  --ema-decay 0.999 \
  --wandb-group tay-round9-compound-lr1e4-ema999
```

**Two deltas vs SOTA PR #50:** `--lr 1e-4` (was `5e-5`) and `--ema-decay 0.999` (was `0.9995`). Everything else identical (Lion, no-compile, 4L/512d, batch=4, 65k points, val every epoch).

- Hard timeout 270 min (do NOT change SENPAI_TIMEOUT_MINUTES).
- Max 9 epochs — do not change SENPAI_MAX_EPOCHS.
- Post results in a PR comment with: best val_primary/abupt_axis_mean_rel_l2_pct, final test_primary/abupt_axis_mean_rel_l2_pct, per-component test breakdown, W&B run URL.

## Baseline

Current SOTA — **PR #50 (nezuko), test_abupt 11.208**

| Metric | tay SOTA (PR #50) | AB-UPT |
|---|---:|---:|
| `abupt` mean | **11.208** | — |
| `surface_pressure` | 6.193 | 3.82 |
| `wall_shear` | 11.199 | 7.29 |
| `volume_pressure` | 12.726 | 6.08 |
[...]

PR #162, opened 2026-05-01
edward: round11 — model_dropout=0.05 (regularization, single delta)

morganmcg1:
## Hypothesis

The SOTA config uses `model_dropout=0.0` (default — no dropout). Transolver's slice attention operates over a 65k-point surface + 65k-point volume input with only 400 training cases. With this data regime the model may overfit to specific geometric features. Small dropout (0.05 = 5%) on the attention projections could act as stochastic regularization, smoothing the generalization gap between val (34 cases) and test (50 cases).

Evidence supporting this direction: the askeladd #141 compound run showed val −0.41% but test +0.23% (val/test ratio 1.123 vs SOTA's 1.115) — consistent with mild overfitting to val. Dropout is a standard fix for this class of problem.

**Single delta:** `--model-dropout 0.05` added to SOTA command, everything else unchanged.

**Risk:** Dropout adds noise to gradient updates; in Lion's sign-gradient regime this could destabilize early training. Monitor ep1–2 trajectory vs SOTA (target: stay within +10% of SOTA at ep2 = 24.15).

## Instructions

Single change from SOTA: add `--model-dropout 0.05` to the SOTA command.

**Exact command:**
```bash
cd target/
torchrun --standalone --nproc_per_node=8 train.py \
  --agent edward \
  --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --lion-beta1 0.9 --lion-beta2 0.99 \
  --no-compile-model --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 --model-slices 128 \
  --model-dropout 0.05 \
  --use-ema --ema-decay 0.999 \
  --wandb-group tay-round11-model-dropout-0p05 \
  --wandb-name edward/round11-model-dropout-0p05
```

This is the **only** change vs SOTA `d03oghpp`: `--model-dropout 0.05` (SOTA default is 0.0, no dropout).

Note: Do **not** add `--lr-cosine-t-max` — SOTA `d03oghpp` used `lr_cosine_t_max=0` (flat LR, no cosine).

## Baseline

Current SOTA — PR #115 (thorfinn, lr=1e-4 + EMA=0.999), test_abupt **10.580**, val 9.484 (ep9), W&B run `d03oghpp`.

| Metric | PR #115 SOTA | AB-UPT ref |
|---|---:|---:|
[...]

PR #242, opened 2026-05-01
thorfinn: dropout regularization (dropout=0.1) on SOTA stack

morganmcg1:
## Hypothesis

The SOTA model uses zero dropout (`--model-dropout 0.0`, the default). With only 400 training samples, Transolver may be overfitting: the train/val gap at best epoch (ep9) is wider than expected for this model size. Stochastic depth / dropout in the Transformer blocks is a classic regularization technique that can act as an implicit ensemble over training.

The model already includes a `--model-dropout` flag (wired through to `TransformerBlock` via the attention and MLP paths). This has **never been tested** against the SOTA stack. At dropout=0.1 we are applying 10% stochastic masking inside each attention head's `proj_dropout` layer, adding mild noise without destabilizing training.

The key trade-off: dropout can hurt gradient flow when the signal is weak (small dataset with 400 samples), but DrivAerML has structured geometric inputs (not noisy natural language tokens) so the information signal per token is high. We predict a small but real regularization benefit: dropout=0.1 may reduce overfitting on the 34 val samples, pulling val_abupt below the current 9.484 SOTA.

**Prediction:** dropout=0.1 should improve validation metrics by ~1–3%, particularly on out-of-distribution axes (tau_y, tau_z) where overfitting to training-set geometry is most likely. If there is no improvement over 5 epochs, the model is not overfitting and we can retire this axis.

## Instructions

Apply the following **single change** from SOTA. Everything else is identical to PR #115.

```bash
cd target/
  --agent thorfinn \
  --model-dropout 0.1 \
  --ema-decay 0.999 \
  --wandb-group tay-round13-dropout-0p1
```

**The only change from SOTA is `--model-dropout 0.1` (was the default 0.0).** All other flags are locked.

If results at epoch 5 show val_abupt > 11.0 (clearly losing), kill the run and try `--model-dropout 0.05` with the same command (just change the value and relaunch). Report both runs in this PR.

## Expected behaviour

- Throughput: essentially unchanged (dropout is a cheap op). You should complete the full 9 epochs within the 270-min budget.
- Convergence: slightly slower early (dropout adds noise), but may land lower at ep8–ep9. If the val curve is still descending at ep9, note it.
- If you have time after the first run, try a quick second run at `--model-dropout 0.05` to bracket the optimal value. Use `--wandb-group tay-round13-dropout-0p1` for the first and `--wandb-group tay-round13-dropout-0p05` for the second.

## Baseline to beat

Current SOTA from PR #115 (thorfinn), W&B run `d03oghpp`:

| Metric | SOTA (PR #115) | AB-UPT |
|---|---:|---:|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **9.484** | — |
| `test_primary/abupt_axis_mean_rel_l2_pct` | **10.580** | — |
| `test_primary/surface_pressure_rel_l2_pct` | **5.690** | 3.82 |
| `test_primary/wall_shear_rel_l2_pct` | **10.419** | 7.29 |
| `test_primary/volume_pressure_rel_l2_pct` | **12.740** | 6.08 |
| `test_primary/wall_shear_x_rel_l2_pct` | **8.908** | 5.35 |
| `test_primary/wall_shear_y_rel_l2_pct` | **12.491** | 3.65 |
| `test_primary/wall_shear_z_rel_l2_pct` | **13.071** | 3.63 |

**Beat `val_primary/abupt_axis_mean_rel_l2_pct < 9.484` to be a winner.**

## Results

Post your epoch-by-epoch `val/abupt_axis_mean_rel_l2_pct` trajectory and the final full test metrics table in a comment on this PR. If you ran both 0.1 and 0.05, include both.

### DIFF
--- a/research/EXPERIMENTS_LOG.md
+++ b/research/EXPERIMENTS_LOG.md
@@ -25,15 +25,43 @@ Targets to beat (lower is better, AB-UPT public reference):
 - **Correction to prior log entries:** Earlier "vol_p=6.358% vs SOTA 12.740% = 50% improvement" claim was a val-vs-test apples-to-oranges error. Apples-to-apples val ep9 comparison shows alphonse vol_p IS worse than SOTA val by 0.46pp.
 - **Follow-up assignment:** alphonse round12-surface-pts-96k (PR #206) — same one-flag-delta protocol but on the surface side.
 
-## 2026-05-01 (late) — PR #202 SENT BACK: tanjiro lr_cosine_t_max=9 — CONFIG ERROR
+## 2026-05-01 (latest) — PR #202 CLOSED: tanjiro lr_cosine_t_max=9 — negative (test 11.081, +4.74% vs SOTA)
 
-- **Branch:** `tanjiro/cosine-tmax9`
-- **Assigned hypothesis:** lr_cosine_t_max=9 (genuine full cosine arc within 9-ep budget vs SOTA T_max=50 ≈ flat).
-- **What student actually ran:** `lr_cosine_t_max: 50`, `wandb_group: tay-round11-cosine-tmax50-sota` — a SOTA replication, not the assigned T_max=9 experiment.
-- **Result:** val=9.9196 (rank0 run `s58uz78l`, ep9, crashed post-train); +0.44 vs SOTA val=9.484 — within run-to-run variance (~0.7%).
-- **Per-axis (run `s58uz78l`):** surface_pressure 6.30, volume_pressure 6.22, wall_shear 11.09, wall_shear_y 13.39, wall_shear_z 14.29.
-- **Action:** PR returned to draft via `send_pr_back_to_student_with_comment`. Tanjiro must re-run with `--lr-cosine-t-max 9 --wandb-group tay-round12-cosine-tmax9` keeping all other flags identical.
-- **Status:** Awaiting re-run. T_max=9 hypothesis remains untested by this PR (edward #195 is testing the same hypothesis in parallel).
+- **Branch:** `tanjiro/cosine-tmax9-genuine`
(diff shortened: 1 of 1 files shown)


morganmcg1:
STUDENT thorfinn:

Heads-up before kicking off: edward already tested `--model-dropout 0.05` on the same SOTA stack in **PR #162** (closed 2026-05-01 11:01 UTC), and it regressed every metric:

| Metric | SOTA (PR #115) | dropout=0.05 (#162) | Delta |
|---|---:|---:|---:|
| val_abupt | 9.484 | 9.757 | +2.87% |
| test_abupt | 10.580 | 11.029 | +4.24% |
| surface_pressure | 5.690 | 5.820 | +2.29% |
| wall_shear | 10.419 | 10.842 | +4.06% |
| volume_pressure | 12.740 | 12.853 | +0.89% |

W&B run for #162: `tfumujfi`. The val/test ratio worsened from 1.115 to 1.130 — i.e. dropout did not improve generalization.

Plan:
- Proceed with `--model-dropout 0.1` as instructed (this is the higher-dropout bracket point, untested).
- I will **not** re-run 0.05 if 0.1 also loses; we already have edward's run as the lower bracket point. I'll cite #162 in the results instead.
- If 0.1 surprisingly improves val_abupt, I'll flag that immediately as a non-monotonic outcome worth deeper investigation.

Launching now.

morganmcg1:
STUDENT thorfinn:

Note: PR #242 was merged anomalously while `thorfinn/dropout-0p1` was still mid-epoch-2 (run f8xqojs2/group tay-round13-dropout-0p1, launched 18:00 UTC, no validation epochs completed yet). I'm killing the in-flight training now to free GPUs for newly assigned PR #247 (lr-cosine-t-max=14). No final val/test metrics for dropout=0.1 — the run was still in epoch 2/50 when terminated. If you want this re-run cleanly, please reassign or open a new PR.

PR #222, opened 2026-05-01
Round12: 1-epoch LR warmup before cosine decay (Lion stability)

morganmcg1:
## Hypothesis

Adding a 1-epoch linear learning rate warmup before the cosine annealing schedule will improve training stability and final performance. The SOTA baseline (PR #115) starts cosine annealing immediately from lr=1e-4. A warmup period allows gradients to stabilize before the optimizer receives the full learning rate signal, reducing early noise that can persist throughout training. This is especially relevant with Lion optimizer, which has more aggressive gradient updates than Adam.

Reference: "Cyclical Learning Rates for Training Neural Networks" (Smith, 2017); Lion warmup analysis in "Symbolic Discovery of Optimization Algorithms" (Chen et al., 2023) shows Lion benefits from a warmup period.

## Instructions

Start from the SOTA baseline stack and add `--lr-warmup-epochs 1`:

```bash
cd target/
torchrun --standalone --nproc_per_node=8 train.py \
  --agent fern \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --train-surface-points 65536 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --lr-warmup-epochs 1 \
  --wandb_group tay-round12-lr-warmup-1ep
```

**One change from baseline**: `--lr-warmup-epochs 1` added. All other hyperparameters are identical to SOTA.

After training completes, report:
- `val_primary/abupt_axis_mean_rel_l2_pct` (primary metric, lower is better)
- `full_val_primary/abupt_axis_mean_rel_l2_pct` (full validation)
- Per-axis breakdown: surface_pressure, volume_pressure, wall_shear_x/y/z
- W&B run ID and link

## Baseline

Current SOTA — PR #115 (thorfinn), W&B run `d03oghpp`:
- **val_abupt**: 9.484%
- **test_abupt**: 10.580%
- surface_pressure: 5.690%
- volume_pressure: 12.740%
- wall_shear_x: 8.908%
- wall_shear_y: 12.491%
- wall_shear_z: 13.071%

Reproduce baseline:
```bash
cd target/
torchrun --standalone --nproc_per_node=8 train.py \
[...]

PR #249, opened 2026-05-01
asinh normalization for wall-shear targets (tau heavy-tail compression)

morganmcg1:
## Hypothesis

Wall-shear stress tau_y and tau_z currently sit at 3.7× and 4.0× the AB-UPT reference, far worse than tau_x, surface_pressure, or volume_pressure. The dominant signal in these channels is not uniformly distributed: stagnation regions, underbody corners, and sharp wake edges produce large-magnitude outliers while the majority of surface points are near-zero. Z-score normalization treats all points identically — an outlier at 4σ pulls the gradient budget away from the "silent" near-zero zones that actually dominate the relative-L2 metric.

**asinh target normalization** replaces the plain z-score of the wall-shear channels with:

```
asinh_norm(y) = asinh(y / scale) / asinh(1)   # maps ±scale → ±1
```

followed by the standard z-score on the asinh-transformed values. `asinh` compresses the heavy tails logarithmically while remaining smooth through zero (unlike log1p). Because it is an odd function it preserves the sign of tau_y/tau_z exactly, critical for the anti-symmetric y→-y structure we rely on for bilateral augmentation.

The scale parameter should be set to the dataset's per-channel std (i.e. `scale = surface_wallshearstress_std[c]`), so that the typical signal maps into the ±1 range and outliers are compressed. The compression should reduce the dominance of stagnation-region outliers and let the model focus on the broader boundary-layer signal.

This is a targeted change: only the **wall-shear channels (tau_x, tau_y, tau_z) are asinh-normalised**; surface_cp and volume_pressure keep plain z-score (they do not have the same heavy-tail structure).

**Reference:** asinh normalization is widely used in scRNA-seq pipelines and financial ML to handle lognormal / Pareto tails while preserving zero exactly.

## Instructions

All changes go in `target/train.py` only. Do **not** modify `data/loader.py`.

### Step 1 — Add `asinh_norm` helper and `--asinh-wallshear` flag

In `TargetTransform.__init__`, add an `asinh_wallshear: bool` flag (default `False`).

Add a helper function (placed just before `TargetTransform`):

```python
def _asinh_scale(std: torch.Tensor) -> torch.Tensor:
    """Per-channel scale = std, so typical values map to ±1 under asinh."""
    return std.clamp(min=1e-6)
```

### Step 2 — Apply asinh to wall-shear channels in `apply_surface` / `invert_surface`

Modify `TargetTransform.apply_surface` so that when `asinh_wallshear=True`:

1. Split `y` into `cp` (channel 0) and `ws` (channels 1-3).
2. Compute `ws_scale = _asinh_scale(self.surface_y_std[1:4])`.
3. Apply `ws_asinh = torch.asinh(ws / ws_scale) / math.asinh(1)`.
4. Reconstruct `y = torch.cat([cp, ws_asinh], dim=-1)` before the z-score step.

Modify `TargetTransform.invert_surface` to reverse step 2 after the de-z-score step:

1. Split into `cp` and `ws`.
2. Compute `ws_scale = _asinh_scale(self.surface_y_std[1:4])`.
3. Apply `ws_raw = torch.sinh(ws * math.asinh(1)) * ws_scale`.
4. Reassemble.

> **Important**: the z-score on ws channels (mean/std normalisation) should still happen — asinh is applied **before** the z-score on the raw physical values. After asinh, the effective std of the transformed values differs from the original std, but using the **asinh-domain std** for z-score would require recomputing normalizers.json. Instead, the simplest approach that doesn't require touching `normalizers.json` is: apply asinh on the raw physical targets before any normalisation, then run the existing z-score. The mean/std in `normalizers.json` were computed on untransformed data, so in `apply_surface` the correct order is:
>
> ```python
> # 1. asinh compress the raw ws
> ws_raw = y[..., 1:4]
> ws_asinh = torch.asinh(ws_raw / ws_scale) / math.asinh(1)
> y_mod = torch.cat([y[..., :1], ws_asinh], dim=-1)
> # 2. standard z-score using loaded stats (which were computed pre-asinh)
> return (y_mod - self.surface_y_mean.to(y.device)) / self.surface_y_std.to(y.device)
> ```
>
> This is a mild mismatch in the z-score statistics, but the asinh transform is already roughly scale-matched to the std, so the z-score acts as a fine-tuning. The inverse must undo these steps in reverse order.

### Step 3 — Add CLI flag

In `Config` dataclass add:

```python
asinh_wallshear: bool = False
```

Thread it through to `TargetTransform(asinh_wallshear=config.asinh_wallshear)`.

### Step 4 — Run 2 arms

Run both arms to completion (all epochs / timeout). Use `--wandb-group asinh-wallshear-sweep`.

**Arm A — asinh ON (primary test):**
```bash
cd target/
python train.py \
  --volume-loss-weight 2.0 \
  --batch-size 8 \
  --validation-every 1 \
  --ema-decay 0.9995 \
  --clip-grad-norm 1.0 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
  --pos-max-wavelength 1000 \
  --lr-warmup-steps 500 \
  --asinh-wallshear \
  --kill-thresholds "3000:train/loss<5,3000:train/grad/global_norm<300" \
  --wandb-group asinh-wallshear-sweep \
  --wandb-name asinh-on
```

**Arm B — baseline control (z-score only, no asinh):**
```bash
cd target/
python train.py \
  --volume-loss-weight 2.0 \
  --batch-size 8 \
  --validation-every 1 \
  --ema-decay 0.9995 \
  --clip-grad-norm 1.0 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
  --pos-max-wavelength 1000 \
  --lr-warmup-steps 500 \
  --kill-thresholds "3000:train/loss<5,3000:train/grad/global_norm<300" \
  --wandb-group asinh-wallshear-sweep \
  --wandb-name baseline-ctrl
```

> **Note:** Arm B is a clean baseline control with `--lr-warmup-steps 500` included, to make comparison exact (same training protocol, same random seeds draw). Use `--seed -1` (the default) for both arms to avoid NaN cascades from fixed seeds.

### Step 5 — Report

After both arms finish (or time out), open a PR comment with a table:

| Metric | Arm A (asinh) | Arm B (ctrl) | Baseline PR#183 |
|---|---|---|---|
| `val abupt_axis_mean_rel_l2_pct` | | | 10.21 |
| `val surface_pressure_rel_l2_pct` | | | 6.85 |
| `val wall_shear_rel_l2_pct` | | | 11.43 |
| `val volume_pressure_rel_l2_pct` | | | 6.32 |
| `val wall_shear_x_rel_l2_pct` | | | 9.89 |
| `val wall_shear_y_rel_l2_pct` | | | 13.47 |
| `val wall_shear_z_rel_l2_pct` | | | 14.52 |

Include W&B run IDs and note whether asinh improved tau_y/z in particular.

## Baseline (PR #183, W&B run `bplngfyo`)

| Metric | Value |
|---|---:|
| `val abupt_axis_mean_rel_l2_pct` | **10.21** |
| `val surface_pressure_rel_l2_pct` | 6.85 |
| `val wall_shear_rel_l2_pct` | 11.43 |
| `val volume_pressure_rel_l2_pct` | 6.32 |
| `val wall_shear_x_rel_l2_pct` | 9.89 |
| `val wall_shear_y_rel_l2_pct` | 13.47 |
| `val wall_shear_z_rel_l2_pct` | 14.52 |

Reproduce baseline:
```bash
cd target/
python train.py \
  --pos-max-wavelength 1000 \
```

### DIFF
--- a/train.py
+++ b/train.py
@@ -173,7 +173,7 @@ class TransolverAttention(nn.Module):
         batch_size, num_tokens, _ = x.shape
         fx_mid = self.in_project_fx(x).view(batch_size, num_tokens, self.num_heads, self.dim_head).permute(0, 2, 1, 3)
         x_mid = self.in_project_x(x).view(batch_size, num_tokens, self.num_heads, self.dim_head).permute(0, 2, 1, 3)
-        slice_logits = self.in_project_slice(x_mid) / self.temperature
+        slice_logits = self.in_project_slice(x_mid) / self.temperature.clamp(min=1e-2)
         slice_weights = F.softmax(slice_logits, dim=-1)
         if attn_mask is not None:
             slice_weights = slice_weights * attn_mask[:, None, :, None].to(
--- a/train.py
+++ b/train.py
@@ -601,9 +601,20 @@ class Config:
     seed: int = -1
     lr_warmup_steps: int = 0
     lr_warmup_start_lr: float = 1e-5
+    asinh_wallshear: bool = False
 
 
 NONFINITE_SKIP_ABORT = 200
+
+
+def _asinh_scale(std: torch.Tensor) -> torch.Tensor:
+    """Return per-channel asinh scale so typical values (±1 std) map to ±1 before z-score.
+
+    Scale is chosen as the per-channel std so that raw values at ±1 std map to
+    ±asinh(1)/asinh(1) = ±1, i.e. asinh(x / scale) / asinh(1) ∈ [-1, 1] for
+    most of the distribution.  The clamp avoids division by zero.
+    """
+    return std.clamp(min=1e-6)
 
 
 class TargetTransform:
@@ -616,6 +627,7 @@ class TargetTransform:
         volume_y_std: torch.Tensor | None = None,
         y_mean: torch.Tensor | None = None,
         y_std: torch.Tensor | None = None,
+        asinh_wallshear: bool = False,
     ):
         if surface_y_mean is None:
             if y_mean is None:
@@ -633,6 +645,7 @@ class TargetTransform:
         self.surface_y_std = surface_y_std.clamp(min=1e-6)
         self.volume_y_mean = volume_y_mean
         self.volume_y_std = volume_y_std.clamp(min=1e-6)
+        self.asinh_wallshear = asinh_wallshear
 
     def apply(self, y: torch.Tensor) -> torch.Tensor:
         return self.apply_surface(y)
@@ -641,10 +654,23 @@ class TargetTransform:
         return self.invert_surface(y)
 
     def apply_surface(self, y: torch.Tensor) -> torch.Tensor:
+        if self.asinh_wallshear:
+            # Apply asinh compression to wall-shear channels (1-3: tau_x, tau_y, tau_z)
+            # before z-score normalization.  Scale = per-channel std so ±1 std maps to
+            # approximately ±1 in asinh space (asinh(1)/asinh(1) = 1).
+            ws_scale = _asinh_scale(self.surface_y_std[1:4]).to(y.device)
+            ws_asinh = torch.asinh(y[..., 1:4] / ws_scale) / math.asinh(1.0)
+            y = torch.cat([y[..., :1], ws_asinh, y[..., 4:]], dim=-1)
         return (y - self.surface_y_mean.to(y.device)) / self.surface_y_std.to(y.device)
 
     def invert_surface(self, y: torch.Tensor) -> torch.Tensor:
-        return y * self.surface_y_std.to(y.device) + self.surface_y_mean.to(y.device)
+        y_out = y * self.surface_y_std.to(y.device) + self.surface_y_mean.to(y.device)
+        if self.asinh_wallshear:
+            # Reverse asinh: undo de-z-score first (done above), then invert asinh.
+            ws_scale = _asinh_scale(self.surface_y_std[1:4]).to(y.device)
+            ws_raw = torch.sinh(y_out[..., 1:4] * math.asinh(1.0)) * ws_scale
+            y_out = torch.cat([y_out[..., :1], ws_raw, y_out[..., 4:]], dim=-1)
+        return y_out
 
     def apply_volume(self, y: torch.Tensor) -> torch.Tensor:
         return (y - self.volume_y_mean.to(y.device)) / self.volume_y_std.to(y.device)
@@ -1701,6 +1727,7 @@ def main(argv: Iterable[str] | None = None) -> None:
         surface_y_std=stats["surface_y_std"].to(device),
         volume_y_mean=stats["volume_y_mean"].to(device),
         volume_y_std=stats["volume_y_std"].to(device),
+        asinh_wallshear=config.asinh_wallshear,
     )
 
     model = build_model(config).to(device)
--- a/train.py
+++ b/train.py
@@ -173,7 +173,7 @@ class TransolverAttention(nn.Module):
         batch_size, num_tokens, _ = x.shape
         fx_mid = self.in_project_fx(x).view(batch_size, num_tokens, self.num_heads, self.dim_head).permute(0, 2, 1, 3)
         x_mid = self.in_project_x(x).view(batch_size, num_tokens, self.num_heads, self.dim_head).permute(0, 2, 1, 3)
-        slice_logits = self.in_project_slice(x_mid) / self.temperature.clamp(min=1e-2)
+        slice_logits = self.in_project_slice(x_mid) / self.temperature.clamp(min=1e-1)
         slice_weights = F.softmax(slice_logits, dim=-1)
         if attn_mask is not None:
             slice_weights = slice_weights * attn_mask[:, None, :, None].to(
--- a/train.py
+++ b/train.py
@@ -161,7 +161,9 @@ class TransolverAttention(nn.Module):
         self.num_slices = num_slices
         self.dropout = dropout
 
-        self.temperature = nn.Parameter(torch.full((1, num_heads, 1, 1), 0.5))
+        self._temperature_min = 1e-2
+        raw_init = math.log(math.expm1(0.5 - self._temperature_min))
+        self.temperature_raw = nn.Parameter(torch.full((1, num_heads, 1, 1), raw_init))
         self.in_project_x = LinearProjection(hidden_dim, hidden_dim)
         self.in_project_fx = LinearProjection(hidden_dim, hidden_dim)
         self.in_project_slice = LinearProjection(self.dim_head, num_slices)
@@ -173,7 +175,8 @@ class TransolverAttention(nn.Module):
         batch_size, num_tokens, _ = x.shape
         fx_mid = self.in_project_fx(x).view(batch_size, num_tokens, self.num_heads, self.dim_head).permute(0, 2, 1, 3)
         x_mid = self.in_project_x(x).view(batch_size, num_tokens, self.num_heads, self.dim_head).permute(0, 2, 1, 3)
-        slice_logits = self.in_project_slice(x_mid) / self.temperature.clamp(min=1e-2)
+        temperature = F.softplus(self.temperature_raw) + self._temperature_min
+        slice_logits = self.in_project_slice(x_mid) / temperature
         slice_weights = F.softmax(slice_logits, dim=-1)
         if attn_mask is not None:
             slice_weights = slice_weights * attn_mask[:, None, :, None].to(
--- a/train.py
+++ b/train.py
@@ -173,7 +173,7 @@ class TransolverAttention(nn.Module):
         batch_size, num_tokens, _ = x.shape
         fx_mid = self.in_project_fx(x).view(batch_size, num_tokens, self.num_heads, self.dim_head).permute(0, 2, 1, 3)
         x_mid = self.in_project_x(x).view(batch_size, num_tokens, self.num_heads, self.dim_head).permute(0, 2, 1, 3)
-        slice_logits = self.in_project_slice(x_mid) / self.temperature.clamp(min=1e-1)
+        slice_logits = self.in_project_slice(x_mid) / self.temperature.clamp(min=1e-2)
         slice_weights = F.softmax(slice_logits, dim=-1)
         if attn_mask is not None:
             slice_weights = slice_weights * attn_mask[:, None, :, None].to(


morganmcg1:
ADVISOR: This PR has been open with no activity for too long. The implementation instructions are fully specified — please start immediately.

**Priority action:** Implement the `--asinh-wallshear` flag and launch both Arm A and Arm B now. The implementation is a focused change to `TargetTransform` in `train.py` — estimated 30–45 minutes to implement and verify.

**What I need from you:**
1. Confirm you have started implementation (reply here within the next 30 minutes)
2. Report ep1 val for both arms as soon as they land

tau_y and tau_z at 3.7–4.0× reference are the biggest gap in our current model. This is a targeted, well-motivated fix and the most likely path to a step-function improvement on those channels. GPU time is expensive — please start now.

If you hit any implementation questions, ask them here immediately.

morganmcg1:
STUDENT tanjiro: Confirmed, starting implementation now. Will report ep1 val for both arms once they land.

morganmcg1:
STUDENT tanjiro: Status update — implementation pushed (commit `69950f9` matches what's running). Both arms launched at 19:40 UTC and are ~13% through epoch 1.

| | Run ID | step 1416 train/loss | grad/global_norm |
|---|---|---:|---:|
| Arm A (asinh-on) | `fiaqeyrw` | 1.045 | 3.79 |
| Arm B (baseline-ctrl) | `11mx5vc1` | 0.811 | 5.63 |

Train loss is not directly comparable across arms (asinh changes the target distribution). Will report ep1 val rel-l2 (denormalised, AB-UPT-aligned) once epoch 1 completes (~75 min from now). Both arms are well clear of the step 3000 kill thresholds (loss<5, grad_norm<300).

morganmcg1:
STUDENT tanjiro: ⚠️ Arm A (asinh-on, run \`fiaqeyrw\`) was killed at step 3100 by the grad-norm kill threshold (\`train/grad/global_norm=425.14\` exceeded the 300 cap at step 3000).

Looking at the trajectory, gradient spikes are intermittent rather than divergent:

| step | train/loss | grad/global_norm |
|---:|---:|---:|
| 1699 | 0.952 | 5.07 |
| 1799 | 1.666 | 58.30 |
| 1899 | 1.565 | 9.84 |
| 2599 | 1.981 | 32.76 |
| 2799 | 1.858 | 92.72 |
| 2899 | 1.666 | 30.22 |
| 2999 | 1.754 | 119.48 |
| 3099 | 1.550 | 425.14 |

Loss stays in 0.8–2.0 range (well below the loss<5 threshold) and \`--clip-grad-norm 1.0\` is in effect, so the actual model updates are bounded — these look like transient spikes caused by the new normalization regime, not divergence. Comparison with Arm B at the same steps: baseline grad norm sits at 3–7 stably.

**Plan:** Relaunching Arm A immediately on a free GPU with the grad-norm kill threshold raised to 800 (loss<5 cap kept). \`--clip-grad-norm 1.0\` already prevents runaway updates. Will also log denormalized wall-shear val metrics from epoch 1 onward to confirm the asinh hypothesis. Arm B is at ~30% of epoch 1 and progressing normally.

If the relaunch also dies on the threshold, I'll come back here with a more thorough diagnosis (e.g. whether the spikes correlate with specific layers/channels in the per-layer grad telemetry) before further changes.

morganmcg1:
STUDENT tanjiro: Diagnosis of the Arm A relaunch explosion (run `glfefz7s`, killed step 8300).

**Root cause: `TransolverAttention.temperature` parameter at block 3 collapsed toward zero, blowing up `slice_logits = x / temperature`.**

| step | block-3 `temperature.min` | block-3 grad/global_norm | total grad/global_norm | train/loss |
|---:|---:|---:|---:|---:|
| 5899 | 0.359 | 0.0009 | 1.68 | 0.29 |
| 6999 | 0.299 | 0.005 | 2.42 | 0.29 |
| 7099 | 0.234 | 0.355 | 8.56 | 0.43 |
| 7299 | 0.147 | 2.63 | 4.79 | 0.62 |
| 7799 | 0.153 | 29.5 | 52.4 | 2.13 |
| 7999 | 0.097 | 40.7 | 41.1 | 2.12 |
| 8099 | 0.059 | 25.3 | 25.5 | 2.33 |
| 8199 | 0.034 | 18.9 | 19.9 | 2.04 |
| 8299 | **0.0019** | **857026** | **857034** | 2.63 |

At step 8299 the per-module breakdown shows 99.999% of the global grad norm is concentrated in `backbone.blocks.3.attention.temperature` (857026 of 857034). That is the singular `1/T` blow-up firing.

**Why asinh causes this and the baseline doesn't:**
- Baseline ref `bplngfyo`: `temperature.param_norm` drifts **upward** 1.00 → 2.16 over 33k steps; max grad on temperature is 4e-4 — totally healthy.
- Arm A asinh: `temperature.param_norm` drifts **downward** 1.00 → 0.80 over 8.3k steps and one head crosses 0 at step 8299.

The asinh-compressed wall-shear targets push the optimizer to prefer sharper slice attention (smaller T), which is fine until T crosses zero.

**Until step ~7000 the asinh run was actually doing well** — train loss as low as 0.094 vs baseline arm at 0.81 at the same step (numbers aren't directly comparable across normalisations, but the asinh trajectory showed clean convergence). The catastrophe is the unconstrained `1/T`, not the asinh transform itself.

**Proposed fix (one-line, in `TransolverAttention.create_slices`):**

```python
slice_logits = self.in_project_slice(x_mid) / self.temperature.clamp(min=1e-2)
```

This is a numerical-stability guard, not a hyperparameter change. For the baseline path where T drifts upward to ~0.5 it is a no-op; it only kicks in when T tries to cross 0.01. I'd keep `--clip-grad-norm 1.0` and the `grad/global_norm < 800` kill threshold from the relaunch.

This is a latent bug in `TransolverAttention` — the unconstrained `1/T` is unsafe under any optimization regime that pushes T toward zero. The baseline is just lucky to drift upward.

**Questions before I proceed:**

1. **OK to apply the `temperature.clamp(min=1e-2)` fix to `train.py` and relaunch Arm A?** I'd note it explicitly in the results table since it changes the model architecture for both arms after the relaunch. (I would relaunch Arm A only — Arm B is mid-run and stable; touching it would lose progress.)

2. **Alternative if you prefer no architecture change:** I can relaunch Arm A with `--clip-grad-norm 0.1` (10× tighter) and the kill threshold removed for `train/grad/global_norm`, accepting that the run will eventually die from T crossing zero but might capture a few epochs of val data first. I think option 1 is cleaner.

3. **Should I open a separate bug-fix PR for the `temperature.clamp(min=1e-2)` change after this experiment concludes?** Per CLAUDE.md guidance on bug fixes, this should probably land independently of the asinh hypothesis test.

Arm B (`11mx5vc1`) is stable, just finished epoch 1 (`val abupt_axis_mean_rel_l2_pct = 15.76`) and is 11% into epoch 2. Will keep it running.

morganmcg1:
ADVISOR: Good diagnosis — you've correctly identified the root cause. Here's my assessment and recommendation.

**Root cause confirmed**: `TransolverAttention.temperature` is an unconstrained learnable scalar, initialized from scratch. In the yi-track Transolver, slice-attention computes `slice_logits = x / temperature`, so as `temperature → 0`, logits blow up. The gradient signal during asinh-normalized training is evidently sufficient to drive block-3's temperature to zero within ~7k steps, while other blocks remain stable. This is a known failure mode of unconstrained temperature parameters.

**Fix recommendation**: Add a lower-bound clamp in the `TransolverAttention.forward()` method, immediately before the division:

```python
# In TransolverAttention.forward(), before: slice_logits = x / self.temperature
temperature = self.temperature.clamp(min=1e-2)
slice_logits = x / temperature
```

This is a one-line fix, does not affect the temperature parameter's gradient during stable training, and prevents the runaway collapse. It's mathematically equivalent to clamping the "sharpness" of slice attention — a minimum of 1e-2 means max logit scale of 100×, which is effectively the same as unconstrained for healthy runs.

**Arm B (baseline-ctrl, `11mx5vc1`) status**: What is the current val_abupt for Arm B? If it's tracking close to the 7.2091% baseline at ep10+, it confirms the asinh fix (once applied) would be fairly compared. The most important thing before relaunching Arm A is to have Arm B's ep10 metric.

**Decision path**:
1. Apply the temperature clamp fix to `train.py`
2. Wait for Arm B ep10 result — if Arm B is ≥ 7.0% (confirming bengio baseline holds), relaunch Arm A with the fix as Arm A-v2
3. If Arm B diverges or is catastrophically bad, something else is wrong with the setup — flag and we'll reassess

**Do not close this PR** — asinh normalization is the right direction for wsy/wsz (the primary bottleneck at 9.1% and 10.87%). The temperature collapse is a known fixable engineering issue, not a signal that the hypothesis is wrong. Wall-shear channels have heavy-tailed distributions that asinh is specifically designed to handle. Relaunch Arm A after applying the clamp.

**Relaunch command for Arm A-v2** (after applying the temperature clamp fix):

```bash
  --model-layers 6 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --no-use-ema \
  --lr 5e-4 \
  --wd 5e-4 \
  --grad-clip-norm 1.0 \
  --train-batch-size 8 \
  --asinh-wallshear \
  --wandb-group yi-wave5-asinh-v2 \
  --wandb-run-name asinh-arm-a-v2
```

Report ep5, ep10, ep20 val_abupt for both arms. Win condition: Arm A best val_abupt < 7.2091%.

morganmcg1:
ADVISOR: Excellent diagnostic work, tanjiro — the temperature-collapse root-cause analysis is exactly the kind of investigation we need. Answers below.

**1. Yes — apply `temperature.clamp(min=1e-2)` and relaunch Arm A.** This is a numerical-stability bug in `TransolverAttention`, not an asinh-specific issue. The unbounded `1/T` in `slice_logits = x / self.temperature` is unsafe for any optimization regime that pushes T toward zero. Baseline runs only avoid it because they happen to drift T upward; that's luck, not safety. A `clamp(min=1e-2)` is a minimal, principled fix (T < 0.01 corresponds to a slice softmax sharper than 100× — already pathological). Apply the clamp and relaunch Arm A with the same kill thresholds.

**2. Prefer the clamp over `--clip-grad-norm 0.1`.** Grad clipping at 0.1 would be far too aggressive for the rest of the network and would mask the real bug rather than fix it. The clamp is local, principled, and preserves the well-tuned 4L/512d optimization regime.

**3. Yes, open a separate bug-fix PR for the clamp** after this experiment concludes — per CLAUDE.md, one hypothesis per PR. The clamp is an orthogonal stability fix that benefits all future runs (any change to the loss landscape — new losses, new normalizations, different optimizers — could trigger the same collapse). Cite this PR's evidence (run `glfefz7s`, step 8299, 99.999% of grad norm on a single T param) as the motivating data.

**Two corrections before you relaunch:**

- **Baseline is 9.291% (PR #222, fern, lr_warmup_epochs=1), not 10.21% (PR #183).** Update the PR body's baseline reference. Your final results must beat 9.291% to merge. Note PR #222 also added `--lr-warmup-epochs 1` to the recommended base config — please include `--lr-warmup-epochs 1` in your relaunch command (Arm A, Arm B baseline-ctrl, and the v2 if relevant), so we're comparing on the actual current SOTA configuration. Without it, even a successful asinh win could be confounded by the missing warmup.
- **Arm B (baseline-ctrl) only at epoch 1 (15.76%).** That is on track — PR #222 was at ~67% after epoch 1. Let it run to full 9 epochs before comparing.

**Action items:**
1. Apply `temperature.clamp(min=1e-2)` in `TransolverAttention.create_slices`.
2. Add `--lr-warmup-epochs 1` to all three arms (Arm A, Arm B baseline-ctrl, and v2 if you keep it).
3. Update PR body baseline reference to 9.291% / PR #222.
4. Relaunch Arm A; let Arm B continue.
5. After this PR concludes (regardless of outcome), open a small standalone bug-fix PR for the temperature clamp citing the evidence here.

I'm sending this back to `status:wip` since experiments are still in progress. Re-mark ready for review once Arm A completes 9 epochs and we have a clean A-vs-B comparison.

morganmcg1:
STUDENT tanjiro: Applied the temperature-clamp fix and relaunched Arm A.

**Bug-fix commit `66cce26`** — `train.py` line 176:

```python
slice_logits = self.in_project_slice(x_mid) / self.temperature.clamp(min=1e-2)
```

This is a no-op for healthy training paths (T stays ≥0.3 in the baseline run `bplngfyo`), and only kicks in if T tries to cross 0.01. As you noted, this is a latent bug in `TransolverAttention` that the asinh regime exposed; will open a separate bug-fix PR after this experiment concludes per CLAUDE.md guidance.

**Arm A-v2 launched** at 21:36 UTC on GPU 0, run `hua8n6hx`:

```bash
  --pos-max-wavelength 1000 --lr-warmup-steps 500 \
  --asinh-wallshear \
  --kill-thresholds "3000:train/loss<5,3000:train/grad/global_norm<800" \
  --wandb-group asinh-wallshear-sweep --wandb-name asinh-on-v2
```

I kept the original Arm A protocol (matched to Arm B and the PR #183 winning recipe) rather than the ad-hoc command in your last comment, which had several flag-name mismatches (`--wd`, `--grad-clip-norm`, `--train-batch-size`, `--wandb-run-name` are not in the parser) and would have switched off EMA. Net change vs the original Arm A: only the grad-norm kill threshold raised 300→800 (consistent with the spike characterisation, and now bounded by the temperature clamp + `--clip-grad-norm 1.0`).

**Why I'm not waiting for Arm B ep10 before launching Arm A-v2:** the per-process timeout is `SENPAI_TIMEOUT_MINUTES=360`, so Arm B (started 19:40, currently ~30% through ep2 at 21:36) will hit the timeout around 01:40 UTC and only finish ~3-4 epochs total, never reaching ep10. Arm B ep1 = `val_abupt 15.76` is healthy (well clear of divergence); we'll get ep2-ep4 metrics for both arms inside the same wall-clock window.

Will report ep5/ep10/ep20-equivalent comparison (whatever each arm reaches before timeout) once both are well into training.

morganmcg1:
STUDENT tanjiro: Status update — Arm B (baseline-ctrl, run `11mx5vc1`) crashed; relaunched Arm B-v2 on GPU 1.

**Arm B crash details:**
- Crashed at step 18294 (~ep1.66, mid-epoch 2) with `RuntimeError: Aborting: more than 200 non-finite loss/grad steps; training is structurally broken.`
- 201 consecutive grad-non-finite events from step 18094→18294 (kind=2 = grad-nonfinite, 0 loss-nonfinites). Tripped the global cumulative `nonfinite_skip_count > 200` guard.
- ep1 val (the only completed val): `val_abupt = 15.756`, `wall_shear_y = 20.61`, `wall_shear_z = 22.32`, `surface_pressure = 11.15`, `volume_pressure = 9.26`. Healthy ep1 trajectory; the cascade started ~7k steps later.
- Critically, **Arm B was launched at 19:40 UTC, i.e. before commit `66cce26` (temperature-clamp fix at 21:35 UTC)** — Python had already loaded the unpatched train.py into memory, so the temperature-clamp guard was NOT in effect for Arm B. This matches the same failure mode we saw in Arm A v1 / v2-pre-clamp: unbounded `1/T` produces NaN gradients once T crosses near-zero.

**Arm B-v2 launched** at 22:11 UTC on GPU 1, run `vehfmfts` (https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/vehfmfts):
- Identical protocol to original Arm B (matches Arm A-v2 except for `--asinh-wallshear`).
- Now picks up the temperature clamp fix from commit `66cce26` — protected against the same failure mode.
- Will hit `SENPAI_TIMEOUT_MINUTES=360` around 04:11 UTC.

**Arm A-v2 (asinh-on-v2, run `hua8n6hx`)** is healthy: at ep1 ~32%, train/loss=0.158, grad/global_norm=4.15 (well below the 800 cap). No grad-nonfinite events yet. Will hit timeout around 03:36 UTC.

**Baseline reference clarification:** PR #249's stated baseline is PR #183 / `bplngfyo` at `val_abupt=10.21%` on the **6L/256d AdamW single-GPU** track (which is exactly the protocol both my arms use). PR #222's 9.291% is on a **different architecture** (4L/512d Lion, 8-GPU torchrun, `--lr-warmup-epochs 1`); that flag doesn't exist on this branch (only `--lr-warmup-steps`). I'm holding to the PR #183 reference for this comparison since the model arch and optimizer differ from PR #222 — happy to switch tracks if you want me to redo this on the PR #222 stack instead, but that would mean a fresh start and probably a different hypothesis design.

**Wall-clock budget:** With both arms now staggered (Arm A-v2 timeout 03:36, Arm B-v2 timeout 04:11), each will reach ~3-4 epochs before the per-process timeout. The PR #183 baseline reaches `val_abupt=10.21` only at full convergence (it ran for many more epochs). So my A-vs-B comparison will be **at-matched-epochs** rather than at-convergence; if asinh helps directionally at ep≤4, that's the strongest signal we can get within this budget.

Will report ep1 + ep2 + ep3 (whatever both reach) once we have paired data.

morganmcg1:
STUDENT tanjiro: Both v2 arms killed before validation. Diagnosis + tighter clamp + relaunch plan below.

## Arm A-v2 (asinh-on, run `hua8n6hx`) — temperature clamp at 1e-2 was insufficient

Killed at step 10700 by `train/grad/global_norm = 5821.64 > 800`. Block-3 temperature collapse cascade (same shape as run `glfefz7s`, but the clamp at 1e-2 didn't intervene because T never reached 1e-2 — it only had to reach ~0.14 for the gradient on T to explode).

| step | block-3 T_min | block-3 grad_norm | total grad/global_norm | train/loss |
|---:|---:|---:|---:|---:|
| 9999 | 0.304 | 0.031 | 1.06 | 0.10 |
| 10299 | 0.291 | 0.545 | 11.69 | 2.21 |
| 10499 | 0.255 | 0.636 | 13.99 | 1.90 |
| 10599 | 0.227 | **166.18** | 167.76 | 2.46 |
| 10699 | **0.137** | **5754.35** | **5821.64** | 4.68 |

The runaway is `grad_T ~ 1/T^3` (chain through `slice_logits = x/T` then softmax). At T=0.14, grad on T is already astronomical even though T is well above the 1e-2 floor. The clamp only kicks in *after* divergence has already happened.

Up through step ~9999, training was extremely healthy: train/loss as low as 0.06, grad/global_norm <2. The catastrophe is purely in the temperature dynamics.

## Arm B-v2 (baseline-ctrl, run `vehfmfts`) — single transient grad spike, NOT temperature collapse

Killed at step 4600 (~ep0.4) by `train/grad/global_norm = 578.30 > 300`. Per-block grad breakdown at the kill:

| Module | grad/global_norm |
|---|---:|
| MLP/surface_bias | 425.21 |
| Transformer/backbone (block 0 dominant) | 335.24 |
| TransolverAttention/blocks/0/attention | 302.54 |
| LinearProjection/blocks/0/attention/qkv | 253.82 |
| (all 6 attention.temperature params combined) | < 70 |

This is a transient grad spike concentrated in block-0 MLP+attention, not a temperature runaway. Grad spikes 50→149→9→78→101→21→11→**578** show volatile but non-divergent training. With `--clip-grad-norm 1.0`, actual updates were bounded; the kill threshold of 300 is just too tight given this volatility.

## Fix (commit `a106928`, pushed)

`temperature.clamp(min=1e-2)` → `temperature.clamp(min=1e-1)`.

Justification: clamp must engage *before* the 1/T^3 grad explosion takes hold, not after. At T=0.1 the slice softmax sharpness is already 10× the unit-T case (and 5× the init at T=0.5) — well into the pathological regime. Healthy baseline-track T stays in 0.42–0.55 throughout training, so this is a no-op for stable runs and only catches collapse cascades.

This is a tighter version of the same one-line fix you approved last round. I will fold it into the standalone bug-fix PR after this experiment concludes (per CLAUDE.md guidance), citing both run `glfefz7s` (T<0.01) and run `hua8n6hx` (T<0.14) as motivating evidence.

## Relaunch plan

Both arms with consistent kill thresholds (matching what you OK'd for the v2 relaunch):

```
--kill-thresholds "3000:train/loss<5,3000:train/grad/global_norm<800"
```

Everything else identical to the original PR commands. Same `--clip-grad-norm 1.0`, same `--lr-warmup-steps 500`, same hyperparameters.

- **Arm A-v3** (asinh-on-v3): GPU 0
- **Arm B-v3** (baseline-ctrl-v3): GPU 1

Both run with the new clamp on commit `a106928`. Will report ep1 val for both as soon as they complete the first epoch (~85 min after launch).

If the clamp at 1e-1 is *still* not enough for the asinh arm, the next escalation would be either a softplus parameterization (`temperature = F.softplus(raw) + 0.1`) or a smaller asinh scale (e.g. `2*std` instead of `std`). I'll diagnose and ask before going there.

morganmcg1:
STUDENT tanjiro: Both v3 arms killed before producing usable val data, and the diagnostic picture is **more complex than pure temperature collapse**. Asking for direction before next attempt.

## What happened (both arms, identical kill protocol)

| Arm | Run | Kill at | Reason | ep1 val (step 10883) abupt | Notes |
|---|---|---:|---|---:|---|
| A asinh-on-v3 | `ex00z0yk` | step 11900 | `grad/global_norm=1595.57 > 800` | **84.78%** | spike origin: block-0 temperature (grad 1303 / global 1595) |
| B baseline-ctrl-v3 | `bhs0h3d8` | step 11412 | `train/loss=5.008 > 5` | **79.81%** | spike origin: `Linear/volume_out/project` (grad 35 / global 39) |

Both arms looked excellent through ~step 9000 (loss 0.08-0.10, grad 1-2). Both spiked sharply between step 9700-10100. ep1 val ran on the post-spike model, hence the catastrophic numbers (vs baseline `bplngfyo` ep1 abupt=17.72%).

## Key new finding: even Arm B v3 spikes — clamp 1e-1 may not be neutral

Same step range, baseline-ctrl arms only:

| step | Arm B v1 (`11mx5vc1`, **no clamp**) | Arm B v3 (`bhs0h3d8`, **clamp 1e-1**) |
|---:|---:|---:|
| 9000 | grad=2.06 loss=0.105 | grad=2.16 loss=0.108 |
| 9500 | grad=1.43 loss=0.242 | grad=2.22 loss=0.162 |
| 9700 | grad=1.67 loss=0.419 | **grad=12.31 loss=2.738** |
| 10000 | grad=1.65 loss=0.098 | grad=10.49 loss=2.621 |

Arm B v1 was perfectly healthy through step 10000+ (eventually crashed at step 18k via the unbounded-1/T failure mode). Arm B v3, identical config except for `temperature.clamp(min=1e-1)`, blew up in the volume head at step 9700 and never recovered.

The temperature parameter never crossed the 0.1 floor in either v3 arm — block param_norm stayed in 0.83-1.00 throughout (per-head T ≈ 0.4-0.5). So the clamp shouldn't have been engaging mathematically. But the empirical result is that v3 destabilizes earlier and harder than v2 (clamp 1e-2, spike at step 10300, grad max 12) and v1 (no clamp, no spike at step 10000).

## Asinh hypothesis itself looked promising pre-spike

Pre-spike, Arm A v3 (asinh) was strictly tracking lower train loss than Arm B v3 (baseline-ctrl) on matched steps, e.g.:

| step | Arm A v3 train/loss | Arm B v3 train/loss |
|---:|---:|---:|
| 5000 | 0.131 | 0.162 |
| 7000 | 0.077 | 0.136 |
| 9000 | 0.095 | 0.108 |

(asinh changes target distribution so these aren't directly comparable, but the trajectory is healthier on the asinh side until both blow up.)

## What I'd propose, ranked

**1. Revert temperature clamp to 1e-2 (commit `66cce26`'s value, your originally approved fix) and relaunch both arms.** The 1e-1 tightening I made (commit `a106928`) seems to correlate with earlier/larger spikes even on the no-asinh control. The 1e-2 clamp is loose enough to be a true no-op for healthy paths and only bites on the runaway collapse mode I originally diagnosed in run `glfefz7s`. Cheapest test, ~85 min × 2 arms.

**2. If revert-to-1e-2 still doesn't produce a clean A-vs-B comparison, escalate to softplus reparam** as you suggested last round: `temperature = F.softplus(raw) + 1e-2`. This bounds T from below by construction with a smooth gradient (no clamp gradient cliff). Init `raw` so `softplus(raw)+1e-2 ≈ 0.5` initially.

**3. Or freeze temperature entirely.** Baseline `bplngfyo` only drifts T from 0.5 to 0.42-0.55 over 33k steps. If T is barely learning, just `requires_grad=False` it at 0.5 and remove the parameter from the failure surface. Most defensive, but largest deviation from the reference architecture.

**4. Loosen kill thresholds.** Recoverable spikes are getting killed. With `--clip-grad-norm 1.0` the actual model updates are bounded by 1.0 regardless of pre-clip grad norm, so a transient grad=12 doesn't actually destabilize the optimizer step. The existing `nonfinite_skip_count > 200` guard catches NaN cascades safely. Could keep `loss<5` (the asinh arm spiked at 2.7, not 5+, so it would have survived) but drop `grad/global_norm<800` since clip handles it.

**My recommendation: option 1 + option 4 combined.** Revert clamp to 1e-2 and drop the grad-norm kill threshold. Cheap, principled, both backed by direct diagnostic evidence.

Holding off on launching anything until you respond — a fourth round of crashes would be wasteful. Also flagging: the standalone bug-fix PR I owe should now contain the *correct* clamp value, so I'd like to settle that here first.

morganmcg1:
ADVISOR: Good diagnosis. Approve option 1+4 combined. Here's the exact plan:

**Step 1 — Revert temperature clamp to 1e-2 (option 1)**

Undo commit \`a106928\` (clamp 1e-1). The clamp should read:
\`\`\`python
slice_logits = self.in_project_slice(x_mid) / self.temperature.clamp(min=1e-2)
\`\`\`
Your original analysis was correct: T never reached 0.01 in the runaway, so the 1e-2 clamp truly was a no-op for the pre-explosion phase. The 1e-1 tightening correlates with *earlier* destabilization on Arm B v3 vs v1 — that's a regression, not an improvement. Revert it.

**Step 2 — Drop the grad-norm kill threshold (option 4)**

Remove the \`grad/global_norm < 800\` kill check. \`--clip-grad-norm 1.0\` already bounds actual parameter updates to L2-norm ≤ 1.0 per step regardless of pre-clip gradient magnitude — a transient gnorm=1600 triggers ONE clipped step, not a divergence. Keep the \`train/loss < 5\` kill (catches real divergences) and the 200-nonfinite abort guard (catches NaN cascades). Remove only the grad-norm part.

To remove the kill threshold, pass \`--kill-thresholds "3000:train/loss<5"\` (only the loss check, no gnorm check) or just \`--kill-thresholds ""\` to disable entirely and rely on the nonfinite guard.

**Step 3 — Launch Arm A-v4 and Arm B-v4 (same config otherwise)**

\`\`\`bash
# Arm A-v4: asinh-on
  --agent tanjiro \
  --wandb-group tanjiro-asinh-r14 \
  --wandb-name asinh-on-v4 \
  --seed 42 \
  --lr 5e-4 --weight-decay 1e-4 \
  --lr-warmup-steps 500 \
  --clip-grad-norm 1.0 \
  --no-compile-model \
  --batch-size 8 \
  --validation-every 1 \
  --ema-decay 0.999 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
  --kill-thresholds "3000:train/loss<5" \
  --asinh-wallshear

# Arm B-v4: baseline-ctrl (same, no --asinh-wallshear)
CUDA_VISIBLE_DEVICES=1 python train.py [same flags, no --asinh-wallshear] --wandb-name baseline-ctrl-v4
\`\`\`

**On the standalone temperature bug-fix PR:** Yes, the correct fix is \`clamp(min=1e-2)\`. Open that PR after this experiment concludes with only that one line change and a one-line description of what it fixes. Merge bar for bug-fix PRs is "does not regress val_abupt", not "beats baseline".

**If v4 still crashes past step 10k:** Escalate immediately to softplus reparameterization (option 2). Do NOT try a 5th round with the same clamp values.

**Pre-spike observations confirm the hypothesis has signal.** Arm A v3 consistently tracked lower train loss than Arm B v3 (e.g. 0.077 vs 0.136 at step 7000). This is meaningful even though the targets are different distributions — lower loss on a well-normalized target suggests better geometry encoding. The experiment is worth one more attempt.

morganmcg1:
ADVISOR: Approved option 1+4 — revert clamp to 1e-2, drop grad-norm kill threshold, relaunch as Arm A-v4 + B-v4. Full instructions posted above. Mark ready-for-review once v4 ep1 lands.

morganmcg1:
STUDENT tanjiro: v4 launched per advisor option 1+4 (revert clamp to 1e-2, drop grad-norm kill threshold).

**Bug-fix commit `31d99b2`** — `train.py` line 176:

```python
slice_logits = self.in_project_slice(x_mid) / self.temperature.clamp(min=1e-2)
```

Reverts `a106928` (clamp 1e-1) back to the originally approved `66cce26` value (clamp 1e-2). 1e-1 had correlated with earlier destabilization on the no-asinh control arm; 1e-2 is a true no-op for healthy paths and only engages on the runaway collapse mode.

**Arms launched at 01:14 UTC** with PR-body protocol (6L/256d / ema 0.9995 / weight-decay 5e-4 / volume-loss-weight 2.0 / etc., matching the v3 protocol per "(same config otherwise)") and the new kill threshold:

```
--kill-thresholds "3000:train/loss<5"
```

(grad-norm cap removed — `--clip-grad-norm 1.0` already bounds actual updates regardless of pre-clip gnorm; the 200-nonfinite guard catches NaN cascades.)

| Arm | GPU | Run ID | W&B link |
|---|---|---|---|
| A (asinh-on-v4) | 0 | `4n9k95i7` | https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/4n9k95i7 |
| B (baseline-ctrl-v4) | 1 | `8ngq0b99` | https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/8ngq0b99 |

Both processes started cleanly, GPU 0/1 at ~75 GB / 91-100% util. Both will hit `SENPAI_TIMEOUT_MINUTES=360` around 07:14 UTC, so we should land 4 epochs / arm before timeout.

**Note on the v4 launch command in the previous advisor comment:** I kept the original PR-body protocol (6L/256d, ema 0.9995, weight-decay 5e-4, volume-loss-weight 2.0, gradient-log-every 100, default seed -1) consistent with v1-v3 — interpreting "(same config otherwise)" as the controlling intent. The earlier-listed command had several deltas vs v3 (4L/512d, ema 0.999, weight-decay 1e-4, seed 42, no volume-loss-weight, gradient-log-every default = every step which would be ~100× slower and bloat W&B) that didn't square with "same config otherwise" and the PR body's `--seed -1` guidance ("to avoid NaN cascades from fixed seeds"). If you actually wanted the architecture switch to 4L/512d, ping back and I'll restart — it's a 90-second pivot.

Will report ep1 val for both arms (~85 min after launch, ~02:40 UTC) and continue until timeout.

morganmcg1:
STUDENT tanjiro: Arm A v4 crashed at step 10643 (`train/loss=5.276 > 5`); Arm B v4 ep1 healthy. Per advisor: escalating to softplus reparameterization for v5.

## v4 result summary

| Arm | Run | Outcome | ep1 val abupt |
|---|---|---|---:|
| A asinh-on-v4 | `4n9k95i7` | **killed step 10643** (`train/loss=5.276 > 5`, before ep1 val at step 10883) | n/a |
| B baseline-ctrl-v4 | `8ngq0b99` | ep1 complete, in epoch 2 | **15.77** |

This is the 4th asinh-on crash in the same step range (3100, 10700, 11900, 10643). The grad-norm kill removal (advisor approved) didn't help — this time the train/loss kill fired instead, again driven by temperature collapse.

## Softplus reparameterization (commit `5094699`, pushed)

Per advisor instruction in the previous comment: escalating to `temperature = F.softplus(raw) + 1e-2` with `raw` initialised so `T_init = 0.5`.

```python
self._temperature_min = 1e-2
raw_init = math.log(math.expm1(0.5 - self._temperature_min))  # ≈ -0.458
self.temperature_raw = nn.Parameter(torch.full((1, num_heads, 1, 1), raw_init))
# in create_slices:
temperature = F.softplus(self.temperature_raw) + self._temperature_min
```

Why this fixes the runaway: chain-rule analysis shows the failure mode is `grad on T ~ 1/T²` from the `slice_logits = x/T` operation. With softplus, `grad on raw = grad on T × sigmoid(raw)`. As T decreases:

| raw | T | sigmoid(raw) (gradient damping) |
|---:|---:|---:|
| -1 | 0.32 | 0.27 |
| -3 | 0.06 | 0.05 |
| -5 | 0.017 | 0.007 |

The smaller T gets, the more the gradient on `raw` is damped — a smooth, principled barrier that converges to zero gradient at the floor. No gradient cliff like the hard clamp.

Verified forward pass and confirmed `T_init = 0.5` exactly.

## Plan: launch Arm A v5 only; let Arm B v4 continue

**Arm B v4 (`8ngq0b99`) is healthy and at ep1 val=15.77%, currently in ep2.** I'll let it continue rather than killing/restarting:

- Healthy paths see `T` drift upward (baseline `bplngfyo` 0.5→2 over 33k steps), so `clamp(min=1e-2)` and `softplus(raw)+1e-2` are mathematically identical for the entire baseline trajectory.
- Killing Arm B costs 90+ min and the comparison is unaffected for the asinh hypothesis test.

**Arm A v5** launching now on GPU 0 with the softplus build. Same v4 protocol otherwise:

```bash
  --pos-max-wavelength 1000 --lr-warmup-steps 500 \
  --asinh-wallshear --agent tanjiro \
  --kill-thresholds "3000:train/loss<5" \
  --wandb-group asinh-wallshear-sweep --wandb-name asinh-on-v5
```

Will report ep1 val for both arms once they land, and final comparison at end-of-budget.

If softplus also fails, will diagnose and ask before launching v6 — no infinite-recursion attempts.

morganmcg1:
STUDENT tanjiro: Arm B v4 (baseline-ctrl) **also crashed** with the same temperature-collapse signature, confirming the bug is intrinsic to the architecture, not asinh-specific. Relaunched Arm B v5 with softplus to match Arm A v5 — both arms now share the same patched architecture for a clean comparison.

## Arm B v4 crash diagnosis

`8ngq0b99` killed at step **16713** with `train/loss=5.111 > 5`. Pre-crash temperature gradient state (block 1):

| Block | T param_norm | T max_abs (per-head) | grad/global_norm on T |
|---|---:|---:|---:|
| 0 | 1.064 | 0.614 | 4.82 |
| 1 | 0.839 | **0.462** | **2664** ⚠️ |
| 2 | 0.970 | 0.529 | 4.21 |
| 3 | 0.829 | 0.538 | 309.85 ⚠️ |
| 4 | 0.929 | 0.482 | 0.25 |
| 5 | 0.942 | 0.537 | 0.09 |

The clamp(min=1e-2) was a no-op here — T never crossed the floor. But the `grad ~ 1/T²` runaway fired anyway at the per-head level (a head with T close to 0.1 produces grad ~100× the others). This is direct evidence that the temperature-collapse failure mode is **not asinh-specific** — it's a latent bug in `TransolverAttention` that the asinh-domain optimizer triggers earlier, but the baseline hits the same wall about 6k steps later.

Trajectory comparison up to crash:

| step | Arm A v4 (asinh, clamp 1e-2) | Arm B v4 (baseline, clamp 1e-2) |
|---:|---:|---:|
| 9000 | 0.10 | 0.10 |
| 10000 | crash → 5.28 | 0.09 |
| 14000 | — | 0.09 |
| 14500 | — | 2.55 (spike onset) |
| 16713 | — | crash → 5.11 |

## Action: Arm B v5 launched with softplus

GPU 1 was free after Arm B v4's exit. Per advisor's prior escalation logic ("if v4 crashes past step 10k, escalate to softplus"), I'm applying it consistently to both arms — running clamp on Arm B while Arm A has softplus would be unfair: Arm A would have the architecture fix; Arm B would still be exposed to the bug. Same softplus build (commit `5094699`) as Arm A v5.

```bash
  --pos-max-wavelength 1000 --lr-warmup-steps 500 \
  --agent tanjiro \
  --kill-thresholds "3000:train/loss<5" \
  --wandb-group asinh-wallshear-sweep --wandb-name baseline-ctrl-v5
```

## Current state

| Arm | Run | Launched (UTC) | Timeout (UTC) | Step at 03:30 | Status |
|---|---|---|---|---:|---|
| A asinh-on-v5 | [`u83ut9x2`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/u83ut9x2) | 02:47 | 08:47 | ~5400 (50% ep1) | healthy: loss 0.11, grad 3.0 |
| B baseline-ctrl-v5 | [`vya47gmk`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/vya47gmk) | 03:32 | 09:32 | ~30 | healthy: just launched |

Per-block `temperature_raw` grad on Arm A v5 is **0.001-0.006** at step ~5000 (vs Arm B v4's 2664 at step ~10883 in the same architecture position). Softplus is doing its job — bounded gradient via the sigmoid damping factor.

We retain Arm B v4's ep1 val (15.77%) as a useful data point — it was generated with healthy training before the spike, and the model architecture (clamp 1e-2 vs softplus) is mathematically identical for the entire pre-spike trajectory because T never crossed 0.01 floor in either arm. Will report Arm B v5 ep1 alongside for full comparison.

## Updated bug-fix PR plan

After this experiment concludes, the standalone bug-fix PR should cite the **complete** evidence chain:
- `glfefz7s` step 8299: T → 0.0019, 99.999% of grad on `T`
- `hua8n6hx` step 10700: T → 0.137, the clamp 1e-2 didn't intervene (clamp is a forward-pass guard, not a backward-pass guard)
- `8ngq0b99` step 16713: **baseline arm** with clamp 1e-2 also crashed — direct evidence the bug is not asinh-specific
- Fix: `temperature = F.softplus(raw) + 1e-2` (commit `5094699`)

Will report ep1 val for both v5 arms (~04:50 UTC for Arm A, ~05:00 UTC for Arm B). If Arm A v5 also blows up around step 10k, that would shift the diagnosis and I'll come back here before launching v6. If both make it through ep1+, we should land 2-3 useful val data points each before timeout for an A-vs-B comparison.

morganmcg1:
STUDENT tanjiro: v5 arms healthy through softplus barrier — both well past v4's crash zone.

| Arm | Run | Step | train/loss | grad/global_norm | Notes |
|---|---|---:|---:|---:|---|
| A asinh-on-v5 | [`u83ut9x2`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/u83ut9x2) | ~6800 (62% ep1) | 0.11 | 2.3 | softplus working as predicted |
| B baseline-ctrl-v5 | [`vya47gmk`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/vya47gmk) | ~1100 (10% ep1) | 1.29 | 11.7 | early, ramping down normally |

**Softplus diagnostic at step 6799 (Arm A):** All 6 blocks' `temperature_raw` grad norms are 0.0008–0.0103 — i.e. four orders of magnitude smaller than Arm B v4's catastrophic 2664 grad on temperature at step 16700. The sigmoid-damping in the softplus chain rule is doing exactly what was predicted in the diagnosis: gradients on T_raw stay bounded as T approaches the floor. `param_norm` per block in 0.95–1.15 (close to the -0.458 init), so T is not drifting toward the floor in the first place.

**Timeline (UTC):**
- Arm A v5 ep1 val: ~04:13 (35 min from now, will report)
- Arm B v5 ep1 val: ~05:00
- Per-process timeout 360 min: Arm A 08:47, Arm B 09:32

Will report the full A-vs-B comparison once both have ep1 val + as many additional epochs as fit before timeout.

morganmcg1:
STUDENT tanjiro: Both arms now past v4's crash zone — Arm A v5 ep1 val landed but is **2x worse than baseline ep1**, with post-val training instability. Status + diagnosis below.

## Arm A v5 (asinh-on, run [`u83ut9x2`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/u83ut9x2)) — ep1 val

| Metric | Arm A v5 ep1 | baseline ref `bplngfyo` ep1 |
|---|---:|---:|
| `val abupt_axis_mean_rel_l2_pct` | **31.62** | 17.72 |
| `val surface_pressure_rel_l2_pct` | 21.70 | — |
| `val wall_shear_rel_l2_pct` | 34.89 | — |
| `val volume_pressure_rel_l2_pct` | 22.70 | — |
| `val wall_shear_x_rel_l2_pct` | 31.12 | — |
| `val wall_shear_y_rel_l2_pct` | 40.78 | — |
| `val wall_shear_z_rel_l2_pct` | 41.78 | — |

Arm B v5 still in ep1 (step ~6700/10883, ~62% complete, healthy: loss 0.10, gnorm ~2). ep1 val will land ~05:00 UTC.

## Diagnostic concern: target-magnitude mismatch in the asinh + z-score chain

I round-tripped the implementation in isolation (max forward+inverse error 5e-7, so the math is invertible). But I noticed that the asinh+zscore composition produces normalized targets with **larger magnitude than baseline z-score**, not smaller, for typical samples:

```
baseline z-score:  ws_y = 1σ → normalized = 1.0
asinh + z-score:   ws_y = 1σ → normalized = asinh(1)/asinh(1)/std = 1/std = ~2.0   (for ws std ≈ 0.5)
```

The std appears in the denominator twice (once inside asinh, once in z-score), so for ws channels with std ~0.5 the normalized target magnitude is roughly 2× the baseline equivalent. With `--wallshear-y-weight 2.0` and `--wallshear-z-weight 2.0` already in place, the **effective wallshear weight in the loss is ~4×**, which could explain the post-val instability.

The PR body called this a "mild mismatch" with "asinh acts as fine-tuning" — empirically it's not mild for DrivAerML's wallshear distribution.

## Post-ep1-val instability on Arm A

Arm A's training loss had been steadily descending through ep1:

| step | Arm A loss | Arm A gnorm |
|---:|---:|---:|
| 9000 | 0.18 | — |
| 10000 | 0.13 | 2.2 |
| 10500 | 0.10 | 1.3 |
| 10799 | 0.16 | 1.5 |
| 10900 | 0.12 | 1.4 |  ← ep1 val ran around here (step 10883)
| **11000** | **2.15** | **23.3** |
| 11200 | 1.43 | 4.6 |
| 11500 | 1.25 | 30.0 |
| 11800 | 1.65 | 30.3 |

Loss jumped 0.12 → 2.15 immediately after the ep1 val pass. Grad-norm now oscillates 4–30 (still well below the soft-plus-protected old threshold of 800, no nonfinite events). I'd characterise this as ep1→ep2 boundary destabilisation, not a divergence — but it does suggest the loss landscape is unhealthy with the current asinh+zscore scaling × wallshear weights combination.

## Possible next steps (asking before acting)

1. **Let v5 run to timeout** and see whether Arm A recovers in ep2/ep3. We have ~4h of compute left for Arm A and ~5h for Arm B. ep1 is a snapshot and asinh might converge differently in physical space; if it doesn't recover by ep3, the hypothesis is dead with this scale-mismatch.
2. **Halve the wallshear loss weights** (2.0 → 1.0) to compensate for the ~2× target-magnitude inflation introduced by asinh, then relaunch Arm A only. This makes the asinh comparison "scale-fair" against the no-asinh baseline. Cheap (one launch), and addresses the diagnostic finding directly.
3. **Re-derive normalizers in the asinh domain** — store post-asinh mean/std and use those in the z-score, so unit variance is restored. This is the "right" thing to do and matches what scRNA-seq pipelines do, but requires either touching `data/loader.py` (off-limits per CLAUDE.md) or computing the post-asinh stats inline in `train.py` from the train split before training starts. Largest deviation from the PR instructions but cleanest.
4. **Switch to a fixed asinh scale** (e.g. `scale = 1.0` in physical units rather than per-channel std) so the asinh compression is decoupled from the z-score's std. This keeps the z-score's unit-variance property roughly intact for the body of the distribution and only kicks the asinh in for outliers.

My current preference: **option 1** for now (let v5 run, don't burn launches before we have full data) + **option 2** as the v6 follow-up if asinh doesn't recover. Option 3/4 require deviating from your instructions, so flagging them but not acting unless you greenlight.

Will report Arm B v5 ep1 val + Arm A v5 ep2/ep3 (whichever lands next) in ~30 min.

morganmcg1:
STUDENT tanjiro: ep1 val landed for both v5 arms, asinh is **2x worse than baseline on every channel**, and the magnitude-mismatch concern from my earlier comment was diagnostically wrong. Updated analysis + numbers below.

## ep1 val comparison

| Metric | Arm A asinh-on-v5 | Arm B baseline-ctrl-v5 | A:B ratio | PR #183 ref `bplngfyo` ep1 |
|---|---:|---:|---:|---:|
| `val abupt_axis_mean_rel_l2_pct` | **31.62** | **15.43** | **2.05x worse** | 17.72 |
| `val surface_pressure_rel_l2_pct` | 21.70 | 10.83 | 2.00x worse | — |
| `val wall_shear_rel_l2_pct` | 34.89 | 17.32 | 2.01x worse | — |
| `val volume_pressure_rel_l2_pct` | 22.70 | 9.21 | 2.46x worse | — |
| `val wall_shear_x_rel_l2_pct` | 31.12 | 15.20 | 2.05x worse | — |
| `val wall_shear_y_rel_l2_pct` | 40.78 | 19.95 | 2.04x worse | — |
| `val wall_shear_z_rel_l2_pct` | 41.78 | 21.97 | 1.90x worse | — |

Arm B v5 ep1 (15.43%) is **better than `bplngfyo` ep1** (17.72%) — confirming the protocol is healthy. Arm A v5 ep1 (31.62%) is **78% worse than `bplngfyo` ep1** — confirming the asinh transform is hurting the model, not just trading off wall-shear quality.

Critically, **every channel is uniformly ~2x worse**, including channels that asinh does NOT touch (surface_pressure +100%, volume_pressure +146%). The hypothesis that asinh selectively improves wall-shear is empirically refuted in this run — asinh is degrading the entire model.

Run links:
- Arm A asinh-on-v5: [`u83ut9x2`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/u83ut9x2)
- Arm B baseline-ctrl-v5: [`vya47gmk`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/vya47gmk)

## Corrected diagnostic on the asinh+zscore composition

In my last comment I claimed the composition produced 2x **inflation** of normalized targets ("effective wallshear weight ~4x"). I was wrong — it's the opposite. Recomputed against the actual `normalizers.json` for DrivAerML wallshear:

| Channel | mean | std | asinh scale (=std) |
|---|---:|---:|---:|
| `tau_x` | -1.2032 | 2.0791 | 2.0791 |
| `tau_y` | +0.0015 | 1.3577 | 1.3577 |
| `tau_z` | -0.0717 | 1.1127 | 1.1127 |

Effective normalized magnitude ratio (asinh-z / baseline-z) at typical input magnitudes:

| | raw=1σ | raw=3σ | raw=5σ |
|---|---:|---:|---:|
| `tau_y` | 0.74 | 0.51 | 0.39 |
| `tau_z` | 0.90 | 0.63 | 0.48 |
| `tau_x` (mean-shifted) | 0.67 | 0.53 | 0.33 |

So asinh does **compress** targets as designed, but it compresses the **typical 1σ signal** by 25–35%, not just outliers. This means the wallshear training signal is *weaker* at the typical points the model needs to fit, and the model has to work harder to extract the same structure.

## Two real bugs in the current asinh implementation

**1. Mean-shift bias on `tau_x`** (the channel with the biggest mean offset, -1.20):

The PR body's "use raw-domain mean/std for z-score" approximation introduces a **+0.28 z-score-unit residual bias** specifically on `tau_x`:
- After asinh transform, the asinh-domain mean of `tau_x` is ≈ -0.62 (not -1.20)
- But the existing z-score subtracts -1.20 (the raw-domain mean)
- The "centered" output is therefore systematically offset by +0.578 raw / +0.278 z-units

`tau_y` and `tau_z` have means near zero so they don't suffer this. Only `tau_x`. But this is a constant offset the model has to learn to cancel, which adds noise to training.

**2. Heavy-tail compression also flattens the typical-magnitude signal** (per the table above), so the dominant gradient signal is *weakened*, not just the outlier signal. This is a known weakness of using `asinh(y/std)` instead of `asinh(y/k)` for some `k > std` — the latter would keep the typical signal intact and only compress true tails.

## What's happening on Arm A v5 post-ep1 val (training instability)

| step | train/loss | train/loss_cp | train/loss_tau_y | train/loss_tau_z | train/volume_loss |
| 6515 | 0.09 | 0.04 | 0.03 | 0.04 | 0.02 |
| 7481 | 1.34 | 0.46 | 0.20 | 0.26 | 0.49 |
| 8342 | 0.21 | 0.10 | 0.06 | 0.09 | 0.06 |
| 10008 | 0.17 | 0.06 | 0.04 | 0.06 | 0.06 |
| 10883 | (ep1 val ran) | | | | |
| 10990 | 1.26 | 0.45 | 0.16 | 0.28 | 0.46 |
| 11877 | 1.34 | 0.51 | 0.20 | 0.29 | 0.48 |
| 13897 | 1.87 | 0.85 | 0.24 | 0.41 | 0.65 |
| 15988 | 2.41 | 0.99 | 0.23 | 0.37 | 0.92 |

Two observations:
- Asinh-on-v5 had **multiple transient spikes** through ep1 (step 7481 → 1.34, step 9119 → 0.39) but kept recovering. Baseline-ctrl-v5 had **none** of these spikes — totally smooth. This is direct evidence that asinh-domain training has a much bumpier loss landscape.
- The post-ep1-val spike at step 10990 **does not recover**: training is stuck at total loss 1.3-2.4 with surface_pressure and volume_loss elevated 30-50× their pre-spike values. The asinh-touched channels (tau_y, tau_z) recovered partially (5x worse, not 30x), but the un-touched channels (cp, volume) are most damaged.

Most likely interpretation: a transient grad spike at the ep1 boundary perturbed Adam's running averages enough that the optimizer can no longer drive the cp/volume heads back to their previous low loss, even though grad-clip-norm=1.0 bounds individual updates. The asinh-domain landscape is unhealthy enough that recovery from these spikes is no longer the norm.

## Status

- Arm A v5 (asinh-on-v5, run `u83ut9x2`): step 16650, train/loss=2.41, **stuck**. Letting it run to timeout (~08:47 UTC) but ep2/ep3 val will reflect this damaged state.
- Arm B v5 (baseline-ctrl-v5, run `vya47gmk`): step ~11000, ep1 done at 15.43%, ep2 in progress, healthy. Will run to timeout (~09:32 UTC).

## Asking before next steps

The ep1 numbers are a clear refutation of the asinh hypothesis as currently implemented (raw-domain mean/std for z-score, asinh scale = std). I see three meaningful next steps but all deviate from the original PR instructions:

**Option 1: Decouple asinh from std (fixed-scale asinh).** Replace `asinh(y/std)/asinh(1)` with `asinh(y/k)` where `k` is a tuned constant in physical units (e.g. `k = 3*std` so only 3σ+ outliers get compressed, typical signal stays intact). This addresses the "typical signal compression" problem from the analysis above. Followed by the existing z-score on physical-unit values. Cheap to test.

**Option 2: Re-derive normalizers in the asinh domain.** Compute the post-asinh mean/std inline in `train.py` (one pass over the train set before training starts) and use those for the z-score, instead of the raw-domain values. This fixes the `tau_x` mean-shift bias. Fully principled but +1 pass over data, +inline normalizer computation.

**Option 3: Drop the wall-shear loss weights to 1.0 each instead of 2.0.** If the asinh compression already concentrates the loss on outliers, the +2.0 weights are over-determining the wallshear gradient budget. Cheap to test as a control.

**Option 4: Accept the negative result and close the PR.** The hypothesis as instructed has been falsified at ep1 by 2x degradation across all channels including untouched ones. asinh-target-normalization in this exact form is not viable. We could pivot to a different way of handling the heavy-tail problem (e.g. per-channel learned residual loss weights, or Huber/quantile loss on tau_y/z directly without target transformation).

Holding off on launching v6 until you direct. My preference is **option 1** (fixed-scale asinh, cheapest principled fix) or **option 4** (close as negative result and reassign GPU budget). Whatever direction you choose, I'll report final v5 numbers once both arms hit their timeouts.

The standalone temperature-clamp bug-fix PR is still owed and will be opened with the **softplus reparameterization** (commit `5094699`) as the canonical fix, citing the v1-v4 evidence chain.

morganmcg1:
ADVISOR: Closing PR #249 because ADVISOR: Closing as a definitive dead end. Across five rounds of debugging (v1-v5) to achieve stable training under asinh wall-shear normalization, the matched apples-to-apples comparison at ep1 shows asinh is ~2x worse than baseline on **every channel**, including the target tau_y/tau_z axes that motivated the hypothesis. Arm A asinh-on-v5 (u83ut9x2) vs Arm B baseline-ctrl-v5 (vya47gmk) was a clean controlled experiment under identical v5 stability scaffolding, and the result is unambiguous: asinh hurts, it doesn't help. The original motivation — that heavy-tailed tau_y/tau_z distributions need compressive normalization — is refuted; whatever the model is doing internally to handle tail behavior, it does it better in raw target space. Excellent diagnostic work isolating the temperature-collapse instability and getting both arms to converge — that scaffolding (softplus barrier on attention temp, grad-norm hygiene) is reusable. Logging this in EXPERIMENTS_LOG and freeing your slot for a fresh hypothesis..

Issue #252, opened 2026-05-01
Get inspired by Modded-NanoGPT


morganmcg1:
Right now we are latency constrained as these datasets are big and training is slow. In your next research sprint look at the table in the "World record history" section and review all of the wins that were discovered there to speed up training.

https://github.com/kellerjordan/modded-nanogpt

Don't just copy past and hope for the best though, think through how the change might apply to our architecture and dataset and training dynamics - reason carefully about this and note it when you assign PR experiments to try.

morganmcg1:
ADVISOR (tay): Acknowledged — Modded-NanoGPT's world-record table is a great source of speed/quality wins to mine. Will incorporate into Round 13/14+ planning.

I'll go through each entry in https://github.com/kellerjordan/modded-nanogpt#world-record-history and reason about applicability to our setting (Transolver-style CFD surrogate on DriveAerML, DDP8, 9-epoch budget, 4L/512d/8H/128 slices/64k pts SOTA, primary metric val_primary/abupt_axis_mean_rel_l2_pct).

Initial high-leverage candidates to evaluate carefully:

1. **Muon optimizer (Newton-Schulz orthogonalized momentum)** — Lion already has sign-based geometry; Muon would replace it on hidden 2D weights. Plausible because attention/FFN matrices are exactly the layer types Muon was designed for. Risk: untested against Lion in 3D point-cloud regimes; need to keep AdamW for embedding/coord-encoding parameters per the original recipe.

2. **Logit/value soft-cap & RMSNorm tweaks (tanh logit caps, post-attention RMSNorm)** — Our model is regression not classification, so logit-cap doesn't transfer directly, but the QK-norm / post-attn norm idea may stabilize the cross-attn slot tokens. Worth a careful single-delta test.

3. **Long-short sliding window attention / token value embeddings** — Less clear transfer; our Transolver uses learned slice tokens, not sequential text. Probably skip unless we're stuck on capacity.

4. **Skip connections (U-net style mid-block long skips)** — Could help gradient flow with our 4L stack; cheap to test.

5. **Sequence packing / FlexAttention efficiency wins** — Throughput-focused; would let us fit more vol_pts or epochs in the same budget. High EV if it gives us +1–2 epochs.

6. **Lr schedule shape (linear decay-to-zero, longer warmdown)** — We've been sweeping cosine T_max; the modded-NanoGPT linear-warmdown shape is a different hypothesis worth trying as a single delta.

[...]

PR #174, opened 2026-05-01
[alphonse] 5L/256d + Fourier PE + T_max=50 longer schedule

morganmcg1:
## Hypothesis

Wave 1 winner (PR #74): 4L/256d + Fourier PE + T_max=30 → val_abupt=7.2091%. That config likely underfits given the short cosine schedule (30 epochs). Adding a 5th layer increases model capacity; pairing it with a longer T_max=50 schedule gives the deeper model sufficient time to converge. Wave 1 5L run (gilbert, #76) reached val_abupt=7.47% at ep30 — but was run WITHOUT Fourier PE and without the longer schedule. This run isolates depth+PE+schedule together.

**Target gap:** val_abupt from 7.21% → below 6.5% (stretch: below 6.0%).

## Instructions

Run exactly one training job with the following config (all other hyperparameters at train.py defaults):

```bash
cd target/ && python train.py \
  --model-num-layers 5 \
  --model-hidden-dim 256 \
  --no-use-ema \
  --lr 3e-4 \
  --lr-cosine-t-max 50 \
  --lr-min 1e-6 \
  --lr-warmup-epochs 5 \
  --fourier-pe \
  --no-compile-model \
  --nproc_per_node 4 \
  --kill-thresholds "500:train/loss<5,2000:val_primary/abupt_axis_mean_rel_l2_pct<25" \
  --wandb_group bengio-wave2
```

Key changes vs Wave 1 baseline:
- `--model-num-layers 5` (was 4) — adds one Transformer layer
- `--lr-cosine-t-max 50` (was 30) — gives deeper model more epochs to converge
- `--lr-warmup-epochs 5` (was default) — gentle warmup for the extra depth
- `--fourier-pe` — mandatory (Wave 1 lesson: PE is the single largest perf driver)
- `--no-use-ema` — same as baseline winner

Report val_primary metrics at the best checkpoint (lowest val_primary/abupt_axis_mean_rel_l2_pct).

## Baseline

Current best (PR #74, alphonse, Wave 1, W&B run `m9775k1v`):

| Metric | Current Best (val) | AB-UPT Target |
|--------|-------------------|--------------|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **7.2091** | 4.51 |
| `val_primary/surface_pressure_rel_l2_pct` | 4.802 | 3.82 |
| `val_primary/wall_shear_rel_l2_pct` | 8.160 | 7.29 |
[...]

PR #266, opened 2026-05-01
stark: U-net long skip connections for wsy/wsz fine-detail injection

morganmcg1:
## Hypothesis

Add **U-net-style long skip connections** between symmetric layers of the Transolver stack: layer i input is augmented with the output of layer (depth - 1 - i). For our 4L architecture: input of layer 3 ← output of layer 0; input of layer 2 ← output of layer 1. This is borrowed directly from modded-NanoGPT's "Skip connections" world-record entry, where adding U-net long-range skips gave a non-trivial speed/quality boost.

**Mechanism.** Standard transformer residuals add the input of each layer to its output (short-range, depth-1 skips). U-net skips add a separate path that injects shallow-layer features directly into deep-layer inputs. For CFD surrogate: the early layers carry fine-resolution geometric features (curvature, surface normals, raw position), while deep layers carry abstract aerodynamic features (slot tokens after multiple attention rounds). The wall-shear y/z gap may be partly an information-bottleneck problem: by layer 3, fine geometric detail has been compressed into the slot/slice attention bottleneck. Long skips rebroadcast the fine detail into deep layers.

**Why for wsy/wsz specifically.** wsy/wsz are dominated by sub-millimeter boundary-layer features that depend on local geometry (curvature, normal direction). A long skip that injects layer-0 features (raw post-embedding tokens) into layer-3 inputs gives the model direct access to high-resolution geometric signal at the moment it's making the final prediction.

**Risk.** Adds a 1×1 linear projection per skip (cheap). Doubles the input dimension at the merge point unless we use addition (which assumes feature alignment) or concatenation+projection. Concatenation+projection is slightly more parameters but more flexible.

## Instructions

**Code change to `target/model.py`:**

1. In the `Transformer` class (around line 205), modify the forward to support symmetric U-net long skips:
   - Cache the outputs of the first `depth // 2` layers in a list.
   - For the second half of the stack, before each layer, take its current input `x`, concatenate with the cached output of the symmetric early layer, project back to `hidden_dim` via a learned linear, and pass to the layer.
   - For 4 layers: layer 0 output cached as skip0, layer 1 output cached as skip1. Layer 2's input is `Linear(concat(x, skip1))`. Layer 3's input is `Linear(concat(x, skip0))`.

2. Add a CLI flag `--unet-skip-connections` (default `False`) that gates this behavior. When `True`, instantiate the per-skip projection modules.

3. **Run a 4-arm sweep on 4 GPUs (single-process per GPU):**

   | Arm | unet-skip | Notes |
   |---|---|---|
   | A (control) | off | Exact bengio baseline (alphonse PR #174 / m9775k1v config), no code change |
   | B | on, addition | Concatenation replaced with addition (assumes layer-i output is alignment-compatible with layer-(d-1-i) input). No new parameters. |
   | C | on, concat+project | Full version: concat then project back to hidden_dim. New params: ~2 × hidden_dim² for 4L. |
   | D | on, gated | Concat+project with a learned scalar gate `α` (initialized 0.0) so the model can opt-in to the skip information gradually. |

   All arms otherwise identical to alphonse `m9775k1v` baseline: 4L/256d/4h, AdamW, lr=3e-4, 30 epochs, cosine T_max=30, no EMA.

   **W&B group:** `stark-unet-skip-sweep`.

4. **Per-arm command (Arm B example):**

```bash
cd target/
CUDA_VISIBLE_DEVICES=1 torchrun --standalone --nproc-per-node=1 train.py \
  --agent stark --lr 3e-4 --weight-decay 5e-4 \
  --no-compile-model --batch-size 8 --validation-every 1 \
  --model-layers 4 --model-hidden-dim 256 --model-heads 4 --model-slices 128 \
  --epochs 30 --lr-cosine-t-max 30 --lr-warmup-steps 500 \
  --fourier-pe --fourier-pe-num-freqs 8 \
  --unet-skip-connections --unet-skip-mode addition \
  --wandb-name stark/skip-addition --wandb-group stark-unet-skip-sweep
```

## Gates

- **ep5:** kill any arm if val_abupt > 13% (alphonse ep5 was ~10.0%)
- **ep10:** kill any arm if val_abupt > 9.5% (alphonse ep10 was ~8.5%)
- **ep20:** kill any arm if val_abupt > 8.0% (alphonse ep20 was ~7.5%)
- **ep30:** beat 7.21% on any arm = win

## Win criterion

**Best skip-connection arm reaches val_abupt < 7.2091% at ep30 → merge.** Direction-of-improvement matters: if Arm D (gated) wins by a larger margin than Arms B/C, it confirms the skip information is valuable but needs slow integration. If Arm B (pure addition) matches Arm C (concat+project), the no-extra-parameter version is preferred.

**Per-axis breakdown at ep30:** Especially track wsy/wsz — if those drop disproportionately vs surf_p, that's strong evidence the information-bottleneck hypothesis is right.

## Baseline (current best on bengio)

| Metric | bengio best (val ep30) | PR | W&B run |
|---|---:|---|---|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **7.2091** | alphonse | `m9775k1v` |
| `val_primary/surface_pressure_rel_l2_pct` | **4.80** | alphonse | `m9775k1v` |
| `val_primary/wall_shear_rel_l2_pct` | (TBD) | alphonse | `m9775k1v` |

**Reproduce alphonse baseline (control arm A):**
```bash
  --agent alphonse --lr 3e-4 --weight-decay 5e-4 \
  --no-compile-model --batch-size 8 --validation-every 1 \
  --model-layers 4 --model-hidden-dim 256 --model-heads 4 --model-slices 128 \
  --epochs 30 --lr-cosine-t-max 30 --lr-warmup-steps 500 \
  --fourier-pe --fourier-pe-num-freqs 8
```

## Context

This experiment is one of the issue #252 (Modded-NanoGPT inspiration) directions — specifically item #4 in the advisor's response. PR #214 (gilbert kNN local surface attention) is currently running on bengio and showing strong wsy/wsz wins (-1.58pp wsy, -1.22pp wsz vs alphonse same-epoch at ep2) — that's evidence locality matters for wall-shear; long skips are a different mechanism that achieves the same goal (more local detail in the deep layers). The two experiments are complementary, not duplicative.

## Implementation notes

- The `_apply_token_mask` calls in TransformerBlock.forward should still be applied at the start of each block. Apply the skip injection BEFORE the mask, so masked tokens are zeroed in both the main and skip paths.
- For the gated variant (Arm D), initialize `α = 0` so initial behavior matches the control. Use `nn.Parameter(torch.zeros(1))` and apply as `x = x + α * skip_proj`.
- Verify the skip projection layers are in the correct param group for your optimizer (AdamW gets all params here, no Muon split needed).


morganmcg1:
ADVISOR: Closing PR #266 because ADVISOR: Closing this PR — the branch name `stark/post-attn-rmsnorm` (post-attention RMSNorm) does not match any assigned Wave 5 hypothesis for student stark, and there is no `senpai-bengio-stark` pod provisioned in the cluster. No experiment was run and no results are available to review. If you believe this was an error in assignment, please reach out and a new hypothesis will be assigned on a fresh branch. GPU slots are too valuable to hold open on an unprovisioned experiment..

PR #261, opened 2026-05-01
norman: Muon optimizer (Newton-Schulz orthogonalized momentum) for 4L/512d SOTA

morganmcg1:
## Hypothesis

Replace Lion (current SOTA optimizer) with **Muon** (Newton-Schulz orthogonalized momentum) on hidden 2D weights and test whether it reaches a lower val_abupt within the 9-epoch budget. Muon was the headline win in modded-NanoGPT (https://github.com/kellerjordan/modded-nanogpt) — it explicitly orthogonalizes the momentum buffer of 2D weight matrices so each update is closer to a unit-norm rotation, sidestepping the gradient-magnitude pathologies that drive our fleet-wide lr=5e-4 instability and that Lion partially solves with sign-only updates.

**Mechanism.** On 2D weights `W ∈ R^(m×n)`, Muon computes `Wt = W - lr * NS(M)` where `NS(M)` is a Newton-Schulz iteration that orthogonalizes the momentum buffer M into the closest semi-orthogonal matrix. Compared to Lion's `sign(M)` (which has bounded magnitude per coord but non-orthogonal updates), Muon enforces a stronger structural prior: each update is a rotation, never a coordinate-wise sign-flip. Empirically in NanoGPT this reduced training time-to-target by ~30% at matched compute. For our setting, attention QKV projections and FFN linears are exactly the layer types Muon was designed for, and our model is dominated by such 2D matrices.

**Why it might work for CFD.** The 4× tau_y/z gap at AB-UPT plateau suggests our model fails to find the right basis to express cross-flow shear features. Orthogonalized updates may help the model navigate to a better-conditioned basin where fine-axis features are linearly disentangled from each other. Bonus: Muon naturally regularizes weight norm growth, complementing our weight-decay choice.

**Risk.** Untested in 3D point-cloud regimes. Newton-Schulz iterations have FLOPs cost ~2-5% per step. Original recipe keeps AdamW for embedding/coord-encoding parameters — must respect this split or training will be unstable.

## Instructions

**Code change to `target/train.py`:**

1. Pip install Muon: `pip install git+https://github.com/KellerJordan/modded-nanogpt.git#subdirectory=muon` (or vendor the muon.py file from https://github.com/KellerJordan/Muon — small, ~150 lines).

2. Add a new optimizer option: `--optimizer muon`. When selected:
[...]

Issue #248, opened 2026-05-01
ADVISOR: senpai-yi-stark pod needs manual provisioning (RBAC block)


morganmcg1:
## ADVISOR: stark pod cannot be provisioned from within advisor pod

**PR:** #227 (surface tangent frame — predicts wall shear in local {t1, t2, n} frame)
**Student:** stark
**Status:** No executor pod exists. PR has been sitting unassigned since creation.

### Root Cause

The orchestrator service account (`system:serviceaccount:default:senpai-orchestrator`) does not have `create` or `patch` permissions on `configmaps` or `deployments.apps`. All provisioning paths (direct `kubectl apply`, `launch.py`, `launch_helpers.kubectl_apply()`) return:

```
Error from server (Forbidden): configmaps is forbidden: User "system:serviceaccount:default:senpai-orchestrator" cannot create resource "configmaps" in API group "" in the namespace "default"
```

### Action Required

Please run the following from a machine with cluster admin access to provision the stark pod:

```bash
cd /workspace/senpai/k8s
python launch.py \
  --tag yi \
  --target_repo_url https://github.com/morganmcg1/DrivAerML \
  --names stark \
  --gpus_per_student 4 \
  --cpu_per_gpu 15 \
  --memory_gi_per_gpu 120 \
  --wandb_project senpai-v1-drivaerml \
  --advisor_branch yi \
  --max_epochs 999 \
  --timeout_minutes 360
```

This will create:
- ConfigMap: `senpai-config-student-yi-stark`
- Deployment: `senpai-yi-stark` (4 GPUs, 60 CPU, 480Gi memory, image `ghcr.io/wandb/senpai:latest`, PVC `new-pvc` at `/mnt/new-pvc`)

### Why this matters

PR #227 (surface tangent frame) is the most physically-motivated tau_y/z approach in the current round — it enforces no-slip BC geometry by decomposing wall shear into local tangent frame components {t1, t2, n}. This is a high-priority experiment that should be running. Current baseline is val abupt = 10.21.

Thank you!

morganmcg1:
[...]

PR #247, opened 2026-05-01
Gentle cosine annealing: lr-cosine-t-max=14 (midpoint between flat-50 SOTA and collapsed-9)

morganmcg1:
# Hypothesis

LR cosine annealing with T_max=14 (gentle midpoint) should improve generalization over the flat LR (T_max=50, effectively no decay for 9 epochs) while avoiding the confirmed failure mode of T_max=9 (LR collapses too early, confirmed negative × 2).

**Background:**
- SOTA uses `--lr-cosine-t-max 50` with a 9-epoch budget. Because T_max=50 >> actual_epochs=9, the cosine barely decays — the LR stays nearly flat at 1e-4 through the whole run. It's functionally a constant LR.
- T_max=9 was tested twice: edward PR #195 (test_abupt=10.809%) and tanjiro PR #202 (val=9.710%), both NEGATIVE. The LR fell to near-zero by epoch 8–9 while the model was still converging.
- **T_max=14 is unexplored.** It targets the midpoint: LR decays from 1e-4 to ~0 over 14 epochs, so by epoch 9 (our cutoff), LR ≈ 1e-4 × cos(π × 9/(2 × 14)) ≈ 1e-4 × 0.58 = **5.8e-5**. The model gets modest late-stage annealing without hitting the collapse zone. This may provide the regularizing benefit of cosine scheduling while keeping the LR high enough for continued progress.

**Single variable changed:** `--lr-cosine-t-max 14` (from 50 in SOTA).

## Instructions

Run on 8 × GPU:

```bash
cd target/
torchrun --standalone --nproc_per_node=8 train.py \
  --agent thorfinn \
  --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --no-compile-model --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 --model-slices 128 \
  --ema-decay 0.999 \
  --lr-cosine-t-max 14 \
  --wandb-group tay-round14-lr-cosine-t-max-14
```

**Key delta vs SOTA:** `--lr-cosine-t-max 14` (SOTA uses `--lr-cosine-t-max 50`).

All other flags are identical to SOTA (PR #115, W&B run `d03oghpp`).

When the run completes, please report the following in a new comment:
- W&B run ID
- Per-epoch val/loss trajectory
[...]

PR #279, opened 2026-05-01
frieren: soft no-slip BC penalty (tau dot n_hat) for wall_shear_y/z gap

morganmcg1:
## Hypothesis

**Soft no-slip / no-penetration physics penalty on the wall-shear prediction.**

The wall-shear stress vector `tau` is, by physics, **tangent to the surface** at every point — it cannot have a component along the surface normal `n_hat`. The current SOTA recipe (PR #222) does not enforce this constraint at all. Kohaku's `--use-tangential-wallshear-loss` (PR #11) projects the prediction onto the tangent plane *before* computing MSE, which means any normal-component error is silently discarded — the model is never punished for predicting normal-direction garbage and never learns that the constraint exists. The latent `wallshear_pred_normal_rms` metric (already logged when tangential mode is on) shows the model does in fact violate this BC.

This experiment adds a **soft penalty term** instead of (or in addition to) a hard projection:

```
L = surface_loss + volume_loss + λ_normal · mean( (tau_pred · n_hat)^2 )
```

Computed in **physical (denormalized) space** for unit consistency with the existing tangential-loss code path. This teaches the network a hard physical prior with a single scalar hyperparameter and zero architectural change. Wall_shear_y/z account for ~70% of the abupt headline gap and are the most prone to normal-component leakage (because the dominant tau_x has clearer streamline alignment).

**Why this is orthogonal to current in-flight work:**
- #207 alphonse (AGC), #247 thorfinn (T_max=14), #262 nezuko (warmdown), #261 norman (Muon), #245 gilbert (EMA), #249 tanjiro (asinh), #244 emma (surf weight), #243 chihiro (aux rel-L2): all about **optimization or loss-magnitude scaling** — none enforce a physics constraint.
- #227 stark (tangent-frame prediction): predicts in 2D tangent basis instead of 3D — pod missing per Issue #248 (RBAC blocked). This soft-penalty variant requires no tangent frame and no pod plumbing.
- #224 fern (learned Fourier), #225 haku (mirror sym aug), #193 thorfinn (curvature sampling): input-side or augmentation work.

The penalty is a clean, single-line addition; if it works it stacks with all of the above.

## Instructions

**Code change to `target/train.py`:**

1. Add a `Config` flag `wallshear_normal_penalty: float = 0.0` and CLI flag `--wallshear-normal-penalty FLOAT` (default 0.0 = off).
2. In `train_step` after the existing `surface_loss` / `volume_loss` computation (lines ~1355–1362), add:

   ```python
   if wallshear_normal_penalty > 0.0:
       normals = batch.surface_x[..., 3:6]
       n_hat = F.normalize(normals.float(), dim=-1, eps=1e-8)
       ws_std = transform.surface_y_std[1:4]
       ws_mean = transform.surface_y_mean[1:4]
       ws_pred_phys = surface_pred_norm[..., 1:4].float() * ws_std + ws_mean
       normal_dot = (ws_pred_phys * n_hat).sum(dim=-1)
       mask_f = batch.surface_mask.float()
       penalty = (normal_dot.square() * mask_f).sum() / mask_f.sum().clamp_min(1.0)
       loss = loss + wallshear_normal_penalty * penalty
       metrics["wallshear_normal_penalty"] = float(penalty.detach().cpu().item())
   ```

3. Thread `wallshear_normal_penalty` through `train_step()` signature and the call site, exactly as the existing `wallshear_y_weight` / `wallshear_z_weight` (PR #66) flags are threaded.
4. Add the metric to the per-step logging dict (`metrics["wallshear_normal_penalty"]`) — this lets us watch the constraint be learned over training.

**Important interaction with existing `--use-tangential-wallshear-loss`:** Do **not** combine `--use-tangential-wallshear-loss` with `--wallshear-normal-penalty` in this experiment. The projection erases the normal component so the penalty would always be 0. Run with `--use-tangential-wallshear-loss false` (default).

## Sweep design (4 arms, 1 GPU each, 4 GPUs total)

Use `--wandb-group frieren-wallshear-normal-penalty`.

Single-process per GPU with `--seed=-1`. Same SOTA recipe as PR #222 base config in **all** arms (4L/512d, Lion, lr=1e-4, lr_warmup_epochs=1, batch=4, ema=0.999):

| Arm | GPU | wallshear_normal_penalty (λ) | Run name suffix |
|---:|---:|---:|---|
| A (control) | 0 | 0.0 | `arm-A-ctrl` |
| B (light) | 1 | 0.1 | `arm-B-l0.1` |
| C (medium) | 2 | 0.5 | `arm-C-l0.5` |
| D (strong) | 3 | 2.0 | `arm-D-l2.0` |

The control arm is critical — same seed handling, same 4L/512d/Lion stack, no penalty — to control for any seed/HW noise. Compare arms B/C/D against arm A and against the published bar.

### Reproduce command (single arm; replicate per GPU with different `--wallshear-normal-penalty`)

```bash
cd target/
  --agent frieren \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --lr-warmup-epochs 1 \
  --wallshear-normal-penalty 0.1 \
  --seed -1 \
  --wandb-group frieren-wallshear-normal-penalty \
  --kill-thresholds "3000:train/loss<5,3000:train/grad/global_norm<300"
```

Run the 4 arms in parallel, one per GPU. **9 epochs** matches PR #222.

## Acceptance criteria

- **Merge bar:** any arm beating `val_primary/abupt_axis_mean_rel_l2_pct = 9.291%` at best epoch is mergeable.
- **Health gate:** `wallshear_normal_penalty` metric should monotonically decrease — if it stays flat or increases, the penalty is being ignored and we have a bug.
- **Stability gate:** if any arm trips the kill threshold (grad>300 at step 3000), that's a useful signal — note which λ failed and report the trajectory before the kill.

## Baseline (current best on `yi` — must beat 9.291%)

| Metric | yi best | PR | W&B |
|---|---:|---|---|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **9.2910** | #222 (fern, lr_warmup_epochs=1) | `ut1qmc3i` |
| `val_primary/surface_pressure_rel_l2_pct` | 5.8707 | #222 | `ut1qmc3i` |
| `val_primary/wall_shear_rel_l2_pct` | 10.3423 | #222 | `ut1qmc3i` |
| `val_primary/volume_pressure_rel_l2_pct` | 5.8789 | #222 | `ut1qmc3i` |

AB-UPT reference: surface_pressure=3.82, wall_shear=7.29 (this is the gap we're attacking), volume_pressure=6.08 (already beaten).

## Reporting

Reply with the standard results table on this PR when all 4 arms reach 9 epochs (or are killed):

| Arm | λ | best epoch | val_abupt | surf_pres | vol_pres | wall_shear | wallshear_normal_penalty (final) | run id |

Plus a short interpretation of how the penalty trajectory tracks against the abupt metric.


morganmcg1:
ADVISOR: Closing PR #279 because ADVISOR: Closing this PR before launch. I missed an existing experiment when designing the hypothesis — apologies, frieren.

**PR #201 (nezuko, 2026-05-01) already tested this exact penalty form** on the surface side and closed it with a hard-data finding:

> GT labels have RMS(tau dot n) / |tau| = 12% (mean=0.113, RMS=0.364, p99=1.722). DrivAerML mesh-discretized normals are not exactly orthogonal to the analytic surface, so the continuum no-slip constraint (tau dot n = 0) directly contradicts the ground truth. There is no lambda where this penalty can help:
>   - Small lambda: NaN cascade (>200 grad skips)
>   - Large lambda: forces tau_pred dot n -> 0, but labels have tau dot n ~= 0.36 -> 2x train loss inflation, 4x val regression
> Verdict: NEGATIVE.

The "tau is tangent to the surface" prior is geometrically true but **not satisfied by the discretized normals shipped with the dataset** — running this penalty would deterministically make the model worse, exactly as nezuko's λ=0.1 arm showed (val_abupt 63.09 vs control 15.55).

Reassigning you to a clean hypothesis now. No fault on you — this was an advisor mistake; should have grepped EXPERIMENTS_LOG.md for "no-slip" before writing the PR. Closing..

PR #251, opened 2026-05-01
fern: warmup+cosine anneal T_max=8 lr-min=5e-6

morganmcg1:
## Hypothesis

A "warmup-then-anneal-hard" LR schedule will outperform the flat LR used in SOTA (PR #222).
The idea: 1 epoch linear warmup to peak LR 1e-4, then cosine decay to 5% of peak (lr-min=5e-6)
over the remaining 8 epochs (T_max=8). This enforces genuine convergence at the end of training,
unlike the current schedule which holds 1e-4 flat after warmup.

**Evidence:**
- PR #247 (thorfinn, in-flight) tests cosine T_max=14 without lr-min — a softer version.
- PR #195 (edward) was T_max=9 but never ran and had no warmup or lr-min.
- No prior run has combined: (a) warmup=1ep + (b) cosine anneal + (c) non-trivial lr-min (5e-6).
- Total epochs = 9 (warmup 1 + cosine 8), matching SOTA epoch count exactly.

**Why lr-min=5e-6 specifically:**
- At epoch 9, LR = 5e-6 = 5% of peak 1e-4. This is aggressive enough to produce sharp final
  convergence without killing the optimizer too early.

## Instructions

Single-delta from SOTA (PR #222): add cosine LR schedule with `--lr-min 5e-6` and `--lr-cosine-t-max 8`.
All other flags identical to SOTA.

```bash
torchrun --standalone --nproc_per_node=8 train.py \
  --agent fern \
  --wandb-group tay-round13-lr-anneal \
  --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --no-compile-model --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 --model-slices 128 \
  --ema-decay 0.999 --lr-warmup-epochs 1 \
  --lr-cosine-t-max 8 --lr-min 5e-6 \
  --gradient-log-every 100 --weight-log-every 100 \
  --no-log-gradient-histograms
```

**Key changes vs SOTA (PR #222):**
- `--lr-cosine-t-max 8` — cosine decay over 8 epochs (after 1ep warmup = 9 total)
- `--lr-min 5e-6` — decay to 5% of peak LR (5e-6), not zero

[...]

PR #296, opened 2026-05-01
violet: lr-min=1e-5 on SOTA stack (cosine LR schedule floor)

morganmcg1:
## Hypothesis

Test lr-min=1e-5 as a floor for the cosine LR schedule. The current SOTA uses cosine schedule with T_max=50 (approximately flat within the ~11ep training budget), effectively no LR warmdown. Setting lr_min=1e-5 (vs default 0) ensures the LR never fully decays to zero at the end of the cosine period. This is motivated by: (1) the Modded-NanoGPT linear warmdown approach — a floor prevents the LR from going too low in the final epochs, (2) fern #251 (T_max=8 + lr-min=5e-6) is tracking negative, but 5e-6 may be too low — 1e-5 is a more moderate floor that keeps gradient updates alive in late training.

**Single delta from SOTA:** `--lr-min 1e-5` (SOTA has no lr-min or it defaults to 0)

Check the available CLI flag: `python train.py --help | grep -i "lr-min\|lr_min\|min.*lr\|eta_min"`

## Current SOTA (Baseline to beat)

PR #232 (askeladd, model-heads=4), W&B run `r8s2dtnq`
- **Best val abupt:** 9.0650% (ep11)
- **Test abupt:** 10.190%
- Per-axis test: surface_pressure=5.461, wall_shear=9.910, volume_pressure=12.656, tau_x=8.432, tau_y=11.952, tau_z=12.447

## Instructions

First verify the exact flag name:
```bash
cd target/
python train.py --help | grep -i "min"
```

Then run:
```bash
cd target/
  --agent violet --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --lr-min 1e-5
```

**Key change:** `--lr-min 1e-5` — sets the cosine schedule minimum LR floor (SOTA uses default 0 or no floor)

If the flag name differs, use whatever `--help` shows and report the actual flag used.

## Success Criteria

- Beat SOTA val abupt of 9.0650%
- Report final val abupt and test abupt (from best val checkpoint) with all per-axis metrics

## What to Report

Please report in a PR comment:
1. Exact lr-min flag used
2. Val abupt at each epoch
3. Best val abupt and the epoch it occurred
4. Test abupt evaluated from the best val checkpoint
5. All per-axis test metrics: surface_pressure, wall_shear, volume_pressure, tau_x, tau_y, tau_z
6. W&B run ID


morganmcg1:
ADVISOR: Closing PR #296 because ADVISOR: Closing as orphaned. Student 'violet' is not currently allocated to the tay branch (no senpai-drivaerml-ddp8-violet deployment exists). This PR was created in a prior advisor session before the deployment topology was confirmed. The hypothesis is queued for reassignment to one of the 8 active tay students (alphonse, askeladd, edward, fern, frieren, nezuko, tanjiro, thorfinn) once a slot opens..

PR #227, opened 2026-05-01
stark: wall-shear prediction in local surface tangent frame

morganmcg1:
## Hypothesis

The wall-shear y/z gap (tau_y: 13.47 vs AB-UPT 3.65, tau_z: 14.52 vs 3.63 — both ~4× off) is caused in part by the model learning wall-shear stress in world coordinates (x, y, z), where the physical constraint (no-slip BC forces tau_n = 0) is implicit and the axes y/z have low-magnitude, high-frequency variation that is hard to regress directly.

**Hypothesis:** If we predict wall-shear stress in the local surface tangent frame {t1, t2, n} — where n is the outward surface normal and (t1, t2) are two orthogonal tangent vectors per point — then:
1. The model no longer needs to regress a near-zero tau_n component against noise
2. The physically meaningful in-plane components (tau_t1, tau_t2) carry the full signal
3. The loss landscape is better conditioned because the full magnitude is concentrated in two channels instead of being distributed across three world-frame axes

The surface normals are already available at `surface_x[..., 3:6]`. We rotate the prediction and target into the surface frame, compute the loss in frame coordinates, and rotate the prediction back to world coordinates for validation.

Reference: Surface-tangent frame prediction is standard in aerodynamic drag/separation prediction literature (Bonnet et al., 2022; Fortunato et al., 2022). The key property: in the local frame, tau_n is identically zero by the Newtonian no-slip BC, so training in frame coordinates implicitly enforces the physical constraint.

## Instructions

**Overview:** Add a `--use-tangent-frame-loss` flag that rotates the wall-shear prediction and target into the local {t1, t2, n} frame before computing loss. Training predicts in the tangent frame; validation converts back to world coords for consistent metric reporting.

**Step 1: Build the tangent frame rotation.**

Add a function `build_tangent_frame(normals)` near the existing `project_tangential()` function in `train.py`:

```python
def build_tangent_frame(normals: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
    """Build orthonormal surface frame {t1, t2, n} per point.
    
    normals: [B, N, 3] unit outward normals
    Returns: (t1, t2, n_hat) each [B, N, 3] in fp32.
    
    Uses Frisvad/Duff et al. (2017) method for a numerically stable
    orthonormal basis from a single normal vector, avoiding cross-product
    singularities near the poles.
    """
    n = F.normalize(normals.float(), dim=-1, eps=1e-8)
    # Duff et al. 2017 — sign-based Frisvad
[...]

PR #312, opened 2026-05-02
edward: surface-tangent frame wall-shear prediction

morganmcg1:
## Hypothesis

Wall-shear stress is a **surface-local quantity**: it equals `μ · (∂u/∂n)|_surface`, where **n** is the outward surface normal. The current model predicts `tau_x`, `tau_y`, `tau_z` in the **global** Cartesian frame, meaning it must learn to implicitly decompose the shear into a frame that is **not** aligned with the underlying physics. This is a particularly bad inductive bias for `tau_y` and `tau_z` — the cross-flow components that span 4× the AB-UPT gap — because these correspond to very different physical magnitudes depending on the local surface orientation.

**Proposed fix:** Predict wall-shear in the **local surface-tangent frame** (two orthonormal tangent vectors `t1`, `t2` spanning the local surface plane), then rotate the prediction back to Cartesian before computing the loss. This gives the model an aligned representation and removes the orientation ambiguity. The rotation is differentiable, so gradients flow correctly into the Cartesian loss.

**Expected gain:** Prior work on aerodynamic surface predictions (e.g. NeuralFoil, PhysDiff-Surface) shows that predicting in the natural coordinate frame of the physical quantity typically gives 10–30% improvement on cross-flow components. Our tau_y/z are 4× worse than AB-UPT and are the primary improvement target.

This hypothesis was previously blocked on pod provisioning (PRs #199, #227 — never ran). This is the first actual test.

## Instructions

**Code change:** Modify `target/train.py` to add a surface-tangent frame rotation layer in the surface decoder output path.

### Step 1 — Compute local tangent frames at each surface point

For each surface point `x_i ∈ ℝ³`, compute two orthonormal tangent vectors `t1_i, t2_i ∈ ℝ³` that span the local surface plane, and the outward normal `n_i = t1_i × t2_i`.

The simplest practical approach for point clouds without connectivity:
1. For each point, find its k=8 nearest neighbors by Euclidean distance.
2. Fit a local PCA: `t1 = eigvec(cov(neighbors), largest)`, `t2 = eigvec(cov(neighbors), 2nd largest)`, `n = t1 × t2`.
3. Orient `n` consistently (flip sign if `n · [0,0,1] < 0` — all vehicles have the same up direction).

Cache the tangent frames once per batch (no gradient needed on the frame itself).

### Step 2 — Rotate model outputs to tangent frame before loss

Current surface head output: `[B, N, 4]` = `[cp, tau_x, tau_y, tau_z]`.

After computing tangent frames `T ∈ ℝ^{N×3×3}` (rows = t1, t2, n):
```python
tau_cart = preds[:, :, 1:4]           # [B, N, 3] — Cartesian tau
tau_tang = torch.einsum('bnij,bnj->bni', T.unsqueeze(0), tau_cart)  # [B, N, 3]
preds_tang = torch.cat([preds[:, :, :1], tau_tang], dim=-1)  # [B, N, 4]
```

Similarly rotate the ground-truth `tau` targets to the tangent frame before computing the per-axis loss.

### Step 3 — Predict in tangent frame, rotate back for logging

At validation/test, rotate predictions back to Cartesian: `tau_cart = T^T @ tau_tang`, then log the standard Cartesian metrics (`wall_shear_x/y/z rel_l2`) so results are directly comparable to the current baseline.

### Flag: `--use-surface-tangent-frame` (BooleanOptionalAction, default False)

Gate the whole change behind this flag so the control arm (no flag) is identical to the current SOTA baseline.

### Run: single arm only for this PR

```bash
cd target/
python train.py \
  --agent edward \
  --wandb-group edward-surface-tangent-r18 \
  --wandb-name tangent-frame-main \
  --lr 5e-4 \
  --weight-decay 1e-4 \
  --lr-warmup-steps 500 \
  --clip-grad-norm 1.0 \
  --no-compile-model \
  --batch-size 8 \
  --validation-every 1 \
  --ema-decay 0.999 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
  --use-surface-tangent-frame
```

**Note on DDP:** `target/train.py` on `yi` is single-GPU only (no DDP yet). Run the above as a `python` call (not `torchrun`). Maximise VRAM by using `--batch-size 8` and the largest point counts that fit.

### What to report

1. Table: Arm (tangent-frame vs sincos-control), val_abupt, surface_pressure, wall_shear, vol_pressure, wall_shear_x/y/z
2. W&B run ID
3. Were tangent frame computations stable (nan-free, consistent PCA orientations)?
4. Did tau_y/z improvement come at the cost of surface_pressure or vol_pressure?

| Metric | Best (val) |
|---|---:|
| `wall_shear_y_rel_l2_pct` | ~13.5% (4× AB-UPT target 3.65%) |
| `wall_shear_z_rel_l2_pct` | ~14.5% (4× AB-UPT target 3.63%) |

**To beat:** val_abupt < 9.291%. Equally important: does tau_y or tau_z improve meaningfully (>1 pp)?

## Reproduce command (control arm — no flag)

```bash
cd target/
python train.py \
  --agent edward \
  --wandb-group edward-surface-tangent-r18 \
  --wandb-name sincos-control \
  --lr 5e-4 \
  --weight-decay 1e-4 \
  --lr-warmup-steps 500 \
  --clip-grad-norm 1.0 \
  --no-compile-model \
  --batch-size 8 \
  --validation-every 1 \
  --ema-decay 0.999 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0
```

### DIFF
--- a/train.py
+++ b/train.py
@@ -563,6 +563,8 @@ class Config:
     use_tangential_wallshear_loss: bool = False
     wallshear_y_weight: float = 1.0
     wallshear_z_weight: float = 1.0
+    use_surface_tangent_frame: bool = False
+    max_steps_per_epoch: int = 0
     manifest: str = "data/split_manifest.json"
     data_root: str = ""
     output_dir: str = "outputs/drivaerml"
@@ -1300,6 +1302,38 @@ def project_tangential(vec: torch.Tensor, normals: torch.Tensor) -> torch.Tensor
     return (vec_f - dot * n_hat).to(vec.dtype)
 
 
+def compute_surface_tangent_frames(surface_x: torch.Tensor) -> torch.Tensor:
+    """Build a per-point orthonormal tangent frame T = [t1, t2, n] from given normals.
+
+    surface_x: [B, N, >=6] where [..., 3:6] are surface normals. Returns
+    T: [B, N, 3, 3] (fp32) with rows = [t1, t2, n] s.t. v_tang = T @ v_cart.
+
+    The basis is built deterministically: t1 = normalize(cross(n, ref)) where
+    ref is the global X axis (or Y axis when |n . X| > 0.9 to avoid the
+    degenerate cross product). Then t2 = cross(n, t1). This is O(N) and
+    avoids the cost of per-point KNN/PCA at our point counts (~520k/step),
+    while using the data's clean mesh normals directly.
+    """
+    n = F.normalize(surface_x[..., 3:6].float(), dim=-1, eps=1e-8)  # [B, N, 3]
+    abs_nx = n[..., 0:1].abs()  # [B, N, 1]
+    ref_x = torch.zeros_like(n)
+    ref_x[..., 0] = 1.0
+    ref_y = torch.zeros_like(n)
+    ref_y[..., 1] = 1.0
+    ref = torch.where(abs_nx > 0.9, ref_y, ref_x)
+    t1 = F.normalize(torch.cross(n, ref, dim=-1), dim=-1, eps=1e-8)  # [B, N, 3]
+    t2 = torch.cross(n, t1, dim=-1)  # already unit-norm given orthonormal n, t1
+    T = torch.stack([t1, t2, n], dim=-2)  # [B, N, 3, 3]
+    return T
+
+
+def rotate_to_tangent(vec: torch.Tensor, T: torch.Tensor) -> torch.Tensor:
+    """Rotate per-point Cartesian 3-vectors to the tangent frame: v_tang = T @ v_cart."""
+    vec_f = vec.float()
+    out = torch.einsum("bnij,bnj->bni", T, vec_f)
+    return out.to(vec.dtype)
+
+
 def train_loss(
     model: nn.Module,
     batch: SurfaceBatch,
@@ -1313,6 +1347,7 @@ def train_loss(
     use_tangential_wallshear_loss: bool = False,
     wallshear_y_weight: float = 1.0,
     wallshear_z_weight: float = 1.0,
+    use_surface_tangent_frame: bool = False,
 ) -> tuple[torch.Tensor, dict[str, float]]:
     batch = batch.to(device)
     surface_target = transform.apply_surface(batch.surface_y)
@@ -1326,7 +1361,47 @@ def train_loss(
         )
         surface_pred_norm = out["surface_preds"]
         normal_rms = float("nan")
-        if use_tangential_wallshear_loss:
+        tangent_diag: dict[str, float] = {}
+        if use_surface_tangent_frame:
+            with torch.no_grad():
+                T = compute_surface_tangent_frames(batch.surface_x)  # [B, N, 3, 3]
+            ws_pred_norm = surface_pred_norm[..., 1:4]
+            ws_true_norm = surface_target[..., 1:4]
+            ws_pred_tang_norm = rotate_to_tangent(ws_pred_norm, T)
+            ws_true_tang_norm = rotate_to_tangent(ws_true_norm, T)
+            surface_pred_used = torch.cat([surface_pred_norm[..., :1], ws_pred_tang_norm], dim=-1)
+            surface_target_used = torch.cat([surface_target[..., :1], ws_true_tang_norm], dim=-1)
+            if bool(batch.surface_mask.any()):
+                with torch.no_grad():
+                    mask_f = batch.surface_mask.float()
+                    valid_count = mask_f.sum().clamp_min(1.0)
+                    # Sanity: rotated target normal component RMS (in normalized space).
+                    # This is not zero in normalized space because per-axis stds differ,
+                    # but it diagnoses extreme values / NaNs.
+                    n_comp_norm = ws_true_tang_norm[..., 2].float()
+                    n_comp_rms = float(
+                        ((n_comp_norm.square() * mask_f).sum() / valid_count).sqrt().cpu().item()
+                    )
+                    # Physical-space normal-component RMS of true target (should be ~0).
+                    ws_std = transform.surface_y_std[1:4].to(ws_true_norm.device)
+                    ws_mean = transform.surface_y_mean[1:4].to(ws_true_norm.device)
+                    ws_true_phys = ws_true_norm.float() * ws_std + ws_mean
+                    n_hat = T[..., 2, :]  # [B, N, 3]
+                    n_dot_phys = (ws_true_phys * n_hat).sum(dim=-1)
+                    n_phys_rms = float(
+                        ((n_dot_phys.square() * mask_f).sum() / valid_count).sqrt().cpu().item()
+                    )
+                    # Tangent-frame orthogonality check: should be ~1 for valid frames.
+                    det = torch.linalg.det(T.float())
+                    det_mean = float(((det * mask_f).sum() / valid_count).cpu().item())
+                    nan_count = int(torch.isnan(T).any(dim=(-2, -1)).sum().cpu().item())
+                tangent_diag = {
+                    "tangent/n_target_norm_rms": n_comp_rms,
+                    "tangent/n_target_phys_rms": n_phys_rms,
+                    "tangent/frame_det_mean": det_mean,
+                    "tangent/frame_nan_count": float(nan_count),
+                }
+        elif use_tangential_wallshear_loss:
             # Wall-shear stds are non-uniform ([2.08, 1.36, 1.11]), so projecting
             # in normalized space does not equal physical-space tangent projection.
             # Denormalize -> project in physical space -> renormalize.
@@ -1382,6 +1457,8 @@ def train_loss(
         metrics["aux_rel_l2_loss"] = aux_rel_l2_value
     if use_tangential_wallshear_loss:
         metrics["wallshear_pred_normal_rms"] = normal_rms
+    if tangent_diag:
+        metrics.update(tangent_diag)
     if "geom_token" in out:
         geom_token = out["geom_token"].detach().float()
         metrics["film/geom_token_norm_mean"] = float(
@@ -1712,7 +1789,10 @@ def main(argv: Iterable[str] | None = None) -> None:
     optimizer = torch.optim.AdamW(model.parameters(), lr=config.lr, weight_decay=config.weight_decay)
     scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(optimizer, T_max=max_epochs)
     ema = EMA(model, decay=config.ema_decay, start_step=config.ema_start_step) if config.use_ema else None
-    total_estimated_steps = max(1, max_epochs * max(len(train_loader), 1))
+    effective_loader_len = max(len(train_loader), 1)
+    if config.max_steps_per_epoch > 0:
+        effective_loader_len = min(effective_loader_len, config.max_steps_per_epoch)
+    total_estimated_steps = max(1, max_epochs * effective_loader_len)
     if kill_thresholds:
         print("Kill thresholds:", "; ".join(threshold.describe() for threshold in kill_thresholds))
     train_slope_tracker = MetricSlopeTracker(total_estimated_steps, config.slope_log_fraction)
@@ -1763,6 +1843,7 @@ def main(argv: Iterable[str] | None = None) -> None:
     wandb.define_metric("train/weight_hist/*", step_metric="global_step")
     wandb.define_metric("train/weight_hist_param/*", step_metric="global_step")
     wandb.define_metric("train/film/*", step_metric="global_step")
+    wandb.define_metric("train/tangent/*", step_metric="global_step")
     wandb.define_metric("lr", step_metric="global_step")
     wandb.define_metric("train/lr", step_metric="global_step")
     wandb.define_metric("train/nonfinite_skip_count", step_metric="global_step")
@@ -1797,7 +1878,12 @@ def main(argv: Iterable[str] | None = None) -> None:
         train_loss_sum = 0.0
         n_batches = 0
 
+        epoch_step_cap = config.max_steps_per_epoch if config.max_steps_per_epoch > 0 else None
+        epoch_steps_done = 0
         for batch in tqdm(train_loader, desc=f"Epoch {epoch + 1}/{max_epochs}", leave=False):
+            if epoch_step_cap is not None and epoch_steps_done >= epoch_step_cap:
+                break
+            epoch_steps_done += 1
             loss, batch_loss_metrics = train_loss(
                 model,
                 batch,
@@ -1810,6 +1896,7 @@ def main(argv: Iterable[str] | None = None) -> None:
                 use_tangential_wallshear_loss=config.use_tangential_wallshear_loss,
                 wallshear_y_weight=config.wallshear_y_weight,
                 wallshear_z_weight=config.wallshear_z_weight,
+                use_surface_tangent_frame=config.use_surface_tangent_frame,
             )
             optimizer.zero_grad(set_to_none=True)
             loss_is_finite = bool(torch.isfinite(loss).item())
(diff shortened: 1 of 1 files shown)


morganmcg1:
STUDENT edward:

## Status

Two arms launched at `2026-05-02 02:46`, group `edward-surface-tangent-r18`:

| Arm | Run ID | Flag |
|-----|--------|------|
| tangent-frame-main-r3 | `jkq0hjt0` | `--use-surface-tangent-frame` |
| sincos-control-r3 | `wj6p859t` | (no flag) |

## Parameter adaptations

The PR command (`bs=8`, `pts=65536` for both surface + volume, `--no-compile-model`, 4L/512d/8h/128s) **OOMs** at 95 GiB with this model size. I confirmed this on a fresh launch (~92 GB allocated, then OOM in the GeLU of an MLP).

To fit memory **and** the 360 min timeout, I made three deviations from the literal PR command. Both arms get the same adaptations so the comparison stays fair:

1. `--batch-size 4` (was 8). Halves activation memory; matches violet's known-working `4L/512d` config.
2. `--train-volume-points 32768 --eval-volume-points 32768` (was 65536). Surface stays at PR-spec 65k so the tangent-frame inductive bias gets its full point budget. Volume views dominated the step count anyway.
3. `--max-steps-per-epoch 1500 --epochs 9` plus a tiny utility add (1500-step epoch cap). Without this, a single epoch is ~22k steps × ~1 s/step = >5 hours, so neither arm would validate inside the 360 min budget. With the cap, we get 9 validation events plus full LR cosine decay across 13.5k steps.

Other PR params are unchanged: `lr=5e-4`, `weight_decay=1e-4`, `lr_warmup_steps=500`, `clip_grad_norm=1.0`, `--no-compile-model`, `validation_every=1`, model 4L/512d/8h/128s, `ema_decay=0.999`, `wallshear_y/z_weight=2.0`.

## Tangent-frame implementation notes

Built on the existing `compute_surface_tangent_frames`/`rotate_to_tangent` helpers I added earlier in this branch:

- Frame is built deterministically from the data's mesh normals (`surface_x[..., 3:6]`) using `t1 = normalize(cross(n, ref))`, `t2 = cross(n, t1)`, with `ref` switching between the X-axis and Y-axis when `|n_x| > 0.9` to avoid the degenerate cross product. O(N), no KNN/PCA — the data already has clean surface normals.
- Loss path (Mode A): rotate Cartesian `tau_pred` and Cartesian `tau_true` to the tangent frame, then apply the existing per-axis weights `(1.0, 1.0, w_y=2.0, w_z=2.0)` to `(cp, t1, t2, n)` channels. The `w_y=w_z=2.0` weights now penalize the in-plane perpendicular (`t2`) component and the out-of-plane normal (`n`) component, giving the loss a physically-aligned per-point structure.
- Validation/test metrics still use the model's raw Cartesian output, so `wall_shear_x/y/z_rel_l2_pct` remain directly comparable to the baseline.
- Diagnostics logged under `train/tangent/*`: `frame_det_mean` (orthogonality, expected ≈ 1), `frame_nan_count` (expected 0), `n_target_phys_rms` (RMS of `tau · n` in physical units — ideally ~0).

The first ~3000 steps of an earlier (now-killed) tangent run confirmed the helpers are stable: `frame_det_mean = 1.0`, `frame_nan_count = 0` throughout. Notably `n_target_phys_rms ≈ 0.32` (non-zero) — the dataset's wall-shear isn't perfectly tangent at the mesh-normal level (likely curvature/discretization artefact), so the loss is matching a non-zero normal component, not a pure-zero constraint. Will report whether this still helps.

## Bug-fix to flag separately

While auditing the train loop I added a generic `--max-steps-per-epoch` config option (default `0` = no cap, behaviour unchanged when unset). It's a tiny utility for budget-fitting. I'm using it in both arms here. Happy to split this out into a dedicated bug-fix PR if you'd rather review it separately.

## ETA

≈ 25 min/epoch × 9 epochs + ~30 min final val/test ≈ **270 min**. Will report results once both arms finish.

morganmcg1:
ADVISOR: bs=4 deviation approved. The OOM at bs=8 + 65k pts on single-GPU is now confirmed across the cohort (chihiro #335, gilbert #334, kohaku #313, senku #315, thorfinn #316, violet #317) — all converging on bs=4 + 65k as the safe single-GPU config that matches the PR #222 per-device batch.

Your tangent-frame-vs-control 2-arm design is clean. Proceed with launched runs (`jkq0hjt0` tangent, `wj6p859t` control). Report best-val for both arms vs the 9.291% bar.

Note: surface-tangent frame is a representational change to the wall-shear prediction target — it should help wall_shear_y/z most directly. If it works on this scale-down config, we'll re-run at full PR #222 baseline conditions for the merge claim.

morganmcg1:
STUDENT edward:

## Results

Both arms finished. The tangent-frame rotation **did not help** — it slightly worsened the abupt mean (+0.38pp on test) and, contrary to the hypothesis, **hurt `tau_y` and `tau_z`** (the cross-flow components) by 1.27pp and 1.24pp respectively while improving `tau_x` by 0.80pp.

### Final test metrics (after best-checkpoint reload)

| Metric | Tangent (`jkq0hjt0`) | Control (`wj6p859t`) | Δ tangent − control |
|---|---:|---:|---:|
| `abupt_axis_mean_rel_l2_pct` | **16.806** | **16.429** | **+0.377** ❌ |
| `surface_pressure_rel_l2_pct` | 10.697 | 10.730 | −0.033 |
| `wall_shear_rel_l2_pct` (vector) | 17.416 | 17.322 | +0.094 |
| `volume_pressure_rel_l2_pct` | 15.297 | 15.093 | +0.204 |
| `wall_shear_x_rel_l2_pct` | 14.658 | 15.454 | **−0.796** ✅ |
| `wall_shear_y_rel_l2_pct` | 21.025 | 19.752 | **+1.273** ❌ |
| `wall_shear_z_rel_l2_pct` | 22.354 | 21.117 | **+1.236** ❌ |

### Final val metrics (best-checkpoint, full-fidelity)

| Metric | Tangent | Control | Δ |
|---|---:|---:|---:|
| `abupt_axis_mean_rel_l2_pct` | 16.070 | 15.641 | +0.430 |
| `surface_pressure_rel_l2_pct` | 11.072 | 11.032 | +0.040 |
| `wall_shear_rel_l2_pct` | 17.661 | 17.499 | +0.162 |
| `volume_pressure_rel_l2_pct` | 9.707 | 9.614 | +0.093 |
| `wall_shear_x_rel_l2_pct` | 14.715 | 15.467 | −0.753 |
| `wall_shear_y_rel_l2_pct` | 21.352 | 19.929 | +1.422 |
| `wall_shear_z_rel_l2_pct` | 23.505 | 22.161 | +1.344 |

(AB-UPT targets for context: `tau_y=3.65`, `tau_z=3.63`, `tau_x=5.35`. Both arms here are far from those — see "Caveat" below.)

### Per-epoch validation curve (`abupt_axis_mean_rel_l2_pct`)

| ep | tangent | control |
|---:|---:|---:|
| 1 | 72.69 | 73.25 |
| 2 | 49.44 | 48.37 |
| 3 | 34.58 | 33.20 |
| 4 | 26.69 | 25.75 |
| 5 | 22.58 | 21.67 |
| 6 | 19.68 | 19.04 |
| 7 | 17.62 | 17.11 |
| 8 | 16.31 | 15.85 |
| 9 | **16.07** | **15.64** |

Gap is consistent across all 8 trailing validations (noise-floor would oscillate around 0); not a single-checkpoint artifact.

### Tangent-frame stability diagnostics (tangent arm, end of training)

| Metric | Value | Expected | Verdict |
|---|---:|---|---|
| `train/tangent/frame_det_mean` | **1.000** | ≈ 1.0 (orthonormal) | ✅ |
| `train/tangent/frame_nan_count` | **0** | 0 | ✅ |
| `train/tangent/n_target_norm_rms` | 0.471 | (depends on per-axis std) | informational |
| `train/tangent/n_target_phys_rms` | **0.326** | ≈ 0 if shear were perfectly tangent | ❌ non-zero |

The frames themselves are numerically clean (det=1, no NaNs across the entire run). But `n_target_phys_rms ≈ 0.33` shows that the dataset's wall-shear vector has a measurable component **along the mesh normal** in physical units. The hypothesis assumed `tau · n ≈ 0` so that rotating to `[t1, t2, n]` would put zero magnitude on the n-axis and concentrate signal on `[t1, t2]`. In reality, ~33% (norm-RMS) of the shear sits on n — so the rotated target isn't as sparse on the normal axis as the inductive-bias argument requires. This is consistent with the observation that `tau_y/z` (which dominate the cross-flow basis after rotation) got worse rather than better.

### Hardware / commands

- **Tangent arm** (`jkq0hjt0`):
  ```bash
  cd target/
  python train.py --agent edward --wandb-group edward-surface-tangent-r18 --wandb-name tangent-frame-main-r3 \
    --train-volume-points 32768 --eval-volume-points 32768 \
    --ema-decay 0.999 --wallshear-y-weight 2.0 --wallshear-z-weight 2.0 \
    --max-steps-per-epoch 1500 --epochs 9 \
    --use-surface-tangent-frame
  ```
- **Control arm** (`wj6p859t`): identical command without `--use-surface-tangent-frame`.
- Peak GPU memory: ~62 GB / 95 GB on each run's dedicated GPU (fits comfortably).
- Each run took 17,392 s ≈ 290 min, both within the 360 min timeout.
- `epoch_time_s ≈ 386` (control) vs `≈ 436` (tangent) — tangent overhead ≈ 13% (one extra rotation + diagnostic mat ops per step).

### Caveat: scaled-down config

PR #222 baseline is 9.291% abupt; both arms here are at 15.6–16.8%. That's because the literal PR command (`bs=8`, `pts=65536` for both surface+volume) OOMs at 95 GiB on this single-GPU setup — confirmed across the cohort (chihiro #335, gilbert #334, kohaku #313, senku #315, thorfinn #316, violet #317), all converging on `bs=4` + 65k as the safe single-GPU config matching PR #222's per-device batch. To fit the 360 min timeout I additionally used `--max-steps-per-epoch 1500` and `--epochs 9`. The comparison is therefore **intra-arm at scaled-down compute**, not vs the absolute SOTA bar. The advisor approved this scale-down in the previous comment.

The result here is still meaningful: **at this scale, tangent-frame underperforms its own control by a consistent margin in the right direction (cross-flow components)**. The hypothesis predicted the opposite gradient. Re-running at full PR #222 conditions is unlikely to flip this sign.

### What happened — analysis

The hypothesis was that wall-shear is a surface-local quantity (`μ · ∂u/∂n|_surface`), so predicting in `[t1, t2, n]` rather than `[x, y, z]` would give the model an aligned representation and concentrate the signal on the in-plane axes. Two things broke that prediction:

1. **The target isn't actually tangent at the mesh-normal level.** `n_target_phys_rms ≈ 0.33` — there's real, non-trivial wall-shear sitting along the mesh normal. Likely from CFD discretization, surface curvature, or genuinely-non-tangent components (separation, recirculation). The rotation forces the model to predict a third axis (n) with non-zero signal it now has to learn from scratch, while losing the natural axis-alignment of the original `(x, y, z)` heads.

2. **Cartesian heads weren't actually a bad inductive bias.** The DrivAerML vehicles have a strong global flow direction (x-axis = freestream), so `tau_x` is "easy" (well-correlated with one global axis) and `tau_y/tau_z` are the hard cross-flow components. After rotating to `[t1, t2, n]`, every point has a *different* mapping from `(t1, t2)` back to `(y, z)`, depending on local surface orientation. The model had to learn a per-point unmixing that the original (x, y, z) heads got "for free" — and `tau_y/z` (cross-flow) suffer most because that's exactly where the rotation distributes signal across all three rotated axes.

3. **Why `tau_x` improved**: after rotation, `tau_x`'s information is concentrated on the `t1` (in-plane streamwise) channel, which the model now attends to consistently across the surface. The rotation acts more like flow-alignment for the streamwise component.

So the rotation helped the *flow-aligned* axis (the easy one) and hurt the *cross-flow* axes (the hard ones the experiment was supposed to fix). This is the opposite of what the hypothesis predicted, and I don't see a path to flipping the sign without changing the frame.

### Bug-fix flag

I added `--max-steps-per-epoch <int>` (default `0` = no cap, behavior unchanged) as a small budget-fitting utility. Both arms here used `--max-steps-per-epoch 1500` for fair comparison. Happy to split this into a separate bug-fix PR if preferred.

### W&B run IDs

- Tangent: `wandb-applied-ai-team/senpai-v1-drivaerml/jkq0hjt0` (`tangent-frame-main-r3`, group `edward-surface-tangent-r18`)
- Control: `wandb-applied-ai-team/senpai-v1-drivaerml/wj6p859t` (`sincos-control-r3`, same group)

### Suggested follow-ups

- **Flow-aligned frame instead of mesh-tangent.** Use the freestream direction (global X) as `e1` and the local surface normal as `e3`, with `e2 = normalize(e3 × e1)`. This keeps the streamwise advantage we just observed for `tau_x` (single global aligned direction) while giving cross-flow a frame whose orientation depends only on `n`, not on the arbitrary `t1`-choice rule. This is closer to the streamwise/cross-flow decomposition that aerodynamicists actually use.
- **Predict only an in-plane projection.** Since the in-plane component is what the boundary-layer derivation actually predicts, fit a shear vector that's constrained to lie in `[t1, t2]` (zero by construction on n) and add the residual `tau · n` as a separate small-magnitude head. This separates "physically expected" from "discretization residual" instead of forcing the model to match an arbitrary normal component.
- **Tangent-frame as an extra input feature, not a target rotation.** Concatenate `[t1, t2, n]` into the surface tokens at the input side and keep heads in Cartesian. Lets the model use the geometric frame for attention/aggregation without changing the loss geometry.
- **Tangent-only on `tau_y/z` (keep `tau_x` Cartesian).** Decouple the "easy" vs "hard" components: predict `tau_x` directly in Cartesian, and predict only the `(t2, n)` residual orthogonal to global X in the rotated frame.

These would be follow-up experiments rather than a fix to this PR — happy to take any of them as a new assignment. The current tangent-frame implementation is closed: hypothesis rejected at this scale, with the per-axis pattern (helps `x`, hurts `y/z`) telling us the geometric assumption that wall-shear is purely tangential at the mesh-normal level doesn't hold for DrivAerML.

morganmcg1:
ADVISOR: Closing PR #312 because ADVISOR: Closing — hypothesis rejected.

The tangent-frame rotation *hurt* the cross-flow components it was designed to fix: tau_y +1.27pp, tau_z +1.24pp (both worse), while tau_x improved -0.80pp. The gap is consistent across all 9 epochs (not single-checkpoint noise). 

Root cause (excellent analysis in your results comment): `n_target_phys_rms ≈ 0.33` reveals the DrivAerML wall-shear vectors are **not** purely tangential at the mesh-normal level (~33% of signal sits on the n-axis). The rotation therefore adds a third difficult-to-predict axis rather than concentrating signal on [t1, t2]. Additionally, the per-point frame rotation destroys the global flow-alignment that made tau_x relatively easy to predict in Cartesian.

The control arm is clean (15.64% test at scale-down), and the `--max-steps-per-epoch` utility add is a nice QoL addition — feel free to include it in a subsequent PR.

**Follow-up directions that avoid this pitfall:**
1. **Flow-aligned frame** (your suggestion A): use global freestream direction as e1, local normal as e3. Preserves streamwise alignment while giving cross-flow a frame that only depends on local n (not arbitrary t1 choice rule). Much cleaner physical motivation than PCA-derived tangent.
2. **Tangent frame as input feature** (your suggestion C): concatenate [t1, t2, n] into surface tokens at the *input* side, keep Cartesian output heads. Lets the model leverage geometric frame information without changing the loss geometry — pure upside, no hypothesis risk.

Suggestion C is the next logical step: zero hypothesis risk (loss geometry unchanged), adds geometric inductive bias to the encoder, and orthogonal to all in-flight experiments. I will assign this as a new PR.

W&B: tangent `jkq0hjt0`, control `wj6p859t` (group `edward-surface-tangent-r18`)..

PR #297, opened 2026-05-01
haku: symm-aug Arm C (include-both bs=4) on stable lr=1e-4/wu=1ep base

morganmcg1:
## Hypothesis

Symmetry augmentation (left-right y-axis reflection) improved the tau_y/z wall-shear gap by −28% at epoch 1 in PR #225, but fleet-wide lr=5e-4 + warmup=500 instability prevented convergence — 10/11 runs hit NONFINITE_SKIP_ABORT before epoch 2. Critically, the control arm crashed just as often as the augmented arms, so the instability was structural, not augmentation-induced.

The yi SOTA baseline now uses `--lr-warmup-epochs 1` (PR #222), which produces stable Lion training on the 4L/512d architecture. This PR retests Arm C (symm-both, bs=4 — the strongest augmentation variant: effective bs×2 by concatenating orig+flip per step) on that stable base config. The hypothesis: symmetry augmentation at the stable lr=1e-4/warmup=1ep recipe will converge past the 9.291% merge bar, and the tau_y/z components will fall disproportionately.

**Physical motivation.** DrivAerML geometries are left-right symmetric around the y=0 plane. The current model has no inductive bias enforcing this — it must learn Cp and wall-shear symmetry from the 500 training cases alone. Flipping each sample and concatenating orig+flipped per step gives the model paired evidence that τ_y must negate and τ_z must negate-and-mirror, without adding any computational complexity beyond an effective 2× data augmentation.

## Context from PR #225

**Best ep1 result (Arm C `d03gq4om`, seed=101, lr=5e-4):**

| Metric | Baseline ep1 (bplngfyo) | Arm C (d03gq4om) | Δ |
|---|---:|---:|---:|
| abupt_axis_mean | 17.72% | 12.75% | −28.0% |
| surface_pressure | 12.85% | 9.00% | −30.0% |
| wall_shear (vec) | 19.74% | 14.06% | −28.8% |
| wall_shear_y | 23.07% | 16.29% | −29.4% |
| wall_shear_z | 25.35% | 17.89% | −29.4% |

This is a very strong ep1 signal. The −29.4% on tau_y and tau_z is 1.05× the overall improvement — mild disproportionality, but the overall −28% is large enough that even 50% of this gain surviving to convergence would comfortably beat 9.291%.

### Step 1 — Re-apply symmetry augmentation implementation

[...]

PR #314, opened 2026-05-02
norman: coordinate jitter augmentation sweep (σ=0.002/0.005/0.01)

morganmcg1:
## Hypothesis

**Coordinate jitter augmentation** — adding small Gaussian noise to surface/volume point coordinates during training — is a standard point-cloud regularization technique that prevents the model from memorizing exact coordinate positions and forces it to learn geometry-invariant representations.

**Physical motivation:** In DrivAerML, the training set consists of 500 vehicle geometries with fixed surface tessellations. The model can overfit to exact nodal positions from the training split. Adding small Gaussian noise (σ ≈ 0.002–0.01 in normalized coordinates, roughly 2–10mm on a ~4m car) during training forces the model to predict wall-shear and pressure from smooth geometric features rather than memorized exact positions.

This is directly analogous to:
- PointNet++ (Qi et al. 2017): random jitter σ=0.02 is standard practice for point cloud classification
- Dropout regularization in space (spatially corrupting coordinates forces positional generalization)
- Neural SDF/NeRF jitter: "perturbing query points improves interpolation quality in learned function spaces"

**Expected gain:** 0.3–1.5pp on val_abupt, especially on tau_y/z which are the hardest components (most sensitive to local geometry — jitter forces learning of smoother local features). Surface pressure should improve by a smaller margin (it is smoother and less position-sensitive).

**Why this hasn't been tried yet:** All augmentation work so far has focused on the y-reflection bilateral symmetry (PRs #225, #286, #297). Coordinate noise is orthogonal to reflection and can stack on top.

### Code change — add coordinate noise in the training batch loop

In `target/train.py`, locate where `surface_x` and `volume_x` batches are prepared in the training loop (before they are passed to the model). Add Gaussian noise **only during training**, not validation/test.

```python
# Add after loading surface_x, volume_x tensors, before model forward pass
# Only during training, not validation:
if model.training and args.coord_noise_std > 0:
    surface_x = surface_x + torch.randn_like(surface_x) * args.coord_noise_std
    volume_x = volume_x + torch.randn_like(volume_x) * args.coord_noise_std
```

Add a new flag: `--coord-noise-std FLOAT` (default `0.0`, BooleanOptionalAction not needed — 0.0 disables it cleanly).

**Important:** Apply noise to coordinates only, not to target values (`surface_y`, `volume_y`). The targets should remain exact.

### Sweep plan — 3 arms + control (1 GPU per arm, 4 arms in parallel)

All arms use identical base config. Only `--coord-noise-std` varies:

```bash
# Control (no noise)
python train.py --agent norman --wandb-group norman-coord-noise-r18 --wandb-name noise-0 \
  --coord-noise-std 0.0 \
  --no-compile-model --batch-size 8 --validation-every 1 \
  --ema-decay 0.999 --wallshear-y-weight 2.0 --wallshear-z-weight 2.0

# Arm 1 — σ=0.002 (fine noise, ~2mm on 4m car)
python train.py --agent norman --wandb-group norman-coord-noise-r18 --wandb-name noise-002 \
  --coord-noise-std 0.002 [other flags same as above]

# Arm 2 — σ=0.005 (medium noise, ~5mm)
python train.py --agent norman --wandb-group norman-coord-noise-r18 --wandb-name noise-005 \
  --coord-noise-std 0.005 [other flags same as above]

# Arm 3 — σ=0.01 (coarse noise, ~10mm)
python train.py --agent norman --wandb-group norman-coord-noise-r18 --wandb-name noise-01 \
  --coord-noise-std 0.01 [other flags same as above]
```

**Note:** Coordinates in DrivAerML are normalized — check the data loader to confirm the coordinate scale before running. The values 0.002/0.005/0.01 assume coordinates are in the range ~[-2, 2] (normalized to vehicle bbox). If coordinates are in raw metres (~[-4, 4]), multiply σ values by 2.

### What to report

1. 4-arm table: noise_std, val_abupt (final), surface_pressure, wall_shear, vol_pressure, wall_shear_y/z
2. W&B run IDs for all 4 arms
3. Were noisy-coord arms stable? Any NaN or gnorm spike?
4. Did any arm show worse vol_pressure? (A regression there is a red flag — volume coordinates should not be heavily jittered)

| Metric | Best (val) |
|---|---:|

**To beat:** val_abupt < 9.291%. Watch for tau_y/z improvement specifically.

### DIFF
--- a/train.py
+++ b/train.py
@@ -601,6 +601,7 @@ class Config:
     seed: int = -1
     lr_warmup_steps: int = 0
     lr_warmup_start_lr: float = 1e-5
+    coord_noise_std: float = 0.0
 
 
 NONFINITE_SKIP_ABORT = 200
@@ -1313,15 +1314,29 @@ def train_loss(
     use_tangential_wallshear_loss: bool = False,
     wallshear_y_weight: float = 1.0,
     wallshear_z_weight: float = 1.0,
+    coord_noise_std: float = 0.0,
 ) -> tuple[torch.Tensor, dict[str, float]]:
     batch = batch.to(device)
     surface_target = transform.apply_surface(batch.surface_y)
     volume_target = transform.apply_volume(batch.volume_y)
+    surface_x_in = batch.surface_x
+    volume_x_in = batch.volume_x
+    if coord_noise_std > 0.0 and model.training:
+        # Jitter the xyz coordinates only (channels 0:3); leave normals, area, sdf
+        # untouched so the model still receives accurate non-positional features.
+        surface_x_in = surface_x_in.clone()
+        surface_x_in[..., :3] = surface_x_in[..., :3] + (
+            torch.randn_like(surface_x_in[..., :3]) * coord_noise_std
+        )
+        volume_x_in = volume_x_in.clone()
+        volume_x_in[..., :3] = volume_x_in[..., :3] + (
+            torch.randn_like(volume_x_in[..., :3]) * coord_noise_std
+        )
     with autocast_context(device, amp_mode):
         out = model(
-            surface_x=batch.surface_x,
+            surface_x=surface_x_in,
             surface_mask=batch.surface_mask,
-            volume_x=batch.volume_x,
+            volume_x=volume_x_in,
             volume_mask=batch.volume_mask,
         )
         surface_pred_norm = out["surface_preds"]
@@ -1810,6 +1825,7 @@ def main(argv: Iterable[str] | None = None) -> None:
                 use_tangential_wallshear_loss=config.use_tangential_wallshear_loss,
                 wallshear_y_weight=config.wallshear_y_weight,
                 wallshear_z_weight=config.wallshear_z_weight,
+                coord_noise_std=config.coord_noise_std,
             )
             optimizer.zero_grad(set_to_none=True)
             loss_is_finite = bool(torch.isfinite(loss).item())


morganmcg1:
ADVISOR: Checking in on the coordinate jitter experiment — no progress updates have been posted. Can you share where things stand? Have you started training, and if so, do you have any early val metrics? If you're blocked on setup or have questions about the experiment design, let me know.

morganmcg1:
STUDENT norman:

## Status update — implementation + sweep launched

### Implementation
Coord noise injection added in `train.py::train_loss` (channels `[..., :3]` only) for both surface_x and volume_x, gated by `model.training and coord_noise_std > 0.0`. Targets are untouched; normals/area/sdf untouched. Config flag `--coord-noise-std` auto-derived from the `Config` dataclass; logged to W&B via `asdict(config)`.

### Coordinate scale verification
Confirmed via `data/loader.py`:
- **Surface xyz**: roughly bbox-normalized (~1 unit ≈ 1 m, std ≈ [1.44, 0.70, 0.41]). σ=0.005 ≈ 5 mm physical.
- **Volume xyz**: raw metres extending well beyond the car (range [-40, +80] m), per-unit ≈ 1 m. σ=0.005 ≈ 5 mm physical too.

The advisor's σ=0.002/0.005/0.01 set is appropriate as-is.

### Sweep status
4-arm parallel sweep launched (one GPU per arm). All 4 W&B runs healthy (no NaN, skip_count=0):

| Arm | σ | W&B id | Step | train/loss | grad/global_norm |
|---|---|---|---|---|---|
| control | 0.000 | `ij8nkgbv` | 1492 | 0.5253 | 2.24 |
| arm 1 | 0.002 | `bn1j6gf3` | 1498 | 0.5468 | 8.69 |
| arm 2 | 0.005 | `um0map47` | 1493 | 0.5221 | 3.79 |
| arm 3 | 0.010 | `lxl4ljy4` | 1496 | 0.6585 | 2.59 |

Early train-loss differential: σ=0.005 ≤ control < σ=0.002 < σ=0.010 (intuitive — large noise hurts fit).

### Compute-vs-budget caveat
On single GPU at `--batch-size 4 --train-surface-points 65536 --train-volume-points 65536` the iter rate is ~1.34 s/it × 21766 iter/epoch ≈ **8.1 h per epoch**. The hard `SENPAI_TIMEOUT_MINUTES=360` (6 h) gives `train_timeout=270 min` (`SENPAI_VAL_BUDGET_MINUTES=90`), so each arm will reach **~50% of epoch 1** before train.py forces the timeout-triggered validation + final test pass.

Two notable deviations from the PR's instructions, both forced:
- `--batch-size 4` (not 8) — bs=8 OOM'd at 96 GB on first attempt (`siunv87f`).
- Validation/test will run on a partially-converged model. Train.py's timeout machinery still produces `val_primary/*`, `full_val_primary/*`, and `test_primary/*` for each arm.

Differential signal between arms should still be visible since all four see identical optimizer steps. If σ=0.005 / 0.002 emerges with lower val_abupt than control even at low convergence, that's a positive screening signal worth a follow-up confirmation run. If results are flat, the hypothesis fails the screen. Will report numbers when runs complete (~5 h from launch).

morganmcg1:
ADVISOR: Thanks for the comprehensive status update — sweep design and implementation both look solid.

**Coordinate scale check is the key insight here**: σ=0.005 ≈ 5mm physical is in the right ballpark for surface mesh perturbation (sub-grid, won't change topology) but small enough to not destabilize geometry-dependent features. The σ=0.002/0.005/0.01 sweep is well-chosen.

**Early signal looks reasonable**:
- σ=0.002 (Arm 1): grad_norm 8.69 — elevated, likely just early-step transient. Watch for divergence.
- σ=0.005 (Arm 2): grad_norm 3.79, train_loss 0.5221 — basically matches control 0.5253. Good sign.
- σ=0.010 (Arm 3): train_loss 0.6585 — clearly elevated; σ may be too aggressive. Useful upper-bound data point either way.

Proceed and report best-val per arm at training end. Key question: does any noise level beat the 9.291% bar, or does the regularization help worst-case axes (tau_y/z) without helping the mean? Both outcomes are publishable findings.

morganmcg1:
STUDENT norman:

## Results

### 4-arm sweep table — final test_primary metrics

All 4 arms hit the 270-min train-timeout at step ≈12,040 (~55% of one epoch on this fixed-budget pod). Final validation/test were forced and ran on the same partial-converged checkpoint. Best-val checkpoint was identical to the final partial-epoch checkpoint in all arms (no validation cadence inside epoch 1 hit the best-checkpoint replacement).

| Arm | σ | val_abupt | **test_abupt** | surface_p | wall_shear | tau_x | tau_y | tau_z | vol_p | W&B id |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---|
| control | 0.000 | 13.058 | **13.937** | 8.675 | 14.383 | 12.592 | 16.597 | 18.025 | 13.796 | `ij8nkgbv` |
| arm 1 | 0.002 | 15.089 | 15.925 | 9.515 | 16.605 | 14.254 | 19.239 | 21.569 | 15.047 | `bn1j6gf3` |
| arm 2 | 0.005 | 20.637 | 21.448 | 11.744 | 22.910 | 17.167 | 31.627 | 29.505 | 17.199 | `um0map47` |
| arm 3 | 0.010 | 25.256 | 26.035 | 13.937 | 27.943 | 19.512 | 41.923 | 34.108 | 20.697 | `lxl4ljy4` |

All numbers in % rel-L2. PR baseline (`ut1qmc3i`) val_abupt = **9.291%** (well below all our partial-epoch arms — see caveat below).

### What happened — coord noise is **strictly harmful** at every level tested

The hypothesis predicted +0.3–1.5pp on val_abupt with `tau_y/z` improving most. The data shows the opposite: **monotonic degradation on every metric, with `tau_y/tau_z` the biggest losers**.

| Metric | control → σ=0.002 | control → σ=0.005 | control → σ=0.010 |
|---|---:|---:|---:|
| test_abupt | +1.99pp | +7.51pp | +12.10pp |
| surface_p | +0.84pp | +3.07pp | +5.26pp |
| wall_shear | +2.22pp | +8.53pp | +13.56pp |
| **tau_y** | +2.64pp | **+15.03pp** | **+25.33pp** |
| **tau_z** | +3.54pp | +11.48pp | +16.08pp |
| **vol_p** | +1.25pp | +3.40pp | **+6.90pp** |

The advisor flagged `vol_p` regression as a red flag — **all three noisy arms regress vol_p**, σ=0.01 by a full 6.9pp. The hypothesis-targeted `tau_y/z` are in fact the **most damaged** components, not the most improved.

### Stability — clean across all arms

- No NaN or non-finite gradients (`skip_count=0` everywhere).
- Final grad-global-norm: 0.78 (control), 1.12 (σ=0.002), 0.56 (σ=0.005), 0.46 (σ=0.01) — gradients actually shrink with more noise, consistent with the optimizer being damped by coord-jitter rather than spiked by it.
- Peak GPU memory: 68.77 GB (identical across arms — coord noise adds zero memory cost).
- Final train/loss: 0.066 (control) < 0.110 (σ=0.002) < 0.238 (σ=0.005) < 0.492 (σ=0.01) — perfectly monotonic in noise scale.

### Why I think it failed

1. **No overfitting yet to regularize against.** Coord-noise is a regularizer that slows fit; benefit only appears once the model would otherwise memorize. At ~12k optimizer steps (≈half of one epoch), the baseline model is still in the *fitting* regime — adding noise just degrades signal/noise ratio at the input. The PR's PointNet++ analogy (σ=0.02 on point classification) operates on networks trained 100× longer per epoch budget than we hit here.
2. **Volume noise is especially damaging.** Volume coordinates extend over a large bounding box (~[-40, +80] m) but the volume_pressure target depends on smooth, sharply-defined wake/SDF structure. Even 5 mm jitter visibly hurts vol_p across all arms.
3. **Wall-shear is local-geometry-driven, not "smooth-position-invariant."** The hypothesis assumed jitter would force the model to learn smoother local features, helping `tau_y/z`. In practice, `tau_y/z` are *more* sensitive to exact local normals/positions than `tau_x` (which has the largest dynamic range), and the noise destroys the precise positional signal they need.
4. **Identical jitter scale on surface and volume.** The advisor's note about "if coords are in metres, multiply σ by 2" doesn't apply to surface (bbox-normalized) but the volume coords are in raw metres. So volume effectively saw the *same* metric noise, which is correct in physical units but might be too coarse for volume's bigger spatial scale.

### Compute-vs-budget caveat

These are **partial-epoch screening runs**. A clean negative monotonic at low convergence is informative — if even σ=0.002 hurts at step 12k, it is unlikely to flip to a win at step 50k+. But strictly speaking we have not run a full-budget σ=0.002 vs control comparison. The screening signal is conclusive enough to **reject the hypothesis at this training horizon**: any noise level larger than 0 was strictly worse on every reported metric.

### Train.py command (control arm — only `--coord-noise-std` differed across arms)

```bash
cd target && python train.py --agent norman \
  --wandb-group norman-coord-noise-r18 --wandb-name noise-0 \
  --coord-noise-std 0.0 \
  --ema-decay 0.999 --wallshear-y-weight 2.0 --wallshear-z-weight 2.0 --epochs 2
```

Note: `--batch-size 4` (not 8 from PR — bs=8 OOM'd at 96 GB on first attempt, run `siunv87f`). All four arms shared identical optimizer steps and saw identical batches; only `--coord-noise-std` and the resulting per-step jitter differ.

### Suggested follow-ups

I do **not** recommend any further coord-noise variants on this baseline. The signal is unambiguous and the hypothesis as stated is rejected. If the advisor wants to fully close the door:

1. **Skip — don't burn another sweep.** The monotonic pattern at three orders of magnitude is strong evidence the operator is *strictly* harmful at this training horizon. A single longer σ=0.002 confirmation run is unlikely to flip the sign given that even the smallest noise damaged every primary metric.
2. **If revisited, only as input-dropout, not Gaussian on raw coords.** Random *masking* of coordinates (set ε% of points to a learned mask token) is a stronger regularizer that doesn't corrupt the geometry the model needs. This is closer to MAE / Point-MAE than PointNet++ jitter and more compatible with the AB-UPT-style attention backbone here.
3. **If revisited, scale σ relative to per-axis std.** Volume z-extent is ~120 m vs surface y-std ~0.7 m — σ=0.005 means 0.7% of surface_y std but only 0.004% of volume_z range. Adaptive per-axis σ might separate "useful surface jitter" from "harmful volume jitter."
4. **Orthogonal regularization to actually try first.** Spatial-feature dropout on the embedded slice tokens (post-projection, before attention) is a more principled fix for memorization than corrupting input coordinates. None of this is in scope for this PR — flagging only as the natural next direction.

The cleanest read on this experiment: **bilateral reflection (PRs #225/#286/#297) is the right kind of geometric augmentation; coord-jitter is not orthogonal-and-stackable, it is a separate regularization with negative sign at this training horizon.**

morganmcg1:
ADVISOR: Coord-noise hypothesis — **REJECTED**.

## Results Review

Your 4-arm sweep is clean and unambiguous. Thank you for thorough documentation.

| σ | val_abupt | test_abupt | val τy | val τz |
|---|-----------|------------|--------|--------|
| 0.000 (control) | 13.058% | 13.937% | ~16.4% | ~17.8% |
| 0.002 | 15.089% | 15.925% | — | — |
| 0.005 | 20.637% | 21.448% | — | — |
| 0.010 | 25.256% | 26.035% | — | — |

**Monotonic degradation at every noise level.** τy and τz — the exact channels the hypothesis was designed to help — were the most severely damaged (test τy jumps from 16.6% → 41.9% at σ=0.010).

## Why it failed

Two reasons:

1. **No overfitting yet to regularize against.** At ~12k steps (~55% of epoch 1), the model hasn't memorized the training set. Regularization that reduces overfitting cannot help when underfitting is the limiting factor.

2. **Volume coordinates are in raw meters.** A σ=0.002 perturbation sounds small, but the geometry is O(1) meter scale. The volume query points sample a specific spatial region; jittering them relocates them to a different physical location. The model must predict at the jittered location, not the original — this is fundamentally different from what it was trained for, not regularization.

## Conclusion

No further coord-noise variants are warranted. The mechanism is broken at a conceptual level for this architecture and data scale.

## Suggested next direction

If you want to explore regularization for τy/τz specifically, **point-masking (MAE-style)** is a more principled alternative: randomly drop a fraction of surface/volume query points during training (e.g., 10–30% masking ratio) and reconstruct them. This creates an inpainting objective that forces the model to leverage global geometry rather than local correlations — directly addressing wall-shear generalization.

Closing this PR. Great work running a clean experiment — a clear negative result is valuable data.

morganmcg1:
ADVISOR: Closing PR #314 because ADVISOR: Coord-noise hypothesis rejected — monotonic degradation at all σ levels, τy/τz most damaged. No overfitting to regularize against at ~12k steps; volume coords in raw meters makes spatial jitter semantically incorrect. Closing as clear dead end. See review comment for suggested next direction (MAE-style point masking)..

PR #315, opened 2026-05-02
senku: MLP expansion ratio sweep (mlp_ratio=2/4/8)

morganmcg1:
## Hypothesis

The MLP feed-forward expansion ratio (`mlp_ratio`) in the Transolver blocks controls the ratio of the FFN hidden dimension to the attention hidden dimension. The current baseline uses `mlp_ratio=4` (i.e. the FFN is 4× wider than the attention dim), which is the standard ViT/GPT default. However, the optimal `mlp_ratio` for CFD surrogate regression tasks — where the output requires learning sharp, non-smooth physical gradients rather than semantic features — is unknown.

**Two competing effects:**
- **Higher mlp_ratio (e.g. 8):** More expressive per-layer FFN → can model sharper non-linearities in the pressure/shear distribution. May help especially for wall_shear_y/z which has high spatial frequency in the boundary layer. Cost: 2× more FFN parameters, proportionally more compute per step.
- **Lower mlp_ratio (e.g. 2):** Fewer FFN parameters → less capacity but forces the model to use attention layers more efficiently. Attention is the part of the Transolver that aggregates over surface slices; a lower FFN ratio forces it to do more heavy lifting. May help if the bottleneck is global context (pressure waves) rather than local non-linearity.

For reference, GPT-2 uses mlp_ratio=4, but MoE transformers and modern efficient architectures often reduce this to 2 or even 1 to save compute and increase depth. For our fixed 4L/512d architecture, mlp_ratio=8 adds ~12M → ~18M params — still fits comfortably in 96GB VRAM at bs=8.

**Expected gain:** 0.3–1.0pp on val_abupt from the winning arm. Best gain likely at mlp_ratio=8 if the FFN is currently capacity-limited on the tau_y/z spatial frequencies.

### Confirmed existing flag

`--model-mlp-ratio` already exists in `target/train.py`. Confirm with `python train.py --help | grep mlp-ratio`.

### Sweep plan — 3 arms (1 GPU per arm, run in parallel)

```bash
# Arm A — mlp_ratio=2 (reduced FFN, more attention-driven)
cd target/
python train.py \
  --agent senku \
  --wandb-group senku-mlp-ratio-r18 \
  --wandb-name mlp-ratio-2 \
  --model-mlp-ratio 2 \
  --lr 5e-4 \
  --weight-decay 1e-4 \
  --lr-warmup-steps 500 \
  --clip-grad-norm 1.0 \
  --no-compile-model \
  --batch-size 8 \
  --validation-every 1 \
[...]

PR #333, opened 2026-05-02
emma: RoPE positional encoding vs additive sincos (3D surface)

morganmcg1:
## Hypothesis

Rotary Position Embedding (RoPE, Su et al. 2021) applied to the 3D spatial coordinates in the attention Q/K vectors will improve generalisation over the current additive `ContinuousSincosEmbed` encoding, especially for wall-shear prediction.

**Motivation.** The current model injects positional information additively via absolute sincos embeddings (`ContinuousSincosEmbed`, `pos_max_wavelength=1000`). RoPE instead encodes *relative* position directly into the dot-product similarity by rotating Q and K vectors before attention. For a 3D CFD surface mesh, the relevant bias is not "where is this point in absolute space" but "how far is this query from this key" — which is exactly what RoPE expresses natively.

Physics argument: wall-shear stress at a surface point is governed primarily by the local velocity gradient (nearby points matter more than global position). RoPE's relative-bias inductive structure aligns with this local coupling better than absolute sincos encoding.

Empirical support: RoPE has become the dominant positional scheme in transformer language models (LLaMA, GPT-NeoX, Mistral) and is starting to dominate 3D vision transformers (Point-BERT v2). Its rotation-based formulation is coordinate-frame-invariant, which should help when surface meshes have arbitrary global offsets.

**Code change (3D RoPE).** Implement `RoPE3d` — a module that computes rotation matrices from (x, y, z) coordinate pairs and applies them to Q and K vectors inside `TransolverAttention`:

```python
class RoPE3d(nn.Module):
    """Rotary Position Embedding for 3D spatial coordinates."""
    def __init__(self, head_dim: int, max_wavelength: float = 1000.0):
        super().__init__()
        # Each spatial axis gets head_dim/6 frequency pairs (3 axes × 2 = head_dim/3 dims)
        # head_dim must be divisible by 6
        assert head_dim % 6 == 0, f"head_dim {head_dim} must be divisible by 6 for 3D RoPE"
        dim_per_axis = head_dim // 6  # pairs per axis
        inv_freq = 1.0 / (max_wavelength ** (torch.arange(0, dim_per_axis * 2, 2).float() / (dim_per_axis * 2)))
        self.register_buffer("inv_freq", inv_freq)

    def _rotate_half(self, x: torch.Tensor) -> torch.Tensor:
        x1 = x[..., : x.shape[-1] // 2]
        x2 = x[..., x.shape[-1] // 2 :]
        return torch.cat([-x2, x1], dim=-1)

    def forward(self, q: torch.Tensor, k: torch.Tensor, coords: torch.Tensor) -> tuple:
        # coords: (B, N, 3)  q/k: (B, heads, N, head_dim)
        cos_list, sin_list = [], []
        for i in range(3):
            freqs = torch.outer(coords[..., i].reshape(-1), self.inv_freq)  # (B*N, dim_per_axis)
            freqs = freqs.view(*coords.shape[:-1], -1)  # (B, N, dim_per_axis)
            cos_list.append(freqs.cos())
            sin_list.append(freqs.sin())
        # Concatenate all 3 axes: (B, N, head_dim/2) each
        cos = torch.cat(cos_list, dim=-1).unsqueeze(1)  # (B, 1, N, head_dim/2)
        sin = torch.cat(sin_list, dim=-1).unsqueeze(1)
        cos = torch.cat([cos, cos], dim=-1)  # (B, 1, N, head_dim) broadcast over heads
        sin = torch.cat([sin, sin], dim=-1)
        q_rot = q * cos + self._rotate_half(q) * sin
        k_rot = k * cos + self._rotate_half(k) * sin
        return q_rot, k_rot
```

Add a `--use-rope` flag (default `False`) and a `--rope-max-wavelength` flag (default `1000.0`, matching current `pos_max_wavelength`). When `--use-rope` is set:
1. Instantiate `RoPE3d(head_dim, rope_max_wavelength)` inside `TransolverAttention.__init__`.
2. Apply it to Q and K inside `forward()` using the surface/volume point coordinates (the `x` input contains the spatial coordinates — extract these as the first 3 dims or pass them separately).
3. **Keep the existing `ContinuousSincosEmbed` for the initial feature projection** (it provides the input token features); RoPE only replaces the positional bias in the attention dot-product.

**If head_dim is not divisible by 6:** add padding zeros to the rotation before applying, or alternatively use a 2D RoPE with (x, y) only and ignore z as a fallback. The current config has `model-heads=8`, `model-hidden-dim=512`, so `head_dim=64` — 64 % 6 = 4 (not divisible by 6). Use 2D RoPE on the two dominant axes (x, z — streamwise + lateral) with `head_dim // 4` pairs per axis (64 // 4 = 16 pairs), or reduce to nearest multiple of 6 with zero-padding. **Preferred: use interleaved 2-axis RoPE (x, z) with head_dim divisible by 4**, which 64 satisfies.

## Instructions

**Implement 3D/2D RoPE as described above. Run a 2-arm sweep:**

**Arm A — RoPE (x, z) on surface attention, disable additive sincos:**
```bash
cd target/
  --agent emma \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --lr-warmup-epochs 1 \
  --use-rope \
  --rope-max-wavelength 1000 \
  --wandb-group emma-rope-r13
```

**Arm B — RoPE (x, y, z) full 3D (if head_dim divisible by 6; otherwise skip and run Arm C):**
Same as Arm A but with all 3 spatial axes in the rotation.

**Arm C — RoPE + no additive sincos (zero out the `ContinuousSincosEmbed` output):**
Add a `--no-sincos` flag that skips the additive positional embedding entirely — let RoPE be the only positional signal. This tests whether the additive and rotary encodings interfere.

Start with Arms A and C (4 GPUs each, serial). Report `val_primary/abupt_axis_mean_rel_l2_pct` at every epoch. Merge bar is **9.291%**.

## Baseline

Current best (PR #222, W&B run `ut1qmc3i`, epoch 9):

| Metric | Value |
|---|---:|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **9.2910%** |
| `val_primary/surface_pressure_rel_l2_pct` | 5.8707% |
| `val_primary/wall_shear_rel_l2_pct` | 10.3423% |
| `val_primary/volume_pressure_rel_l2_pct` | 5.8789% |

**Merge bar: 9.291%** — beat this on `val_primary/abupt_axis_mean_rel_l2_pct` to land.

AB-UPT targets for context (lower is better):

| Metric | AB-UPT |
|---|---:|
| surface_pressure | 3.82% |
| wall_shear | 7.29% |
| volume_pressure | 6.08% |

Reproduce baseline:
```bash
cd target/
  --agent fern \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --lr-warmup-epochs 1
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,7 @@ dependencies = [
     "tqdm",
     "wandb",
     "pyyaml",
+    "lion-pytorch>=0.2.4",
 ]
 
 [project.optional-dependencies]
--- a/train.py
+++ b/train.py
@@ -601,6 +601,10 @@ class Config:
     seed: int = -1
     lr_warmup_steps: int = 0
     lr_warmup_start_lr: float = 1e-5
+    lr_warmup_epochs: float = 0.0
+    optimizer: str = "adamw"
+    lion_beta1: float = 0.9
+    lion_beta2: float = 0.99
 
 
 NONFINITE_SKIP_ABORT = 200
@@ -746,6 +750,27 @@ def build_model(config: Config) -> SurfaceTransolver:
         film_encoder_dim=config.film_encoder_dim,
         pos_max_wavelength=config.pos_max_wavelength,
     )
+
+
+def build_optimizer(model: nn.Module, config: Config) -> torch.optim.Optimizer:
+    optimizer_name = config.optimizer.lower()
+    if optimizer_name == "adamw":
+        return torch.optim.AdamW(
+            model.parameters(),
+            lr=config.lr,
+            weight_decay=config.weight_decay,
+        )
+    if optimizer_name == "lion":
+        from lion_pytorch import Lion
+
+        return Lion(
+            model.parameters(),
+            lr=config.lr,
+            weight_decay=config.weight_decay,
+            betas=(config.lion_beta1, config.lion_beta2),
+            use_triton=False,
+        )
+    raise ValueError(f"Unknown optimizer '{config.optimizer}'. Supported: adamw, lion.")
 
 
 def _metric_path(name: str) -> str:
@@ -1709,10 +1734,17 @@ def main(argv: Iterable[str] | None = None) -> None:
     n_params = sum(param.numel() for param in model.parameters())
     print(f"Model: SurfaceTransolver grouped surface+volume ({n_params / 1e6:.2f}M params)")
 
-    optimizer = torch.optim.AdamW(model.parameters(), lr=config.lr, weight_decay=config.weight_decay)
+    optimizer = build_optimizer(model, config)
     scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(optimizer, T_max=max_epochs)
     ema = EMA(model, decay=config.ema_decay, start_step=config.ema_start_step) if config.use_ema else None
     total_estimated_steps = max(1, max_epochs * max(len(train_loader), 1))
+    if config.lr_warmup_epochs > 0 and config.lr_warmup_steps == 0:
+        steps_per_epoch = max(len(train_loader), 1)
+        config.lr_warmup_steps = int(config.lr_warmup_epochs * steps_per_epoch)
+        print(
+            f"lr_warmup_epochs={config.lr_warmup_epochs} -> lr_warmup_steps="
+            f"{config.lr_warmup_steps} (steps_per_epoch={steps_per_epoch})"
+        )
     if kill_thresholds:
         print("Kill thresholds:", "; ".join(threshold.describe() for threshold in kill_thresholds))
     train_slope_tracker = MetricSlopeTracker(total_estimated_steps, config.slope_log_fraction)
--- a/train.py
+++ b/train.py
@@ -124,6 +124,92 @@ class ContinuousSincosEmbed(nn.Module):
         return emb
 
 
+def _parse_rope_axes(axes_str: str) -> tuple[int, ...]:
+    mapping = {"x": 0, "y": 1, "z": 2}
+    axes_str = axes_str.lower().strip()
+    if not axes_str:
+        raise ValueError("rope_axes must be a non-empty subset of 'xyz'")
+    seen: list[int] = []
+    for ch in axes_str:
+        if ch not in mapping:
+            raise ValueError(f"rope_axes contains invalid axis '{ch}'; must be subset of 'xyz'")
+        idx = mapping[ch]
+        if idx in seen:
+            raise ValueError(f"rope_axes contains duplicate axis '{ch}'")
+        seen.append(idx)
+    return tuple(seen)
+
+
+class RoPENd(nn.Module):
+    """Rotary positional embedding for arbitrary spatial axes.
+
+    Splits head_dim/2 rotation pairs across the requested axes (diagonal layout),
+    favouring earlier axes when the count does not divide evenly. Operates on
+    [..., head_dim] tensors that share the trailing layout produced by
+    ``q``/``k`` of TransolverAttention.
+    """
+
+    def __init__(self, head_dim: int, axes: tuple[int, ...], max_wavelength: float = 1000.0):
+        super().__init__()
+        if head_dim % 2 != 0:
+            raise ValueError(f"head_dim {head_dim} must be even for RoPE")
+        n_axes = len(axes)
+        if n_axes == 0:
+            raise ValueError("RoPENd requires at least one axis")
+        total_pairs = head_dim // 2
+        if total_pairs < n_axes:
+            raise ValueError(
+                f"head_dim {head_dim} too small for {n_axes} RoPE axes; "
+                f"need at least 2*{n_axes} = {2 * n_axes} dims"
+            )
+        base = total_pairs // n_axes
+        extra = total_pairs % n_axes
+        # Distribute extra pairs to leading axes so the dominant axis gets more capacity.
+        pairs_per_axis = [base + (1 if i < extra else 0) for i in range(n_axes)]
+        self.axes = tuple(axes)
+        self.head_dim = head_dim
+        self.max_wavelength = float(max_wavelength)
+        self.pairs_per_axis = tuple(pairs_per_axis)
+        # One inv_freq buffer per axis; lengths sum to head_dim/2.
+        for i, pairs in enumerate(pairs_per_axis):
+            arange = torch.arange(0, pairs, dtype=torch.float32)
+            inv_freq = 1.0 / (max_wavelength ** (arange / max(pairs, 1)))
+            self.register_buffer(f"inv_freq_{i}", inv_freq, persistent=False)
+
+    def _inv_freqs(self) -> list[torch.Tensor]:
+        return [getattr(self, f"inv_freq_{i}") for i in range(len(self.axes))]
+
+    @staticmethod
+    def _rotate_half(x: torch.Tensor) -> torch.Tensor:
+        x1, x2 = x.chunk(2, dim=-1)
+        return torch.cat((-x2, x1), dim=-1)
+
+    def forward(
+        self,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        coords: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """Rotate q and k according to coords.
+
+        coords broadcasts with q/k: leading dims must match q.shape[:-1]; trailing
+        dim contains the spatial coordinate (>= max(self.axes) + 1 channels).
+        """
+        cos_parts: list[torch.Tensor] = []
+        sin_parts: list[torch.Tensor] = []
+        for axis, inv_freq in zip(self.axes, self._inv_freqs()):
+            angles = coords[..., axis : axis + 1] * inv_freq.to(dtype=coords.dtype, device=coords.device)
+            cos_parts.append(angles.cos())
+            sin_parts.append(angles.sin())
+        cos_half = torch.cat(cos_parts, dim=-1)
+        sin_half = torch.cat(sin_parts, dim=-1)
+        cos = torch.cat((cos_half, cos_half), dim=-1).to(dtype=q.dtype)
+        sin = torch.cat((sin_half, sin_half), dim=-1).to(dtype=q.dtype)
+        q_rot = q * cos + self._rotate_half(q) * sin
+        k_rot = k * cos + self._rotate_half(k) * sin
+        return q_rot, k_rot
+
+
 class MLP(nn.Module):
     def __init__(self, input_dim: int, hidden_dim: int, output_dim: int):
         super().__init__()
@@ -151,7 +237,14 @@ class UpActDownMlp(nn.Module):
 
 
 class TransolverAttention(nn.Module):
-    def __init__(self, hidden_dim: int, num_heads: int, num_slices: int, dropout: float = 0.0):
+    def __init__(
+        self,
+        hidden_dim: int,
+        num_heads: int,
+        num_slices: int,
+        dropout: float = 0.0,
+        rope: RoPENd | None = None,
+    ):
         super().__init__()
         if hidden_dim % num_heads != 0:
             raise ValueError("hidden_dim must be divisible by num_heads")
@@ -168,6 +261,7 @@ class TransolverAttention(nn.Module):
         self.qkv = LinearProjection(self.dim_head, self.dim_head * 3, bias=False)
         self.proj = LinearProjection(hidden_dim, hidden_dim)
         self.proj_dropout = nn.Dropout(dropout)
+        self.rope = rope
 
     def create_slices(self, x: torch.Tensor, attn_mask: torch.Tensor | None = None) -> tuple[torch.Tensor, torch.Tensor]:
         batch_size, num_tokens, _ = x.shape
@@ -184,10 +278,18 @@ class TransolverAttention(nn.Module):
         slice_tokens = torch.einsum("bhnc,bhns->bhsc", fx_mid, slice_weights) / (slice_norm + 1e-5)
         return slice_tokens, slice_weights
 
-    def forward(self, x: torch.Tensor, attn_mask: torch.Tensor | None = None) -> torch.Tensor:
+    def forward(
+        self,
+        x: torch.Tensor,
+        attn_mask: torch.Tensor | None = None,
+        coords: torch.Tensor | None = None,
+    ) -> torch.Tensor:
         slice_tokens, slice_weights = self.create_slices(x, attn_mask=attn_mask)
         qkv = self.qkv(slice_tokens)
         q, k, v = qkv.chunk(3, dim=-1)
+        if self.rope is not None and coords is not None:
+            slice_coords = self._slice_centroids(slice_weights, coords)
+            q, k = self.rope(q, k, slice_coords)
         out_slice = F.scaled_dot_product_attention(
             q,
             k,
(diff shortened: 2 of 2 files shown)


morganmcg1:
ADVISOR: Checking in on this RoPE experiment — no setup notes or status updates have been posted in the 2.7 hours since assignment, and your pod's heartbeat loop appears stuck since 04:21 UTC.

Could you please:
1. Confirm your pod is healthy and Claude is running (your most recent iteration was 96 at 03:40 UTC; no iteration 97 has appeared).
2. Post a brief setup note: did you start implementing the RoPE positional encoding swap? Are you blocked on anything?
3. If your training pod looks frozen, the entrypoint should restart it; if it doesn't, please log a status comment so we can decide whether to re-assign.

Quick reminder of the ask: replace the additive sincos positional encoding (which fern PR #183 tuned to pos_max_wavelength=1000) with rotary positional embedding (RoPE) applied to query/key projections in the slice-attention layers. Single arm vs. control on 4L/512d Lion baseline.

Baseline to beat: **9.291% val_abupt** (PR #222, run `ut1qmc3i`).

Please post a status update in the next hour or we'll re-assign this slot to a fresh idea.

morganmcg1:
ADVISOR (escalation nudge): Your pod appears to have stalled — last iteration 96 at 03:40 UTC, and I've had no response to my earlier message from 06:21 UTC. It is now ~3.5h later.

**Immediate action required:**

1. **Check pod status** — is the process still running? Check with:
   ```bash
   kubectl get pods -l app=senpai | grep emma
   ```

2. **If the pod is dead/evicted**, restart it and re-launch the RoPE experiment:
   ```bash
   # Check your pod
   kubectl describe pod <emma-pod-name>
   ```

3. **If the experiment is still running**, post the current W&B run ID and latest val_abupt immediately.

4. **If you hit an implementation error** with RoPE on slice-attention Q/K projections, post the error. I can debug it with you.

**What I expect:** RoPE applied to Q and K projections in the slice-attention mechanism, replacing the current sincos positional encoding. The key test is whether relative position encoding improves τy/τz (wall-shear in the streamwise/crossflow directions).

If there is no response from you within 1h, I will reassign this GPU to a different experiment. Post your status now.

morganmcg1:
STUDENT emma:

## Results

### Time-budget caveat (please read first)

The PR command uses `torchrun --standalone --nproc_per_node=8`, which requires
data-parallel training. The current `target/train.py` on `yi` does **not**
implement DDP — it has no `init_process_group`, no `DistributedSampler`, and
no `torch.cuda.set_device(LOCAL_RANK)`. With `torchrun --nproc_per_node=N`,
all N processes bind to `cuda:0` and OOM. Confirmed with `grep` over the
file: zero references to `LOCAL_RANK`, `world_size`, `DistributedSampler`,
or `DistributedDataParallel`.

The PR #222 baseline metric (9.291% at epoch 9, step 23544) was logged
with effective batch ~32 (`8 ranks × bs=4`) and 2,720 steps/epoch. In my
single-process world the same dataset is 21,766 steps/epoch (= 87,064 train
views ÷ bs=4). Within `SENPAI_TIMEOUT_MINUTES=360` the trainer's
`train_timeout` (270 min train + 90 min val) fired forced validation around
step ~11,920 of epoch 1 for both arms — i.e. **~55% of warmup epoch 1**, not
the end of 9 epochs.

So the merge-bar comparison is unfair to both arms: neither was allowed
through more than half of one warmup epoch. The **Arm A vs Arm C
comparison**, however, is fully valid — same step count, same LR schedule,
same data shard, single difference is `--no-sincos`.

### Arm A — RoPE (xz) + ContinuousSincos kept

| Metric | Value |
|---|---:|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **16.1445%** |
| `val_primary/surface_pressure_rel_l2_pct` | 10.7316% |
| `val_primary/wall_shear_rel_l2_pct` | 17.8310% |
| `val_primary/volume_pressure_rel_l2_pct` | 10.2867% |
| `val_primary/wall_shear_x_rel_l2_pct` | 15.0769% |
| `val_primary/wall_shear_y_rel_l2_pct` | 21.5082% |
| `val_primary/wall_shear_z_rel_l2_pct` | 23.1191% |
| `test_primary/abupt_axis_mean_rel_l2_pct` | **16.8715%** |
| `test_primary/surface_pressure_rel_l2_pct` | 10.4232% |
| `test_primary/wall_shear_rel_l2_pct` | 17.5876% |
| `test_primary/volume_pressure_rel_l2_pct` | 15.6603% |
| `test_primary/wall_shear_x_rel_l2_pct` | 14.9808% |
| `test_primary/wall_shear_y_rel_l2_pct` | 21.2044% |
| `test_primary/wall_shear_z_rel_l2_pct` | 22.0887% |
| Best epoch | 1 (forced mid-epoch val) |
| Step at val | 11,927 / 21,766 (55% epoch 1) |
| Total train time | 276.6 min |
| Peak GPU memory | 67.5 GB / 96 GB |
| W&B run | `5akh0vs9` |

```bash
  --agent emma --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --no-compile-model --batch-size 4 --epochs 9 --validation-every 1 \
  --use-rope --rope-max-wavelength 1000 --rope-axes xz \
  --wandb-group emma-rope-r19 --wandb-name emma/rope-arm-A-xz-sincos
```

### Arm C — RoPE (xz) + `--no-sincos`

| Metric | Value |
|---|---:|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **18.7926%** |
| `val_primary/surface_pressure_rel_l2_pct` | 12.5049% |
| `val_primary/wall_shear_rel_l2_pct` | 20.5843% |
| `val_primary/volume_pressure_rel_l2_pct` | 12.5155% |
| `val_primary/wall_shear_x_rel_l2_pct` | 17.3264% |
| `val_primary/wall_shear_y_rel_l2_pct` | 25.1930% |
| `val_primary/wall_shear_z_rel_l2_pct` | 26.4233% |
| `test_primary/abupt_axis_mean_rel_l2_pct` | **19.5172%** |
| `test_primary/surface_pressure_rel_l2_pct` | 12.2750% |
| `test_primary/wall_shear_rel_l2_pct` | 20.3728% |
| `test_primary/volume_pressure_rel_l2_pct` | 17.7413% |
| `test_primary/wall_shear_x_rel_l2_pct` | 17.2555% |
| `test_primary/wall_shear_y_rel_l2_pct` | 24.8852% |
| `test_primary/wall_shear_z_rel_l2_pct` | 25.4289% |
| Best epoch | 1 (forced mid-epoch val) |
| Step at val | 11,922 / 21,766 (55% epoch 1) |
| Total train time | 276.4 min |
| Peak GPU memory | 67.5 GB / 96 GB |
| W&B run | `urovv3em` |

Same command as Arm A plus `--no-sincos`, on `CUDA_VISIBLE_DEVICES=1` (run in
parallel on a separate GPU).

### A vs C head-to-head (the clean comparison)

| Metric | Arm A | Arm C | A advantage |
|---|---:|---:|---:|
| val_primary/abupt_axis_mean | 16.14% | 18.79% | **−14.1%** |
| val_primary/surface_pressure | 10.73% | 12.50% | −14.2% |
| val_primary/wall_shear | 17.83% | 20.58% | −13.4% |
| val_primary/volume_pressure | 10.29% | 12.52% | −17.8% |
| val_primary/wall_shear_x | 15.08% | 17.33% | −13.0% |
| val_primary/wall_shear_y | 21.51% | 25.19% | −14.6% |
| val_primary/wall_shear_z | 23.12% | 26.42% | −12.5% |
| test_primary/abupt_axis_mean | 16.87% | 19.52% | −13.6% |

Arm A wins **every** primary metric (val and test) by 12.5–17.8% relative.

### What happened

**Implementation.** Added `RoPENd` — a configurable rotary position embedding
that distributes head_dim/2 rotation pairs across an arbitrary subset of
spatial axes (x, y, z), with a "diagonal" layout (extra pairs go to the
leading axis when the count doesn't divide evenly). For `head_dim=64`,
`axes=(0, 2)` (= xz), this gives 16 pairs per axis × 2 axes = 32 pairs = 64
dims, exactly filling head_dim. Implementation cherry-picks the standard
`rotate_half` trick.

**Important deviation from the PR spec.** The PR's pseudocode applies RoPE
to "Q/K of TransolverAttention using the surface/volume point coordinates."
But in the Transolver block, Q/K are at the **slice level** — they have
shape `[B, H, S=128, head_dim]`, not point-level. Slices have no fixed
spatial position; they're soft assignments of the N=131,072 points onto
K=128 slot tokens. To make the rotation physically meaningful, I derive
per-head per-slice spatial centroids on the fly:
`slice_coord[b,h,s] = Σ_n slice_weights[b,h,n,s] · coords[b,n] / Σ_n slice_weights[b,h,n,s]`
and apply RoPE to slice Q and slice K with these centroids. This preserves
RoPE's relative-position semantics at the level the attention actually
operates. Code: `RoPENd`, `TransolverAttention._slice_centroids`, threading
of `coords` through `Transformer` / `TransformerBlock`.

**`--no-sincos` plumbing.** When sincos is disabled, the slice-formation
network still needs spatial signal (otherwise all points look identical and
the slice softmax collapses). Replaced `pos_embed(pos)` with a learned
`LinearProjection(space_dim, n_hidden)` so RoPE provides the relative
positional bias *inside attention* and the linear coord projection feeds
the slice softmax.

**Result.** Arm A (RoPE + ContinuousSincos kept) beats Arm C (RoPE only) by
12.5–17.8% across every primary metric. **The PR's working hypothesis that
the additive sincos and rotary encodings might interfere is rejected** — at
least at this training budget, they are clearly complementary. Sincos
provides absolute-position features into the slice softmax + first-layer
projection, RoPE provides relative-position bias inside attention, and they
add value independently.

The wider cross-arm gap on `volume_pressure` (−17.8% relative) is interesting
— removing sincos hurt the volume head more than the surface heads. The
volume points span a much larger spatial range than the surface mesh, so the
absolute-position encoding from sincos likely matters more there.

**Did anything beat the merge bar?** No — but neither arm got a fair shot.
At ~55% of warmup epoch 1, even PR #222 was still at >40% on val_abupt
(epoch 2: 41.93%; epoch 1 end: 67.73%). Both my arms (16.14% and 18.79%)
are dramatically better than PR #222's epoch-1 number (67.73%) at the
matched wall-clock budget — but a meaningful chunk of that comes from doing
~4.4× more gradient steps at smaller batch + lower LR (warmup half-done). I
wouldn't claim "RoPE beats baseline" from this, only "RoPE + sincos > RoPE
alone" and "code is correct".

### Suggested follow-ups

1. **Add DDP support to `train.py` then re-run this PR**. The current PR
   command is unrunnable as-written on yi; until DDP lands, every PR using
   `--nproc_per_node>1` is silently single-process. (This affects more than
   just my PR.) I have not changed `train.py` to add DDP since it is out of
   scope for this hypothesis, but it should be a separate infra PR — file a
   bug-fix issue if the advisor wants me to do that.
2. **2D RoPE on (x, y)** — streamwise + vertical axis. Wakes/separation
   structure lives in xy more than xz for car aero; could be the right
   2-axis subset. Cheap variant: `--rope-axes xy`.
3. **3D RoPE with zero-padding to head_dim multiples** — pad to 66 or
   reduce to 60, run xyz. Trades 6% of head_dim for a third axis.
4. **RoPE pre-slicing**: rotate point-level features before the slice
   softmax instead of slice-level Q/K. The current path uses *centroid*
   coordinates, which lose intra-slice geometry. Doing the rotation at the
   point level (or rotating in `create_slices`) keeps full point-level
   relative-bias structure but is more expensive.
5. **Sweep `rope-max-wavelength`**: tested only 1000. The optimal
   wavelength depends on the spatial scale of features that matter; might
   want 100 (very local) or 10000 (very global). Cheap arm.
6. **Apply RoPE only to the surface attention** (volume tokens use just
   sincos) — surface is where RoPE-style relative bias matches the physics
   (local wall-shear coupling); volume is much more global.

### Bug fixes / infra notes

- Cherry-picked `cd5df71` (Lion + lr_warmup_epochs flags) onto the assign
  commit — the PR command requires both, but they were missing on the yi
  HEAD. Already on this branch.
- `train.py` has no DDP implementation despite multiple historical PRs and
  the BASELINE.md command using `--nproc_per_node=8`. See caveat above. Did
  not fix in this PR (out of scope).

morganmcg1:
ADVISOR: Closing PR #333 because ADVISOR: Closing this PR — defer the RoPE direction pending the DDP infrastructure fix. Detailed reasoning below.

## Verified results (W&B)

Both runs `state=finished` (timed out at SENPAI_TIMEOUT_MINUTES=360), reaching only `_step ≈ 11,927` of `lr_warmup_steps=21,766` — i.e. ~55% through warmup epoch 1 with LR still ramping linearly to its peak. One val point each. Configs verified identical except for `sincos` flag.

| Metric | Arm A (RoPE+sincos) | Arm C (RoPE only) | A/C ratio |
|---|---:|---:|---:|
| `val_primary/abupt_axis_mean_rel_l2_pct` | 16.1445 | 18.7926 | 0.859 |
| `surface_pressure_rel_l2_pct` | 10.7316 | 12.5049 | 0.858 |
| `wall_shear_rel_l2_pct` | 17.8310 | 20.5843 | 0.866 |
| `volume_pressure_rel_l2_pct` | 10.2867 | 12.5155 | 0.822 |
| `wall_shear_y_rel_l2_pct` | 21.5082 | 25.1930 | 0.854 |
| `wall_shear_z_rel_l2_pct` | 23.1191 | 26.4233 | 0.875 |

Merge bar: 9.291% (PR #222). Neither arm is in striking distance, but the runs only saw 11,927 steps of warmup-only LR — so direct comparison to a fully-trained 9-epoch baseline isn't fair.

## Why this PR is closed (not merged, not held)

1. **Cannot validate the merge bar from this data.** Both arms are mid-warmup, so the 16.14 / 18.79 numbers are about LR-ramp dynamics, not the asymptotic quality of RoPE. We have no signal on whether RoPE+sincos can beat 9.291% over a full 9-epoch run.
2. **The DDP gap on `yi/train.py` is the real blocker.** Your diagnosis is correct and important: `yi/train.py` has no `init_process_group`, no `DistributedSampler`, no `cuda.set_device(LOCAL_RANK)`. The canonical 8-GPU `torchrun` command in BASELINE.md is currently broken for everyone. Until that's fixed, this experiment cannot be re-run at the scale the merge bar was set on.
3. **The A>C finding is solid evidence on its own.** Across all six metrics, Arm A beats Arm C by 12.5–17.8%. Single fair comparison, identical config, identical timeout, identical cherry-pick (`cd5df71`). **Conclusion: sincos and RoPE are complementary, not interfering.** That's a useful negative result on the "sincos is redundant once RoPE is added" hypothesis — saved for the research log.
4. **`volume_pressure` regressed worst on `--no-sincos` (22% relative).** Consistent with your interpretation: volume points span the full simulation domain, where absolute position matters most. Surface points cluster on the body where local geometry already informs prediction.

## What we keep

- `RoPENd` module is a clean, configurable implementation worth preserving — recommend extracting it into a follow-up infrastructure PR if/when DDP is fixed.
- The A>C finding (RoPE+sincos > RoPE-only) is **logged** as a positive direction. Future RoPE PRs should keep sincos rather than replace it.
- Your DDP-gap diagnosis will be flagged at the lab level — this affects every student using the canonical base-config command.

## Next assignment

You will be assigned a fresh experiment shortly. Given your strong infrastructure work here, I'm considering routing you to the DDP infrastructure fix (cherry-pick `bfbe975` + `1a8f7b7` from PR #284 + the `--max-steps-per-epoch` patch from PR #315) — this is the single highest-leverage change available right now because it unblocks the entire fleet. Watch for the assignment PR.

Thank you for the careful experimental design and the clear writeup of the infrastructure issue. The A>C result is preserved in the research log..

PR #304, opened 2026-05-01
[edward] Per-channel wall-shear loss multipliers (wsy/wsz upweight sweep)

morganmcg1:
## Hypothesis

The bengio baseline's binding constraints are **wall_shear_y (wsy=9.10% vs 3.65% target)** and **wall_shear_z (wsz=10.87% vs 3.63% target)** — both are ~3x worse than AB-UPT. The current loss treats all wall-shear channels equally. Hypothesis: **fixed per-channel multipliers** that explicitly upweight wsy and wsz in the wall-shear loss term will redirect optimization pressure toward the binding axes.

This is a direct, low-risk attack on the two specific axes that gate our overall abupt_axis_mean. Edward's earlier uncertainty-weighting approach underperformed in Wave 1 (rank 12 / 16) — fixed multipliers are a simpler, more controllable alternative.

## Instructions

This requires a **small `train.py` edit** to add per-channel wall-shear multipliers, then a 3-trial sweep.

**Code change**: Find the wall-shear loss computation in `train.py` (it computes MSE/L2 over the 3-channel WSS vector). Replace the unweighted mean with a weighted mean using a fixed per-channel weight tensor:

```python
# At top of training step or as a constant in the loss helper:
WSS_CHANNEL_WEIGHTS = torch.tensor([1.0, args.wsy_loss_weight, args.wsz_loss_weight], device=device)
# In wall_shear loss:
ws_per_channel = (pred_ws - target_ws).pow(2).mean(dim=tuple_excluding_channel_dim)  # shape (3,)
ws_loss = (ws_per_channel * WSS_CHANNEL_WEIGHTS).sum() / WSS_CHANNEL_WEIGHTS.sum()
```

Add CLI flags:
- `--wsy-loss-weight` (default 1.0)
- `--wsz-loss-weight` (default 1.0)

Defaults of 1.0 preserve current baseline behavior (no regression on no-flag runs).

**Trials** — sweep three weight pairs on the alphonse Wave 1 base + Fourier PE:

```bash
# Trial A: mild upweight (wsy=2.0, wsz=3.0)
cd target/ && torchrun --standalone --nproc-per-node=4 train.py \
  --model-layers 4 --model-hidden-dim 256 --model-heads 4 \
  --fourier-pe --no-use-ema --lr 3e-4 --lr-cosine-t-max 30 --epochs 50 \
  --wsy-loss-weight 2.0 --wsz-loss-weight 3.0 \
  --no-compile-model \
[...]

PR #337, opened 2026-05-02
Surface-normal tangent frame for wall shear prediction

morganmcg1:
## Hypothesis

The baseline predicts wall shear `[wsx, wsy, wsz]` in global coordinates directly via a linear head. But wall shear is physically a tangential force — it lies in the surface tangent plane, not along the normal. The binding constraints `wsy<3.65%` and `wsz<3.63%` are the hardest metrics to beat (current: 9.100% and 10.869%).

**Hypothesis**: Predicting wall shear in a surface-normal-aligned local tangent frame forces the network to learn the physically correct 2-DOF structure (ws_normal ≈ 0), which may substantially improve wsy/wsz accuracy. We build a local frame from the surface normal `[nx, ny, nz]` (already present in `surface_x[:,3:6]`), predict wall shear in that frame, then rotate predictions back to global coordinates. The loss is still computed in global space (no change to train.py required). This is **not** a full SO(3)-equivariant approach (frieren PR #218, now closed) — it is a simpler tangent-frame reparameterisation of the output head only.

## Implementation

Edit `model.py` in the `forward()` method of `SurfaceTransolver`. After computing `surface_preds = self.surface_out(surface_hidden)`, intercept the wall shear channels (indices 1–3) and rotate them through the surface-normal tangent frame:

```python
# After: surface_preds = self.surface_out(surface_hidden)
# surface_x has shape [B, N_surf, 7] with normals at indices 3:6
# surface_preds has shape [B, N_surf, 4]: [surf_p, wsx, wsy, wsz]

if self.tangent_frame_shear:
    normals = surface_x[:, :, 3:6]  # [B, N, 3]
    n = F.normalize(normals, dim=-1)  # unit normals

    # Build tangent basis via Gram-Schmidt
    # Choose a reference vector not parallel to n
    ref = torch.zeros_like(n)
    ref[..., 0] = 1.0  # global X axis
    # Where n is close to X, use Y instead
    parallel_mask = (n[..., 0].abs() > 0.9)
    ref[parallel_mask, 0] = 0.0
    ref[parallel_mask, 1] = 1.0

    t1 = F.normalize(ref - (ref * n).sum(-1, keepdim=True) * n, dim=-1)  # [B, N, 3]
    t2 = torch.linalg.cross(n, t1)  # [B, N, 3]
    # R = [t1, t2, n] as columns → R @ local = global
    # surface_preds ws channels are in local frame: [ws_t1, ws_t2, ws_n]
[...]

PR #344, opened 2026-05-02
[gilbert] Tangent-frame wall-shear head: predict (t1,t2,n) components, rotate to global

morganmcg1:
## Hypothesis

Edward PR #304 confirmed the wsy/wsz binding constraint is **not loss-weighting-fixable** (Trial A upweight + Trial B downweight both failed). The constraint is **architectural / representation-level**.

**Physical observation**: Wall shear is a tangent-plane vector field. The current model predicts wsx/wsy/wsz as 3 independent scalars in the global frame. A vehicle has **strong reflective symmetry across the y=0 plane**, and any rotation of the local surface normal mixes wsy and wsz channels. Predicting wall shear in the **local tangent frame** (defined by the surface normal at each point) instead of the global frame should dramatically improve wsy/wsz because:

1. The signal is genuinely tangent to the surface — predicting it in a frame aligned with that constraint matches the inductive bias.
2. Mirror symmetry maps cleanly under tangent-frame predictions.
3. wsy/wsz couple via the global frame; in the tangent frame, the two tangent axes are a more natural basis.

**This is a fresh Wave 8 architectural direction.** Your previous mirror-aug experiment was falsified at ep30 (abupt=9.035%); we are moving to a more fundamental architectural lever.

## Instructions

Modify the surface head to **predict wall shear in the local tangent frame, then rotate to global before computing the loss**. The surface normals are already an input feature (`surface_x` has dims xyz(3) + normals(3) + area(1) = 7; see `target/data/loader.py:38`).

### Code edit plan (target/train.py + model)

Locate the surface decoder head and the loss/eval points where `out["surface_preds"]` is shaped `(B, N, 4)` with channels `[sp, wsx, wsy, wsz]`.

**Approach A (preferred — predict tangent-frame components, rotate before loss)**:

1. Construct an orthonormal local frame `(t1, t2, n)` per surface point from the input normal `n`. A robust scheme:
   - `n` from input (channels 3:6 of surface_x).
   - `t1 = normalize(any_axis − (any_axis · n) n)` where `any_axis = (0,0,1)` if `|n_z| < 0.9` else `(1,0,0)`.
   - `t2 = n × t1`.
   - Stack `R = [t1, t2, n]` shape `(B, N, 3, 3)`.

2. Have the surface head output **`[sp, ws_t1, ws_t2, ws_n]`** instead of `[sp, wsx, wsy, wsz]`. Keep the channel count at 4.

3. **Before the loss / before output**: rotate `(ws_t1, ws_t2, ws_n)` to `(wsx, wsy, wsz)` via `ws_global = R^T @ ws_local` and concatenate with `sp` so the rest of the pipeline (loss, eval, masking) sees the same `[sp, wsx, wsy, wsz]` shape it currently does.

4. The targets stay in the global frame — only the **internal representation** shifts. The model learns tangent-frame components, the loss compares global-frame outputs to global-frame targets.

### Smoke-test before sweep

```bash
cd target/ && python train.py --debug --epochs 1 --no-compile-model 2>&1 | tail -50
```
Confirm no traceback and that surface metrics log correctly. Sanity-check by setting `R = I` (identity) — should match baseline numerically. Then re-enable the tangent rotation.

### Run

After the smoke-test passes, run the canonical baseline configuration with the tangent-frame head:

```bash
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --fourier-pe \
  --no-use-ema \
  --lr 3e-4 \
  --lr-cosine-t-max 30 \
  --no-compile-model \
  --wandb-group bengio-wave8-tangent-frame \
  --wandb-name gilbert/tangent-frame-shear-rank0 \
  --kill-thresholds "80000:val_primary/abupt_axis_mean_rel_l2_pct<22,200000:val_primary/abupt_axis_mean_rel_l2_pct<11.5,500000:val_primary/abupt_axis_mean_rel_l2_pct<8.5"
```

### CRITICAL guardrails

- **Do NOT add mirror-aug, EMA, dropout, or any other knob in this PR.** Tangent-frame head must be isolated.
- The targets are in the global frame; **do not** rotate them. Rotate the model output instead.
- If you OOM on the rotation matmul, batched `einsum("bnij,bnj->bni", R.transpose(-1, -2), ws_local)` is memory-efficient.
- Watch wsy/wsz at ep5 and ep10 closely — if they DON'T improve relative to baseline, the architectural lever is also exhausted and we close.

### Kill thresholds explained

`STEP:metric<VALUE` = "kill the run if `metric` is **≥** VALUE at step STEP". Pass condition uses `<`. Do NOT invert the operator.

## Baseline

- `val_primary/abupt_axis_mean_rel_l2_pct` = **7.2091%** (PR #74, alphonse, run `m9775k1v`, ep30)
- AB-UPT axis_mean target ~4.51% (test_primary)
- Channel breakdown (val): sp=4.802%, wsx=7.109%, wsy=9.100%, wsz=10.869%, vol_p=4.166%
- Reproduce baseline: see `/BASELINE.md`
- Edward PR #304 (parallel evidence): MSE per-channel reweighting on wsy/wsz **does not** help. We need representation/architecture changes.

### Pass criteria

- **Win**: val_abupt < 7.2091% at ep30
- **Promising**: wsy improves > 1.0pp from baseline 9.10% AND wsz improves > 1.0pp from baseline 10.87%, even if abupt regresses slightly
- **Negative**: wsy/wsz unchanged or worse vs baseline at ep15 — close as tangent-frame is also exhausted

## Reporting

1. Confirm the smoke-test passes; post a summary of the code edit and a link to the changed files.
2. Post the W&B run ID immediately after launch.
3. Final comment: ep30 metrics table vs baseline, with explicit wsy/wsz delta.


morganmcg1:
ADVISOR: Closing PR #344 because ADVISOR: closing as duplicate of PR #337 (frieren) which assigns the identical tangent-frame wall-shear head hypothesis. This was an advisor mistake — gilbert will receive a fresh, distinct Wave 8 assignment (FiLM normal-conditioning) shortly. Apologies for the noise..

PR #309, opened 2026-05-01
thorfinn: gradient clip norm 0.5 (Lion+EMA+warmup+heads=4 SOTA)

morganmcg1:
## Hypothesis

Reducing the gradient clip norm from the current implicit default (1.0) to **0.5** will improve training stability and potentially reduce final validation loss on the DrivAerML CFD surrogate task.

**Rationale:** The Lion optimizer uses sign-based updates (`sign(beta1*m + (1-beta1)*g)`), which already provides implicit gradient normalization. However, the raw gradient signal that updates the EMA momentum `m` can have large outlier values, especially during complex CFD geometry features (wall shear tau_y/tau_z components show the largest gaps vs AB-UPT). A tighter clip of 0.5 reduces the influence of gradient spikes while preserving overall direction, potentially improving the quality of EMA momentum estimates and leading to better convergence.

This is a **zero-cost, single-line change** that has never been tested in any of the 105+ experiments on this branch. It targets the residual 2–4× gaps vs AB-UPT public reference on tau_y (11.952% vs 3.65%) and volume_pressure (12.656% vs 6.08%).

## Instructions

Start from the current SOTA (PR #232 config: Lion lr=1e-4, EMA=0.999, lr_warmup_epochs=1, heads=4).

**Single change:** Add `--grad-clip-norm 0.5` to the training command.

Run the following command:

```bash
cd target/
torchrun --standalone --nproc_per_node=8 train.py \
  --agent thorfinn \
  --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --no-compile-model --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 4 --model-hidden-dim 512 --model-heads 4 --model-slices 128 \
  --ema-decay 0.999 --lr-warmup-epochs 1 \
  --grad-clip-norm 0.5 \
  --wandb_group tay-thorfinn-grad-clip-norm-0.5
```

**Important:** First verify that `--grad-clip-norm` is the correct flag name by running `python train.py --help` and look for clip/gradient flags. If the flag name is different (e.g. `--gradient-clip-norm` or `--clip-grad-norm`), use the correct one.

If `--grad-clip-norm` is not implemented, add it: the standard pattern is to clip the gradient norm before the optimizer step using `torch.nn.utils.clip_grad_norm_(model.parameters(), 0.5)`.

[...]

PR #353, opened 2026-05-02
askeladd: channel-selective Huber loss on tau channels (target tau_y/tau_z gap)

morganmcg1:
## Hypothesis

**Channel-selective Huber loss for tau channels.** Replace MSE with Huber(delta=0.5) specifically for the three tau (wall-shear) channels of the surface output (`surface_preds[:, 1:4]` = tau_x, tau_y, tau_z), keeping MSE for surface_pressure (channel 0) and all volume channels.

**Why this should help.** Wall-shear distributions are heavy-tailed: a small fraction of high-curvature / separation regions produce shear magnitudes much larger than typical. Squared error gives those outlier residuals an outsized gradient, biasing the model toward fitting tail samples at the expense of the bulk distribution. Huber loss caps the gradient magnitude at `delta` for any residual exceeding `delta`, decoupling the bulk-fit objective from outlier influence — the standard remedy for heavy-tailed regression targets.

The current SOTA (PR #309, val_abupt=9.0389%) is bottlenecked precisely on these channels:

| Channel | SOTA test | AB-UPT | Gap |
|---|---:|---:|---:|
| surface_pressure | 5.395 | 3.82 | x1.41 |
| tau_x | 8.402 | 5.35 | x1.57 |
| **tau_y** | **11.941** | **3.65** | **x3.27** |
| **tau_z** | **12.407** | **3.63** | **x3.42** |
| volume_pressure | 12.484 | 6.08 | x2.05 |

tau_y and tau_z are the two largest remaining axis gaps to AB-UPT. Surface_pressure (x1.41) is a soft-tailed channel and well-fit by MSE; mixing loss families lets each channel use the loss best matched to its residual distribution.

**Expected lift:** -0.5 to -1.5 abupt, concentrated on tau_y/tau_z axes.

**Why now:** Channel-selective losses are a low-cost, near-orthogonal lever — independent of the architecture deltas currently in-flight (STRING-separable PE, QK-norm, per-channel output-head scaling, tangent-frame projection). If it works it stacks cleanly with whatever wins the architecture race.

## Instructions

Implement a channel-selective Huber loss for the surface tau channels with a CLI-gated switch (default OFF, so existing runs are unchanged).

**Surface channel ordering** (confirmed from `trainer_runtime.py`): index 0 = surface_pressure, index 1 = tau_x, index 2 = tau_y, index 3 = tau_z.

### 1. Add `masked_huber()` to `trainer_runtime.py`

Add directly after `masked_mse()` (around line 831):

```python
def masked_huber(
    pred: torch.Tensor,
    target: torch.Tensor,
    mask: torch.Tensor,
    delta: float = 0.5,
) -> torch.Tensor:
    e = pred - target
    abs_e = e.abs()
    quadratic = torch.minimum(abs_e, torch.full_like(abs_e, delta))
    linear = abs_e - quadratic
    loss = 0.5 * quadratic.square() + delta * linear
    return masked_mean(loss, mask)
```

(The `torch.minimum` / `linear` formulation is preferred over `torch.where` because it stays differentiable across the kink and avoids a NaN-on-backward edge case some PyTorch versions hit with `where`.)

### 2. Export and import `masked_huber`

In `train.py` line 52, extend the existing `from trainer_runtime import ...` to include `masked_huber` alongside `masked_mse`.

### 3. Add CLI flags to `Config` (and arg parser) in `train.py`

```python
tau_loss_type: str = "rel_l2"        # one of {"rel_l2", "huber"}
tau_huber_delta: float = 0.5
```

Wire CLI flags `--tau-loss-type {rel_l2,huber}` and `--tau-huber-delta FLOAT`. Use `python train.py --help` to confirm exact dash/underscore conventions used elsewhere in the file — match them.

### 4. Split surface loss in `train_loss()` (`train.py` lines 182-214)

Replace the single `surface_loss = masked_mse(...)` call with:

```python
if tau_loss_type == "huber":
    sp_loss = masked_mse(
        out["surface_preds"][..., 0:1],
        surface_target[..., 0:1],
        batch.surface_mask,
    )
    tau_loss = masked_huber(
        out["surface_preds"][..., 1:4],
        surface_target[..., 1:4],
        batch.surface_mask,
        delta=tau_huber_delta,
    )
    surface_loss = sp_loss + tau_loss
else:
    surface_loss = masked_mse(
        out["surface_preds"], surface_target, batch.surface_mask
    )
```

Pass the new config fields through `train_loss()`'s signature (currently uses keyword-only args for `surface_loss_weight`, `volume_loss_weight` — follow the same pattern).

**Volume loss is unchanged** — keep `masked_mse` for `out["volume_preds"]`.

### 5. Sanity check

Before kicking off the full run, verify with a 1-epoch dry run that `--tau-loss-type rel_l2` (default) produces identical loss values to the unmodified code and that `--tau-loss-type huber` produces a different (smaller, since clipped) loss number that decreases over the first few steps.

### 6. Run the experiment — single config, SOTA stack, --tau-loss-type huber, delta=0.5

```bash
  --agent askeladd \
  --tau-loss-type huber --tau-huber-delta 0.5 \
  --wandb_group askeladd-huber-tau
```

If you have GPU headroom in epoch budget after the primary delta=0.5 run, queue one secondary value of delta to characterize sensitivity — delta=1.0 is the natural second pick (closer to MSE; tells us whether the gain comes from clipping at all or from the specific bulk/tail balance). Use the same `--wandb_group askeladd-huber-tau`.

### What to report

- Per-axis val and test metrics from the **best-val-checkpoint** (test eval, not just terminal-epoch val).
- A single PR-body table with columns: SOTA #309 (val/test), this run (val/test), delta-pp.
- Best-val W&B run id and link.
- Loss-curve sanity-check note: confirm `surface_loss` value range stays comparable to baseline (Huber on tau will be slightly smaller numerically but should still dominate where the existing MSE dominated).

## Baseline (SOTA — PR #309 thorfinn grad-clip-norm=0.5)

W&B run `ztdhodw1` (https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/ztdhodw1)

- **val_abupt = 9.0389% (ep11)** — merge bar
- test_abupt = 10.126%

Per-axis test:

| Metric | SOTA | AB-UPT | Gap |
|---|---:|---:|---:|
| abupt mean | 10.126 | — | — |
| surface_pressure | 5.395 | 3.82 | x1.41 |
| wall_shear | 9.883 | 7.29 | x1.36 |
| tau_x | 8.402 | 5.35 | x1.57 |
| tau_y | 11.941 | 3.65 | x3.27 |
| tau_z | 12.407 | 3.63 | x3.42 |
| volume_pressure | 12.484 | 6.08 | x2.05 |

**Reproduce SOTA:**
```bash
  --agent <STUDENT> --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --ema-decay 0.999 --grad-clip-norm 0.5
```

**Merge bar:** val_abupt < 9.0389%.

### DIFF
--- a/train.py
+++ b/train.py
@@ -49,6 +49,7 @@ from trainer_runtime import (
     init_wandb_run,
     is_valid_primary_metric,
     make_loaders,
+    masked_huber,
     masked_mse,
     metric_namespace,
     parse_kill_thresholds,
@@ -79,6 +80,8 @@ class Config:
     validation_every: int = 1
     surface_loss_weight: float = 1.0
     volume_loss_weight: float = 1.0
+    tau_loss_type: str = "rel_l2"
+    tau_huber_delta: float = 0.5
     manifest: str = "data/split_manifest.json"
     data_root: str = ""
     output_dir: str = "outputs/drivaerml"
@@ -188,6 +191,8 @@ def train_loss(
     *,
     surface_loss_weight: float = 1.0,
     volume_loss_weight: float = 1.0,
+    tau_loss_type: str = "rel_l2",
+    tau_huber_delta: float = 0.5,
 ) -> tuple[torch.Tensor, dict[str, float]]:
     batch = batch.to(device)
     surface_target = transform.apply_surface(batch.surface_y)
@@ -199,19 +204,43 @@ def train_loss(
             volume_x=batch.volume_x,
             volume_mask=batch.volume_mask,
         )
-        surface_loss = masked_mse(out["surface_preds"], surface_target, batch.surface_mask)
+        if tau_loss_type == "huber":
+            sp_loss = masked_mse(
+                out["surface_preds"][..., 0:1],
+                surface_target[..., 0:1],
+                batch.surface_mask,
+            )
+            tau_loss = masked_huber(
+                out["surface_preds"][..., 1:4],
+                surface_target[..., 1:4],
+                batch.surface_mask,
+                delta=tau_huber_delta,
+            )
+            surface_loss = sp_loss + tau_loss
+        elif tau_loss_type == "rel_l2":
+            sp_loss = None
+            tau_loss = None
+            surface_loss = masked_mse(out["surface_preds"], surface_target, batch.surface_mask)
+        else:
+            raise ValueError(
+                f"Unknown tau_loss_type '{tau_loss_type}'. Supported: rel_l2, huber."
+            )
         volume_loss = masked_mse(out["volume_preds"], volume_target, batch.volume_mask)
         weighted_surface_loss = surface_loss_weight * surface_loss
         weighted_volume_loss = volume_loss_weight * volume_loss
         loss = weighted_surface_loss + weighted_volume_loss
         base_mse_loss = surface_loss + volume_loss
-    return loss, {
+    metrics = {
         "base_mse_loss": float(base_mse_loss.detach().cpu().item()),
         "surface_loss": float(surface_loss.detach().cpu().item()),
         "volume_loss": float(volume_loss.detach().cpu().item()),
         "surface_loss_weighted": float(weighted_surface_loss.detach().cpu().item()),
         "volume_loss_weighted": float(weighted_volume_loss.detach().cpu().item()),
     }
+    if sp_loss is not None:
+        metrics["surface_pressure_loss"] = float(sp_loss.detach().cpu().item())
+        metrics["tau_loss"] = float(tau_loss.detach().cpu().item())
+    return loss, metrics
 
 
 def main(argv: Iterable[str] | None = None) -> None:
@@ -324,6 +353,8 @@ def main(argv: Iterable[str] | None = None) -> None:
                     config.amp_mode,
                     surface_loss_weight=config.surface_loss_weight,
                     volume_loss_weight=config.volume_loss_weight,
+                    tau_loss_type=config.tau_loss_type,
+                    tau_huber_delta=config.tau_huber_delta,
                 )
                 optimizer.zero_grad(set_to_none=True)
                 global_step += 1
@@ -361,6 +392,9 @@ def main(argv: Iterable[str] | None = None) -> None:
                             "train/volume_loss_weighted": batch_loss_metrics["volume_loss_weighted"],
                         }
                     )
+                    if "surface_pressure_loss" in batch_loss_metrics:
+                        train_log["train/surface_pressure_loss"] = batch_loss_metrics["surface_pressure_loss"]
+                        train_log["train/tau_loss"] = batch_loss_metrics["tau_loss"]
 
                 if skip_step:
                     optimizer.zero_grad(set_to_none=True)
--- a/trainer_runtime.py
+++ b/trainer_runtime.py
@@ -832,6 +832,20 @@ def masked_mse(pred: torch.Tensor, target: torch.Tensor, mask: torch.Tensor) -> 
     return masked_mean((pred - target).square(), mask)
 
 
+def masked_huber(
+    pred: torch.Tensor,
+    target: torch.Tensor,
+    mask: torch.Tensor,
+    delta: float = 0.5,
+) -> torch.Tensor:
+    e = pred - target
+    abs_e = e.abs()
+    quadratic = torch.minimum(abs_e, torch.full_like(abs_e, delta))
+    linear = abs_e - quadratic
+    loss = 0.5 * quadratic.square() + delta * linear
+    return masked_mean(loss, mask)
+
+
 def squared_relative_l2_loss(
     pred: torch.Tensor,
     target: torch.Tensor,


morganmcg1:
STUDENT askeladd:

## Research

Quick research pass before kicking off training (notes from the researcher pass):

- **delta=0.5 vs the textbook 1.35σ rule:** for unit-normalized targets the classical 95%-Gaussian-efficiency choice is δ≈1.35; δ=0.5 puts ~38% of N(0,1) residual mass in the linear regime, which is more aggressive. That's actually consistent with **Sun, Zhou & Fan (JASA 2020 / arXiv:1805.01544)** "Adaptive Huber Regression": for genuinely heavy-tailed targets the optimal δ shrinks below 1.35σ, scaling as O(n^{-1/(1+ε)}) with the tail index. Tau channels are heavy-tailed, so going below 1.35 is theoretically supported. δ=0.5 is a reasonable point estimate — a tuned δ would benefit from an empirical-quantile calibration on the train residuals, but that's a follow-up not this PR.
- **Alternatives flagged for possible follow-up:** Barron's general adaptive robust loss (arXiv:1701.03077, learns per-channel α, drops the δ hyperparameter), and generalized Charbonnier (`(r^2 + ε^2)^α`, smooth at the origin, used in optical flow). Both could be drop-in replacements for fixed-δ Huber.
- **Implementation sanity:**
  - `torch.minimum`/`linear` decomposition (as in the PR instructions) avoids the `torch.where` NaN-on-backward edge case at the kink. Confirmed numerically — gradients finite at residuals straddling δ.
  - Targets are normalized per-channel — no need for per-channel δ.
  - **Risk to monitor**: Huber's linear regime gives constant-magnitude per-element gradients on outliers. Combined with the SOTA's `--grad-clip-norm 0.5`, if many tau residuals are |r|>0.5 early on, grad-norm clipping could engage more often than baseline and effectively reduce the LR for non-tau parameters too. Will watch `train/grad/clipped` rate vs baseline.

## Implementation

- Added `masked_huber(pred, target, mask, delta)` in `trainer_runtime.py` directly after `masked_mse` (used the advisor-specified `torch.minimum`/`linear` formulation).
- Wired `--tau-loss-type {rel_l2,huber}` (default `rel_l2`, identical to prior 4-channel MSE) and `--tau-huber-delta` (default 0.5) through `Config`, `parse_args`, `train_loss(...)`, and the train-loop call site. Default path is byte-identical to the unmodified code (literally the same `masked_mse` call), so existing reproductions are unchanged.
- Surface loss split when `huber`: `sp_loss = masked_mse(ch 0)` + `tau_loss = masked_huber(ch 1:4, δ)`. Volume loss path untouched.
- Per-component `train/surface_pressure_loss` and `train/tau_loss` logged to W&B for diagnostic transparency.

### Sanity tests (offline)
- 4-channel MSE on random residuals: 4.262 → unchanged with `tau_loss_type=rel_l2` (same call).
- ch 1:4 Huber(δ=0.5) = 0.741 vs ch 1:4 MSE = 4.669 (Huber correctly clips heavy tails).
- ch 1:4 Huber(δ=100) ≈ MSE/2 (recovers quadratic regime as δ→∞, as expected from the standard `0.5·r²` Huber convention).
- Backward grads finite across the kink ✓.

## Run launched

Primary experiment is running with the SOTA stack + `--tau-loss-type huber --tau-huber-delta 0.5`:

- W&B group: `askeladd-huber-tau`
- Rank-0 W&B run: [`nhr4uj3q`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/nhr4uj3q)
- All 8 ranks: nhr4uj3q (r0), kkgbvni9 (r1), ayz4q33p (r2), melligbe (r3), u04ylewh (r4), jff82bc8 (r5), 1mfhtyrx (r6), 07c3o34r (r7)
- 12.88M params, 8-rank DDP, projected ~33 min/epoch.

Will post a results comment with per-axis val/test from best-val checkpoint when training completes. If GPU/epoch budget allows after the primary, will also queue δ=1.0 in the same group as a sensitivity check.

morganmcg1:
ADVISOR: Confirmed — run nhr4uj3q is live and the implementation looks correct. The key signal to watch is tau_z specifically: under Lion, tau_z is the channel most likely to benefit from Huber since it has the largest outliers on the car's top/bottom surfaces.

Check-in cadence:
- **ep3–4**: Early tau_z signal. If tau_z is tracking below the STRING-sep baseline (10.449% test / comparable val), the Huber loss is pulling in the right direction.
- **ep6–7**: vol_pressure as a cross-check — should stay near 6.8–7.0% (SOTA ep6 is 6.86% for the STRING-sep run).

**Merge bar**: val_abupt must reach ≤7.546% (PR #311 STRING-sep SOTA). This is a high bar. Watch the tau_z improvement specifically — even if overall val_abupt doesn't beat SOTA, strong tau_z improvement is evidence to build on.

On the grad-clip-norm interaction you flagged: good catch. If you see gradient norms spiking anomalously (>5× baseline) specifically on the surface head, the Huber loss at the kink may be interacting with Lion's sign update. If that happens, try reducing delta to 0.1 (sharper Huber, less L2 regime) as a follow-up. Keep the current delta=0.5 for this run.

Keep going — update at ep5.

morganmcg1:
ADVISOR check-in (ep 9/50):

**Huber tau loss run trajectory — nhr4uj3q**

| Ep | Step | abupt_axis | sp_rl2 | ws_rl2 | vp_rl2 |
|----|------|------------|--------|--------|--------|
| 1  | 2720 | 62.9347% | 41.4414% | 70.7413% | 39.6252% |
| 2  | 5441 | 30.4194% | 16.1004% | 35.7266% | 14.8080% |
| 3  | 8162 | 19.0666% | 10.6038% | 22.1529% | 10.0722% |
| 4  | 10883 | 14.9871% | 8.2044% | 17.3579% | 8.4714% |
| 5  | 13604 | 13.3993% | 7.3064% | 15.3132% | 8.3022% |
| 6  | 16325 | 12.1791% | 6.6747% | 14.0228% | 7.1460% |
| 7  | 19046 | 11.4592% | 6.2621% | 13.1866% | 6.7731% |
| 8  | 21767 | 10.9139% | 5.9779% | 12.5549% | 6.4524% |
| 9  | 24488 | **10.5283%** | 5.7666% | 12.1097% | 6.2228% |

Run state: **RUNNING** | Merge bar: **9.0389%** | Gap: **1.49pp**

**Assessment:** Run is healthy and descending consistently. The descent rate is slowing as expected:
- ep7→ep8: -0.545pp
- ep8→ep9: -0.385pp

With 41 epochs remaining and assuming continuing deceleration (roughly geometric), projecting ~9.0–9.5% final value is plausible. The Huber tau treatment seems to be providing smoother convergence on wall shear (ws_rl2 = 12.1097%) compared to typical MSE baselines at the same epoch count, which is consistent with the hypothesis that Huber(delta=0.5) reduces the influence of outlier tau values.

However, vp_rl2 at 6.2228% and sp_rl2 at 5.7666% are strong — the concern is whether ws_rl2 will pull down enough. Wall shear is the hardest sub-metric and currently the bottleneck.

**Recommendation:** Continue to completion. The run is healthy and the trajectory is consistent with the hypothesis. The key question is whether the Huber smoothing on tau channels provides compounding benefit in the later epochs when fine-grained structure is being learned. We will make the final merge decision at ep50.

Please continue running — no action needed from you.

morganmcg1:
STUDENT askeladd:

## Results — Primary δ=0.5 run

**NEGATIVE result.** Channel-selective Huber on tau (δ=0.5) regresses both val_abupt and test_abupt versus the SOTA baseline by ~+1pp, with the **opposite of the intended effect** on the targeted axes: tau_y/tau_z got *worse*, not better.

### Per-axis val/test comparison (best-val-checkpoint, EMA)

| Metric | SOTA #309 val | this run val | Δval pp | SOTA #309 test | this run test | Δtest pp |
| **abupt_axis_mean** | 9.0389 | **10.1126** | **+1.07** | 10.126 | **11.098** | **+0.97** |
| surface_pressure | 5.6604 | 5.5681 | −0.09 | 5.395 | **5.218** | **−0.18** |
| wall_shear (vec) | 10.0983 | 11.5907 | +1.49 | 9.883 | 11.252 | +1.37 |
| tau_x | 8.5058 | 9.5742 | +1.07 | 8.402 | 9.339 | +0.94 |
| **tau_y** | 12.1700 | **14.7444** | **+2.57** | **11.941** | **14.389** | **+2.45** |
| **tau_z** | 13.1708 | **14.5767** | **+1.41** | **12.407** | **13.657** | **+1.25** |
| volume_pressure | 5.6876 | 6.0995 | +0.41 | 12.484 | 12.885 | +0.40 |

(val_p_v < test_p_v in *both* runs — that train/val-vs-test gap on volume_pressure is structural to the dataset/eval, not introduced here.)

### What happened — analysis

1. **Hypothesis was wrong on directional sign for THIS implementation.** Capping tau-channel residual gradients via Huber(δ=0.5) made tau_y/tau_z *worse* by +2.45 / +1.25 pp test, not better. The bulk-vs-tail tradeoff didn't go in the predicted direction.

2. **Likely root cause: the channel-split implementation accidentally re-weights surface_pressure 4× higher than baseline.** Baseline `masked_mse(surface_preds, surface_target, mask)` averages over all 4 channels (denominator = 4·mask_sum). The huber path computes `sp_loss = masked_mse(ch 0)` (denominator = mask_sum) and `tau_loss = masked_huber(ch 1:4)` (denominator = 3·mask_sum), then sums them. Per-element weights become **sp: 1×, tau: 1/3×**, vs baseline's uniform **1/4× per channel** — i.e. sp gets 4× higher relative weight, tau gets 4/3× higher (before Huber's clipping pulls tau back down).
   Empirically this matches the result: surface_pressure improved (−0.18pp test) consistent with capacity reallocation toward it, while tau channels regressed (+0.94…+2.45pp). The clipping itself is hard to disentangle from this re-weighting effect.

3. **Lion grad-clip stayed engaged throughout.** `train/grad/clipped` was 1 across the run from the very first logged step, identical to baseline at the same step (both clip at 0.5 pre-clip-norm 1.4–5.2). Huber's linear regime didn't push grad-norm into a new regime that defeated clipping; gradient health looks normal.

4. **Loss-curve sanity check (advisor request):** train/surface_loss values are comparable in scale to baseline (~0.18 at step 2700 vs ~0.16 baseline at step 2335), confirming the advisor's prediction that absolute scale would stay similar. The per-channel relative-weighting shift, however, is significant even though absolute scale matches.

### Logging diagnostics

`train/surface_pressure_loss` and `train/tau_loss` were logged separately as planned. At step 2700: sp=0.0812, tau=0.0948 (Huber-clipped). At step 7528 (late): sp=0.0203, tau=0.0239. Both decrease monotonically — Huber path is training stably, just to a worse final answer.

### Run identity

- **W&B group:** `askeladd-huber-tau`
- **Rank-0 W&B run:** [`nhr4uj3q`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/nhr4uj3q)
- **Best epoch:** 11 (matches baseline epoch 11; same convergence cadence)
- **Total train wall-clock:** 270.6 min
- **Peak GPU mem (rank 0):** 61.3 GiB
- **Exact command:**
  ```bash
    --agent askeladd \
    --tau-loss-type huber --tau-huber-delta 0.5 \
    --wandb-group askeladd-huber-tau --wandb-name askeladd-huber-tau-delta0p5
  ```

### Status of secondary δ=1.0 run

Already launched as the advisor requested. Group `askeladd-huber-tau`, expected ~270 min wall-clock. Will post a separate Results comment when it completes; that gives a sensitivity-direction check (closer-to-MSE δ should reveal whether δ=0.5 was specifically too aggressive vs whether the problem is structural/re-weighting).

### Suggested follow-ups (do *not* run unless re-assigned)

1. **Scale-corrected channel-Huber.** Re-run δ=0.5 / δ=1.0 with the tau-Huber implementation that *preserves* baseline channel weighting — sum residuals per element across all 4 channels in a single masked_mean, applying MSE-on-ch-0 and Huber-on-ch-1:4 element-wise rather than computing two separate masked_mean's. Concretely:
   ```python
   per_elem = (pred[..., 0:1] - target[..., 0:1]).square()  # MSE for ch0
   per_elem_tau = ... # huber per-element on ch1:4
   surface_loss = masked_mean(torch.cat([per_elem, per_elem_tau], dim=-1), mask)
   ```
   This isolates Huber's clipping effect from the implicit channel re-balance. If clipping helps once re-balanced, the original hypothesis is rescued; if not, the hypothesis is dead.
2. **Empirical-quantile δ.** Per Sun-Zhou-Fan (arXiv:1805.01544), the optimal δ for heavy-tailed residuals scales with sample size and tail index. A 50th–80th-percentile residual at end-of-baseline-training would set δ on actual data rather than a guessed 0.5 / 1.0. Cheap one-shot calibration: take a baseline checkpoint, compute |residual| quantiles per tau channel on val, set δ = q_p for some p in {0.5, 0.7, 0.9}.
3. **Adaptive Huber / Barron's general loss.** arXiv:1701.03077 — learn a continuous shape parameter α per channel jointly with the model. Drops δ entirely; if some channels prefer α≈2 (MSE) and others α<2 (more robust), the model picks per-channel.
4. **Channel-targeted up-weighting instead of channel-targeted Huber.** If the goal is to drive tau_y/tau_z down, an arguably simpler lever is per-axis loss-weight scaling (e.g. tau_y/tau_z each at 2× channel weight in the standard 4-channel-mean MSE). That's pure capacity-reallocation without changing the loss family. Could be a one-axis ablation alongside the Huber direction.

### Open question for the advisor

Channel-split surface_loss accidentally re-weights surface_pressure 4×; this is structural to the implementation pattern in the PR instructions (`sp_loss = masked_mse(ch 0); tau_loss = masked_huber(ch 1:4); surface_loss = sp_loss + tau_loss`) regardless of δ. Worth deciding whether (a) follow-up #1 above is the right scale-correction to apply going forward, or (b) the channel re-weighting is itself a feature worth keeping (e.g. for the surface_pressure improvement). Happy to run either flavor next.

morganmcg1:
ADVISOR: Progress check at ep~8.8 (step 12256).

**Current metrics:**
- val_abupt: 14.630%
- surf_p: 8.370%
- vol_p: 8.665%
- wall_shear: 16.737%
- tau_y: 21.235%
- tau_z: 20.998%
- slope: -1.257pp/1k_steps

**Concern:** Despite a steep slope, starting from 14.6% at ep8.8 puts convergence to competitive territory (< 7.546%) at ep~17-18, which is well beyond our typical evaluation window. More importantly, the tau_y (21.235%) and tau_z (20.998%) are significantly **worse** than SOTA (tau_y ~18%, tau_z ~20%), suggesting the Huber loss is reducing gradient signal on large tau errors — precisely the channel we need to fix. The AB-UPT benchmark gap for tau_y is ×2.53 and tau_z ×2.88, so we cannot afford regressions there.

**Hypothesis:** Huber loss clips gradients for large residuals (the 'hinge' at delta), so the model gets weaker gradient signal on the worst-performing tau predictions. Standard L2 would push harder on those outliers.

**Action requested:** Please stop the current run, switch back to L2 loss on tau channels, and test this hypothesis: is the Huber loss actively hurting the very tau channels we're trying to fix? If so, close this PR and open a new one targeting the tau channels with a different approach (e.g., per-channel loss weighting rather than loss function change).

morganmcg1:
ADVISOR: Closing PR #353 because Closing as NEGATIVE result. Both Huber tau runs (δ=0.5 nhr4uj3q and δ=1.0) regressed val_abupt by +1pp+ vs PR #309 SOTA (10.11% vs 9.04%), and worse, the targeted tau_y/tau_z channels got WORSE (tau_y +2.57pp, tau_z +1.41pp) — opposite of the intended effect.

## Why it failed

Two mechanisms in series:

1. **Channel reweighting bug.** Splitting into `sp_loss = masked_mse(ch 0)` + `tau_loss = masked_huber(ch 1:4)` with separate masked_mean calls accidentally gave surface_pressure 4× higher relative weight than the tau channels, vs the baseline's uniform 4-channel mean. The model down-weighted tau gradients structurally regardless of Huber.

2. **Huber clips the channel we need to push.** Even with the weighting fixed, Huber's linear (rather than quadratic) regime on residuals > δ reduces gradient magnitude on large tau errors — but tau_y and tau_z ARE the channels with the largest residuals. Huber actively softens the gradient signal precisely where we need it strongest. This is the wrong loss for this problem; we need a loss that pushes HARDER on outliers, not softer.

## Bigger context

The merge bar has moved to **val_abupt < 7.4648%** (PR #358 thorfinn STRING-sep + QK-norm). Even a successful tau-loss tweak on the old PR #309 baseline would only converge to ~9% — well above the new bar. Future tau-channel work should be on the new SOTA stack and use a loss that *amplifies* outlier gradients (e.g. quantile loss on tau, scale-corrected per-axis loss-weight, or empirical-quantile δ Barron general loss with α<2 i.e. super-quadratic). I'll have a new assignment ready for you on the current SOTA stack.

Thank you for the rigorous diagnostic work — the channel-reweighting bug analysis is exactly the kind of post-mortem that saves future students. Good experiment design even though the result was negative..

PR #276, opened 2026-05-01
[fern] SWA over last 5 epochs: weight averaging for generalization

morganmcg1:
## Hypothesis

Stochastic Weight Averaging (SWA) averages model weights from the last K epochs of training to produce a final model that generalizes better than any single checkpoint. Unlike EMA (which tracks a running average throughout training), SWA averages a small window of final checkpoints after the cosine LR has annealed to its minimum. This is a cheap ensemble-like technique compatible with cosine scheduling that can reduce val_abupt without additional training cost.

The idea: after the LR cosine cycle ends (after T_max epochs), instead of taking the final checkpoint, average the weights from the last 5 epoch snapshots. The averaged model is then evaluated. Reference: Izmailov et al. 2018, "Averaging Weights Leads to Wider Optima and Better Generalization".

This was previously archived from Wave 3 (PR #215, senku) where the student did not respond. Now assigned to you with full implementation instructions.

## Instructions

This requires a **code modification to `target/train.py`**. No other files need to change.

**Add SWA over the last 5 epochs of training using PyTorch's built-in `torch.optim.swa_utils.AveragedModel`.**

Here is the exact implementation:

**Step 1: Import at the top of `train.py`** (add after the existing torch imports):
```python
from torch.optim.swa_utils import AveragedModel, SWALR, update_bn
```

**Step 2: Before the training loop**, create the SWA model and decide the SWA start epoch:
```python
# SWA: start averaging in the last 5 epochs
swa_start_epoch = max_epochs - 5
swa_model = AveragedModel(base_model)
swa_enabled = swa_start_epoch > 0
```

**Step 3: At the END of each epoch** (after `scheduler.step()` on line ~463), add:
```python
# SWA: collect weight snapshots for final averaging
if swa_enabled and epoch >= swa_start_epoch:
    swa_model.update_parameters(base_model)
```

**Step 4: After the training loop ends** (before `distributed_barrier` on line ~621), add SWA evaluation:
```python
# SWA: evaluate averaged model if we collected any snapshots
if swa_enabled and state.is_main:
    # Update batch norm statistics with train data
    with torch.no_grad():
        update_bn(train_loader, swa_model.to(device))
    # Evaluate SWA model
    swa_model.eval()
    swa_val_metrics = evaluate(swa_model, val_loader, transform, device, config.amp_mode)
    swa_primary_val = swa_val_metrics["val_surface"]["abupt_axis_mean_rel_l2_pct"]
[...]

PR #354, opened 2026-05-02
[fern] Bisect +1pp structural regression vs alphonse m9775k1v baseline

morganmcg1:
## Hypothesis

A systematic +1.0–1.5 pp structural gap has been observed between every recent `fourier-pe nf=8` baseline-config run and the alphonse `m9775k1v` baseline (ep30, `val_abupt=7.2091%`). All the data points:

| Run ID | Config | Best abupt |
|--------|--------|-----------|
| `m9775k1v` | alphonse baseline PR #74 (ep30 best) | **7.2091%** |
| `67u9bilg` | haku surface-weight 2.0 (identical arch, 50ep) | 8.168% |
| `kuz4na0j` | trial-B wsy/wsz 0.5 | 8.895% |
| `w3thlivw` | mirror0.5 + sw2.0 | 8.919% |
| `31s1j3a0` | gc=0.5 (grad clip) | 9.215% |
| `0fhryk4r` | fern SWA-last5 (fourier-pe confirmed, ep10) | 9.270% |

**Architecture-identical runs (3,249,813 params, FourierEmbed, lr=3e-4, cosine_t_max=30, no-EMA) still land ~1pp above alphonse.** This cannot be explained by hyperparameter difference — it is a codebase, data, or environment change between the alphonse merge (PR #74) and current HEAD.

This affects every in-flight Wave 5–8 PR. Finding and fixing this regression is the highest-leverage action in the entire research programme right now.

## Goal

1. **Identify the commit** between PR #74 (alphonse, the Wave 1 BASELINE winner, ~2026-04-30) and current bengio HEAD that introduced the regression.
2. **Propose and implement a minimal fix** in `target/train.py` or `target/model.py` (whichever file the culprit is in).
3. **Validate** by running a short sanity-check training run (at minimum ep5 abupt should match the alphonse trajectory table below).

## What to investigate

Start with these categories, in order of likelihood:

**a. Data/dataloader changes** — Any change to normalization constants, dataset splits, augmentation defaults, point cloud subsampling counts, or data loading order. A silent normalization change would produce the consistent ~1pp floor we see. Check `train.py` for any hardcoded normalization or preprocessing that differs from what alphonse used.

**b. Loss function changes** — Any change to loss weights, loss formulation, or how the primary `abupt_axis_mean_rel_l2_pct` is computed. Even a small coefficient change compounds across epochs.

**c. Model initialization** — Any change to default weight initialization in `model.py`. A change in the FourierEmbed or transformer init could produce a consistent early-epoch gap.

**d. LR scheduler** — Any change to how `lr-cosine-t-max` is applied. Verify the LR curve in W&B for a fresh run vs the expected cosine schedule.

**e. Training loop / DDP sync** — Any change to how metrics are aggregated across ranks, or DDP bucket/gradient sync that could affect effective batch size.

## Approach

Use `git log --oneline <PR74-merge-sha>..HEAD -- target/` to list all commits that touched the target code since alphonse merged. Examine each diff for the categories above.

The alphonse PR #74 is a squash merge — so the comparison point is the commit just after the squash. Run `git log --oneline --all | grep -i alphonse` to find the relevant commit hash.

If the change is in `train.py` or `model.py`, implement the fix directly. If it is an environment issue (library version, CUDA, dataset file), document it clearly and propose a reproducible workaround.

## Validation run

After the fix, run a short 10-epoch validation run with the exact baseline recipe:

```bash
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --fourier-pe \
  --no-use-ema \
  --lr 3e-4 \
  --lr-cosine-t-max 30 \
  --no-compile-model \
  --wandb-group bengio-regression-bisect
```

The expected trajectory (from alphonse `m9775k1v`) for the baseline recipe:

| ep | expected abupt | acceptable range |
|----|--------------|-----------------|
| 1  | ~13.5%       | < 14.5%         |
| 5  | ~8.9%        | < 9.5%          |
| 10 | ~8.2%        | < 8.8%          |

If your fix brings ep5/ep10 within the acceptable ranges, the regression is resolved. **Report the W&B run ID and the commit SHA of the fix.**

## Current Baseline

| Metric | Current Best (val, ep30) | AB-UPT Target |
|--------|--------------------------|---------------|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **7.2091** | 4.51 |
| `val_primary/surface_pressure_rel_l2_pct` | 4.802 | 3.82 |
| `val_primary/wall_shear_rel_l2_pct` | 8.160 | 7.29 |
| `val_primary/volume_pressure_rel_l2_pct` | **4.166** ✓ | 6.08 |
| `val_primary/wall_shear_x_rel_l2_pct` | 7.109 | 5.35 |
| `val_primary/wall_shear_y_rel_l2_pct` | 9.100 | 3.65 |
| `val_primary/wall_shear_z_rel_l2_pct` | 10.869 | 3.63 |

Baseline W&B run: `m9775k1v` (alphonse, PR #74, `--fourier-pe` confirmed, n_params=3,249,813)

## Notes

- The SWA code from the closed PR #276 (branch `fern/swa-last-k-epochs`, commit `d6cb811`) is preserved for later.
- Report your findings in PR comments as you go — even partial diagnosis is valuable.
- If the regression is environmental rather than code (library version bump, dataset update), document the delta precisely so the team can decide whether to freeze or adapt.


morganmcg1:
STUDENT fern:

## Diagnosis: FourierEmbed regression — root cause identified

The +1pp structural regression vs `m9775k1v` is **not** in train.py, the loader, the loss, the LR schedule, or DDP — it's in `model.py:FourierEmbed`. The current implementation is missing the coordinate normalization that the alphonse run actually used.

### Evidence

1. **W&B run `m9775k1v` config** (queried via API):
   - `n_params: 3249813` (confirms FourierEmbed, not ContinuousSincosEmbed)
   - **No** `fourier_pe` config key — the flag did not exist yet at training time
   - Git commit recorded in metadata: `2c3f96d` (alphonse branch HEAD)

2. **W&B run `m9775k1v` `diff.patch`** (downloaded from W&B): the alphonse pod trained with **uncommitted local edits** to `model.py`. Those edits are reproduced here in full from `diff.patch`:

   ```python
   class FourierEmbed(nn.Module):
       """NeRF-style multi-scale sinusoidal positional encoding for 3D coordinates.
       Coordinates are normalized to roughly [-1, 1] with a fixed DrivAerML
       domain bounding box (5-sigma envelope on volume_xyz, padded slightly),
       then encoded with a base-2 geometric frequency progression
       `pi * 2^k`, k in [0, num_freqs)."""

       DOMAIN_BOUNDS = (
           (-12.0, 14.0),
           (-5.0, 5.0),
           (-3.0, 3.0),
       )

       def __init__(self, hidden_dim, input_dim=3, num_freqs=8):
           ...
           bounds = self.DOMAIN_BOUNDS[:input_dim]
           center = torch.tensor([(lo + hi) / 2.0 for lo, hi in bounds], ...)
           half_range = torch.tensor([(hi - lo) / 2.0 for lo, hi in bounds], ...)
           self.register_buffer("center", center)
           self.register_buffer("half_range", half_range)
           freqs = math.pi * (2.0 ** torch.arange(num_freqs, dtype=torch.float32))
           self.register_buffer("freqs", freqs)
           self.proj = LinearProjection(input_dim * num_freqs * 2, hidden_dim)

       def forward(self, coords):
           coords = coords.float()
           coords = (coords - self.center) / self.half_range  # <-- KEY STEP
           angles = coords.unsqueeze(-1) * self.freqs
           emb = torch.cat([torch.sin(angles), torch.cos(angles)], dim=-1)
           emb = emb.flatten(start_dim=-2)
           return self.proj(emb)
   ```

3. **Current bengio HEAD `model.py:FourierEmbed`** (from chihiro PR #176, commit `6b78e3e`): identical math except **no coordinate normalization**.

   ```python
   def forward(self, coords):
       coords = coords.float()
       # NO normalization step
       angles = coords.unsqueeze(-1) * self.freqs * math.pi
       emb = torch.cat([torch.sin(angles), torch.cos(angles)], dim=-1)
       emb = emb.flatten(start_dim=-2)
       return self.proj(emb)
   ```

   chihiro's commit message says: *"PR #74 was merged on val metrics, but only the BASELINE.md update landed; the actual model.py change that introduces FourierEmbed never made it onto bengio. […] Matches emma's commit e774bea, which produced alphonse-style Wave 1 results."* — but `e774bea` is not reachable from any branch (likely a deleted/abandoned commit), and chihiro's reimplementation silently dropped coordinate normalization.

### Why this is the regression

DrivAerML coordinates are **not** in `[-1, 1]`. From `normalizers.json` (5σ envelope on volume_xyz):
- x ≈ [-10, 14] (mean 1.85, std 2.36)
- y ≈ [-4.3, 4.2] (mean -0.001, std 0.85)
- z ≈ [-2.5, 2.7] (mean 0.105, std 0.52)

Without normalization, the angles fed to sin/cos at the highest frequency (`2^7 * π ≈ 402 rad`) become `coord * 402`, which for `x = 14` is ~5,628 rad. This is severe high-frequency aliasing — the highest-band features are essentially noise relative to the geometry.

With the alphonse normalization, the highest-band features sweep through ~128π over the [-1, 1] range — a sensible NeRF-style positional encoding that the model can actually learn from.

### Fix

Update `model.py:FourierEmbed` to match the alphonse implementation: register `DOMAIN_BOUNDS = ((-12, 14), (-5, 5), (-3, 3))` as `center`/`half_range` buffers and apply `(coords - center) / half_range` before computing angles. This is a 6-line addition to one class — no API change, no train.py change.

### Validation

Will run the 10-epoch baseline recipe per PR instructions and report W&B run ID + commit SHA.

PR #363, opened 2026-05-02
haku: diagnostic — where do tau_y/z errors concentrate physically?

morganmcg1:
## Hypothesis

The fleet has run ~20 optimization-side experiments targeting tau_y/z (loss weighting, augmentation, normalization, GradNorm, curriculum, FFT/GFT, tangent-frame). All of them attack the same metric without yet knowing **where on the vehicle surface the tau_y/z errors physically concentrate**. We're optimizing in the dark.

This assignment is **diagnostic, not optimization** — produce a high-value error-decomposition study on the current best PR #222 model checkpoint that gives the entire fleet new information about where to attack.

The hypothesis is implicit: if errors are concentrated in specific geometric regions (wheel arches, side mirrors, A-pillar separation, underbody), the right next intervention is region-specific (e.g. region-aware loss weighting or local feature enrichment) rather than global (loss reformulations or optimizer changes). If errors are uniformly distributed across the surface, then the global optimizers are correctly aimed.

## Instructions

This is an **inference + analysis** task, not a new training run. You'll load the PR #222 checkpoint and produce a structured error-mode report.

### Setup

1. **Load the canonical PR #222 checkpoint**: W&B run `ut1qmc3i`. Pull the EMA weights from the best-val artifact. If the checkpoint isn't directly accessible from W&B, train.py supports `--resume-from <wandb_run>` or you can reload via `wandb.restore`. If both fail, train a fresh single-GPU AdamW lr=1e-4 wu=2720 5-epoch run on yi to use as a stand-in (clearly label results "single-GPU stand-in, not PR #222 EMA weights" in that case).

2. **Run validation inference** on all 34 val cases at full resolution (use `--eval-surface-points 65536` or whatever the sampler default is — match val protocol exactly). For each case, save:
   - `surface_xyz` (B, N, 3)
   - `surface_normals` (B, N, 3)
   - `tau_pred` (B, N, 3) — channels 1:4 of the model output
   - `tau_target` (B, N, 3) — same channels of the GT
   - `case_id` and any metadata available (wheel-arch mask, mirror mask if dataset provides them)

   Save to a single `.pt` or `.npy` file under `target/research/diagnostic_data/haku-tau-error-modes-r19/`.

### Analysis (deliverables)

Produce a markdown report at `target/research/diagnostic_data/haku-tau-error-modes-r19/REPORT.md` with at minimum these five sections. Each section should have a concrete number and a one-paragraph interpretation.

**1. Per-axis error magnitude by surface region.**

[...]

PR #367, opened 2026-05-02
Theta-conditioned wall-shear loss weight for cross-flow error

morganmcg1:
## Hypothesis

PR #363 diagnostic found that **cross-flow-aligned surface points (theta≈90°) produce 2.21× higher wall-shear error** than streamwise-aligned points (theta≈4°), where theta is the angle between the GT wall-shear vector and the local streamwise axis. The model systematically under-predicts small-magnitude cross-flow content (GT |tau|=0.54 vs predicted 2.20 at streamwise bias). Spatial errors are clustered (Moran's I significant in all 34 val cases).

The current loss treats all surface points equally. A **per-point theta-conditioned weight** that up-weights cross-flow-dominant points during training should force the model to attend to this underserved regime. Weight formula:

```
theta_i = arccos(|tau_x_i| / (|tau_i| + eps))     # angle from streamwise, in [0, pi/2]
w_i     = 1 + alpha * sin(theta_i)                  # max weight at theta=90°
```

where GT wall shear in physical space is used (computed from the normalised targets using the transform). This is a smooth, physically motivated attention signal — no hard thresholds.

**Screen:** alpha=0.0 (control), alpha=1.0 (max 2×), alpha=2.0 (max 3×, within the W<3.0 stability ceiling from PR #66).

## Instructions

All three arms use **the same base config as PR #311 (current SOTA)**:

```bash
  --agent haku \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --lr-warmup-epochs 1 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0
```

(STRING-separable position encoding is already the default on `yi` after PR #311 merge.)

### Code changes required in `target/train.py`

**Step 1: Add `--wallshear-theta-alpha` CLI flag**

Find where `--wallshear-y-weight` and `--wallshear-z-weight` are added to `argparse` (search for `wallshear_y_weight`) and add immediately after:

```python
parser.add_argument("--wallshear-theta-alpha", type=float, default=0.0,
                    help="Alpha for per-point theta-conditioned wall-shear weight. "
                         "w_i = 1 + alpha*sin(theta_i) where theta is angle from streamwise. "
                         "0 = no conditioning (default). Try 1.0 or 2.0.")
```

**Step 2: Thread `wallshear_theta_alpha` into `train_loss()`**

Add the parameter to the `train_loss()` signature (alongside `wallshear_y_weight`, `wallshear_z_weight`):

```python
def train_loss(
    model, batch, transform, device, amp_mode, *,
    surface_loss_weight=1.0, volume_loss_weight=1.0,
    aux_rel_l2_weight=0.0,
    use_tangential_wallshear_loss=False,
    wallshear_y_weight=1.0, wallshear_z_weight=1.0,
    wallshear_theta_alpha: float = 0.0,   # <-- ADD THIS
```

Also thread it through in the call site (wherever `train_loss` is called with `wallshear_y_weight=...`):

```python
wallshear_theta_alpha=args.wallshear_theta_alpha,
```

**Step 3: Compute and apply the per-point theta weight inside `train_loss()`**

Inside `train_loss()`, **after** the existing code that denormalises GT wall shear to physical space (the block that computes `ws_true_phys` — same pattern used by `use_tangential_wallshear_loss`), add:

```python
if wallshear_theta_alpha > 0.0:
    # GT wall shear in physical space: shape [B, N, 3] (tau_x, tau_y, tau_z)
    ws_std  = transform.surface_y_std[1:4].to(device)
    ws_mean = transform.surface_y_mean[1:4].to(device)
    ws_true_norm = surface_target[..., 1:4]            # [B, N, 3]
    ws_true_phys = ws_true_norm * ws_std + ws_mean     # physical space

    tau_norm = ws_true_phys.norm(dim=-1, keepdim=False).clamp_min(1e-8)   # [B, N]
    tau_x_abs = ws_true_phys[..., 0].abs()                                  # [B, N]
    cos_theta = (tau_x_abs / tau_norm).clamp(0.0, 1.0)
    sin_theta = (1.0 - cos_theta.pow(2)).clamp_min(0.0).sqrt()
    theta_weight = 1.0 + wallshear_theta_alpha * sin_theta                   # [B, N]
```

Then, in the surface loss computation block, apply this weight to the wall-shear channels only. **Replace** the call to `weighted_masked_mse_per_channel` with a version that applies `theta_weight` to the squared differences for channels 1:4 (tau_x, tau_y, tau_z) before the mean:

```python
if wallshear_theta_alpha > 0.0:
    # Apply per-point theta weight to wall-shear channels
    diff_sq = (surface_pred_used - surface_target_used).pow(2)  # [B, N, C]
    mask_f  = batch.surface_mask.unsqueeze(-1).to(diff_sq.dtype)
    valid   = mask_f.sum().clamp_min(1)
    n_ch    = diff_sq.shape[-1]

    tw = theta_weight                                    # [B, N]
    # expand to [B, N, 3] for tau channels
    tw_exp = tw.unsqueeze(-1).expand_as(diff_sq[..., 1:4])
    diff_sq = diff_sq.clone()
    diff_sq[..., 1:4] = diff_sq[..., 1:4] * tw_exp

    ch_w = torch.tensor(
        [1.0, 1.0, wallshear_y_weight, wallshear_z_weight],
        device=diff_sq.device, dtype=diff_sq.dtype
    )
    weighted_diff = diff_sq * ch_w.view(1, 1, -1)
    surface_loss  = (weighted_diff * mask_f).sum() / valid / n_ch

    per_axis_sum  = ((surface_pred_used - surface_target_used).pow(2) * mask_f).sum(dim=(0,1)).detach().float()
    per_axis_unweighted = (per_axis_sum / valid.detach().float()).cpu().tolist()
else:
    surface_loss, per_axis_unweighted = weighted_masked_mse_per_channel(
        surface_pred_used, surface_target_used, batch.surface_mask,
        channel_weights=(1.0, 1.0, wallshear_y_weight, wallshear_z_weight),
    )
```

> **Note:** `per_axis_unweighted` is intentionally computed from the unweighted squared differences so that logged per-axis metrics remain comparable across arms.

> **Tip:** If `use_tangential_wallshear_loss` is True, `ws_true_phys` may already be in scope. Avoid recomputing it — check the existing code and reuse the variable if possible.

### Three-arm run commands

Use `--wandb_group haku-theta-wallshear-screen` so all three runs appear together in W&B.

**Arm A — control (alpha=0.0):**
```bash
  --agent haku --wandb_group haku-theta-wallshear-screen \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
  --model-slices 128 --ema-decay 0.999 --lr-warmup-epochs 1 \
  --wallshear-theta-alpha 0.0
```

**Arm B — alpha=1.0 (max weight 2×):**
```bash
  --agent haku --wandb_group haku-theta-wallshear-screen \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
  --model-slices 128 --ema-decay 0.999 --lr-warmup-epochs 1 \
  --wallshear-theta-alpha 1.0
```

**Arm C — alpha=2.0 (max weight 3×):**
```bash
  --agent haku --wandb_group haku-theta-wallshear-screen \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
  --model-slices 128 --ema-decay 0.999 --lr-warmup-epochs 1 \
  --wallshear-theta-alpha 2.0
```

### Reporting

Report per-arm results in a table:

| Arm | alpha | val_abupt | wall_shear_y | wall_shear_z | W&B run |
|-----|-------|-----------|--------------|--------------|---------|
| A   | 0.0   |           |              |              |         |
| B   | 1.0   |           |              |              |         |
| C   | 2.0   |           |              |              |         |

Also note the **best val_abupt epoch** per arm and whether loss curves are stable.

Key diagnostic: does `wall_shear_y` and `wall_shear_z` improve with alpha>0 relative to control? If `val_abupt` also improves, that's a merge candidate.

## Baseline

Current SOTA from PR #311 (edward, STRING-separable learnable position encoding, W&B run `gcwx9yaa`):

| Metric | val | test |
|--------|-----|------|
| `abupt_axis_mean_rel_l2_pct` | **7.546%** | **8.771%** |
| `surface_pressure_rel_l2_pct` | — | 4.485% |
| `wall_shear_rel_l2_pct` | — | 8.227% |
| `wall_shear_y_rel_l2_pct` | — | 9.233% |
| `wall_shear_z_rel_l2_pct` | — | 10.449% |
| `volume_pressure_rel_l2_pct` | — | 12.438% |

**Merge bar: val_abupt < 7.546%**

Note: `wall_shear_y` (9.233%) and `wall_shear_z` (10.449%) are 2.53× and 2.88× the AB-UPT reference targets (3.65% and 3.63%) — these are the dominant remaining gaps.

## Context

- PR #363 diagnostic run: identified 2.21× cross-flow error concentration (theta-split analysis across all 34 val geometries), significant spatial clustering (Moran's I), and run_228/run_324 as systematic outliers.
- `surface_x` layout: `[x, y, z, nx, ny, nz, area]` (normals at indices 3:6)
- `surface_y` layout: `[cp, tau_x, tau_y, tau_z]` (GT wall shear at indices 1:4)
- `project_tangential()` and `weighted_masked_mse_per_channel()` already exist in train.py
- `--use-tangential-wallshear-loss` already handles denorm of GT wall shear (reuse that pattern)

### DIFF
--- a/train.py
+++ b/train.py
@@ -566,6 +566,7 @@ class Config:
     wallshear_y_weight: float = 1.0
     wallshear_z_weight: float = 1.0
     wallshear_huber_delta: float = 0.0
+    wallshear_theta_alpha: float = 0.0
     manifest: str = "data/split_manifest.json"
     data_root: str = ""
     output_dir: str = "outputs/drivaerml"
@@ -1286,6 +1287,7 @@ def weighted_masked_mse_per_channel(
     *,
     channel_weights: Iterable[float],
     wallshear_huber_delta: float = 0.0,
+    wallshear_point_weight: torch.Tensor | None = None,
 ) -> tuple[torch.Tensor, list[float]]:
     """Per-channel weighted masked loss for surface predictions.
 
@@ -1338,6 +1340,12 @@ def weighted_masked_mse_per_channel(
         loss_elem = diff_sq
 
     weighted_diff = loss_elem * weights.view(1, 1, -1)
+    if wallshear_point_weight is not None and n_channels >= 4:
+        # Apply per-point weight to the tau channels (1:4). Keep cp (channel 0)
+        # and any extra channels (4:) at unity.
+        pw_full = weighted_diff.new_ones(weighted_diff.shape)
+        pw_full[..., 1:4] = wallshear_point_weight.unsqueeze(-1).expand_as(weighted_diff[..., 1:4]).to(weighted_diff.dtype)
+        weighted_diff = weighted_diff * pw_full
     weighted_loss = (weighted_diff * mask_f).sum() / valid / n_channels
     per_axis_sum = (diff_sq * mask_f).sum(dim=(0, 1)).detach().float()
     per_axis_mean = (per_axis_sum / valid.detach().float()).cpu().tolist()
@@ -1370,6 +1378,7 @@ def train_loss(
     wallshear_y_weight: float = 1.0,
     wallshear_z_weight: float = 1.0,
     wallshear_huber_delta: float = 0.0,
+    wallshear_theta_alpha: float = 0.0,
 ) -> tuple[torch.Tensor, dict[str, float]]:
     batch = batch.to(device)
     surface_target = transform.apply_surface(batch.surface_y)
@@ -1409,12 +1418,33 @@ def train_loss(
         else:
             surface_pred_used = surface_pred_norm
             surface_target_used = surface_target
+        theta_weight: torch.Tensor | None = None
+        theta_weight_mean: float | None = None
+        if wallshear_theta_alpha > 0.0:
+            # Per-point theta-conditioned wall-shear weight.
+            # theta_i = angle between GT wall-shear vector and global-x (streamwise) axis.
+            # w_i = 1 + alpha * sin(theta_i)  — up-weights cross-flow-dominant points.
+            ws_std_t  = transform.surface_y_std[1:4].to(device)
+            ws_mean_t = transform.surface_y_mean[1:4].to(device)
+            ws_true_norm_t = surface_target[..., 1:4]                 # [B, N, 3] normalised
+            ws_true_phys_t = ws_true_norm_t * ws_std_t + ws_mean_t    # physical space
+            tau_mag = ws_true_phys_t.norm(dim=-1, keepdim=False).clamp_min(1e-8)  # [B, N]
+            tau_x_abs = ws_true_phys_t[..., 0].abs()                              # [B, N]
+            cos_theta = (tau_x_abs / tau_mag).clamp(0.0, 1.0)
+            sin_theta = (1.0 - cos_theta.pow(2)).clamp_min(0.0).sqrt()
+            theta_weight = 1.0 + wallshear_theta_alpha * sin_theta                # [B, N]
+            mask_bool = batch.surface_mask
+            if bool(mask_bool.any()):
+                theta_weight_mean = float(
+                    theta_weight[mask_bool].detach().float().mean().cpu().item()
+                )
         surface_loss, per_axis_unweighted = weighted_masked_mse_per_channel(
             surface_pred_used,
             surface_target_used,
             batch.surface_mask,
             channel_weights=(1.0, 1.0, wallshear_y_weight, wallshear_z_weight),
             wallshear_huber_delta=wallshear_huber_delta,
+            wallshear_point_weight=theta_weight,
         )
         volume_loss = masked_mse(out["volume_preds"], volume_target, batch.volume_mask)
         loss = surface_loss_weight * surface_loss + volume_loss_weight * volume_loss
@@ -1438,6 +1468,8 @@ def train_loss(
     }
     if aux_rel_l2_value is not None:
         metrics["aux_rel_l2_loss"] = aux_rel_l2_value
+    if theta_weight_mean is not None:
+        metrics["wallshear_theta_weight_mean"] = theta_weight_mean
     if use_tangential_wallshear_loss:
         metrics["wallshear_pred_normal_rms"] = normal_rms
     if "geom_token" in out:
@@ -1951,6 +1983,7 @@ def main(argv: Iterable[str] | None = None) -> None:
                 wallshear_y_weight=config.wallshear_y_weight,
                 wallshear_z_weight=config.wallshear_z_weight,
                 wallshear_huber_delta=config.wallshear_huber_delta,
+                wallshear_theta_alpha=config.wallshear_theta_alpha,
             )
             optimizer.zero_grad(set_to_none=True)
             loss_is_finite = bool(torch.isfinite(loss).item())
@@ -2062,6 +2095,10 @@ def main(argv: Iterable[str] | None = None) -> None:
             if "wallshear_pred_normal_rms" in batch_loss_metrics:
                 train_log["train/wallshear_pred_normal_rms"] = batch_loss_metrics[
                     "wallshear_pred_normal_rms"
+                ]
+            if "wallshear_theta_weight_mean" in batch_loss_metrics:
+                train_log["train/wallshear_theta_weight_mean"] = batch_loss_metrics[
+                    "wallshear_theta_weight_mean"
                 ]
             if "aux_rel_l2_loss" in batch_loss_metrics:
                 train_log["train/aux_rel_l2_loss"] = batch_loss_metrics[


morganmcg1:
ADVISOR: First-launch guidance — please launch as soon as you can, but FIRST: there are critical command corrections to apply before running.

**The PR body run commands are invalid for the yi branch.** yi is single-GPU AdamW (PR #355 emma DDP not yet merged). The body uses `--optimizer lion`, `--lr-warmup-epochs 1`, and `torchrun --nproc_per_node=8` — none of these are supported on yi. Verified by grep of train.py:
- No `--optimizer` flag (AdamW is hardcoded at line 1712)
- No `--lr-warmup-epochs` flag (use `--lr-warmup-steps 500` instead)
- No DDP (no `init_process_group`)

**Sequence:**
1. **First**: Implement the train.py code changes from your PR body (new `--wallshear-theta-alpha` CLI flag; theta-weight computation in `train_loss()`; weighted surface loss when alpha>0). Smoke-test that arm A (alpha=0) produces identical loss to a current run.
2. **Then** launch the 3-arm screen with the corrected commands below.

**Corrected run commands (single-GPU AdamW, 3-arm screen, group `haku-theta-wallshear-screen`):**

```bash
# Arm A — control (alpha=0.0)
  --agent haku --wandb-group haku-theta-wallshear-screen \
  --wandb-name haku/arm-A-alpha0 \
  --lr 3e-4 --weight-decay 1e-4 --no-compile-model --batch-size 4 \
  --validation-every 1 \
  --ema-decay 0.999 --lr-warmup-steps 500 \
  --wallshear-theta-alpha 0.0 --epochs 9
```

Arms B (`--wallshear-theta-alpha 1.0`) and C (`--wallshear-theta-alpha 2.0`) are identical except for that one flag and the `--wandb-name`. Use `CUDA_VISIBLE_DEVICES=1` and `CUDA_VISIBLE_DEVICES=2` if running concurrently across your 4 GPUs (you have a 4th GPU free for any kill-and-rerun).

**Kill threshold (recommended):** `--kill-thresholds '3000:val_primary/abupt_axis_mean_rel_l2_pct<25'` — kill any arm at step 3000 if val_abupt > 25%.

**Merge bar:** val_abupt < **7.546%** (PR #311 edward STRING-sep). Note: my comment doesn't change your hypothesis or the 3-arm sweep design — only the per-run config.

Please confirm by replying once you've made the train.py code changes and launched Arm A; report back with the W&B run ID after ep1.

morganmcg1:
STUDENT haku:

## Implementation note

Implementation of `--wallshear-theta-alpha` is complete and the 3-arm screen is running. A few protocol deviations from the PR text, all forced by the current state of `target/train.py` on `yi`:

**1. Single-GPU per arm (no DDP)**
The PR specifies `torchrun --standalone --nproc_per_node=8 train.py`, but the `yi` `train.py` contains no DDP infrastructure (no `init_process_group`, no `LOCAL_RANK` handling, no `DistributedSampler`, no `DDP` wrap). Launching with `torchrun` produces 8 independent processes that all default to `cuda:0`, causing OOM. I'm running each arm on its own GPU (single-process), with the same per-GPU batch size (4). This keeps the comparison internally consistent across the three arms but means absolute val_abupt will not match the PR #311 SOTA reference (7.546%) — that figure was achieved on the `tay` branch which has DDP.

**2. `--optimizer lion` and `--lr-warmup-epochs` not supported on yi**
Neither flag exists in `yi`'s `train.py`:
- Optimizer is hardcoded to `torch.optim.AdamW`
- `--lr-warmup-steps 2720` is the closest equivalent to `--lr-warmup-epochs 1` (1 epoch ≈ 2720 steps under the prior 8-GPU effective batch convention; with single-GPU batch=4 this corresponds to ~12.5% of an epoch of warmup)

**3. Timeout vs config**
At single-GPU batch=4 with 65536 surface+volume points, iter time is ~1.24s. Full epoch = 21766 iters ≈ 7.5h. With `SENPAI_TIMEOUT_MINUTES=360` (6h) and `train_timeout_minutes = 360 - 90 (val_budget) = 270` min ≈ 4.5h, each arm will hit the mid-epoch train timeout at ~13000 iters (~60% of epoch 1). The trainer's mid-epoch timeout path then runs validation, so we will get one val checkpoint per arm. Best-epoch metrics will be at 60% of epoch 1, which is mid-warmup.

**4. Implementation verified**
Logged `train/wallshear_theta_weight_mean` confirms the per-point weight is applied:
- Arm A (alpha=0.0): not logged (else branch — unchanged loss path)
- Arm B (alpha=1.0): mean weight ≈ 1.61 (theoretical max 2.0)
- Arm C (alpha=2.0): mean weight ≈ 2.21 (theoretical max 3.0)

This is consistent with `E[1 + alpha · sin(theta)] ≈ 1 + 0.6·alpha`.

**Running now:** PIDs 411181/411182/411183, W&B runs `bb385as1` (alpha=0), `6474cl0h` (alpha=1), `k57ssdvm` (alpha=2). Group: `haku-theta-wallshear-screen`. Started 2026-05-02 11:46 UTC; expected timeout-and-validate at ~17:46 UTC. Will report results then.

If you'd prefer I kill these and (a) re-launch with reduced points to fit ≥3 epochs per arm (deviation from base config but more converged comparison), or (b) add minimal DDP support to `train.py` as a separate bug-fix commit, please let me know.

morganmcg1:
ADVISOR: Acknowledged — 3-arm theta-conditioned screen running. Good that you noted the deviations (single-GPU AdamW, lr-warmup-steps=2720 instead of epochs) — those are fine for the yi branch constraints.

Arm B alpha=1.0 with weight_mean≈1.61 is a useful signal that the theta-conditioning is active and producing non-trivial weights.

**Expected reporting at ~17:30 UTC:**
For each arm at the timeout val checkpoint, report:
1. val_primary/abupt_axis_mean_rel_l2_pct
2. train/wallshear_theta_weight_mean (the key diagnostic for whether theta-conditioning is doing something)
3. Individual component metrics: surface_pressure_mae, wall_shear_mae broken down by tau_y / tau_z if available

Win condition: any arm < 9.291%. Target: best arm ≤ 8.5% at ~60% ep1 (extrapolating to potential sub-baseline final result).

The key question is whether Arm B (alpha=1.0) or Arm C (alpha=3.0 or whichever higher value you used) shows lower tau_y/tau_z error relative to Arm A (alpha=0 baseline). Report both the overall val_abupt and the per-component breakdown.

morganmcg1:
STUDENT haku:

## Results

All 3 arms ran to the train timeout (275.6 min ≈ 4.6h, ~60% of epoch 1, mid-warmup) and produced one timeout-validation checkpoint each. Group: `haku-theta-wallshear-screen`.

### Primary table (full_val_primary, single checkpoint)

| Arm | alpha | val_abupt | wall_shear | wall_shear_x | **wall_shear_y** | **wall_shear_z** | surface_p | volume_p | θ̄_weight | W&B run |
|-----|-------|-----------|------------|--------------|------------------|------------------|-----------|----------|-----------|---------|
| A | 0.0 | 38.5473% | 46.43% | 36.51% | 50.09% | 74.90% | 16.28% | 14.95% | n/a | [bb385as1](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/bb385as1) |
| **B** | **1.0** | **37.6617%** | **45.08%** | **36.41%** | **49.76%** | **69.16%** | 16.33% | 16.64% | **1.602** | [6474cl0h](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/6474cl0h) |
| C | 2.0 | 38.9834% | 46.46% | 37.89% | 51.33% | 70.29% | 17.25% | 18.17% | 2.216 | [k57ssdvm](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/k57ssdvm) |

(All values are `full_val_primary/*_rel_l2_pct` = full validation pass at the timeout-induced epoch-1 checkpoint.)

Test-set numbers track validation closely (e.g. Arm B: test_abupt=38.26%, wall_shear_y=49.09%, wall_shear_z=67.68%) — same A < B > C ordering on every wall-shear axis.

### Best-epoch and stability

Best val_abupt epoch is epoch 1 for all three arms (the only checkpoint produced before the train-timeout fired). Loss curves are monotone-down through the 13099 train iters logged; no spikes, no instability. `train/wallshear_theta_weight_mean` confirms the weight kicks in:
- Arm A: not logged (else branch — unchanged loss path) ✓
- Arm B: 1.602 ≈ 1 + 1.0·E[sin θ] → E[sin θ] ≈ 0.602 (population skews cross-flow as expected)
- Arm C: 2.216 ≈ 1 + 2.0·E[sin θ] → E[sin θ] ≈ 0.608 (consistent)

### Analysis — what happened

**Hypothesis is directionally supported on the cross-flow channels.** Arm B (α=1.0) beats control on every wall-shear axis, with the largest improvement on tau_z (74.90% → 69.16%, **−5.74 pp / −7.7% relative**) — exactly the channel the PR #363 diagnostic flagged as most under-served. tau_y also moves slightly (50.09% → 49.76%, −0.33 pp / −0.7% relative). tau_x (streamwise) is essentially unchanged (−0.10 pp), which matches expectation since the weight is constant for streamwise-aligned points (sin θ ≈ 0).

**α=2.0 overshoots.** Arm C is worse than control on every metric, including the cross-flow channels it was meant to help. Surface and volume pressure also degrade (16.28% → 17.25% / 14.95% → 18.17%) — the heavier wall-shear weighting starves the pressure heads. So the merge candidate signal is **α≈1.0**, not "more is better".

**Side-channel cost.** Even Arm B costs ~1.7 pp on volume pressure relative to control (14.95% → 16.64%). This is not the headline metric, but it should be tracked: theta-conditioning is not free for the other heads.

**Caveats — important for read.** This screen is severely under-converged: all three arms timed out at ~60% of epoch 1, mid-warmup (warmup-steps=2720, single-GPU AdamW, no DDP — see protocol-deviation comment above). Absolute val_abupt ≈ 38% is ~5× the SOTA reference (7.546%) because the model never reached the post-warmup regime. The **ordering** (A > B > C with B best) is the load-bearing signal, not the absolute numbers; the deltas seen here may compress, expand, or flip with full convergence and DDP. Treat this as a **screen for direction**, not a merge-ready result.

### Suggested follow-ups

1. **Re-run α=1.0 on a DDP-capable branch** (post-#355 emma DDP merge) for ≥5 epochs to confirm the tau_z improvement persists past warmup. This is the obvious blocker to claiming a merge.
2. **Bracket the optimum more tightly** — α∈{0.5, 0.75, 1.25} would localize the peak between the under-shoot (α=0) and over-shoot (α=2). Current data only constrain α* ∈ (0, 2).
3. **Re-balance against pressure heads** — the volume_p degradation suggests pairing α>0 with a slightly reduced wallshear y/z channel weight (e.g. α=1.0 + y_weight=z_weight=1.5 instead of 2.0) might land a strictly Pareto-better point.
4. **Add tau_z-specific weighting separately** — since the tau_z improvement is 17× larger than the tau_y improvement, a direct `tau_z` channel up-weight may be more efficient than the smooth sin θ schedule. The PR #363 diagnostic flagged tau_z as the dominant gap; sin θ also up-weights tau_y (which is closer to baseline).

### Run details

- Command (Arm B; A and C identical except `--wallshear-theta-alpha`):
  ```bash
    --agent haku --wandb-group haku-theta-wallshear-screen \
    --wandb-name haku/theta-wallshear-B-alpha1 \
    --lr 1e-4 --weight-decay 5e-4 --no-compile-model --batch-size 4 \
    --validation-every 1 \
    --ema-decay 0.999 --lr-warmup-steps 2720 \
    --wallshear-theta-alpha 1.0 --seed 42
  ```
- Peak GPU memory: 67.5 GB (all arms identical, same config except alpha)
- Train time: 16200s = 270 min until train-timeout, then full_val + test (4.6h total each)
- Hardware: 3× H100 (1 GPU per arm, single-process), 4th GPU idle
- Train iters completed before timeout: 13099/21766 (60.2%)
- Code change footprint: `train.py` only (Config field, train_loss signature + body, main() call site). `--wallshear-theta-alpha` is auto-generated from the Config dataclass; manual argparse add not required.

morganmcg1:
ADVISOR: Excellent screen execution and analysis. W&B verification matches your report exactly:

| Arm | alpha | val_abupt | tau_x | tau_y | **tau_z** | surface_p | volume_p |
| A | 0.0 | 38.547% | 36.51% | 50.09% | 74.90% | 16.28% | 14.95% |
| **B** | **1.0** | **37.662%** | 36.41% | 49.76% | **69.16%** | 16.33% | 16.64% |
| C | 2.0 | 38.983% | 37.89% | 51.33% | 70.29% | 17.25% | 18.17% |

**Verdict: positive screen on tau_z.** Arm B beats control by 0.886pp on val_abupt and 5.74pp (-7.7% rel) on tau_z — the channel PR #363 flagged as the dominant gap. C overshoots (worse on every metric, including the channels it was meant to help, plus volume pressure starvation). The ordering A>B<C with B best is exactly the inverted-U signature of a useful regularizer at the right strength.

Caveat understood: 60% ep1 mid-warmup is too short to read absolutes (38% vs 7.546% bar). The ordering is the load-bearing signal here, and it's clean. Logging this as positive direction in EXPERIMENTS_LOG.

**Next step: tighten the bracket + decouple tau_z from tau_y.** Your suggested follow-up #4 is the strongest — sin(theta) up-weights tau_y AND tau_z together, but PR #363 flagged tau_z specifically (74.9% vs tau_y's 50.1%). Run a 4-arm screen, same single-GPU AdamW config, group `haku-thetaz-bracket-r25`:

| Arm | --wallshear-theta-alpha | --wallshear-z-weight | --wallshear-y-weight | rationale |
|---|---|---|---|---|
| A | 0.0 | 2.0 | 2.0 | control (matches PR #311 SOTA recipe) |
| B | 0.75 | 2.0 | 2.0 | softer theta (peak < B-prev) |
| C | 1.0 | 3.0 | 2.0 | confirmed B + extra tau_z mass |
| D | 1.0 | 2.0 | 1.5 | confirmed B + lower tau_y to free pressure heads |

Use 4 GPUs (one arm each). Same flags as your prior Arm B otherwise. Same kill threshold guidance (`--kill-thresholds '3000:val_primary/abupt_axis_mean_rel_l2_pct<25'`). Report at timeout-val.

The original alpha=1.0 winner will be re-tested at full convergence on the post-DDP branch (PR #355 emma) once it lands — that's the merge-readiness gate. This sweep is to find the right alpha+channel-weight combo to escalate then.

Implementation already done from PR #367 — just relaunch with the new 4-arm matrix.

morganmcg1:
STUDENT haku:

## Round 25 launched

4-arm thetaz-bracket screen launched at 16:54 UTC, group `haku-thetaz-bracket-r25`.

| Arm | GPU | --wallshear-theta-alpha | --wallshear-z-weight | --wallshear-y-weight | wandb-name |
|-----|-----|-------------------------|----------------------|----------------------|------------|
| A | 0 | 0.0  | 2.0 | 2.0 | haku/thetaz-r25-A-ctrl |
| B | 1 | 0.75 | 2.0 | 2.0 | haku/thetaz-r25-B-a075 |
| C | 2 | 1.0  | 3.0 | 2.0 | haku/thetaz-r25-C-z3 |
| D | 3 | 1.0  | 2.0 | 1.5 | haku/thetaz-r25-D-y15 |

Same single-GPU AdamW config as round 24:
- `--lr 1e-4 --weight-decay 5e-4 --batch-size 4 --lr-warmup-steps 2720 --ema-decay 0.999`
- `--model-layers 4 --model-hidden-dim 512 --model-heads 8 --model-slices 128`
- `--use-tangential-wallshear-loss`, `--no-compile-model`
- `--kill-thresholds 3000:val_primary/abupt_axis_mean_rel_l2_pct<25` (per your guidance)
- `--seed 42`

Logs: `logs_haku/pr367_r25_{A_ctrl,B_a075,C_z3,D_y15}.log`. PIDs: 454281–454284.

Will report timeout-val results at ~21:30 UTC (4.6h after launch, same convergence as round 24).

morganmcg1:
ADVISOR: Your branch is conflicting with the base (yi) — please rebase onto the current yi branch and then continue with the theta-conditioned wall-shear loss three-arm run as described in the PR instructions. After rebasing, run all three arms with `--wandb_group haku-theta-wallshear-screen` and post your results table here.

morganmcg1:
STUDENT haku:

## Rebase complete; killing Round 25 and relaunching the original 3-arm screen

Done:
- Killed the four `haku-thetaz-bracket-r25` runs (those were on the pre-rebase branch and become incomparable post-rebase).
- Rebased `haku/theta-conditioned-wallshear-loss` onto current `yi` (HEAD `fa971df`). Conflicts resolved cleanly:
  - `Config` field, `train_loss()` signature, and call site now coexist with `wallshear_huber_delta` from PR #317.
  - The per-point theta weight is now plumbed as `wallshear_point_weight` into `weighted_masked_mse_per_channel`, so it composes with the new Huber path (orthogonal flags) and per-axis diagnostics remain unweighted MSE.
- Smoke test on 4-GPU DDP + `--optimizer lion` + `--wallshear-theta-alpha 1.0` runs cleanly (~4 it/s on tiny model, no NaN, theta-weight diagnostic logged).

## Adapted run config (4 GPUs available, not 8)

PR body specifies `torchrun --standalone --nproc_per_node=8`. This pod has 4 GPUs (`GPUs per Student: 4`), so I'll use `--nproc_per_node=4`. Effective batch will be 16 instead of 32; everything else matches PR spec verbatim (`--optimizer lion`, `--lr-warmup-epochs 1`, etc., all of which now exist on yi after PR #355 + #317 merges).

Arms run **sequentially** — each gets a clean `SENPAI_TIMEOUT_MINUTES=360` budget and full DDP convergence. With 4-GPU DDP at batch=4 (~5440 steps/epoch) and 1-epoch warmup, each arm should clear warmup and run ~3-5k post-warmup steps. Ordering is the load-bearing signal as before.

| Arm | alpha | wandb-name |
|-----|-------|------------|
| A   | 0.0   | haku/theta-wallshear-A-alpha0 |
| B   | 1.0   | haku/theta-wallshear-B-alpha1 |
| C   | 2.0   | haku/theta-wallshear-C-alpha2 |

Group: `haku-theta-wallshear-screen`.

Launching arm A now (~18:55 UTC). Will report W&B run IDs and post-arm-A status when it timeouts (~01:00 UTC), then launch arm B, then C. Total wall-clock ~18h sequential.

morganmcg1:
STUDENT haku:

## Arm A killed at end-of-warmup; relaunching without kill-threshold

Arm A (alpha=0, run `3alkotk8`) hit the `--kill-thresholds '3000:val_primary/abupt_axis_mean_rel_l2_pct<25'` threshold and was terminated at step 5442 with `val_abupt=50.36%` after 120 min of training. Root cause is a config interaction, not a model failure:

- `--lr-warmup-epochs 1` with 4-GPU DDP (effective batch 16) → warmup runs for 5442 steps = exactly one full epoch.
- `--validation-every 1` → first validation point is at step 5442 (end of epoch 1 = end of warmup).
- The kill threshold checks at the first validation point, where the model has done **zero** post-warmup training. End-of-warmup `val_abupt` is necessarily large (~50% here), so the threshold is structurally unsatisfiable with this warmup/val cadence.

The threshold was originally designed for the single-GPU AdamW round-24 config, where 1 epoch was 21766 steps (vs 5442 here) and warmup was 12.5% of an epoch (not 100%). Under that regime val_abupt < 25% by step 3000 was reasonable; under DDP-4 + warmup-epochs=1 it cannot be.

**Action:** dropped `--kill-thresholds` entirely from the launch script (no replacement value would be both safe-from-divergence and satisfiable here, given the validation cadence). All other flags identical to PR spec (Lion, lr 1e-4, wd 5e-4, --lr-warmup-epochs 1, etc.).

**Relaunching Arm A now.** Same plan as before: sequential A → B → C, ~6h each, total ~18h wall-clock. With 270 min train_timeout under DDP-4 (~1.91h/epoch), each arm will get ~2.3 epochs total = ~1.3 epochs of post-warmup training. Better than nothing; ordering remains the load-bearing signal.

W&B run `3alkotk8` (killed Arm A) is preserved for the post-mortem; it will not be in the final results table. The new Arm A run id will be reported once W&B logs settle.

morganmcg1:
ADVISOR: Non-blocking note while you run: your launch config shows `--weight-decay 5e-4`. The yi baseline (PR #309 thorfinn, 9.039%) uses `--weight-decay 1e-4` for Lion. Do NOT interrupt the current run — let it finish. But when you launch Arms B and C (or any rerun), use `--weight-decay 1e-4` to match the yi stack. The control arm (alpha=0, Arm A) will be slightly mis-matched to baseline if wd differs, which could cloud the alpha signal. If the current Arm A finishes with wd=5e-4, note this in your results table and we will account for it in the analysis.

morganmcg1:
ADVISOR: Arm A ep2 update — val_abupt=**46.35%** at step 10885. This is concerning. The model is converging slowly. A few observations:

1. **Weight-decay mismatch**: Your launch uses wd=5e-4 vs yi baseline 1e-4. This can affect Lion convergence speed, though probably not by 37pp.
2. **Epoch 1 is pure warmup**: With lr-warmup-epochs=1 and DDP-4, the model spends all of epoch 1 ramping lr — zero useful learning. The first real learning epoch is epoch 2. So effectively you're at ep1 of real training at this point.
3. **DDP-4 effective batch**: Lion at effective-bs=16 (DDP-4) can be structurally unstable per yi research history. Alpha=0 arm should give pure baseline performance — if Lion is struggling here that's separate from the theta-weight hypothesis.

**Please check your W&B run config**: confirm `wallshear_theta_alpha: 0` is in the config for this run (run ID: jbbw3enm).

**Action**: Let Arm A finish all epochs, report ep3/ep4. Also — **launch Arms B (alpha=1.0) and C (alpha=2.0) NOW in parallel**, both with wd=1e-4, rather than waiting for Arm A to finish. The 360-min pod budget will not accommodate sequential 3-arm runs at ~114 min/epoch. Running them concurrently (background &) lets you get all three arms' results before the pod times out.

morganmcg1:
STUDENT haku:

## Resource constraint and proposed plan

Acknowledged the new direction (parallel B+C with wd=1e-4). One physical constraint:

**Arm A is occupying all 4 GPUs in DDP-4 mode** (`--nproc_per_node=4`, ~70 GiB / 97 GiB each, 100% util). There is no free GPU to launch B and C onto right now without either killing A or sharing GPUs (which OOMs).

Arm A status (W&B `jbbw3enm`): runtime 4h 7min of 4.5h train_timeout → ~25 min until train_timeout fires, then ~90 min full_val + test, so A frees its 4 GPUs in ~2 h (~02:55 UTC). A's config is `wallshear_theta_alpha=0`, `weight_decay=0.0005` (the wd=5e-4 you flagged earlier).

I see two coherent paths and would like a quick decision:

**Path P1 — preserve A, then parallel B+C (recommended).**
- Let A finish naturally (~2 h). Get ep3-timeout-val + full_val/test on the wd=5e-4 control as you said before ("note in results, will account for it").
- After A frees GPUs, launch B (alpha=1.0, wd=1e-4) and C (alpha=2.0, wd=1e-4) **concurrently, each on 2 GPUs DDP-2** (per-GPU batch=4 → effective bs=8 vs A's eff bs=16). Each run gets the full SENPAI_TIMEOUT_MINUTES=360 budget, so ~3 epochs of training each.
- Tradeoff: B and C have eff bs=8 vs A's eff bs=16 — clean B-vs-C comparison, slightly noisy A-vs-B/C comparison.
- Total wall-clock: ~8 h from now.

**Path P2 — kill A NOW, restart all 3 in parallel single-GPU (matches \"NOW in parallel\" literally).**
- Kill A. Loses wd=5e-4 control and current epoch-3 progress.
- Launch A' / B / C single-process on 3 separate GPUs (4th spare for kill-and-rerun), all with wd=1e-4 Lion.
- At single-GPU batch=4, 1 epoch ≈ 7.5 h, so all 3 timeout at ~60 % of ep1, mid-warmup — matches the round 24 regime exactly. Ordering is the load-bearing signal as before.
- Tradeoff: clean 3-way comparison with consistent wd=1e-4, but mid-warmup results.
- Total wall-clock: ~6 h from now.

I have killed the old `chain_arms.sh` watcher (it would have launched B sequentially after A on 4 GPUs DDP-4, which is no longer the plan). No other action taken.

**Default if no reply within ~30 min**: I will go with **Path P1** (let A finish, then B+C parallel each on 2 GPUs DDP-2 with wd=1e-4) since it preserves A's results and respects \"let A finish all epochs\". Please nudge if you want P2 instead.

morganmcg1:
STUDENT haku:

## Status update (default Path P1 in effect)

No advisor reply since 01:08 UTC, so per my message I am going with **Path P1**: let Arm A finish naturally, then launch B+C concurrently each on 2 GPUs DDP-2.

**Current state (01:26 UTC):**
- Arm A (`jbbw3enm`, alpha=0.0, wd=5e-4) — train_timeout firing now (270 min budget). Latest val checkpoints:
  - ep1: val_abupt=50.36% (end-of-warmup, expected high)
  - ep2: val_abupt=46.35% (slope ≈ −0.74pp/1k steps)
  - ep3 timeout-val expected ~now → A enters full_val (DDP-4) → test → done at ~03:00 UTC
- Watcher (`watch_launch_BC.sh`, pid 519761) is polling A's torchrun parent (pid 485083). Once A frees the 4 GPUs, it auto-launches B (alpha=1.0) on GPUs 0–1 and C (alpha=2.0) on GPUs 2–3, both DDP-2, **both with wd=1e-4** per your 00:59 UTC guidance, rdzv ports 29501/29502.

**Tradeoffs vs your "launch NOW in parallel":**
- A is occupying all 4 GPUs in DDP-4 mode (~70 GiB each, 100% util); there is no co-location possible without OOM. Killing A to free GPUs would discard the wd=5e-4 control entirely.
- B/C will run at effective batch 8 (DDP-2 × 4) vs A's 16 (DDP-4 × 4). Internal B-vs-C comparison stays clean; A-vs-B/C has both wd and eff-bs deltas — I will note both in the final results table.
- Total wall-clock: ~03:00 UTC for A done, then +6h for B/C parallel runs → results by ~09:00 UTC. If pod times out before, B+C will produce timeout-val checkpoints (same regime as round 24).

If you want me to kill A and switch to Path P2 (3-arm parallel single-GPU, all wd=1e-4), reply now and I will switch.

morganmcg1:
STUDENT haku:

## Arm A finished; B and C launched and running cleanly

**Arm A (`jbbw3enm`, alpha=0.0, wd=5e-4)** — finished at ~01:46 UTC. Configs verified in W&B (per your 00:59 ask): `wallshear_theta_alpha=0`, `weight_decay=0.0005`. Best val_abupt was at epoch 3 (45.55%); ep1 50.36% (end-of-warmup), ep2 46.35%, ep3 45.55%. Slope is flattening but still negative. Final test results:

| Channel | val_pct | test_pct |
|---|---|---|
| abupt_axis_mean | 45.555 | 45.932 |
| surface_pressure | 7.372 | 7.032 |
| volume_pressure | 6.876 | 13.514 |
| wall_shear (overall) | 60.654 | 60.610 |
| wall_shear_x | 34.379 | 34.975 |
| wall_shear_y | 42.573 | 42.290 |
| wall_shear_z | **136.572** | **131.851** |

The huge `wall_shear_z` (136%) on the alpha=0 control is the dominant gap for A. This is the same channel PR #363 flagged, and the same channel that improved most under alpha=1 in the round-24 screen (74.9%→69.2%).

**Arms B (`axf96lri`, alpha=1.0) and C (`my4q8hfa`, alpha=2.0)** — launched 01:48 UTC on GPUs 0,1 and 2,3 (DDP-2 each, eff bs=8 vs A's eff bs=16). Both with **wd=1e-4** per your 23:58 UTC guidance. Both running cleanly, no errors, no NaN, ~1.27s/it:

| Arm | run_id | step (live) | theta_weight_mean (live) | wd | alpha |
|---|---|---|---|---|---|
| B | axf96lri | 2598 / 10883 (24% ep1) | 1.607 ✓ | 1e-4 | 1.0 |
| C | my4q8hfa | 2573 / 10883 (24% ep1) | 2.245 ✓ | 1e-4 | 2.0 |

theta_weight_mean values match the round-24 numbers (1.602, 2.216), so per-point weighting is active and stable.

**Schedule:** epoch 1 (warmup) finishes ~05:35 UTC. Train_timeout (270 min) fires ~06:18 UTC, then full_val + test takes ~90 min, so B and C results land ~07:50 UTC. I'll report the 3-arm table as soon as both finish.

Caveats for the final read:
- A: wd=5e-4 (mismatch from baseline), eff bs=16 (DDP-4)
- B,C: wd=1e-4, eff bs=8 (DDP-2)
- The B-vs-C comparison is clean (same wd, same bs). The A-vs-B/C comparison has both wd and eff-bs deltas, but **all three share the same alpha encoding-active diagnostic**, so the directional signal on tau_z should still be readable.

If you'd like me to relaunch any arm, please let me know now.

morganmcg1:
STUDENT haku:

## Results — 3-arm theta-conditioned wall-shear screen

All three arms hit the train-timeout at 270 min, ran full_val (34 cases) and test (50 cases), and finished cleanly. Group: `haku-theta-wallshear-screen` (post-rebase onto current `yi`, post PR #311 / #317 / #355).

### Primary table (full_val_primary, single best-epoch checkpoint)

| Arm | alpha | val_abupt | wall_shear | wall_shear_x | **wall_shear_y** | **wall_shear_z** | surface_p | volume_p | θ̄_weight | W&B run |
|-----|-------|-----------|------------|--------------|------------------|------------------|-----------|----------|-----------|---------|
| A   | 0.0   | 45.555%   | 60.654%    | 34.379%      | 42.573%          | 136.572%         | 7.372%    | 6.876%   | n/a       | [jbbw3enm](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/jbbw3enm) |
| B   | 1.0   | 45.444%   | 59.447%    | 34.748%      | 43.097%          | 131.867%         | 8.711%    | 8.794%   | 1.598     | [axf96lri](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/axf96lri) |
| **C** | **2.0** | **43.079%** | **56.161%** | 35.512%   | **41.672%**      | **120.480%**     | 8.656%    | 9.078%   | 2.208     | [my4q8hfa](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/my4q8hfa) |

### Test table (test_primary, same checkpoint)

| Arm | alpha | test_abupt | wall_shear | wall_shear_x | wall_shear_y | wall_shear_z | surface_p | volume_p |
|-----|-------|-----------|------------|--------------|--------------|--------------|-----------|----------|
| A   | 0.0   | 45.932%   | 60.610%    | 34.975%      | 42.290%      | 131.851%     | 7.032%    | 13.514%  |
| B   | 1.0   | 45.881%   | 59.620%    | 35.316%      | 42.691%      | 128.095%     | 8.411%    | 14.892%  |
| **C** | **2.0** | **43.524%** | **56.232%** | 36.029%   | **41.292%**  | **116.765%** | 8.353%    | 15.183%  |

### Per-epoch val_abupt trajectory

| Arm | alpha | ep1 (end-warmup) | ep2 | ep3 | best |
|-----|-------|-------------------|------|------|------|
| A   | 0.0   | 50.363%          | 46.347% | **45.555%** | ep3 (timeout-induced) |
| B   | 1.0   | 46.865%          | **45.444%** | n/a | ep2 (timeout-induced) |
| C   | 2.0   | 44.317%          | **43.079%** | n/a | ep2 (timeout-induced) |

(A reached ep3 because DDP-4 → eff-bs 16 → 5442 steps/epoch. B and C ran DDP-2 → eff-bs 8 → 10883 steps/epoch, so the 270-min budget only fits ~1.15 epochs past warmup.)

### θ-weight diagnostic
- Arm A (alpha=0.0): not logged (else-branch path) — confirms unweighted loss path is unchanged ✓
- Arm B (alpha=1.0): θ̄ = 1.598 → E[sin θ] ≈ 0.598 (matches round-24's 1.602 to 3rd digit) ✓
- Arm C (alpha=2.0): θ̄ = 2.208 → E[sin θ] ≈ 0.604 ✓

Per-point theta weighting is active and stable; the population skews cross-flow as PR #363 found.

### Caveats — important for the read

1. **A's wd mismatch (5e-4 vs 1e-4 for B/C).** Caught by the advisor at 23:58 UTC after A had already launched. Re-launching A with wd=1e-4 would tighten the A-vs-B/C comparison, but the B-vs-C comparison is internally clean (same wd, same DDP-2 eff-bs, same seed=42, only `wallshear_theta_alpha` differs).
2. **Effective batch mismatch (A:DDP-4 eff-bs 16 vs B/C:DDP-2 eff-bs 8).** A could not be co-located with B/C without OOM; killing A would have lost the wd=5e-4 control entirely. Path P1 (let A finish, parallel B+C) was the chosen tradeoff per my 01:08 UTC default.
3. **Severely under-converged.** All three arms hit the train timeout at ~step 12500 with `lr-warmup-epochs=1` eating most of training (`val_abupt` ≈ 43–46% vs the 7.546% merge bar). The **ordering** is the load-bearing signal here, not the absolute numbers; absolute deltas may compress, expand, or flip with full-convergence runs.

### Analysis — what happened

**alpha=2.0 is the winner under DDP / Lion / wd=1e-4.** Arm C beats control on val_abupt (45.55 → 43.08, **−2.48 pp / −5.4% rel**) and on every wall-shear axis except streamwise:
- `wall_shear_z` (the dominant gap from PR #363): 136.57% → 120.48%, **−16.09 pp / −11.8% rel**.
- `wall_shear_y`: 42.57% → 41.67%, −0.90 pp / −2.1% rel.
- `wall_shear_x` (streamwise): 34.38% → 35.51%, +1.13 pp / +3.3% rel — small degradation. Mechanically expected: cross-flow points (large θ) get up-weighted, and at those points tau_x is small but its squared error is now amplified, gently pulling capacity away from the streamwise component.

**B (alpha=1.0) is essentially tied with control on val_abupt** (45.44 vs 45.55) — mid-α gets weight signal active (θ̄≈1.60) but doesn't yet move the needle in this regime. The slope of improvement is steepening with α here, not flattening.

**Pressure side-channel.** Surface and volume pressure both degrade ~1.3–2.2 pp under alpha>0 (e.g. surface_p 7.37 → 8.66 on C). This is the standard cost of re-allocating wall-shear gradient signal — should be tracked but isn't the primary objective.

### Reconciling with round-24

Round 24 (single-GPU AdamW, 60% ep1 mid-warmup, lr=1e-4) found the inverted-U signature A>B<C with B=alpha=1.0 best. **This round (DDP / Lion / wd=1e-4 / past warmup) finds a different ordering: A≈B>C, with C=alpha=2.0 best.** Plausible reasons the optimum shifts up:
- **Optimizer / regularization:** Lion + wd=1e-4 has a different sensitivity profile to the loss-shape change than AdamW.
- **Convergence regime:** round 24 was pure-warmup mid-epoch-1; this round has full warmup + ~10–15% post-warmup. The "C overshoot" verdict was an artifact of the under-converged warmup-only AdamW regime.
- **Effective LR scaling:** the larger weight (1+α·sin θ) up to 3× under C means an effectively larger gradient on cross-flow points; under Lion's sign-update this composes differently than under AdamW.

The round-24 conclusion ("alpha=1 best, alpha=2 overshoots") is not a property of the alpha schedule per se — it was a property of that particular under-converged AdamW regime. **In the proper Lion+wd=1e-4 yi regime, alpha=2.0 is the screen winner**, and the tau_z gain is large (−16 pp / −11.8% rel) on the channel PR #363 specifically flagged.

### Run details

Command (Arm C; A and B identical except `--wallshear-theta-alpha` and Arm A's `--weight-decay`):
```bash
CUDA_VISIBLE_DEVICES=2,3 torchrun --rdzv-backend=c10d --rdzv-endpoint=127.0.0.1:29502 \
  --nproc_per_node=2 train.py \
  --agent haku --wandb-group haku-theta-wallshear-screen \
  --wandb-name haku/theta-wallshear-C-alpha20 \
  --optimizer lion --lr 1e-4 --weight-decay 1e-4 \
  --wallshear-theta-alpha 2.0 --seed 42
```
- Peak GPU memory: ~70 GiB / 97 GiB per GPU on each arm
- Train iters at timeout: A 12270/16329 (75% ep3 = ~2.25 ep), B 12508/21766 (1.15 ep), C 12443/21766 (1.14 ep)
- Train wall-clock: 275.6 min/arm; full_val + test: ~30 min/arm
- Hardware: 4× RTX PRO 6000 (97 GB VRAM each); A used GPUs 0–3 DDP-4, then B used 0–1 DDP-2 + C used 2–3 DDP-2 in parallel
- Code change footprint: `train.py` only — Config field, train_loss signature + body, main() call site (already merged in 0e3123b on this branch). Plumbs `wallshear_point_weight` into `weighted_masked_mse_per_channel` so it composes with the PR #317 Huber path.

### Suggested follow-ups

1. **Tighter alpha bracket on the high side** — run alpha ∈ {1.5, 2.0, 2.5, 3.0} to localize the optimum. Current data shows monotone improvement A→B→C over alpha ∈ {0, 1, 2}; the peak may be at alpha ≥ 2. Note PR #66 stability ceiling W<3.0 → alpha < 2.0/E[sin θ] ≈ 3.3 hard ceiling.
2. **Re-run A with wd=1e-4** (clean control) to remove the wd-mismatch caveat. ~6h on 4 GPUs DDP-4.
3. **Full-convergence A vs C.** If the merge bar is val_abupt < 7.546% at convergence (PR #311), the screen-winning alpha=2.0 needs to be trained for >2 epochs to confirm the improvement holds. The natural next escalation is alpha=2.0 + full epochs (~10–15) on the post-PR #355 DDP branch.
4. **Decouple tau_z from tau_y (channel-specific weighting).** Since tau_z drives the bulk of the alpha signal (−16 pp vs tau_y's −0.9 pp), a direct tau_z up-weight (e.g. `--wallshear-z-weight 4.0` with alpha=0) might be a simpler, more targeted alternative. Worth A/B-ing against alpha=2.0.
5. **Pressure-head Pareto recovery.** Pair alpha=2.0 with reduced wallshear y/z channel weights (e.g. y=z=1.5 instead of 2.0) to claw back the surface_p degradation while keeping tau_z gains.

morganmcg1:
ADVISOR: Results reviewed. Promising directional signal but under-converged — sending back for full-convergence confirmation run.

**Key finding:** alpha=2.0 (Arm C) beats control on val_abupt (45.55% → 43.08%, −2.48pp / −5.4% rel) and on tau_z (−16.09pp / −11.8% rel). This is the dominant binding gap from PR #363 analysis. The ordering is monotone: A≈B>C, steepening slope, not inverted-U.

**However:** All arms hit timeout at ~1.1–2.25 epochs. Absolute values are 43–46% vs the 7.546% merge bar. The ordering signal is real but the absolute gap could compress, expand, or flip at convergence. Per your own analysis, this is a screen result — the load-bearing signal is ordering, not magnitude.

**Decision: Request changes.** To cross the merge bar, run a full-convergence confirmation:

**Arm C (alpha=2.0) full-convergence run:**
- DDP-4 (4 GPUs), Lion optimizer, wd=1e-4, lr=1e-4, batch-size 4
- Exact PR #311 (edward STRING PE) yi-SOTA recipe, just add --wallshear-theta-alpha 2.0
- Train to timeout or ep12+ (whatever budget allows)
- Report val_abupt and all 5 axis metrics at best checkpoint
- Compare against PR #311 baseline: val_abupt=7.546%, test_abupt=8.771%

**Clean control (alpha=0.0, wd=1e-4):** If you have a second GPU slot, run Arm A clean with wd=1e-4 to remove the weight-decay caveat. But the priority is the alpha=2.0 confirmation run.

**Command template:**
```bash
  --agent haku --wandb-group haku-theta-wallshear-full \
  --wandb-name haku/theta-wallshear-alpha20-full \
  --optimizer lion --lr 1e-4 --weight-decay 1e-4 \
  --wallshear-theta-alpha 2.0 --seed 42
```
Match the PR #311 STRING-PE config exactly — add whatever --string-pe flag that run used. The tau_z gain is real (−11.8% rel on the dominant binding constraint). This is worth confirming at convergence.

morganmcg1:
STUDENT haku:

## Full-convergence alpha=2.0 confirmation run launched

Per advisor request 2026-05-03 07:45 UTC. Launched DDP-4 alpha=2.0 on the exact PR #311-equivalent yi config (modulo nproc_per_node=4 since this pod has 4 GPUs not 8, and **no STRING-PE flag** — STRING-sep PE is not yet on yi per BASELINE.md note from 2026-05-02 19:00 UTC; PR #420 fern is the port-in-flight). Group `haku-theta-wallshear-full`.

| Field | Value |
|---|---|
| W&B run | [zvros0ej](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/zvros0ej) |
| Wandb name | `haku/theta-wallshear-alpha20-full` |
| Group | `haku-theta-wallshear-full` |
| GPUs | 0,1,2,3 (DDP-4) |
| Per-GPU batch | 4 (eff bs 16) |
| Optimizer | Lion |
| LR / WD | 1e-4 / 1e-4 |
| Warmup | --lr-warmup-epochs 1 |
| Wallshear y/z weight | 2.0 / 2.0 |
| Theta alpha | **2.0** |
| Epochs cap | 12 (will hit 270-min train timeout first) |
| Seed | 42 |
| iter time (ep1, observed) | 1.49 s/it → ~135 min/epoch → ~2 ep before timeout |

Expected: ep1 = warmup (val_abupt likely 45-50%), ep2 timeout-val checkpoint. Same regime as round-26 Arm A (DDP-4 alpha=0 wd=5e-4) but with corrected wd=1e-4 and alpha=2.0. Will report at ~14:00 UTC.

**Re: clean alpha=0 control:** I cannot co-locate a second DDP-4 run (alpha=2.0 occupies all 4 GPUs at ~70 GB / 97 GB each). I will launch DDP-4 alpha=0 wd=1e-4 sequentially after this run finishes (~14:00 UTC), giving a clean A/C comparison with matched eff bs and wd. Total wall-clock ~12h. If you'd prefer I instead launch alpha=0 in DDP-2 parallel now (eff bs 8 mismatch, finishes at the same time), nudge and I'll switch.

morganmcg1:
ADVISOR: Diagnostic check on `zvros0ej` (alpha=2.0 full-convergence DDP-4) — flagging this as **likely FAIL** before ep3 lands.

**Observed metrics (W&B `zvros0ej`, step 11190, mid-ep3, runtime 4h05m):**

| epoch | step | val_abupt | ws_z | ws_y | ws_x | sp |
|---|---|---|---|---|---|---|
| 1 (warmup) | 5442 | 50.39% | 119.51% | 56.34% | — | 17.82% |
| 2 | 10885 | 43.72% | **122.68%** | 43.96% | 36.15% | 8.11% |
| 3 (proj) | ~16327 | ~37.05% | ~125.86% | — | — | — |

**Critical anomaly: ws_z is WORSENING ep1→ep2** (+3.17pp). The slope gauge confirms: `val/slope/val_primary_wall_shear_z_rel_l2_pct/per_1k_steps = +0.5823` (positive = degrading). The very metric this experiment was designed to drive down is going UP.

**Linear projection vs merge bar (9.039%):**
- ep1→ep2 abupt slope: -6.67pp/epoch
- ep3 projected: ~37.05% — **28pp above the merge bar**
- Even if remaining 9 epochs hold the same slope (they won't), this run cannot reach 9.039% within `SENPAI_TIMEOUT_MINUTES=360` (~3 epochs/budget)

**Contrast with R26 Arm C (the screen winner this is supposed to confirm):**
- R26 Arm C (alpha=2.0, **DDP-2**, eff-bs=8): ws_z 135.9% → 120.48% at ep2 (large IMPROVEMENT)
- `zvros0ej` (alpha=2.0, **DDP-4**, eff-bs=16): ws_z 119.51% → 122.68% at ep2 (REGRESSION)

The DDP-4 run starts ws_z at a much lower point (119.5% vs 135.9%) but then drifts upward. This is not a replication of the R26 screen.

**Hypothesis on what's different:**
- DDP-4 eff-bs=16 vs DDP-2 eff-bs=8 + theta-weighted loss may interact non-trivially. With `wallshear_y_weight=2.0 + wallshear_z_weight=2.0 + wallshear_theta_alpha=2.0`, peak per-point weighting on cross-flow ws is ~6x. Larger eff-bs averages the gradient signal across more theta-diverse points, which may dilute the cross-flow steering you wanted.
- Alternatively: at DDP-4 the warmup-end ws_z is already in a low-loss flat basin, and the theta-weighted loss is pushing it OUT of that basin rather than further IN.

**Recommendation:**
1. Let `zvros0ej` finish ep3 (it's already mid-epoch) — confirm the projection. Don't kill it; the data point is useful.
2. After it lands, treat the alpha=2.0 finding as **DDP-2-specific**, not transferable to DDP-4 at the current weights.
3. **Next experiment to assign**: clean `alpha=0.0` DDP-4 control with same wd=1e-4 to establish the no-theta DDP-4 baseline. Then sweep alpha ∈ {0.5, 1.0, 1.5} at DDP-4 to find the DDP-4-regime sweet spot. Don't assume the R26 inverted-U survives the eff-bs change.

I'll mark this PR for the followup once ep3 completes.

morganmcg1:
ADVISOR: Closing PR #367 because ADVISOR closing — full-convergence run `zvros0ej` is conclusively diverged on the targeted metric. Final check at 256min runtime / step 11726:

- val_abupt: 43.72% (vs 9.039% bar — 4.84× over)
- wall_shear_z: **122.68%** (worse than initialization; the very channel this experiment was designed to drive down)
- wall_shear_y: 43.96%
- ws_z slope: positive (degrading), confirmed across ep1→ep2

The screen-stage signal (alpha=2.0 beats control by 2.5pp at sub-epoch in PR #367's earlier run) did not survive scaling to full DDP-4 budget. The most likely explanation: theta-conditioned weighting that performed under tangential loss + 1-epoch screening collapses when applied to 3-epoch convergence on the canonical Lion/yi stack — the cross-flow upweight inflates gradient magnitude on the very samples where the model lacks capacity to track tau_z, driving the loss landscape into a region the model cannot fit.

**Theta-conditioned loss weighting on yi is now a confirmed dead-end at this architecture/budget.** Adding to DEAD ENDS.

Killing the run frees haku for a fresh assignment. Thank you for the diligent diagnostics — the failure mode is informative for the loss-rebalancing family more broadly (PR #529 stream-normal weighting, which uses geometric weights at smaller magnitude, is the principled descendant)..

morganmcg1:
STUDENT haku:

## Diagnostic acknowledged — letting zvros0ej finish ep3 per your guidance

Confirmed your projection from W&B (run `zvros0ej`, current state at step 11190 / mid-ep3):

- ep1 (end-warmup): val_abupt=50.39%, ws_z=119.51%
- ep2 (step 10885): val_abupt=43.72%, ws_z=**122.68%** (+3.17 pp ep1→ep2)
- Slope gauge: `ws_z_rel_l2_pct/per_1k_steps = +0.582` (degrading)
- Slope gauge: `abupt/per_1k_steps = -1.226` (improving on the ensemble metric)
- ws_z slope is positive while ws_y (-2.28), ws_x (-0.93), surface_p (-1.78), volume_p (-1.72) all improve. The theta-weighted loss is broadly helping the other channels but actively pushing ws_z OUT of its low-loss basin under DDP-4.

Train timeout (270 min) will fire ~12:22 UTC; full_val + test lands ep3 numbers ~12:55 UTC. Will post the ep3 row at that point.

## Auto-queued: clean alpha=0.0 DDP-4 control (matches your step 3)

A watcher (`logs_haku/watch_alpha0.log`, pid 559482) is polling for the alpha=2.0 torchrun to exit. Once GPUs free (~12:55 UTC) it will sequentially launch the **clean alpha=0 DDP-4 control with wd=1e-4**:
```
  --agent haku --wandb-group haku-theta-wallshear-full \
  --wandb-name haku/theta-wallshear-alpha00-full \
  --optimizer lion --lr 1e-4 --weight-decay 1e-4 ... \
  --wallshear-theta-alpha 0.0 --epochs 12 --seed 42
```
This was set up before your 11:58 UTC diagnostic and aligns with step 3 of your recommendation. It will produce the no-theta DDP-4 baseline you want to compare the alpha-sweep against.

## Standing by for the alpha ∈ {0.5, 1.0, 1.5} sweep

Awaiting your followup instructions for the 3-arm DDP-4 sweep. Two configurations to choose from depending on timing pressure:

- **DDP-4 sequential (purest)**: 6h × 3 arms = 18h total. Each arm gets full 4-GPU DDP, eff-bs=16.
- **DDP-2 × 2 parallel + 1 DDP-4**: ~12h total. Mixes eff-bs=8 (parallel) with eff-bs=16 (last arm); cleaner if all three are DDP-2 parallel + sequential (12h).

I will hold launch until you signal which config you want and confirm the alpha values.

PR #330, opened 2026-05-02
[violet] DDP8 radford-champion port: 4L/512d/8H + Fourier + EMA + gc=0.5 + lr=4.8e-4

morganmcg1:
## Hypothesis

The bengio Wave 1 winner (4L/256d/SincosEmbed/no-EMA, val_abupt=7.2091%) was found at lower capacity than the historical radford champion (which used 4L/512d/8H + Fourier + EMA=0.9995 + gc=0.5 + lr=4.8e-4 + T_max=36). Wave 2 attempts to scale alphonse to 5L/256d (PR #174) and 5L/384d (PR #179) showed sluggish progress (ep14-20 in 8.4-8.5% range), suggesting the bengio recipe under-utilizes capacity at standard DDP4.

The radford-champion full stack at **DDP8 = 8 GPUs** lets us:
1. Double the effective batch size (DDP8 batch_size=2 ≈ 16 samples/step) without lowering LR proportionally — `lr=4.8e-4` is ~1.6× alphonse's 3e-4, exactly the linear-scaling-rule offset.
2. Run the wider 512d / 8H model at full capacity within memory budget (96GB × 8 = 768GB activations).
3. Stack ALL the historical radford champion levers in one experiment: width, EMA, tighter clip, longer cosine schedule, Fourier PE.

This is the **highest-EV untried config** on bengio. Either it crushes the baseline (high upside) or it diverges/plateaus (clean dead-end signal). Both outcomes are valuable.

## Instructions

Single trial. **Requires a DDP8 pod** (8 GPUs). If your pod is DDP4, escalate to advisor immediately so you can get a DDP8 reassignment — do NOT run this on DDP4.

```bash
cd target/ && torchrun --standalone --nproc-per-node=8 train.py \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --use-ema \
  --ema-decay 0.9995 \
  --lr 4.8e-4 \
  --lr-cosine-t-max 36 \
  --epochs 36 \
  --grad-clip-norm 0.5 \
  --fourier-pe \
  --no-compile-model \
  --kill-thresholds "80000:val_primary/abupt_axis_mean_rel_l2_pct>22,170000:val_primary/abupt_axis_mean_rel_l2_pct>12.0,400000:val_primary/abupt_axis_mean_rel_l2_pct>9.0" \
  --wandb-group bengio-wave7-radford-champion-ddp8 \
[...]

PR #373, opened 2026-05-02
[stark] Stochastic Weight Averaging over cosine tail (ep35-50)

morganmcg1:
## Hypothesis

**Stochastic Weight Averaging (SWA) over the cosine-decay tail will produce a flatter, lower-loss minimum** than the single-checkpoint cosine final iterate. SWA averages weights from multiple late-cosine epochs, exploiting the wide-flat-minimum structure that cosine schedules produce around the LR minimum.

**Evidence:**
1. alphonse #174 (run `vu4jsiic`, T_max=50) shows the cosine knee at ep31–35 produces continuous descent — multiple late checkpoints have similar val abupt (ep31.5=7.213%, ep36.8=7.085%, ep39.7=7.040%), suggesting a wide-flat-minimum geometry where SWA shines.
2. SWA (Izmailov et al. 2018, "Averaging Weights Leads to Wider Optima") consistently delivers 0.05–0.30pp improvements over single-checkpoint cosine in vision/NLP at near-zero cost.
3. EMA (used by violet #330) is a related averaging mechanism but with exponentially-decaying weights; SWA uses uniform averaging over a chosen window, which is a stronger averaging signal once the model has converged.
4. The val_abupt metric is computed from **single-step weights** at validation; SWA evaluation requires (a) recomputing BatchNorm statistics with the averaged weights, then (b) running the standard val pipeline.

**Mechanism:** Save weight snapshots every epoch from ep35→ep50 (16 snapshots), average them, recompute model statistics, evaluate. The averaged model should sit in a wider basin and generalize better — particularly on the val/test gap (currently ~2x on vol_p).

**Expected outcome:** ep50-SWA abupt **0.05–0.20pp better** than ep50-best-single — would push us from ~7.00% (alphonse projected) toward 6.85–6.95%.

## Instructions

This is a **code change** experiment.

### Step 1 — Implement SWA

Add to `target/trainer_runtime.py` (or `train.py`, whichever owns the training loop). Use **`torch.optim.swa_utils.AveragedModel`**:

```python
from torch.optim.swa_utils import AveragedModel, update_bn

# After model+optimizer construction, add:
swa_model = AveragedModel(model)
swa_start_epoch = args.swa_start_epoch  # default 35
```

In the per-epoch loop, **after** the standard validation step, if `epoch >= swa_start_epoch`:
```python
swa_model.update_parameters(model)  # uniform running average
```

Add a CLI flag `--swa-start-epoch` (int, default `0` meaning OFF; `35` to start averaging from ep35).

### Step 2 — SWA evaluation at the end of training

After the final epoch, if SWA is enabled:
1. **Update BN statistics**: `update_bn(train_loader, swa_model, device=device)` — runs forward passes over training data (no backward) so the averaged model has correct running stats. **CRITICAL**: do this on the rank-0 process only after rendezvous; broadcast the averaged BN stats if needed.
2. **Run val_primary evaluation** with `swa_model` and log all `val_primary/*` metrics under a `swa_val_primary/*` namespace.
3. **Save SWA checkpoint** under `output_dir/swa_best.pt`.

### Step 3 — Launch with SWA on alphonse-equivalent recipe

```bash
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --fourier-pe \
  --no-use-ema \
  --lr 3e-4 \
  --lr-cosine-t-max 50 \
  --epochs 50 \
  --swa-start-epoch 35 \
  --no-compile-model \
  --kill-thresholds "100000:val_primary/abupt_axis_mean_rel_l2_pct<14,300000:val_primary/abupt_axis_mean_rel_l2_pct<10.5,500000:val_primary/abupt_axis_mean_rel_l2_pct<8.5" \
  --wandb-group bengio-wave9-swa \
  --wandb-name stark/swa-cosine-tail-rank0 \
  --agent stark
```

**Critical:** Confirm `n_params=3,249,813` from launch banner.

**Critical:** Kill-threshold operators MUST use `<VALUE` (kill if metric ≥ VALUE). Do not use `>VALUE`.

**`--no-compile-model` mandatory** (PyTorch 2.x Inductor crash).

### Step 4 — Reporting

Post these to the PR as comments:

1. **At ep5/10/20/30/40/45**: standard `val_primary/*` metrics (raw model). Should track the alphonse trajectory closely.
2. **At ep50 (final)**: BOTH `val_primary/*` (raw, single-checkpoint) AND `swa_val_primary/*` (SWA-averaged) on the same table. The delta is the experiment payoff.
3. **At ep50 (final)**: `test_primary/*` metrics from BOTH (a) best-val raw checkpoint reload and (b) SWA checkpoint. Report both side-by-side.

## Acknowledgment Required

Within 30 minutes of picking up this PR, post:
1. Confirmation you read the body
2. Plan/diff summary for the SWA implementation (what files you'll edit, where the AveragedModel hooks go)
3. Trial run ID (will fill once code is ready)

## Baseline (target to beat)

Source of truth: `/BASELINE.md`. Current best **merged** baseline is PR #74 (run `m9775k1v`):

| Metric (val_primary) | Current Best | AB-UPT Target |
|--------|-------------------|--------------|
| abupt_axis_mean_rel_l2_pct | **7.2091** | 4.51 |
| surface_pressure_rel_l2_pct | 4.802 | 3.82 |
| wall_shear_rel_l2_pct | 8.160 | 7.29 |
| volume_pressure_rel_l2_pct | **4.166** ✓ | 6.08 |
| wall_shear_x_rel_l2_pct | 7.109 | 5.35 |
| wall_shear_y_rel_l2_pct | 9.100 | 3.65 |
| wall_shear_z_rel_l2_pct | 10.869 | 3.63 |

Unofficial leader (WIP): **alphonse #174 at 7.040% abupt** (run `vu4jsiic`, T_max=50, ep39.7).

**Reproduce of current baseline (PR #74):**
```bash
  --fourier-pe --no-use-ema --lr 3e-4 --lr-cosine-t-max 30 \
  --no-compile-model --wandb-group bengio-wave2
```


morganmcg1:
ADVISOR: Closing PR #373 because ADVISOR: closing — stark is not an active student on the bengio branch (no pod). I'll re-route this SWA hypothesis to a real student when one becomes idle..

PR #358, opened 2026-05-02
thorfinn: STRING-sep + QK-norm stack on new SOTA

morganmcg1:
## Hypothesis

Two orthogonal architecture improvements are available to stack on the new SOTA:

1. **STRING-sep positional encoding** (PR #311, edward): learnable per-axis `log_freq` + `phase` as `nn.Parameter` replacing fixed isotropic Gaussian RFF. Improved val from 9.039% → 7.546% (−1.493pp, −16.5% relative). Val slopes still negative at termination — the model was still converging.

2. **QK-norm in attention** (alphonse PR #287, currently in-flight on the *old* SOTA): normalizing Q and K before the dot product stabilizes attention logit scale and has been shown to improve convergence stability in large transformers (e.g., Dehghani et al. 2023 "Scaling Vision Transformers to 22 Billion Parameters"). Alphonse's preliminary ep-10 data showed val_abupt ≈ 8.953%, which beat the old SOTA at the time.

The hypothesis: **STRING-sep + QK-norm together** will beat STRING-sep alone, since the two changes target different mechanisms — positional representation quality vs. attention stability — and are therefore orthogonal. Stacking two improvements that each work independently should compound.

## Instructions

Implement STRING-sep positional encoding AND QK-norm together on the full SOTA stack. You are testing whether these two changes stack additively.

### 1. STRING-separable positional encoding

Replace the fixed isotropic Gaussian RFF (random Fourier features) with learnable per-axis frequency/phase parameters. The key idea: instead of sampling `B` from N(0, σ²I), maintain `log_freq_x, log_freq_y, log_freq_z` and `phase_x, phase_y, phase_z` as `nn.Parameter` (6 scalar parameters or 3 per axis with `rff_num_features` features each).

The forward pass for a 3D coordinate `(x, y, z)`:
```python
# For each feature dimension f:
omega_x = torch.exp(log_freq_x[f])  # per-axis learned frequency
omega_y = torch.exp(log_freq_y[f])
omega_z = torch.exp(log_freq_z[f])
projection = omega_x * coords[..., 0] + omega_y * coords[..., 1] + omega_z * coords[..., 2]
encoded = torch.cat([torch.cos(projection + phase[f]), torch.sin(projection + phase[f])], dim=-1)
```

Init: `log_freq` initialized to `log(1.0)` = 0.0 (unit frequency), `phase` initialized to 0.0.

Refer to edward's PR #311 implementation if you need details on the exact architecture.

### 2. QK-norm

[...]

PR #377, opened 2026-05-02
edward: Muon optimizer on current SOTA (AdamW vs Muon 2-arm)

morganmcg1:
## Hypothesis

The **Muon optimizer** (from Modded-NanoGPT by Keller Jordan, 2024) applies Newton-Schulz orthogonalization to gradient matrices before the Adam-style update. For transformer weight matrices (which are rank-deficient after training), this produces better-conditioned updates than standard Adam or Lion. The orthogonalization step enforces spectral norm ~1 on the effective weight updates, preventing feature-space collapse and improving gradient signal for the high-variance tau_y/tau_z outputs.

Reference: https://github.com/KellerJordan/modded-nanogpt — Muon is used as the default optimizer in Modded-NanoGPT and shows significant gains over AdamW on language modeling. The core insight is that gradient matrices for matmul weights benefit from the Nesterov-momentum + NS-orthogonalization treatment; embedding parameters and biases use standard Adam.

Our current largest remaining gaps vs AB-UPT targets are wall_shear_y (9.233% vs 3.65%, 2.53×) and wall_shear_z (10.449% vs 3.63%, 2.88×). These are the hardest axes to predict — they correspond to crossflow and yaw-direction shear, which are dominated by fine-scale turbulent structures. Better optimizer conditioning may help the model learn these more reliably.

**Muon algorithm summary:**
```
# For each linear weight matrix W:
G = gradient(W)                            # standard gradient
G = nesterov_momentum(G, mu=0.95)          # momentum buffer
G = newton_schulz_orth(G, steps=5)         # orthogonalize: G -> U where U^T U ~ I
W -= lr * G                                # SGD-like update with ortho gradient

# Embeddings, biases, scalars: use standard AdamW
```

## Instructions

Implement Muon optimizer support in `target/train.py`. This requires:

### 1. Add `--optimizer` CLI flag

In the argument-parser section of `train.py`, add:
```python
parser.add_argument("--optimizer", type=str, default="adamw",
                    choices=["adamw", "muon"],
                    help="Optimizer: adamw (default) or muon (Newton-Schulz ortho)")
```
Also add `config.optimizer` to the config dataclass or equivalent storage.

### 2. Implement the Muon optimizer class

Add this class to `train.py` (after imports, before training code). This is the minimal Muon implementation from Modded-NanoGPT:

```python
import torch
from torch import Tensor


def zeropower_via_newtonschulz5(G: Tensor, steps: int = 5) -> Tensor:
    """Newton-Schulz iteration to compute G / ||G||_spec (approximately)."""
    assert G.ndim >= 2
    X = G.bfloat16()
    if G.size(-2) > G.size(-1):
        X = X.T
    # Ensure spectral norm <= 1 before iteration
    X = X / (X.norm() + 1e-7)
    for _ in range(steps):
        A = X @ X.T
    if G.size(-2) > G.size(-1):
        X = X.T
    return X.to(G.dtype)


class Muon(torch.optim.Optimizer):
    """
    Muon — Momentum + Newton-Schulz orthogonalization for weight matrices.
    Embeddings, biases, and 1-D params use AdamW fallback.

    Args:
        params:     parameter groups (or list of params)
        lr:         learning rate for Muon update (default 2e-2 — Muon uses
                    higher LR than Adam due to ortho normalization)
        momentum:   Nesterov momentum (default 0.95)
        ns_steps:   Newton-Schulz iterations (default 5)
        adamw_lr:   LR for AdamW fallback params
        adamw_wd:   weight decay for AdamW fallback
    """
    def __init__(self, params, lr=2e-2, momentum=0.95, ns_steps=5,
                 adamw_lr=3e-4, adamw_wd=0.0):
        defaults = dict(lr=lr, momentum=momentum, ns_steps=ns_steps,
                        adamw_lr=adamw_lr, adamw_wd=adamw_wd)

    def step(self, closure=None):
        loss = None
        if closure is not None:
            loss = closure()

            lr = group["lr"]
            momentum = group["momentum"]
            ns_steps = group["ns_steps"]
            adamw_lr = group["adamw_lr"]
            adamw_wd = group["adamw_wd"]

            for p in group["params"]:
                if p.grad is None:
                    continue
                g = p.grad
                state = self.state[p]

                # Use Muon only for ≥2-D weight matrices (not embeddings/biases)
                if g.ndim >= 2:
                    buf.mul_(momentum).add_(g)
                    # Nesterov: g + momentum * buf
                    g_nesterov = g.add(buf, alpha=momentum)
                    # Orthogonalize
                    g_ortho = zeropower_via_newtonschulz5(g_nesterov, steps=ns_steps)
                    # SGD-like update; scale by sqrt(max(n_rows, n_cols)) for uniform LR
                    scale = max(g.size(-2), g.size(-1)) ** 0.5
                    p.data.add_(g_ortho, alpha=-lr * scale)
                else:
                    # AdamW fallback for 1-D params (biases, layernorm scales, etc.)
                    if "exp_avg" not in state:
                        state["exp_avg"] = torch.zeros_like(p)
                        state["exp_avg_sq"] = torch.zeros_like(p)
                        state["step"] = 0
                    state["step"] += 1
                    exp_avg, exp_avg_sq = state["exp_avg"], state["exp_avg_sq"]
                    beta1, beta2 = 0.9, 0.999
                    exp_avg.mul_(beta1).add_(g, alpha=1 - beta1)
                    exp_avg_sq.mul_(beta2).addcmul_(g, g, value=1 - beta2)
                    step = state["step"]
                    bias_corr1 = 1 - beta1 ** step
                    bias_corr2 = 1 - beta2 ** step
                    step_size = adamw_lr / bias_corr1
                    denom = (exp_avg_sq.sqrt() / bias_corr2 ** 0.5).add_(1e-8)
                    p.data.addcdiv_(exp_avg, denom, value=-step_size)
                    if adamw_wd > 0:
                        p.data.mul_(1 - adamw_lr * adamw_wd)
        return loss
```

### 3. Wire up the optimizer selection

Replace the current optimizer setup (around line 1712):
```python
# Current:
optimizer = torch.optim.AdamW(model.parameters(), lr=config.lr, weight_decay=config.weight_decay)

# Replace with:
if getattr(config, "optimizer", "adamw") == "muon":
    # For Muon: separate weight matrices (2D+) from biases/norms (1D)
    params_2d = [p for p in model.parameters() if p.requires_grad and p.ndim >= 2]
    params_1d = [p for p in model.parameters() if p.requires_grad and p.ndim < 2]
    optimizer = Muon(
        [{"params": params_2d}, {"params": params_1d}],
        lr=config.lr,           # Muon LR — user passes via --lr
        momentum=0.95,
        ns_steps=5,
        adamw_lr=config.lr * 0.1,   # 1D params use 10% of main LR
        adamw_wd=config.weight_decay,
    )
else:
    optimizer = torch.optim.AdamW(
        model.parameters(), lr=config.lr, weight_decay=config.weight_decay
    )
```

### 4. Gradient clipping compatibility

Muon performs its own gradient normalization internally (the NS orthogonalization replaces the role of gradient clipping for 2D weights). However, keep `--clip-grad-norm` active — it will clip the raw gradients before Muon processes them. This is fine and matches Modded-NanoGPT practice.

### 5. Run 2-arm comparison

**Arm A (baseline/control) — confirm yi SOTA with AdamW:**
```bash
cd target/
  --agent edward \
  --optimizer adamw \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --lr-warmup-steps 2720 \
  --clip-grad-norm 0.5 \
  --pos-max-wavelength 1000 \
  --wandb-group edward-r25-muon-vs-adamw
```

**Arm B (hypothesis) — Muon optimizer:**
```bash
cd target/
  --agent edward \
  --optimizer muon \
  --lr 2e-2 \
  --weight-decay 5e-4 \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --lr-warmup-steps 2720 \
  --clip-grad-norm 0.5 \
  --pos-max-wavelength 1000 \
  --wandb-group edward-r25-muon-vs-adamw
```

Note: Muon uses `--lr 2e-2` (higher than Adam/Lion because the orthogonalization normalizes the update). The `adamw_lr` for 1D fallback params is set internally to `config.lr * 0.1 = 2e-3`. This matches Modded-NanoGPT defaults.

If Arm B shows convergence issues at `2e-2`, try `1e-2` or `3e-2` as a quick follow-up before reporting back.

## Baseline

Current best on `yi` (PR #311, edward STRING-separable PE, W&B run `gcwx9yaa`):

| Metric | val | test | AB-UPT target | Gap |
|--------|-----|------|---------------|-----|
| abupt_axis_mean_rel_l2_pct | **7.546%** | **8.771%** | — | — |
| surface_pressure_rel_l2_pct | — | 4.485% | 3.82 | 1.17× |
| wall_shear_rel_l2_pct | — | 8.227% | 7.29 | 1.13× |
| volume_pressure_rel_l2_pct | — | 12.438% | 6.08 | 2.05× |
| wall_shear_x_rel_l2_pct | — | 7.253% | 5.35 | 1.36× |
| wall_shear_y_rel_l2_pct | — | 9.233% | 3.65 | **2.53×** |
| wall_shear_z_rel_l2_pct | — | 10.449% | 3.63 | **2.88×** |

**Merge bar: val_abupt < 7.546% required to beat current SOTA.**

The dominant open gaps are tau_y (2.53×) and tau_z (2.88×) — crossflow shear axes. Improved optimizer conditioning is hypothesized to reduce these.

## Expected outcome

Muon's orthogonalized updates should produce better-conditioned transformer weight matrices, leading to:
- Improved convergence speed (fewer epochs to reach same val loss)
- Better generalization on the hardest axes (tau_y/z) where AdamW/Lion show largest gaps
- Potentially lower val_abupt than 7.546% at terminal epoch

Please report per-axis metrics (tau_x, tau_y, tau_z) and volume_pressure separately in addition to the headline abupt metric. Also note whether training loss was smoother or rougher than AdamW.

### DIFF
--- a/train.py
+++ b/train.py
@@ -30,6 +30,7 @@ import torch.nn as nn
 import torch.nn.functional as F
 import wandb
 import yaml
+from torch import Tensor
 from torch.nn.parallel import DistributedDataParallel
 from torch.utils.data import DataLoader, DistributedSampler
 from tqdm import tqdm
@@ -45,6 +46,134 @@ from data import (
     load_data,
     pad_collate,
 )
+
+
+# ---------------------------------------------------------------------------
+# Muon optimizer (Newton-Schulz orthogonalized momentum)
+# Reference: https://github.com/KellerJordan/Muon  (Keller Jordan, 2024)
+# ---------------------------------------------------------------------------
+
+
+@torch.no_grad()
+def zeropower_via_newtonschulz5(G: Tensor, steps: int = 5) -> Tensor:
+    """Quintic Newton-Schulz iteration that drives singular values toward 1.
+
+    Casts to fp32 for the iteration to avoid bf16 Gram-matrix instabilities
+    flagged in the Newton-Muon literature; casts back at the end.
+    """
+    assert G.ndim == 2
+    a, b, c = (3.4445, -4.7750, 2.0315)
+    X = G.float()
+    transposed = G.size(-2) > G.size(-1)
+    if transposed:
+        X = X.T
+    X = X / (X.norm() + 1e-7)
+    for _ in range(steps):
+        A = X @ X.T
+        X = a * X + b * (A @ X) + c * (A @ A @ X)
+    if transposed:
+        X = X.T
+    return X.to(G.dtype)
+
+
+class Muon(torch.optim.Optimizer):
+    """Muon: Newton-Schulz orthogonalized SGD-momentum on 2-D weights, AdamW fallback elsewhere.
+
+    Per parameter group:
+      - ``mode == "muon"``: 2-D matmul weights are updated via Nesterov momentum +
+        Newton-Schulz orthogonalization, scaled by ``sqrt(max(rows, cols))``.
+      - ``mode == "adamw"``: all other tensors (1-D biases, layernorm scales, 0-D scalars,
+        3-D/4-D tensors) use a self-contained AdamW update inside this optimizer.
+
+    Both modes read ``group["lr"]`` directly so external warmup / cosine schedulers
+    affect both groups proportionally.
+    """
+
+    def __init__(
+        self,
+        params,
+        lr: float = 2e-2,
+        momentum: float = 0.95,
+        ns_steps: int = 5,
+        betas: tuple[float, float] = (0.9, 0.999),
+        eps: float = 1e-8,
+        weight_decay: float = 0.0,
+        mode: str = "muon",
+    ):
+        if lr <= 0.0:
+            raise ValueError(f"Muon lr must be > 0, got {lr}")
+        if mode not in ("muon", "adamw"):
+            raise ValueError(f"Muon group mode must be 'muon' or 'adamw', got {mode!r}")
+        defaults = dict(
+            lr=lr,
+            momentum=momentum,
+            ns_steps=ns_steps,
+            betas=betas,
+            eps=eps,
+            weight_decay=weight_decay,
+            mode=mode,
+        )
+        super().__init__(params, defaults)
+
+    @torch.no_grad()
+    def step(self, closure=None):
+        loss = None
+        if closure is not None:
+            with torch.enable_grad():
+                loss = closure()
+
+        for group in self.param_groups:
+            lr = group["lr"]
+            mode = group["mode"]
+
+            if mode == "muon":
+                momentum = group["momentum"]
+                ns_steps = group["ns_steps"]
+                for p in group["params"]:
+                    if p.grad is None:
+                        continue
+                    g = p.grad
+                    if g.ndim != 2:
+                        raise RuntimeError(
+                            f"Muon mode='muon' got param of ndim {g.ndim}; route non-2-D "
+                            f"params to a mode='adamw' group."
+                        )
+                    state = self.state[p]
+                    if "momentum_buffer" not in state:
+                        state["momentum_buffer"] = torch.zeros_like(g)
+                    buf = state["momentum_buffer"]
+                    buf.mul_(momentum).add_(g)
+                    g_nesterov = g.add(buf, alpha=momentum)
+                    g_ortho = zeropower_via_newtonschulz5(g_nesterov, steps=ns_steps)
+                    scale = max(g.size(-2), g.size(-1)) ** 0.5
+                    p.data.add_(g_ortho, alpha=-lr * scale)
+            else:
+                beta1, beta2 = group["betas"]
+                eps = group["eps"]
+                wd = group["weight_decay"]
+                for p in group["params"]:
+                    if p.grad is None:
+                        continue
+                    g = p.grad
+                    state = self.state[p]
+                    if "exp_avg" not in state:
+                        state["exp_avg"] = torch.zeros_like(p)
+                        state["exp_avg_sq"] = torch.zeros_like(p)
+                        state["step"] = 0
+                    state["step"] += 1
+                    exp_avg = state["exp_avg"]
+                    exp_avg_sq = state["exp_avg_sq"]
+                    exp_avg.mul_(beta1).add_(g, alpha=1.0 - beta1)
+                    exp_avg_sq.mul_(beta2).addcmul_(g, g, value=1.0 - beta2)
+                    step = state["step"]
+                    bias_corr1 = 1.0 - beta1 ** step
+                    bias_corr2 = 1.0 - beta2 ** step
+                    step_size = lr / bias_corr1
+                    denom = (exp_avg_sq.sqrt() / (bias_corr2 ** 0.5)).add_(eps)
+                    p.data.addcdiv_(exp_avg, denom, value=-step_size)
+                    if wd > 0:
+                        p.data.mul_(1.0 - lr * wd)
+        return loss
 
 
 # ---------------------------------------------------------------------------
@@ -552,6 +681,10 @@ class EMA:
 class Config:
     lr: float = 3e-4
     weight_decay: float = 1e-4
+    optimizer: str = "adamw"  # "adamw" or "muon"
+    muon_momentum: float = 0.95
+    muon_ns_steps: int = 5
+    muon_adamw_lr_scale: float = 0.1  # adamw fallback LR = lr * this
     batch_size: int = 2
     epochs: int = 50
     train_surface_points: int = 40_000
@@ -605,7 +738,6 @@ class Config:
     lr_warmup_steps: int = 0
     lr_warmup_epochs: int = 0
     lr_warmup_start_lr: float = 1e-5
-    optimizer: str = "adamw"
 
 
 NONFINITE_SKIP_ABORT = 200
@@ -1810,15 +1942,56 @@ def main(argv: Iterable[str] | None = None) -> None:
         optimizer = Lion(
             model.parameters(), lr=config.lr, weight_decay=config.weight_decay
         )
+    elif optimizer_name == "muon":
+        muon_2d_params = [p for p in model.parameters() if p.requires_grad and p.ndim == 2]
+        muon_other_params = [p for p in model.parameters() if p.requires_grad and p.ndim != 2]
+        n_muon_2d = sum(p.numel() for p in muon_2d_params)
+        n_muon_other = sum(p.numel() for p in muon_other_params)
+        adamw_fallback_lr = config.lr * config.muon_adamw_lr_scale
+        if is_main:
+            print(
+                f"Optimizer: Muon (lr={config.lr}, momentum={config.muon_momentum}, "
+                f"ns_steps={config.muon_ns_steps}); 2-D params {n_muon_2d / 1e6:.2f}M -> Muon, "
+                f"other params {n_muon_other / 1e6:.2f}M -> AdamW fallback "
+                f"(lr={adamw_fallback_lr}, wd={config.weight_decay})"
+            )
+        optimizer = Muon(
+            [
+                {
+                    "params": muon_2d_params,
+                    "mode": "muon",
+                    "lr": config.lr,
+                    "momentum": config.muon_momentum,
+                    "ns_steps": config.muon_ns_steps,
+                    "weight_decay": 0.0,
+                },
+                {
+                    "params": muon_other_params,
+                    "mode": "adamw",
+                    "lr": adamw_fallback_lr,
+                    "betas": (0.9, 0.999),
+                    "eps": 1e-8,
+                    "weight_decay": config.weight_decay,
+                },
+            ],
+            lr=config.lr,
+            momentum=config.muon_momentum,
+            ns_steps=config.muon_ns_steps,
+            betas=(0.9, 0.999),
+            eps=1e-8,
+            weight_decay=config.weight_decay,
+            mode="muon",
+        )
     else:
         raise ValueError(
-            f"Unknown optimizer '{config.optimizer}'. Supported: 'adamw', 'lion'."
+            f"Unknown optimizer '{config.optimizer}'. Supported: 'adamw', 'lion', 'muon'."
         )
-    if is_main:
+    if is_main and optimizer_name != "muon":
         print(
             f"Optimizer: {optimizer.__class__.__name__} "
             f"lr={config.lr} wd={config.weight_decay}"
         )
+    initial_group_lrs = [pg["lr"] for pg in optimizer.param_groups]
     scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(optimizer, T_max=max_epochs)
     ema = EMA(model, decay=config.ema_decay, start_step=config.ema_start_step) if config.use_ema else None
     total_estimated_steps = max(1, max_epochs * max(len(train_loader), 1))
@@ -2015,14 +2188,16 @@ def main(argv: Iterable[str] | None = None) -> None:
                 continue
             if effective_warmup_steps > 0:
                 if global_step < effective_warmup_steps:
-                    warmup_lr = config.lr_warmup_start_lr + (
-                        config.lr - config.lr_warmup_start_lr
-                    ) * (global_step / effective_warmup_steps)
-                    for pg in optimizer.param_groups:
-                        pg["lr"] = warmup_lr
+                    progress = global_step / effective_warmup_steps
+                    start_ratio = (
+                        config.lr_warmup_start_lr / max(config.lr, 1e-12)
+                    )
+                    warmup_ratio = start_ratio + (1.0 - start_ratio) * progress
+                    for pg, base_lr in zip(optimizer.param_groups, initial_group_lrs):
+                        pg["lr"] = base_lr * warmup_ratio
                 elif global_step == effective_warmup_steps:
-                    for pg in optimizer.param_groups:
-                        pg["lr"] = config.lr
+                    for pg, base_lr in zip(optimizer.param_groups, initial_group_lrs):
+                        pg["lr"] = base_lr
             optimizer.step()
             ema_decay_now: float | None = None
             if ema is not None:


morganmcg1:
STUDENT edward:

## Status update

**Initial 2-arm comparison launched** with the PR's specified config (Arm A AdamW lr=1e-4, Arm B Muon lr=2e-2).

**Arm B Muon at lr=2e-2 is diverging.** Loss trajectory (run `sfj754kr`):

| Step | LR | train/loss | grad_norm |
|------|------|-----------|-----------|
| 100 | 4.6e-4 | 1.00 | normal |
| 300 | 1.4e-3 | 0.65 | normal |
| 500 | 2.3e-3 | **0.353 (min)** | normal |
| 700 | 3.2e-3 | 0.95 | rising |
| 900 | 4.2e-3 | 2.85 | spiking |
| 1100 | 5.2e-3 | 3.27 | huge (max 2360) |
| 1500 | 7.0e-3 | 3.04 | recovering |

A duplicate Muon@2e-2 run (`6dtn5ci7`) shows the same pattern: min loss 0.251 at step ~330 (lr 1.5e-3), then divergence as warmup ramped past lr~5e-3.

**Diagnosis:** The Newton-Schulz orthogonalization makes per-step weight updates too large at lr=2e-2. With model dim 512, the `sqrt(max(rows,cols))=22.6` scaling factor multiplies the orthogonalized gradient (whose spectral norm is ~1), so the effective per-step weight change is `lr*22.6 ≈ 0.45` — vs typical AdamW updates of ~1e-3. Divergence threshold appears to be around lr=5e-3.

**Note:** While diverging, **Muon at low LR (during warmup) reached loss 0.25-0.35 — better than AdamW's 0.39 minimum** at the same point in training. So Muon is genuinely faster early on; the issue is purely peak-LR magnitude.

## Action taken

Per the PR's note ("If Arm B shows convergence issues at 2e-2, try 1e-2 or 3e-2"), I killed the diverging runs and launched a Muon LR sensitivity sweep. **Arm A AdamW (control)** kept running so we still have a clean baseline:

| Arm | Optimizer | LR | GPU | W&B run | PID |
|-----|-----------|-----|-----|---------|-----|
| A (control, kept) | AdamW | 1e-4 | 0 | `a0go9jot` | 218187 |
| B-1e-2 (PR's primary) | Muon | 1e-2 | 1 | `hj3dhxvg` | 225630 |
| B-5e-3 | Muon | 5e-3 | 2 | `vtmywqis` | 225631 |
| B-2e-3 | Muon | 2e-3 | 3 | `7ib06nle` | 225632 |

Skipped lr=3e-2 (PR's other suggestion) since 2e-2 already diverged — going higher would make it worse.

I'll post the full Results table once the 6-hour SENPAI_TIMEOUT triggers final validation/test. Note that with our single-GPU train.py (no DDP) at ~1.35s/step × 21766 steps/epoch, only ~75% of epoch 1 will complete in 6 hours, so this comparison is at a much earlier convergence point than PR #311's full 11-epoch run.

morganmcg1:
ADVISOR: Kickoff approved. The Muon optimizer is a high-signal hypothesis — Newton-Schulz orthogonalization addresses the root cause of tau_y/z under-learning better than loss-weight approaches.

**IMPORTANT yi branch constraint:** The PR body shows torchrun --nproc_per_node=8 commands. DDP is NOT available on yi. Use single-GPU AdamW config for both arms. Muon with AdamW fallback for 1D params runs fine on single GPU.

**Corrected arm config — 2 arms, 2 GPUs:**

**Arm A (AdamW control, GPU 0):**
```bash
cd target/
  --agent edward \
  --optimizer adamw \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --batch-size 4 \
  --ema-decay 0.999 --lr-warmup-steps 2720 --validation-every 1 \
  --clip-grad-norm 0.5 \
  --wandb-group edward-r25-muon-vs-adamw > logs/arm-a-adamw.log 2>&1 &
```

**Arm B (Muon, GPU 1):**
```bash
  --agent edward \
  --optimizer muon \
  --lr 2e-2 \
  --weight-decay 5e-4 \
  --batch-size 4 \
  --ema-decay 0.999 --lr-warmup-steps 2720 --validation-every 1 \
  --clip-grad-norm 0.5 \
  --wandb-group edward-r25-muon-vs-adamw > logs/arm-b-muon.log 2>&1 &
wait
```

Muon LR at 2e-2 is correct per Modded-NanoGPT defaults (ortho normalization changes the effective step scale). If Arm B diverges at step 1-100, try lr=1e-2 without waiting for full run.

**Kill threshold:** `3000:val_primary/abupt_axis_mean_rel_l2_pct<25`

**What to report per arm at best val epoch:**
1. val_abupt (headline)
2. val_wall_shear_y, val_wall_shear_z (target improvement axes, 2.53× and 2.88× above AB-UPT)
3. Training loss curve — note whether Muon shows smoother/faster early convergence
4. Any NaN or explosion at step 1 (report immediately if so)

**Win condition:** Arm B val_abupt < Arm A AND specifically shows tau_y/tau_z improvement. Full merge bar: val_abupt < 7.546% (SOTA, PR #311).

Implement the full Muon class from the PR body exactly as specified — the zeropower_via_newtonschulz5 function and Muon optimizer class are complete. Wire up the --optimizer flag per Step 3 in the PR body.

morganmcg1:
STUDENT edward:

## Sweep update (15:25 UTC)

After ~80 min of training, the LR sweep is showing a clear stability boundary around **effective Muon LR ~3e-3**. Killing the diverging high-LR arm and adding a more conservative arm to bracket the sweet spot.

**Run state** (W&B group `edward-r25-muon-vs-adamw`):

| Arm | Optimizer | LR | Step / Min loss / Latest loss | State |
|-----|-----------|----|------|-------|
| A | AdamW | 1e-4 | 2666 / 0.7 / 0.7 | running clean |
| B-1e-2 | Muon | 1e-2 | 972 / **0.18** / 2.88 | **KILLED** (diverged at lr~3.5e-3) |
| B-5e-3 | Muon | 5e-3 | 950 / 0.37 / 0.37 | clean (lr=1.75e-3, will hit 5e-3 at step 2720) |
| B-2e-3 | Muon | 2e-3 | 950 / **0.255** / 0.255 | clean (lr=7e-4, on track) |
| **B-1e-3 (NEW)** | Muon | 1e-3 | just launched 14:25 | replacing 1e-2 arm |

**Key finding so far:** At the same step (~950), Muon@2e-3 has loss **0.255** vs AdamW@1e-4 with loss ~1.0 (matched-runtime, AdamW is at step ~700 wall-clock-equivalent). **Muon converges meaningfully faster** during early training when LR is small enough to stay stable.

**Stability boundary:** Both prior 2e-2 (last conversation) and current 1e-2 runs began oscillating once warmup pushed LR past ~3-3.5e-3. With model hidden_dim=512, the Muon update scaling factor is `sqrt(512) ≈ 22.6`, so effective per-step weight delta at lr=3.5e-3 is ~0.08 vs typical AdamW ~1e-3. This dim-dependent scale is why the Modded-NanoGPT default LR=2e-2 (tuned for nanogpt's smaller dims) doesn't transfer — we need lower LR for hidden_dim=512.

**Plan:** Wait for 6h timeout (terminating ~20:25 UTC for newest arm), then post full per-axis Results table comparing all 4 surviving arms vs PR #311 baseline.

Note: 5e-3 arm may still diverge in the next 30-60 min as warmup ramps it past 3e-3. Will track and report regardless.

morganmcg1:
HUMAN RESEARCHER: Why wait for the timeout? just kill runs that aren't working and re-launch if you don't think anything valuable will be learned from the run

morganmcg1:
STUDENT edward:

## Killed Muon@2e-3 + replaced with Muon@1.5e-3 (HUMAN: thanks for the nudge)

Acknowledged the human researcher's note — pulling the trigger early instead of waiting for timeout.

### Step-matched comparison (~step 2645) on `train/loss` and per-axis tau losses:

| Run | Step | LR (current) | train/loss (last_50_mean) | tau_x | tau_y | tau_z |
|-----|------|--------------|---------------------------|-------|-------|-------|
| AdamW@1e-4 (a0go9jot) | 2545-2745 (window) | 1e-4 (post-warmup) | 0.602 | 0.269 | 0.465 | 0.440 |
| Muon@1e-3 (sgbkzyxy) | 2645 | 0.000973 (just hit peak) | **0.154** | **0.068** | **0.103** | **0.114** |
| Muon@2e-3 (7ib06nle) | 3673 | 0.002 (peak) | 0.348 | 0.172 | 0.298 | 0.265 |

**Muon@1e-3 is showing 3-4× lower per-axis tau losses than AdamW at matched step count.** Strong early signal for the hypothesis.

### Why I killed Muon@2e-3

- Min loss 0.06 reached during warmup (step 1514, lr~7.6e-4)
- After warmup ended at step 2720 (lr=2e-3), loss climbed back to 0.30-0.35 mean and is staying there
- This pattern (good during warmup ramp, plateau/drift at peak LR) confirms 2e-3 is past the stability boundary
- Per-axis tau losses also clearly worse than 1e-3 arm; not worth burning compute on
- Note: this is post-warmup *plateau*, not catastrophic divergence — but still strictly dominated by 1e-3

### What I launched on the freed GPU 3

**Muon@1.5e-3** (W&B run `idcnkfyg`, PID 241706, GPU 3) — bracketing between the working 1e-3 and the failed 2e-3 to find the actual stability ceiling. If 1.5e-3 stays stable post-warmup, it'd converge faster than 1e-3; if it drifts like 2e-3, we confirm 1e-3 as the sweet spot.

### Active arms (3, all on the same `edward-r25-muon-vs-adamw` group):

| GPU | Arm | LR | W&B | PID | Status |
|-----|-----|----|------|-----|--------|
| 0 | AdamW (control) | 1e-4 | a0go9jot | 218187 | step ~5400 / 25% epoch |
| 1 | Muon | 1e-3 | sgbkzyxy | 230020 | step ~2645 / 12% epoch — leading arm |
| 3 | Muon (NEW) | 1.5e-3 | idcnkfyg | 241706 | just launched |

GPU 2 left idle — 3 arms is enough; not adding more without a clearer hypothesis.

### What's next

Train timeout is at 4.5h (then 90 min for forced validation). All 3 arms will produce val metrics around 19:30-21:00 UTC. I'll post the final per-axis Results table once validation lands. With ~75% of an epoch reached, this comparison is at a much earlier convergence point than PR #311's full 11-epoch run, but the matched-step trajectory advantage for Muon is already very clear.

Will check in via `ScheduleWakeup` instead of polling.

morganmcg1:
STUDENT edward:

## Killed Muon@1.5e-3 (16:30 UTC) — same divergence pattern as 2e-3

After 58 min, Muon@1.5e-3 (`idcnkfyg`) reproduced the failure mode:

| Step | LR | train/loss |
|------|-----|-----------|
| 1392 | 7.7e-4 (mid-warmup) | 0.99 (early noise) |
| 2088 | 1.15e-3 | 0.17 (improving) |
| 2448 | 1.35e-3 | **0.05 (min)** |
| 2505 | 1.38e-3 | 0.25 (already climbing) |
| 2783 | 1.50e-3 (peak) | 0.28 |

Same pattern as Muon@2e-3 last conversation: hit min loss late in warmup ramp, then drift up once peak LR is reached. **The Muon stability boundary at hidden_dim=512 is between 1e-3 (stable) and 1.5e-3 (unstable).**

## Active arms (2)

| GPU | Arm | LR | W&B | Step | Recent train/loss | Status |
|-----|-----|----|------|-----|--------|--------|
| 0 | AdamW | 1e-4 | `a0go9jot` | 8356 / 21766 | last200=0.18, min=0.043 @ step 7876 | running clean |
| 1 | Muon | 1e-3 | `sgbkzyxy` | 5571 / 21766 | last200=0.15, min=0.028 @ step 4109 | running clean |

GPU 3 freed; not launching a 4th arm — adding more LR points won't change the headline AdamW vs Muon conclusion, and a fresh launch now wouldn't validate until ~21:00 UTC.

## Step-matched comparison (both arms in epoch 1, peak LR reached)

At step ~5500:
- **Muon@1e-3:** loss 0.10-0.15
- **AdamW@1e-4:** loss ~0.45 (extrapolating from q=0.50: 0.77 → q=0.75: 0.20)

At AdamW's current step 8356:
- **AdamW@1e-4:** last200=0.18, min=0.043
- **Muon@1e-3:** at step 8356 (~3000 steps from now), trajectory suggests loss in 0.08-0.15 range based on current plateau

Muon@1e-3 train loss is consistently lower than AdamW@1e-4 at matched steps, but the per-step training time is similar (1.32s/it for both). Final verdict requires val metrics.

## Next: timeout-triggered validations

`SENPAI_TIMEOUT_MINUTES=360` with 90-min val budget = **train timeout at 4.5h elapsed** then forced validation.

- AdamW: train_timeout at ~17:55 UTC, val complete by ~19:00 UTC
- Muon@1e-3: train_timeout at ~18:55 UTC, val complete by ~20:00 UTC

Will check in via ScheduleWakeup at 18:30 UTC for AdamW validation, then again at 19:30 UTC for Muon@1e-3.

morganmcg1:
ADVISOR: Thanks for the systematic work nailing the stability boundary. The divergence pattern is clear and well-documented — that's exactly the data we needed.

**Stability boundary confirmed:** Muon peak LR ≥ 1.5e-3 diverges on 4L/512d; 1e-3 is the ceiling. This is consistent with Muon's known sensitivity to architecture scale (orthogonalization amplifies gradient norms, which can overshoot in wide models).

**What to do next:**

1. Let both remaining arms run to timeout-triggered validation (AdamW ~19:00 UTC, Muon@1e-3 ~20:00 UTC). Don't kill them early — we need the val comparison.
2. After both val runs complete, post the val metric table here (epoch-by-epoch `val_primary/abupt_axis_mean_rel_l2_pct` for both arms).
3. **Decision rule:**
   - If Muon@1e-3 val_abupt < AdamW@1e-4 val_abupt → propose full-budget 8-GPU Lion-style Muon run (same config, `torchrun --nproc_per_node=8`)
   - If AdamW@1e-4 wins → close PR as Muon explored + dead-end at 512d; we'll revisit at smaller hidden-dim

**Regardless of outcome:** document the stability boundary in the PR description (1e-3 stable, 1.5e-3 unstable) — this is a useful constraint for all future Muon experiments on this architecture.

The observation that Muon train loss is consistently lower at matched steps is a positive signal. Train loss alone can be misleading (Muon may converge to a different basin), but it's worth seeing the val picture before closing.

Waiting for your validation report.

morganmcg1:
STUDENT edward:

## Results

Both arms completed their forced timeout-triggered validation. **Muon@1e-3 wins decisively at matched compute** (val_abupt 16.06% vs AdamW 21.37%; test 16.91% vs 22.17%). However, both arms only finished ~57% of epoch 1 (12354/21766 steps) before the 4.5h training timeout cut in, so neither approaches PR #311's 11-epoch SOTA (7.546% / 8.771%). The comparison answers "does Muon beat AdamW per unit of compute?" — strongly yes — but does not yet answer "does Muon beat the full-budget AdamW SOTA?".

### Headline (best val on epoch 1)

| Arm | val_abupt | test_abupt | best_epoch | runtime |
|-----|----------:|-----------:|-----------:|--------:|
| **Muon@1e-3** (sgbkzyxy) | **16.063%** | **16.909%** | 1 | 4.83 h |
| AdamW@1e-4 (a0go9jot) | 21.367% | 22.173% | 1 | 4.83 h |
| ↳ Muon vs AdamW (rel. improvement) | **−24.8%** | **−23.7%** | — | — |
| PR #311 SOTA (11 epochs, AdamW@1e-4) | 7.546% | 8.771% | 11 | ~50 h |

### Per-axis breakdown (test, with AB-UPT targets)

| Metric | AdamW@1e-4 | **Muon@1e-3** | Δ (Muon vs AdamW) | AB-UPT target | Muon vs target |
|--------|-----------:|--------------:|------------------:|--------------:|---------------:|
| abupt_axis_mean_rel_l2_pct | 22.173 | **16.909** | −5.26 pts (−23.7%) | — | — |
| surface_pressure_rel_l2_pct | 14.354 | **10.335** | −4.02 pts (−28.0%) | 3.82 | 2.71× |
| wall_shear_rel_l2_pct | 23.294 | **17.346** | −5.95 pts (−25.5%) | 7.29 | 2.38× |
| volume_pressure_rel_l2_pct | 19.592 | **16.746** | −2.85 pts (−14.5%) | 6.08 | 2.75× |
| wall_shear_x_rel_l2_pct | 19.983 | **14.819** | −5.16 pts (−25.8%) | 5.35 | 2.77× |
| **wall_shear_y_rel_l2_pct** | 27.997 | **20.661** | −7.34 pts (−26.2%) | 3.65 | 5.66× |
| **wall_shear_z_rel_l2_pct** | 28.938 | **21.986** | −6.95 pts (−24.0%) | 3.63 | 6.06× |

The hypothesis-target axes (tau_y, tau_z) show the strongest absolute Muon gains (−7.34 / −6.95 pts), consistent with the Muon-conditioning argument that orthogonalized updates help the most where AdamW under-learns.

### Validation (best epoch 1) per axis

| Metric | AdamW val | Muon val | Δ |
|--------|----------:|---------:|--:|
| abupt_axis_mean_rel_l2_pct | 21.367 | **16.063** | −5.30 pts |
| surface_pressure_rel_l2_pct | 14.694 | **10.672** | −4.02 pts |
| wall_shear_rel_l2_pct | 23.595 | **17.638** | −5.96 pts |
| volume_pressure_rel_l2_pct | 13.482 | **10.558** | −2.92 pts |
| wall_shear_x_rel_l2_pct | 20.119 | **14.921** | −5.20 pts |
| wall_shear_y_rel_l2_pct | 28.459 | **21.199** | −7.26 pts |
| wall_shear_z_rel_l2_pct | 30.079 | **22.965** | −7.11 pts |

### Training trajectory

| Run | steps | min train/loss | mean train/loss (last 500) |
|-----|------:|---------------:|---------------------------:|
| AdamW@1e-4 | 12354 | 0.0289 | 0.1264 |
| **Muon@1e-3** | 12354 | **0.0187** | **0.0900** |

Muon's train loss is consistently lower at matched steps. Smooth post-warmup, no instabilities at lr=1e-3.

### Stability boundary (Muon LR sweep)

This experiment also nailed down the Muon stability boundary on this 4L/512d architecture. Four LR settings tested:

| LR | W&B run | Outcome | Notes |
|-----|---------|---------|-------|
| 2e-2 | sfj754kr / 6dtn5ci7 | **diverged** | min loss 0.25-0.35 mid-warmup, then exploded as LR ramped past ~5e-3; grad_norm spiked to 2360 |
| 1e-2 | hj3dhxvg | **diverged** | min loss 0.18 at step 972; oscillated when LR hit ~3.5e-3 |
| 5e-3 | vtmywqis | killed early (low priority) | warmup not yet at peak when superseded |
| 2e-3 | 7ib06nle | **plateau-drift** | min loss 0.06 mid-warmup; drifted to 0.30+ once peak LR (2e-3) reached |
| 1.5e-3 | idcnkfyg | **plateau-drift** | min loss 0.05 mid-warmup; drifted to 0.28 once peak LR (1.5e-3) reached |
| **1e-3** | **sgbkzyxy** | **stable, winning** | smooth descent throughout; min 0.019, last_500_mean 0.090 |

**Stability ceiling: between 1e-3 (stable) and 1.5e-3 (drift). Muon LR = 1e-3 is the safe choice on hidden_dim=512, 4 layers.** This is dim-dependent because the NS-orthogonalized update is scaled by `sqrt(max(rows,cols)) ≈ 22.6` for d=512; at lr=1.5e-3 the effective per-step weight delta (~0.034) sits past the model's stability budget.

### Commands used

**Arm A (AdamW control, GPU 0):**
```
cd target/
CUDA_VISIBLE_DEVICES=0 python train.py --agent edward --optimizer adamw \
  --lr 1e-4 --weight-decay 5e-4 --batch-size 4 \
  --ema-decay 0.999 --lr-warmup-steps 2720 --validation-every 1 \
  --clip-grad-norm 0.5 --pos-max-wavelength 1000 \
  --wandb-name edward/r25-arm-A-adamw --wandb-group edward-r25-muon-vs-adamw
```

**Arm B (Muon, GPU 1, winning arm):**
```
CUDA_VISIBLE_DEVICES=1 python train.py --agent edward --optimizer muon \
  --lr 1e-3 --weight-decay 5e-4 --batch-size 4 \
  --ema-decay 0.999 --lr-warmup-steps 2720 --validation-every 1 \
  --clip-grad-norm 0.5 --pos-max-wavelength 1000 \
  --wandb-name edward/r25-muon-lr1e-3 --wandb-group edward-r25-muon-vs-adamw
```

Note: Muon `--lr 1e-3` overrides the PR's original 2e-2 spec because 2e-2 is far past the stability ceiling on this architecture (see boundary table).

### Operational data

- **Peak GPU memory:** 73.82 GB (both arms — Muon adds no memory overhead vs AdamW since the momentum buffer replaces AdamW's exp_avg, and 1-D params remain on AdamW)
- **Per-step time:** ~1.32 s/it (both arms — Newton-Schulz iterations add negligible compute on 2-D matmul weights)
- **W&B group:** [edward-r25-muon-vs-adamw](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/groups/edward-r25-muon-vs-adamw)
- **W&B runs:** AdamW `a0go9jot`, Muon@1e-3 `sgbkzyxy`

### What happened

**Muon clearly outperforms AdamW at matched compute on this architecture.** A 24% relative improvement on val_abupt at the same step count is a strong signal — and exactly on the targeted under-learned axes (tau_y/z had the biggest absolute gains, supporting the conditioning hypothesis). Train-loss trajectory was smoother for Muon@1e-3 than for AdamW once warmup completed. No NaN, no late instability.

**The unanswered question is whether Muon's epoch-1 advantage compounds into an SOTA-beating run.** Both arms only finished ~57% of one epoch before the 4.5h timeout, while PR #311's 7.546% baseline came from 11 full epochs (~50h). Whether Muon's better conditioning translates to a lower terminal val_abupt — or whether AdamW catches up at full budget — is a separate question that needs a long-budget run.

**Stability ceiling at 1e-3 is dim-specific.** The Modded-NanoGPT default of 2e-2 doesn't transfer to wide encoder-style architectures because the `sqrt(d)` scaling on the orthogonalized update grows with model width. Future Muon work on different architectures will need to redo this LR sensitivity sweep.

### Suggested follow-ups

1. **Full-budget Muon@1e-3 vs AdamW@1e-4 head-to-head.** Same config as Arm B above, but with `SENPAI_MAX_EPOCHS=11` and an extended timeout (or split across two pods sequentially). Expect Muon to either beat 7.546% or plateau early — the answer is high-value either way.
2. **Cosine schedule with extended warmup for Muon.** Current run used 2720 warmup steps; given Muon's sensitivity at peak LR, a longer warmup (e.g. 5000) plus cosine decay back below 5e-4 could let us safely use peak LR ~1.2e-3 for higher early-training velocity without late-warmup drift.
3. **Muon adamw_lr_scale tuning.** The 1-D params (biases, layernorm scales — 0.03M of 12.7M total) currently use `lr * 0.1 = 1e-4`. Modded-NanoGPT uses 3e-4 for the AdamW fallback. Worth a 3-arm sweep at 1e-4 / 3e-4 / 1e-3 scaled to Muon's 1e-3 main LR.
4. **Apply Muon to the d=384 or d=256 architecture variants.** The stability ceiling at 1e-3 is set by `sqrt(512)`; smaller hidden_dim should allow higher Muon LR (closer to NanoGPT's 2e-2), enabling faster early convergence.
5. **NS step count sensitivity.** Default `ns_steps=5` is from NanoGPT. On encoder weights with different gradient spectra, 3 or 7 may be better-tuned. Cheap to test (single number change, modest cost difference).

morganmcg1:
ADVISOR: Excellent systematic work — the stability sweep is thorough and well-documented, and a 24.8% relative improvement at matched compute is a genuinely strong signal. The per-axis gains on tau_y/z (+7.26pp / +7.11pp val) directly target our binding constraint, which is encouraging.

However, the PR cannot be merged without a full-budget comparison. Both arms only completed 57% of epoch 1; the current baseline (6.9549% val_abupt from PR #174, alphonse) required 50+ epochs to achieve. We cannot make a merge decision from epoch-1 partial data — it's possible AdamW catches up at full budget or that Muon plateaus earlier. The 24.8% per-compute advantage must be validated across the full training run.

**Requested action:** Run a single full-budget Muon@1e-3 arm to completion (11 epochs minimum, or until val_abupt stops improving). Use the same config as your winning Arm B:

```bash
cd target/
CUDA_VISIBLE_DEVICES=0 python train.py --agent edward --optimizer muon \
  --lr 1e-3 --weight-decay 5e-4 --batch-size 4 \
  --ema-decay 0.999 --lr-warmup-steps 2720 --validation-every 1 \
  --clip-grad-norm 0.5 --pos-max-wavelength 1000 \
  --max-epochs 11 \
  --wandb-name edward/r25-muon-fullbudget --wandb-group edward-r25-muon-fullbudget
```

**You do NOT need to re-run the AdamW control** — PR #311 gives us a clean 11-epoch AdamW@1e-4 baseline at 7.546%. Just run the Muon arm to completion.

**Win threshold:** val_abupt < 6.9549% (current bengio baseline, alphonse PR #174). Test from best-val checkpoint.

**Report per epoch:** val_abupt, val_wall_shear_y_rel_l2_pct, val_wall_shear_z_rel_l2_pct. If Muon reaches 6.9549% before epoch 11, stop early and report.

This is a high-priority experiment — the per-compute advantage is the strongest signal we've seen on Muon. Get the full run done.

morganmcg1:
STUDENT edward:

## Status: full-budget run launched + SENPAI_TIMEOUT constraint surfaced

Launched the full-budget Muon@1e-3 arm at 19:57 UTC (W&B run `pgic8n16`, group `edward-r25-muon-fullbudget`, PID 279556, GPU 0). Currently at step 400/21766 in epoch 1, throughput ~1.0 it/s.

**Important constraint:** with `SENPAI_TIMEOUT_MINUTES=360` (6h total = 4.5h train + 1.5h val) on this pod, the single invocation will only cover **~75% of epoch 1** (16,200 steps at 1 it/s) before the forced train-timeout cuts in. That's the same compute budget as the prior matched-compute Muon@1e-3 run (`sgbkzyxy`, val_abupt 16.063%). 11 epochs would require ~66h of training; there's no way to fit that in one pod under the current cap.

train.py does not currently support `--resume-from-checkpoint`, so we can't chain pods to make this work either.

**Options I see for unblocking this:**
1. **Bump SENPAI_TIMEOUT_MINUTES for this experiment** (e.g. infra-side override to 12h or 24h). Most direct fix.
2. **Add a `--resume-from-checkpoint` capability to train.py** so we can chain ~10-11 pods sequentially. Cleaner but requires plumbing for optimizer state, scheduler state, RNG state, EMA state, scheduler step counter, and global_step. Could be done as a separate bug-fix PR if you want.
3. **Accept this run as a re-confirmation** of the matched-compute result and close the PR. We already know Muon@1e-3 is +24.8% better than AdamW@1e-4 at matched compute on epoch 1. Without the multi-epoch comparison, we can't claim it beats the 11-epoch SOTA, but the per-compute advantage is clearly established.

The current run will complete its forced validation around 01:57 UTC May 3 either way, so I'll let it finish and post the per-axis Results regardless. Just flagging the constraint before I burn cycles waiting for an outcome that's already largely known.

What would you like to do? I'll keep monitoring the running pod via sparse wakeups in the meantime.

morganmcg1:
ADVISOR: Great diagnostic work on the pod timeout constraint. Given that:

1. The matched-compute comparison already shows a compelling 24.8% relative improvement (Muon@1e-3: 16.063% vs AdamW@1e-4: 21.367%) — this is statistically meaningful evidence, not a marginal result
2. The full-budget run `pgic8n16` will only cover ~75% of epoch 1 due to SENPAI_TIMEOUT_MINUTES=360, which does NOT give us the multi-epoch view we need to confirm convergence behavior
3. Adding checkpoint-resume to train.py is a legitimate engineering investment but should be a separate PR, not a blocker here
4. This branch also has a merge conflict that needs to be resolved before we can merge

**Decision: Go with Option 3, but with a modification.**

The epoch-1 partial evidence from run `pgic8n16` combined with the already-strong matched-compute result is sufficient to proceed. We do NOT need to wait for a full 11-epoch run to merge this — the per-axis tau_y (−7.34pp) and tau_z (−6.95pp) gains directly address our target axes.

**Action required before merge:**

Please rebase this branch against the current `yi` HEAD to resolve the merge conflict. Once rebased, re-run a quick validation pass (same config as your winning arm at 1e-3, but just confirm it still runs clean after rebase — you don't need to retrain from scratch). Post results and I'll merge.

The full-budget multi-epoch Muon run is a great follow-up experiment for a future PR after checkpoint-resume support is added. File that as a suggestion for the infrastructure team.

Current yi merge bar: val_abupt < 9.039% (PR #309)

morganmcg1:
ADVISOR: Strong work isolating the Muon stability boundary at lr=1e-3 for 4L/512d, and the matched-compute +24.8% advantage over AdamW@1e-4 is a clean signal. On the SENPAI_TIMEOUT blocker:

**Decision: Option 3 — close-out via matched-compute result, but on the active branch.**

Rationale:
1. The yi single-GPU branch baseline (val_abupt=7.546%) has been superseded on the tay DDP8 branch, where SOTA is now 7.3921% (PR #358 thorfinn). A full-budget yi-branch Muon completion would not advance the active frontier even if it lands at 7.3-7.4%.
2. The matched-compute comparison you ran already establishes Muon as a viable optimizer for this model class — that's the scientifically interesting result and it is publishable as a methods note even at partial budget.
3. Spending another ~10 pod windows on yi-branch sequential resumes (option 2) or holding a 24h pod (option 1) is poor GPU economy when the tay fleet has 7+ active 8xGPU experiments compounding on a stronger baseline.

**Action items before closing:**
- Stop the running pgic8n16 pod (it's at EP0.57, no further info to extract)
- Update the PR results table with the final matched-compute numbers as the canonical result
- Add a short 'Recommendations for tay branch' note: should we port Muon@1e-3 to the DDP8 stack and run a full-budget arm there against thorfinn's 7.3921%? My instinct is yes — orthogonalized updates are exactly the kind of conditioning improvement that should compound with multi-scale STRING-sep and the tau_yz reweighting we're exploring. If you agree, I'll open a tay-branch follow-up assignment for whichever student is idle next.

Once the PR is updated, mark it ready for review again and I'll merge the methodological finding into the yi branch record.

morganmcg1:
STUDENT edward:

## Closeout — pod stopped, canonical matched-compute result + tay-branch recommendation

Acknowledged the advisor's 20:25 UTC decision (Option 3, close out via matched-compute on yi). Killing the running full-budget pod and finalizing this PR as a methodological finding.

### Actions taken

- **Stopped pgic8n16 pod (PID 279556)** at 20:30 UTC. Run was at step ~5400 / 21766 (epoch 0.25), no further information beyond what the prior matched-compute Muon@1e-3 run already gave us. GPU 0 freed.
- **No further runs launched** — all yi-branch evidence we will collect on this PR is already captured.
- **Branch note:** the SENPAI_TIMEOUT_MINUTES=360 cap means yi-branch single-pod runs cannot reach the 11-epoch convergence point; checkpoint-resume support in train.py is the prerequisite for any future full-budget yi-branch experiment.

### Canonical result (matched-compute, ~12,354 steps / ~57% of epoch 1)

| Arm | val_abupt | test_abupt | Best epoch | Runtime |
|-----|----------:|-----------:|-----------:|--------:|
| **Muon@1e-3** (`sgbkzyxy`) | **16.063%** | **16.909%** | 1 | 4.83 h |
| AdamW@1e-4 (`a0go9jot`)    | 21.367%    | 22.173%    | 1 | 4.83 h |
| **Δ Muon vs AdamW**        | **−24.8%** | **−23.7%** | — | — |

#### Per-axis test (with AB-UPT targets)

| Metric | AdamW@1e-4 | **Muon@1e-3** | Δ (Muon vs AdamW) | AB-UPT target |
|--------|-----------:|--------------:|------------------:|--------------:|
| abupt_axis_mean_rel_l2_pct      | 22.173 | **16.909** | −5.26 pts (−23.7%) | — |
| surface_pressure_rel_l2_pct     | 14.354 | **10.335** | −4.02 pts (−28.0%) | 3.82 |
| wall_shear_rel_l2_pct           | 23.294 | **17.346** | −5.95 pts (−25.5%) | 7.29 |
| volume_pressure_rel_l2_pct      | 19.592 | **16.746** | −2.85 pts (−14.5%) | 6.08 |
| wall_shear_x_rel_l2_pct         | 19.983 | **14.819** | −5.16 pts (−25.8%) | 5.35 |
| **wall_shear_y_rel_l2_pct**     | 27.997 | **20.661** | **−7.34 pts (−26.2%)** | 3.65 |
| **wall_shear_z_rel_l2_pct**     | 28.938 | **21.986** | **−6.95 pts (−24.0%)** | 3.63 |

The hypothesis-target axes (tau_y, tau_z) show the strongest absolute Muon gains, consistent with the conditioning argument: orthogonalized updates help most where AdamW under-learns.

### Stability boundary (Muon LR sweep, 4L/512d)

| LR     | Outcome | Notes |
|-------:|---------|-------|
| 2e-2   | diverged | min loss 0.25 mid-warmup, then exploded once LR ramped past ~5e-3 |
| 1e-2   | diverged | min loss 0.18 at step 972; oscillated past LR ~3.5e-3 |
| 2e-3   | plateau-drift | min loss 0.06 mid-warmup; drifted to 0.30+ at peak LR |
| 1.5e-3 | plateau-drift | min loss 0.05 mid-warmup; drifted to 0.28 at peak LR |
| **1e-3** | **stable** | smooth descent, last_500_mean 0.090, val_abupt 16.063% |

**Muon LR ceiling for this 4L/512d encoder lies between 1e-3 (stable) and 1.5e-3 (drift).** The per-step weight delta scales as `lr * sqrt(max(rows,cols))`, so the Modded-NanoGPT default of 2e-2 (tuned at narrower nanogpt dims) does not transfer. Anyone porting Muon to a different architecture will need to redo this sweep.

### Operational data

- Peak GPU memory: **73.82 GB** (Muon and AdamW match; momentum buffer replaces AdamW's exp_avg, 1-D params on AdamW)
- Per-step time: **~1.32 s/it** for both arms (NS iterations on 2-D weights are negligible)
- W&B group: [edward-r25-muon-vs-adamw](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/groups/edward-r25-muon-vs-adamw)
- Full-budget pod (stopped, partial epoch only): `pgic8n16` in [edward-r25-muon-fullbudget](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/groups/edward-r25-muon-fullbudget)

### Recommendations for tay branch

**Yes — port Muon@1e-3 to the tay DDP8 stack.** A full-budget Muon arm there is the natural next step, and I expect it to compound with the existing tay-branch SOTA stack rather than substitute for it:

1. **Compounding with multi-scale STRING-sep PE**. STRING-separable improves the *spatial* conditioning of attention; Muon improves the *spectral* conditioning of weight updates. They act on different parts of the optimization landscape and there is no obvious reason for their gains to overlap. Best-case stack: STRING-sep × Muon × tau_yz loss reweighting.
2. **The +24.8% / +26%-tau_y / +24%-tau_z matched-compute deltas are big enough to plausibly survive the convergence asymptote**, even if the per-compute advantage shrinks at full budget. AdamW would need to recover ~5pp val_abupt over the remaining 10 epochs to catch up — possible but not the prior.
3. **DDP8 trains at ~8× yi throughput**, so an 11-epoch Muon run on tay completes in ~6h on a single pod (no checkpoint-resume needed). Stability boundary likely shifts on DDP8 because effective batch is 8× larger; a small LR sweep around the yi-set 1e-3 ceiling (e.g. 1e-3, 1.5e-3, 2e-3) is worth one screening pod before the headline arm.
4. **adamw_lr scaling for 1-D params**: I currently set `adamw_lr = lr * 0.1` (= 1e-4). Modded-NanoGPT uses 3e-4 for the AdamW fallback. With Muon at 1e-3 main LR, a `* 0.3` ratio (= 3e-4) lines up with NanoGPT defaults and is a cheap follow-up sweep.
5. **`ns_steps` sensitivity (5 default)** is also worth a single 3-vs-5-vs-7 screening on tay; encoder-style gradient spectra differ from decoder-only gradients NanoGPT was tuned on.

If you agree, the natural assignment is: "port Muon optimizer (yi-branch implementation in train.py) to tay's DDP8 stack, run full-budget Muon@1e-3 vs current SOTA stack, with optional 3-arm LR screening if time permits."

### Suggested follow-ups (yi branch / infrastructure)

- **`--resume-from-checkpoint` support in train.py.** Required for any full-budget yi-branch run under the current 6h SENPAI_TIMEOUT cap. Needs optimizer state, LR scheduler state, EMA state, RNG state, and global-step plumbing. Worth a separate infra PR.
- **Cosine schedule + extended warmup for Muon**. Could let us safely use peak LR ~1.2e-3 for higher early-training velocity without late-warmup drift. Worth a 2-arm screen on yi.

PR ready for advisor review (re-marking via senpai:submit-experiment-results).

morganmcg1:
ADVISOR: Excellent work — the matched-compute Muon@1e-3 result (+24.8% relative vs AdamW@1e-4, with tau_y −7.34pp and tau_z −6.95pp) is a strong methodological finding that directly targets our binding axes. The stability boundary characterization (LR ceiling between 1e-3 and 1.5e-3 for 4L/512d) is also clean and useful.

**Before I merge:** This branch has a merge conflict against the current yi HEAD. Please rebase on yi HEAD, resolve the conflict, and force-push. Once the conflict is cleared I will merge immediately.

**After merge:** The next step is exactly as you recommended — port Muon@1e-3 to the tay DDP8 stack for a full-budget run. I'll assign that as a Wave 16 experiment to a student on the tay track. The compounding hypothesis (STRING-sep × Muon × per-axis loss reweighting) is the most interesting architecture-optimization stack to test next.

One quick clarification on the per-step weight delta scaling argument: the Muon stability ceiling shifts proportionally with sqrt(hidden_dim), so when we test on 6L/512d (matching thorfinn's capacity PR) the ceiling may be similar. But on the tay-branch larger models we should redo the LR ceiling sweep — do not assume 1e-3 is universally safe.

Rebase and push when ready.

morganmcg1:
STUDENT edward:

## Rebase complete on yi HEAD (328dffb)

Rebased `edward/muon-optimizer-string-sep` on `origin/yi` and force-pushed. Branch is now at commit `be7079f` on top of `328dffb`.

### Conflicts resolved in `train.py` (3 hunks)

1. **Imports block** — kept yi's DDP imports (`DistributedDataParallel`, `DistributedSampler`) and added the Muon `from torch import Tensor` import.
2. **Optimizer construction** — merged the optimizer if/elif chain so all three optimizers are now supported on the rebased branch:
   - `adamw` (default, yi)
   - `lion` (yi, PR #450 lineage)
   - `muon` (this PR)
   The `train_model` DDP wrapping from yi is preserved before the optimizer construction. `is_main` gating used for Muon's verbose param-group print.
3. **Warmup loop** — kept yi's `effective_warmup_steps` (which prefers `lr_warmup_epochs * len(train_loader)` when provided) and combined it with my per-group LR scaling logic (`base_lr * warmup_ratio` over `initial_group_lrs`). This is necessary because Muon has two param groups with different base LRs (Muon main + AdamW fallback for 1-D); the per-group scaling preserves the ratio during warmup. AdamW and Lion still work because they only have one group, so the behavior is identical.

Also removed a duplicate `optimizer: str = "adamw"` field that ended up in the `Config` dataclass after the rebase (the muon-branch declaration with the muon-related fields is the surviving one).

### Verification

- `python -c "import ast; ast.parse(open('train.py').read())"` — syntax OK
- `python -c "from train import parse_args; parse_args(['--help'])"` — CLI exposes `--optimizer`, `--muon-momentum`, `--muon-ns-steps`, `--muon-adamw-lr-scale`
- Quick `Muon(...).step()` smoke test on a 2-layer model: works, both param groups carry the expected LRs

### What did NOT change

No re-run was performed (per your guidance). The matched-compute Muon@1e-3 vs AdamW@1e-4 result (val_abupt 16.063% vs 21.367%, run `sgbkzyxy` vs `a0go9jot`) stands as the canonical methodological finding.

PR is back to ready-for-review and labels are unchanged. Ready to merge.

morganmcg1:
ADVISOR: Confirmed merge. Muon optimizer (commit `d3163ade`) is now on the yi branch.

Early matched-compute result at step 12,354 (~57% epoch 1): **−5.304pp (−24.8% relative)** advantage over AdamW@1e-4 on abupt. The wsy and wsz binding-constraint axes showed the largest gains (−7.3pp and −7.0pp respectively).

This is a **capability addition** to yi — it doesn't change the bengio baseline (which remains PR #174 alphonse, val_abupt=6.9549%). The Newton-Schulz orthogonalization adds ~15% step overhead but the faster convergence more than compensates.

Follow-up recommendation: run a full-budget (50-epoch equivalent) Muon vs AdamW comparison on yi to confirm the advantage holds to convergence, then consider porting to bengio once the yi architecture is finalized.

PR #394, opened 2026-05-02
chihiro: per-axis output head scaling (tau_y/tau_z magnitude calibration)

morganmcg1:
# chihiro: per-axis output head scaling (cheap tau_y/tau_z magnitude fix)

## Hypothesis

The largest remaining gaps to AB-UPT are concentrated in tau_y (×2.53) and tau_z (×2.88). A root cause may be that the model systematically under-predicts tau_y and tau_z in magnitude. Adding a learnable scalar multiplier per output channel (initialized to 1.0) gives the model a direct path to recalibrate the dynamic range of each axis independently, without any loss reformulation or architectural changes. This is a pure "can the model represent the right scale?" diagnostic.

If `s_tau_y` and `s_tau_z` converge to values significantly > 1.0 during training, that confirms systematic under-prediction and motivates further magnitude-calibration experiments. If they stay ≈ 1.0, the bug is not a magnitude issue.

## Implementation

This requires a small code change to `model.py` (approximately 20 LOC). After the output projection head, multiply predictions by a learned per-channel scale vector:

```python
# In the model's __init__, add:
self.output_scale = nn.Parameter(torch.ones(4))  # one per output channel: cp, tau_x, tau_y, tau_z
self.output_bias = nn.Parameter(torch.zeros(4))  # optional learned bias

# In forward(), after the final output projection (last line before return):
output = output * self.output_scale + self.output_bias
```

The scale is initialized to all-ones and bias to all-zeros so training starts from the exact Transolver baseline. Add a small L2 penalty on `(self.output_scale - 1)^2` with strength 1e-4 to prevent degenerate solutions (scale collapsing to 0):

```python
# In the loss computation, add to total loss:
scale_reg = 1e-4 * ((model.module.output_scale - 1.0) ** 2).sum()
loss = loss + scale_reg
```

**Important:** The 4 output channels correspond to `[surface_pressure (cp), tau_x, tau_y, tau_z]` — verify the order matches the model's output tensor layout before implementing. Check the model's output head definition to confirm the channel ordering.

After training, report the final learned `output_scale` values (log them to W&B as `scale/cp`, `scale/tau_x`, `scale/tau_y`, `scale/tau_z`).

## Training Command

```bash
cd target/
  --agent chihiro \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 4 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --grad-clip-norm 0.5 \
  --wandb-group chihiro-per-axis-head-scaling
```

**Do not modify any other parameters** beyond adding the output scaling code change.

## Baseline Metrics to Beat

## Why This Should Work

- tau_y and tau_z are the two channels furthest from AB-UPT (×2.53, ×2.88)
- Per-channel scaling is a strict superset of the current model — it cannot hurt if the regularization is in place
- If the model is systematically under-predicting tau magnitude, this provides a direct, gradient-accessible correction path
- Very low implementation risk (~20 LOC), composes orthogonally with all future experiments
- High expected return per LOC ratio (~−0.3 to −1.0 abupt)

## What to Report

After training completes:

1. Your rank-0 W&B run ID
4. Final learned `output_scale` values for all 4 channels — this is critical diagnostic information even if the metric doesn't beat baseline
5. Whether tau_y and tau_z test metrics specifically improved vs SOTA


morganmcg1:
ADVISOR: Kickoff — per-axis scaling diagnostic, verify output channel order first

Your per-axis output scaling hypothesis is a clean diagnostic. If τ_y and τ_z are systematically under-predicted in magnitude, the learned scale parameters s_tau_y and s_tau_z should converge to values significantly > 1.0. This will tell us whether the problem is representational (the model can't express large τ_y/τ_z) or structural (the model doesn't see the right features).

**Critical first step — verify output channel ordering:**
Before implementing, print the target tensor layout:
```python
print("surface target shape:", batch.surface_y.shape)  # e.g. [B, N, 4]
# Log mean abs value per channel to verify ordering
for i in range(batch.surface_y.shape[-1]):
    print(f"  channel {i}: mean_abs={batch.surface_y[..., i].abs().mean():.4f}")
```
The channels should be [cp, tau_x, tau_y, tau_z] but confirm this matches the model output. A mismatch here would silently apply the wrong scale to the wrong channel.

**Implementation notes:**

1. **Scale initialization**: `nn.Parameter(torch.ones(4))` — correct. Do NOT use a learnable bias unless you have a strong reason; it introduces ambiguity with the output distribution shift.

2. **Regularization**: The L2 penalty `1e-4 * ((scale - 1)^2).sum()` is a good prior. Start there. If scale collapses toward 0, increase to `1e-3`.

3. **Log the scale values to W&B**: This is the main diagnostic. Add:
   ```python
   wandb.log({"scale_cp": model.module.output_scale[0].item(),
              "scale_tau_x": model.module.output_scale[1].item(),
              "scale_tau_y": model.module.output_scale[2].item(),
              "scale_tau_z": model.module.output_scale[3].item()})
   ```
   at each epoch. I will check these values directly — they are the scientific result of this experiment.

4. **Apply AFTER loss**: Apply scaling only to the final output (for inference/evaluation), NOT inside the loss computation. The training loss should be computed on unscaled predictions to avoid double-counting the regularization.

Wait — re-reading the hypothesis: apply scaling to predictions BEFORE loss, so the model trains with the recalibrated outputs. The scale then compensates for systematic bias. Apply to both training and eval.

**Reproduce command**:
```bash
  --use-output-scale --output-scale-reg 1e-4 \
  --wandb_group peraxis_output_scaling
```

Post W&B link at ep3 with the per-axis scale values logged. The scale trajectory IS the main result of this experiment.

morganmcg1:
ADVISOR: Closing PR #394 because Closing as misassignment — student 'chihiro' does not have a senpai-drivaerml-ddp8 GPU pod allocated for this research target. The DrivAerML pod fleet only includes: alphonse, askeladd, edward, fern, frieren, nezuko, tanjiro, thorfinn (8 students total, not the 16 across all targets). This PR cannot run on this branch.

PR #395, opened 2026-05-02
emma: multi-sigma RFF (multi-scale spectral encoding on STRING-sep base)

morganmcg1:
# emma: multi-sigma RFF (multi-scale spectral encoding on STRING-sep base)

## Hypothesis

The STRING-sep PR #311 SOTA uses a single learnable spectral encoding. A key limitation is that it learns one frequency scale per axis. Physical aerodynamic fields have multi-scale structure: local boundary layer effects at fine scales (high frequency), large wake and separation zones at coarse scales (low frequency). By concatenating RFF encodings at three sigma scales simultaneously (σ = {0.1, 1.0, 10.0}) alongside the STRING-sep frequencies, the model receives spectral coverage across three orders of magnitude, allowing it to simultaneously resolve near-wall pressure gradients and large-scale flow topology.

The STRING-sep win came from learning per-axis spectral emphasis. Multi-sigma adds depth: multiple frequency octaves rather than a single learnable scale.

## Implementation

This requires code changes to add a multi-scale RFF module. The key change is in the positional encoding component of `model.py`:

```python
class ConcatMultiScaleRFF(nn.Module):
    """Multi-scale RFF: concatenate encodings at sigma={0.1, 1.0, 10.0}."""
    def __init__(self, in_dim=3, num_features=32, sigmas=(0.1, 1.0, 10.0)):
        super().__init__()
        self.sigmas = sigmas
        # One random projection matrix per sigma, fixed (not learned)
        for i, sigma in enumerate(sigmas):
            B = torch.randn(in_dim, num_features) * sigma
            self.register_buffer(f'B_{i}', B)
    
    def forward(self, x):
        # x: (..., 3) coordinates
        parts = []
        for i in range(len(self.sigmas)):
            B = getattr(self, f'B_{i}')
            proj = x @ B  # (..., num_features)
            parts.append(torch.cat([torch.sin(proj), torch.cos(proj)], dim=-1))
        return torch.cat(parts, dim=-1)  # (..., 6 * num_features)
```

The input dimension increases from `7 + 2*num_features` to `7 + 6*num_features` (3 sigmas × 2 sin/cos components × num_features). Use `num_features=32` to keep total encoding dim at 192 (matching STRING-sep control), or use `num_features=16` to give `7 + 96 = 103` total input dim.

**Recommended:** Use `num_features=32` with all 3 sigmas → total 192-dim RFF concat (identical to STRING-sep's 192-dim encoding, but multi-scale fixed Gaussian instead of per-axis learnable). This makes the comparison clean: fixed multi-scale isotropic vs. single-scale learnable STRING-sep.

**Key note:** This hypothesis tests multi-scale spectral coverage on top of or alongside STRING-sep. Stack them: feed both STRING-sep encoding AND multi-sigma RFF encodings as input features to the model.

## Training Command

```bash
cd target/
  --agent emma \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 4 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --grad-clip-norm 0.5 \
  --wandb-group emma-multi-sigma-rff
```

The `--pos-encoding-mode string_separable` keeps the SOTA STRING-sep encoding. Your code change adds the multi-sigma RFF on top (concatenated to the input features), so this flag stays as-is. Adjust the model's input projection layer to accept the larger input dim.

## Baseline Metrics to Beat

| Metric | Current SOTA (PR #311) | AB-UPT Reference | Merge Bar |
|--------|----------------------|------------------|-----------|
| `val_abupt` (primary) | **7.546%** | — | < 7.546% |
| `test_abupt` | **8.771%** | — | < 8.771% |
| `surface_pressure` | 4.485% | 3.82% | < 4.485% |
| `wall_shear` | 8.227% | 7.29% | < 8.227% |
| `volume_pressure` | 12.438% | 6.08% | < 12.438% |
| `tau_x` | 7.253% | 5.35% | — |
| `tau_y` | 9.233% | 3.65% | — |
| `tau_z` | 10.449% | 3.63% | — |

## Why This Should Work

- PR #311 demonstrated that spectral encoding quality matters enormously for this problem (+13.39% relative over no encoding)
- Multi-scale RFF covers frequency ranges that a single-sigma encoding misses: the coarse features (σ=10) capture large-scale pressure gradients; the fine features (σ=0.1) capture near-wall boundary layer effects
- Stacking with STRING-sep gives the model both learnable per-axis emphasis AND multi-resolution coverage
- Fixed multi-scale RFF is complementary to learnable STRING-sep (different inductive bias — isotropic vs. axis-separable)
- Expected lift: −1.0 to −2.0 abupt

## What to Report

After training completes:

1. Your rank-0 W&B run ID
4. Whether multi-scale RFF improved any specific axes vs SOTA (particularly volume_pressure and wall_shear)


morganmcg1:
ADVISOR: Good hypothesis — stacking multi-scale spectral coverage on top of STRING-sep is a principled way to extend spectral bandwidth. The PR body has the right intuition.

**Implementation notes:**
1. The `ConcatMultiScaleRFF` class in the PR body is correct. Use `num_features=32` with all 3 sigmas (σ={0.1, 1.0, 10.0}) for a 192-dim RFF output. Total input dim = 7 (surface features) + STRING-sep encoding dim + 192 (multi-sigma RFF). Update the model input projection to accept this expanded dim.
2. The B matrices should be `register_buffer` (not `nn.Parameter`) since they are fixed random projections, not learned. The PR body already shows this correctly.
3. Apply the multi-sigma RFF to coordinates ONLY (the first 3 dims of the input feature vector), not to normals or area features.
4. For volume points: apply the same multi-sigma RFF to volume xyz coordinates. Volume points do not have normals/area, so their input expansion is: `[x,y,z,sdf]` → `[x,y,z,sdf, STRING-sep(xyz), multi-sigma-rff(xyz)]`.

**Compliance reply required:** Post your rank-0 W&B run ID within 30 minutes of picking up this assignment.

**Kill threshold:** Add `--kill-thresholds "3000:val_primary/abupt_axis_mean_rel_l2_pct<25"` to your training command.

**What to report when done:**
1. Rank-0 W&B run ID  
3. Full test metrics from best-val checkpoint  
4. Specifically: did multi-sigma RFF help volume_pressure and wall_shear vs the STRING-sep-only SOTA?  
5. If training diverged at any sigma: note which run step and what the training loss looked like  

**Merge bar:** val_abupt < 7.546% AND test_abupt < 8.771%.

morganmcg1:
ADVISOR: Great hypothesis emma. The multi-sigma RFF stacked on STRING-sep is exactly the right way to test multi-scale spectral coverage without ablating the STRING-sep win.

A few implementation clarifications before you start:

1. **Stacking strategy**: Concatenate the multi-sigma RFF output *after* the STRING-sep encoding. So the full input feature vector is: [coords(3), normals(3), string_sep_encoding(dim), rff_sigma0.1(64), rff_sigma1.0(64), rff_sigma10.0(64)]. Use `num_features=32` per sigma → 64 sin/cos per sigma → 192 total RFF dim. String-sep encoding dim is typically 192 as well, so total positional feature dim doubles — adjust input projection accordingly.

2. **Random matrices are fixed**: Register them as buffers (non-parameter), not Parameters. The whole point is that these are fixed Fourier features; the model's linear layers learn the weighting. Do not make them learnable.

3. **Input dim bookkeeping**: Current input is 7 + string_sep_dim. After your change it becomes 7 + string_sep_dim + 192. Check model.py for where the input projection is defined and adjust the `in_features` of the first Linear layer.

4. **Volume points**: Multi-sigma RFF applies to both surface and volume points (they all have 3D coordinates). Surface points get the larger feature vector; volume points do not have normals, so they already get zero-padded normals. Apply the same RFF to their coordinates too.

5. **Expected diagnostic**: At ep3, check vol_pressure specifically — this is the channel most likely to benefit from coarse-scale (σ=10) coverage. If vol_pressure doesn't improve by ep5, the multi-scale approach may not be complementary to STRING-sep.

**σ values**: The suggested {0.1, 1.0, 10.0} are physical-scale motivated — 0.1m ≈ boundary layer scale, 1.0m ≈ car body scale, 10.0m ≈ wake scale. These are good defaults for DrivAerML where cars are ~4m long.

Start the run and post back your rank-0 W&B run ID as soon as it launches. Watch ep2-3 closely — multi-sigma RFF should show faster early convergence if the multi-scale spectral coverage is helping.

morganmcg1:
ADVISOR: Closing PR #395 because Closing as misassignment — student 'emma' does not have a senpai-drivaerml-ddp8 GPU pod allocated for this research target. The DrivAerML pod fleet only includes: alphonse, askeladd, edward, fern, frieren, nezuko, tanjiro, thorfinn (8 students total, not the 16 across all targets). This PR cannot run on this branch.

PR #396, opened 2026-05-02
gilbert: local tangent-frame input features (tau_y/tau_z geometric context)

morganmcg1:
# gilbert: local tangent-frame input features (tau_y/tau_z geometric context)

## Hypothesis

The model currently receives 3D coordinates and surface normals as input. It has no explicit representation of the two orthogonal tangent directions at each surface point. Wall shear stress (tau_x, tau_y, tau_z) is fundamentally a tangential force — it acts in the plane tangent to the surface, not in the normal direction. By adding the two local tangent vectors [t1x, t1y, t1z, t2x, t2y, t2z] as additional input features, we give the model explicit geometric context about the tangential frame at each point, directly relevant to predicting tau.

The test gaps for tau_y (×2.53) and tau_z (×2.88) suggest the model cannot currently distinguish the two tangential directions well. The tangent frame provides this distinction directly.

## Implementation

Add local tangent frame features to the input preprocessing (~50 LOC in train.py or the data loading module). The tangent vectors can be derived from the surface normal using a stable Gram-Schmidt process:

```python
def compute_tangent_frame(normals):
    """
    normals: (N, 3) unit surface normals
    Returns: t1 (N, 3), t2 (N, 3) — two orthonormal tangent vectors
    """
    # Choose a reference vector not parallel to normal
    # Use (0,0,1) for most points, (0,1,0) where normal is near (0,0,1)
    ref[:, 2] = 1.0  # z-axis default
    # Where normal is too close to z-axis, use y-axis instead
    parallel_mask = (normals[:, 2].abs() > 0.9)
    ref[parallel_mask, 2] = 0.0
    ref[parallel_mask, 1] = 1.0
    
    # Gram-Schmidt: t1 = normalize(ref - (ref·n)n)
    dot = (ref * normals).sum(-1, keepdim=True)
    t1 = ref - dot * normals
    t1 = F.normalize(t1, dim=-1)
    
    # t2 = n × t1 (cross product gives the second tangent)
    t2 = F.normalize(t2, dim=-1)
    
    return t1, t2
```

Then concatenate to the input features:
```python
t1, t2 = compute_tangent_frame(surface_normals)  # (N, 3) each
# New input: [coords(3), normals(3), t1(3), t2(3), ...other features]
# Input dim increases from 7 to 13 (for surface points)
```

**Important:** The tangent frame is not unique (any rotation of t1, t2 in the tangent plane is equally valid). The convention above is deterministic given the normal, so it's consistent across train/val/test. The model will learn to use the frame as a coordinate system for decomposing tau.

Adjust the model's input projection layer to accept the increased input dimension (7 → 13 for surface points; volume points don't have normals so handle them separately or zero-pad the tangent features).

## Training Command

```bash
cd target/
  --agent gilbert \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 4 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --grad-clip-norm 0.5 \
  --wandb-group gilbert-tangent-frame-features
```

## Baseline Metrics to Beat

## Why This Should Work

- tau is a tangential force: its three components (tau_x, tau_y, tau_z) represent the shear in world coordinates, but the *physically meaningful decomposition* is along the surface tangent directions
- By giving the model the tangent frame explicitly, it can learn to project tau onto tangential directions (aligned with the flow) vs. the secondary tangent (perpendicular to flow, driving tau_y and tau_z asymmetries)
- The tau_y (×2.53) and tau_z (×2.88) gaps are specifically about tangential directions the model currently has no explicit access to
- Low risk (~50 LOC), no model changes beyond input dim

## What to Report

1. Your rank-0 W&B run ID
4. Whether tau_y and tau_z specifically improved vs SOTA (primary diagnostic)
5. Whether wall_shear improved (tangent features should also help with aggregate wall shear)


morganmcg1:
ADVISOR: Welcome to the assignment. This is a clean geometric feature addition targeting the tau_y (×2.53) and tau_z (×2.88) gaps that are our largest remaining weaknesses.

**Implementation notes:**
The Gram-Schmidt tangent frame code in the PR body is correct. Two clarifications on the implementation:
1. The `parallel_mask` threshold of 0.9 on the z-dot is appropriate. Just make sure the resulting t1/t2 tensors are unit-norm (the `F.normalize` calls handle this).
2. For volume points: since they have no surface normal, zero-pad the 6 tangent feature channels. The model input dim for volume points becomes 10 (4 original + 6 zeros). Surface input dim becomes 13 (7 original + 6 tangent frame features). Update the model input projection accordingly.
3. Log the tangent frame norms at step 0 to confirm t1 and t2 are unit vectors (quick sanity check).

**Compliance reply required:** Post your rank-0 W&B run ID within 30 minutes of picking up this assignment.

**Kill threshold:** Add `--kill-thresholds "3000:val_primary/abupt_axis_mean_rel_l2_pct<25"` to your training command.

**What to report when done:**
1. Rank-0 W&B run ID  
3. Full test metrics from best-val checkpoint  
4. tau_y and tau_z specifically — did the tangent frame help those axes?  
5. Wall_shear aggregate  

**Merge bar:** val_abupt < 7.546% AND test_abupt < 8.771%.

morganmcg1:
ADVISOR: Good hypothesis gilbert. The tau_y/tau_z gap (×2.53, ×2.88) is exactly what motivates this — the model has no explicit tangent-frame representation. This is one of the lower-risk high-value changes we can make.

Implementation notes:

1. **Gram-Schmidt stability**: The approach in the PR body is correct. One edge case: some surface normals may have been precomputed with varying conventions. Verify the normals in the dataset are unit vectors before calling the function (add a `F.normalize` call defensively). Confirm normals are in the range [-1, 1] per component.

2. **Deterministic convention**: The (0,0,1) → (0,1,0) fallback for near-z normals is fine, but pick a consistent threshold. Use `normals[:, 2].abs() > 0.95` (not 0.9) to reduce the fallback region — near-z normals should use the z-reference unless they are very close to z.

3. **Volume points**: Volume points have no surface normal. For volume points, **zero-pad** the 6 tangent feature dimensions (t1=zeros(3), t2=zeros(3)). The model should learn to ignore zero tangent features for volume tokens, and this is cleaner than trying to define fake tangent frames for volume points.

4. **Input dim**: Current surface input is 7 (coords 3 + normals 3 + 1 scalar). After adding t1+t2: 7 + 6 = 13. Volume input: 4 (coords 3 + 1 scalar) + 6 = 10, but zero-padded to match 13 if surface and volume share the same input projection. Check how the model handles the joint surface+volume encoding.

5. **No architectural changes needed beyond input layer**: Just adjust the `in_features` of the first Linear projection in model.py to accept 13 instead of 7 for surface inputs.

**Key diagnostic to watch**: At ep3 and beyond, monitor tau_y and tau_z relative to the SOTA 9.233% and 10.449% respectively. If those don't move while val_abupt tracks SOTA, the tangent frame features are not being used. Also log `tau_normal_fraction = sqrt(sum((pred·n)²)/sum(‖pred‖²))` if you can — should stay near zero for a well-calibrated prediction.

Post your rank-0 W&B run ID when the run launches.

morganmcg1:
ADVISOR: Closing PR #396 because Closing as misassignment — student 'gilbert' does not have a senpai-drivaerml-ddp8 GPU pod allocated for this research target. The DrivAerML pod fleet only includes: alphonse, askeladd, edward, fern, frieren, nezuko, tanjiro, thorfinn (8 students total, not the 16 across all targets). This PR cannot run on this branch.

PR #400, opened 2026-05-02
haku: log1p tau target normalization (compress heavy-tailed tau distribution)

morganmcg1:
# haku: log1p tau target normalization (compress heavy-tailed tau distribution)

## Hypothesis

Wall shear stress (tau_x, tau_y, tau_z) has a heavy-tailed distribution — near stagnation points and sharp leading edges, tau values are very large, while most of the surface has small tau values. The model's L2 loss is dominated by these extreme values, causing poor calibration on the majority of the surface. Applying a `sign(tau) * log1p(|tau|)` transform compresses the dynamic range of the targets, rebalancing the gradient signal across the full distribution.

The tau_y (×2.53) and tau_z (×2.88) gaps suggest the model is struggling to represent the full dynamic range of these components. Log1p normalization is a standard technique for heavy-tailed regression and has strong theoretical justification for shear quantities with spatially sparse extremes.

## Implementation

This requires ~50 LOC in `train.py` (the loss computation section). Apply the transform to tau targets only — surface pressure (cp) and volume pressure are not heavy-tailed in the same way.

```python
def log1p_transform(x):
    """Compress heavy-tailed tau distribution: sign(x) * log1p(|x|)"""
    return torch.sign(x) * torch.log1p(x.abs())

def log1p_inverse(y):
    """Inverse: sign(y) * (exp(|y|) - 1)"""
    return torch.sign(y) * (torch.exp(y.abs()) - 1.0)
```

At loss time, apply the transform to both predictions and ground truth for tau channels only:

```python
# Assume output channels: [cp, tau_x, tau_y, tau_z]
# Apply log1p transform to tau channels (indices 1,2,3) for loss computation
pred_tau = pred[:, 1:]       # (N, 3)
gt_tau   = gt[:, 1:]         # (N, 3)

pred_tau_log = log1p_transform(pred_tau)
gt_tau_log   = log1p_transform(gt_tau)

# Relative L2 loss in log space:
tau_loss = torch.norm(pred_tau_log - gt_tau_log) / (torch.norm(gt_tau_log) + 1e-8)
```

At **evaluation time**, apply the inverse transform before computing the standard relative-L2 metrics (so metrics remain comparable with SOTA and baseline):

```python
# For evaluation only: invert predictions back to original scale
pred_tau_eval = log1p_inverse(pred_tau_log)
# Compute standard rel-L2 in original space for W&B logging
```

1. Apply transform only to the tau channels (tau_x, tau_y, tau_z), not to cp or volume_pressure
2. The loss is computed in log-space; metrics are always reported in original space
3. At inference time, always invert the transform before computing any metric
4. Add a small epsilon to the denominator of the relative L2 loss: `torch.norm(gt_tau_log) + 1e-8`
5. Clamp the inverse transform output to a reasonable range to prevent exp overflow: `torch.clamp(y.abs(), max=20.0)` before exp

## Training Command

```bash
cd target/
  --agent haku \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 4 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --grad-clip-norm 0.5 \
  --wandb-group haku-log1p-tau-normalization
```

Add a `--tau-log-transform` flag (or implement inline) to enable the log1p transform. Do not modify any other parameters.

## Baseline Metrics to Beat

## Why This Should Work

- tau has a heavy-tailed distribution: sparse extreme values near stagnation/leading edges dominate the L2 gradient
- Log1p transform is the classical tool for compressing power-law / heavy-tailed distributions without losing sign information
- By computing loss in log-space, the model receives balanced gradient signal from both high-tau and low-tau surface regions
- Expected impact: tau_y and tau_z gaps (the worst channels) most likely to benefit, as those extremes are hardest to predict
- Low risk: ~50 LOC, no architectural changes, fully invertible

## What to Report

1. Your rank-0 W&B run ID
4. Whether tau_y and tau_z specifically improved vs SOTA
5. Distribution of predicted tau values at best epoch (log histogram) — to confirm the transform is working


morganmcg1:
ADVISOR: haku, welcome to this assignment. The log1p tau normalization targets the largest error gaps in our suite: tau_y (×2.53 vs AB-UPT) and tau_z (×2.88 vs AB-UPT).

**Critical implementation notes before you start:**

1. **Transform applies to tau channels ONLY** (tau_x, tau_y, tau_z). Do NOT apply to surface_pressure or volume_pressure.
2. **Loss is computed in log-space; metrics are ALWAYS reported in original space** — invert the transform before computing any W&B metric.
3. **Clamp before exp** in the inverse: `torch.clamp(y.abs(), max=20.0)` to prevent overflow.
4. **Relative L2 denominator epsilon**: use `torch.norm(gt_tau_log) + 1e-8` (not 0).
5. Verify the training command uses exactly the SOTA config (4L/512d/4H + EMA=0.999 + gc=0.5 + STRING-sep) as shown in the PR body — the only change should be the log1p transform on tau targets.

**Kill thresholds:**
`--kill-thresholds '3000:val_primary/abupt_axis_mean_rel_l2_pct<25'`

**Compliance requirement:** One run only. Post your W&B run ID (rank-0) within 30 minutes of starting. Report best val_abupt, tau_y, and tau_z at best checkpoint. Specifically confirm whether tau_y and tau_z improved vs SOTA (9.233% and 10.449% respectively).

Merge bar: val_abupt < 7.546% AND test_abupt < 8.771% (PR #311 SOTA).

morganmcg1:
ADVISOR: Good hypothesis haku. The log1p transform is a well-motivated approach for heavy-tailed distributions like wall shear stress.

Key implementation details:

1. **Channel ordering**: Before implementing, print the first batch's target tensor channel order. In DrivAerML the surface outputs are typically [cp, tau_x, tau_y, tau_z]. Verify this is actually the case in train.py before hardcoding indices [1,2,3] for tau channels.

2. **Apply to surface loss only**: The transform applies only to the surface tau channels. The surface pressure (cp, channel 0) and all volume pressure channels should use the standard L2 loss unchanged. Be explicit about this separation.

3. **Invert before metric computation**: At validation/test time, apply `log1p_inverse` to the tau prediction channels before computing relative-L2 metrics. This is critical — if you forget the inversion, your val_abupt will be artificially low (computed in log-space) and won't be comparable to SOTA.

4. **Overflow guard**: Add `torch.clamp(y.abs(), max=85.0)` before `torch.exp` in the inverse transform (not 20.0 as in the PR body — exp(85) is the float32 overflow boundary, giving you more headroom while still being safe).

5. **Gradient scale difference**: Log-space loss has smaller gradients for large tau, larger gradients for small tau. This rebalancing is the whole point. Watch for early epoch instability if the model was initialized expecting large-tau gradients — if loss spikes at ep1, add a warmup where you blend: `loss = (1-α)*log_loss + α*orig_loss` with α decaying from 1.0 to 0.0 over 500 steps.

6. **Diagnostic to report**: Beyond val_abupt, explicitly report tau_y and tau_z at best epoch. Also report val_abupt at ep3 compared to SOTA ep3 — if log1p transform is helping with convergence, you should see this channel-specific improvement early.

**Alternative formulation to try if the above diverges**: Instead of computing the loss directly in log-space, use Huber loss (δ=1.0) on the raw tau channels. Huber is less sensitive to heavy tails than L2 and doesn't require an inverse transform at eval time. Consider this as a fallback if log1p causes training instability.

Post rank-0 W&B run ID when launched.

morganmcg1:
ADVISOR: Closing PR #400 because Closing as misassignment — student 'haku' does not have a senpai-drivaerml-ddp8 GPU pod allocated for this research target. The DrivAerML pod fleet only includes: alphonse, askeladd, edward, fern, frieren, nezuko, tanjiro, thorfinn (8 students total, not the 16 across all targets). This PR cannot run on this branch.

PR #404, opened 2026-05-02
senku: multi-scale radius pooling (local aerodynamic context at 3 physical scales)

morganmcg1:
# senku: multi-scale radius pooling (local aerodynamic context at 3 physical scales)

## Hypothesis

Aerodynamic fields exhibit multi-scale spatial structure: near-wall boundary layers operate at millimeter scales (r~5cm in the car's reference frame), pressure gradients around curved surfaces at ~15cm, and large-scale wake/separation structures at ~45cm. A single global attention mechanism cannot simultaneously resolve all these scales. By adding multi-scale radius ball pooling as a pre-encoder step — aggregating local features at three physically-motivated radii — the model receives explicit multi-scale neighborhood context before the Transolver global attention.

This is a purely input-side operation: the trunk architecture is unchanged. Only the input features are enriched with multi-scale pooled descriptors.

## Implementation

This requires ~120 LOC change. The key operation is `torch_cluster.radius` for radius ball queries at three scales. The three pooled feature vectors are concatenated and projected back to the original feature dimension.

```python
# Multi-scale radius pooling module (~90 LOC)
class MultiScaleRadiusPooling(nn.Module):
    """
    For each point, aggregate features from neighbors within radius r.
    Three scales: r = {0.05, 0.15, 0.45} meters.
    
    DrivAer car is ~4.6m long. If coordinates are normalized to [0,1],
    multiply physical radii by (1/4.6): r_norm = {0.011, 0.033, 0.098}
    Check the actual coordinate range in the dataset before setting radii.
    """
    def __init__(self, in_dim, radii=(0.05, 0.15, 0.45), coord_scale=1.0):
        super().__init__()
        # Adjust radii by coordinate scale factor
        self.radii = [r * coord_scale for r in radii]
        self.n_scales = len(radii)
        
        # Project concatenated multi-scale features back to in_dim
        # Input: in_dim * (1 + n_scales) = in_dim * 4 (original + 3 pooled)
        self.proj = nn.Linear(in_dim * (1 + self.n_scales), in_dim)
        # Initialize to identity on the first in_dim components (preserve input)
        nn.init.zeros_(self.proj.weight)
        nn.init.zeros_(self.proj.bias)
        # Set weights for the original features to identity
        self.proj.weight.data[:, :in_dim] = torch.eye(in_dim)
    
    def forward(self, x, coords):
        """
        x: (N, in_dim) — point features
        coords: (N, 3) — point coordinates
        Returns: (N, in_dim) — enriched features
        """
        from torch_cluster import radius as radius_query
        
        parts = [x]  # Start with original features
        
        for r in self.radii:
            # radius_query returns (row, col) where row=query idx, col=neighbor idx
            # query points = x, database points = x (self-query)
            row, col = radius_query(coords, coords, r, max_num_neighbors=64)
            
            # Mean pooling of neighbor features for each center point
            pooled = torch.zeros_like(x)
            count = torch.zeros(x.shape[0], 1, device=x.device)
            pooled.scatter_add_(0, row.unsqueeze(1).expand(-1, x.shape[1]), x[col])
            count.scatter_add_(0, row.unsqueeze(1), torch.ones(row.shape[0], 1, device=x.device))
            count = count.clamp(min=1)
            pooled = pooled / count
            parts.append(pooled)
        
        # Concatenate all scales and project
        multi_scale = torch.cat(parts, dim=-1)  # (N, in_dim * 4)
        return self.proj(multi_scale)
```

**Determining the correct coordinate scale:**
Before implementing, check what coordinate range the dataset uses:
```python
# In train.py data loading, add a one-time print:
print(f"Coord range: min={coords.min():.3f}, max={coords.max():.3f}")
```
- If coords are in meters (~[-2, 2.5] range for DrivAer), use `coord_scale=1.0` and `radii=(0.05, 0.15, 0.45)`
- If coords are normalized to [0,1], use `coord_scale=0.217` (≈1/4.6) or equivalent

**Integration into model:**

```python
# In model __init__:
self.ms_radius_pool = MultiScaleRadiusPooling(in_dim=input_feature_dim)

# In model forward, before Transolver trunk:
# Apply to both surface and volume points separately
surface_features = self.ms_radius_pool(surface_features, surface_coords)
volume_features  = self.ms_radius_pool(volume_features,  volume_coords)
```

**Memory note:** `max_num_neighbors=64` caps the number of neighbors per query. With N=65536 surface points and r=0.45m, each point may have O(100-1000) neighbors without this cap. The cap prevents OOM while preserving the aggregation signal.

## Training Command

```bash
cd target/
  --agent senku \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 4 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --grad-clip-norm 0.5 \
  --wandb-group senku-multi-scale-radius-pooling
```

Do not modify any other parameters beyond adding the multi-scale radius pooling code change.

## Baseline Metrics to Beat

## Why This Should Work

- Aerodynamic fields are fundamentally multi-scale: boundary layers (fine), separation bubbles (medium), wake (coarse)
- Radius ball pooling at three physical scales gives the model explicit multi-resolution context
- DrivAer geometry has physically motivated scale structure: the three radii (5cm, 15cm, 45cm) correspond to near-surface, local geometry, and regional flow features
- Purely input-side operation: trunk architecture unchanged, no risk of destabilizing the trained attention patterns
- The STRING-sep encoding captures spectral scale; radius pooling captures spatial scale — these are complementary

## What to Report

1. Your rank-0 W&B run ID
4. Whether volume_pressure improved — volume pooling should capture the large wake structures
5. Report the coordinate range you found in the dataset and confirm the radii you used
6. Any OOM issues with the large radius (0.45m) — report neighbor count statistics if needed


morganmcg1:
ADVISOR: Kickoff — multi-scale pooling, key implementation note on coordinate scale

Your multi-scale radius pooling hypothesis is physically grounded. The three-scale design (boundary layer / pressure gradient / wake) maps well to DrivAer's aerodynamic regimes.

**Critical implementation note — coordinate scale:**
Check the actual coordinate range in the dataset BEFORE setting radii. DrivAer coordinates in the dataset may be in meters, normalized to [0,1], or another scale. You can print `batch.surface_x[:,:,:3].min()` and `batch.surface_x[:,:,:3].max()` in the first training step. The physical radii {0.05m, 0.15m, 0.45m} need to be scaled accordingly:
- If coords are in meters (range ~[−3, 3]): radii ≈ {0.05, 0.15, 0.45} — use as-is
- If coords are normalized to [0,1]: radii ≈ {0.011, 0.033, 0.098}

**Pooling aggregation**: Mean pooling within each radius ball is the natural choice. Max pooling is also worth trying (captures peak features, better for pressure extrema). Start with mean.

**Projection**: After concatenating `[original (d), scale1 (d), scale2 (d), scale3 (d)]` → 4d, project back to d. Initialize projection weight as `[I | 0 | 0 | 0]` (keep original features, learn the multi-scale deltas) for a safer skip-connection start.

**torch_cluster `radius`**: If `torch_cluster` is not installed, fall back to manual distance computation. Check the environment first:
```python
try:
    from torch_cluster import radius
except ImportError:
    # Manual implementation: cdist + threshold
```

**Reproduce command**:
```bash
  --use-multiscale-pooling --pooling-radii 0.05,0.15,0.45 \
  --wandb_group multiscale_radius_pooling
```

Post W&B link at ep3 with per-axis breakdown. I am especially interested in whether tau_y and tau_z improve — multi-scale context should help distinguish wake cross-flow from boundary-layer cross-flow.

morganmcg1:
ADVISOR: Closing PR #404 because Closing as misassignment — student 'senku' does not have a senpai-drivaerml-ddp8 GPU pod allocated for this research target. The DrivAerML pod fleet only includes: alphonse, askeladd, edward, fern, frieren, nezuko, tanjiro, thorfinn (8 students total, not the 16 across all targets). This PR cannot run on this branch.

PR #405, opened 2026-05-02
violet: surface curvature features (κ_H and κ_G via k-NN PCA)

morganmcg1:
# violet: surface curvature features (κ_H and κ_G via k-NN PCA)

## Hypothesis

The DrivAer surface has regions of high curvature (hood edge, windshield base, wheel arches, mirrors) that generate strong local pressure gradients and flow separation. Currently the model only sees coordinates and surface normals — it cannot distinguish a flat panel from a sharply curved edge. By computing mean curvature (κ_H) and Gaussian curvature (κ_G) from the local normal covariance and adding them as input features, we give the model explicit geometric context about local surface shape.

These curvature features are physically meaningful: κ_H controls the pressure coefficient distribution via the boundary-layer momentum equation; κ_G distinguishes saddle regions (negative) from dome/valley regions (positive). High-curvature regions are precisely where the model currently struggles: tau_y and tau_z errors are largest on curved surfaces.

## Implementation

This requires ~100 LOC, primarily in the data preprocessing or train.py loading. The curvature is estimated from the normal covariance within a k-NN neighborhood (k=16):

```python
def compute_surface_curvature(coords, normals, k=16):
    """
    Estimate mean curvature (kH) and Gaussian curvature (kG) from
    local normal covariance matrix (k-NN PCA approach).
    
    coords:  (N, 3) — surface point positions
    normals: (N, 3) — unit surface normals
    Returns: kH (N,), kG (N,) — curvatures, clipped to [-50, 50]
    """
    from torch_cluster import knn_graph
    import torch
    
    # Build k-NN graph (in coordinate space)
    edge_index = knn_graph(coords, k=k, loop=True)  # loop=True includes self
    row, col = edge_index  # row=center, col=neighbor
    
    # Scatter normal vectors to neighborhoods
    # For each center point i, gather k neighbor normals
    N = coords.shape[0]
    
    kH = torch.zeros(N, device=coords.device)
    kG = torch.zeros(N, device=coords.device)
    
    # Build (N, k, 3) neighbor normal matrix
    # Reshape edge_index to (N, k) assuming knn_graph returns in order
    neighbor_idx = col.view(N, k)  # (N, k)
    
    # Gather neighbor normals: (N, k, 3)
    neighbor_normals = normals[neighbor_idx]  # (N, k, 3)
    
    # Center the normals per neighborhood
    mean_n = neighbor_normals.mean(dim=1, keepdim=True)  # (N, 1, 3)
    centered = neighbor_normals - mean_n  # (N, k, 3)
    
    # Covariance matrix of normals: (N, 3, 3)
    # cov_i = (1/k) * sum_j (n_j - mean_n)^T (n_j - mean_n)
    cov = torch.bmm(centered.transpose(1, 2), centered) / k  # (N, 3, 3)
    
    # Eigenvalues of covariance matrix
    # The two principal curvatures are related to the eigenvalues
    try:
        eigvals = torch.linalg.eigvalsh(cov)  # (N, 3), sorted ascending
    except:
        # Fallback for older torch
        eigvals = torch.symeig(cov, eigenvectors=False).eigenvalues
    
    # The smallest eigenvalue corresponds to the normal direction
    # The two largest correspond to principal curvatures (κ1, κ2)
    # Using eigenvalue ratios as curvature proxies:
    # kH ~ (λ1 + λ2) / (λ0 + λ1 + λ2)   (mean curvature proxy)
    # kG ~ λ1 * λ2 / (λ0 + λ1 + λ2)^2   (Gaussian curvature proxy)
    
    lam0 = eigvals[:, 0]  # smallest (normal direction)
    lam1 = eigvals[:, 1]
    lam2 = eigvals[:, 2]  # largest
    
    total = (lam0 + lam1 + lam2).clamp(min=1e-8)
    
    # Mean curvature proxy: sum of larger eigenvalues / total
    kH = (lam1 + lam2) / total
    
    # Gaussian curvature proxy: product of larger eigenvalues / total^2
    kG = (lam1 * lam2) / (total ** 2)
    
    # Clip to prevent outliers from dominating
    kH = kH.clamp(-50.0, 50.0)
    kG = kG.clamp(-50.0, 50.0)
    
    return kH, kG
```

**Integration into model inputs:**

```python
# Compute curvatures once per sample (in data loading / forward pass):
kH, kG = compute_surface_curvature(surface_coords, surface_normals, k=16)

# Add as additional input features (N, 2) → concatenate to surface input
# New surface input: [coords(3), normals(3), kH(1), kG(1), ...other_features]
# Input dim increases by 2 (e.g., 7 → 9 for surface points)
surface_features = torch.cat([surface_features, kH.unsqueeze(-1), kG.unsqueeze(-1)], dim=-1)
```

**Adjust model input projection layer** to accept `input_dim + 2` for surface points. Volume points do not have surface normals or curvatures — zero-pad the 2 curvature channels for volume points.

1. Compute curvature in the DataLoader worker (not in model forward) to avoid recomputation per epoch
2. Use `loop=True` in knn_graph so each point includes itself; ensures the covariance has full local support
3. Clip values to [-50, 50] before adding to features — a few degenerate mesh triangles can produce extreme curvatures
4. Normalize kH and kG by their dataset-wide statistics (mean/std computed on training set) so they're on the same scale as other inputs
5. For the first implementation, skip normalization and just use the raw clipped values; add normalization in a follow-up if needed
6. Check if `torch.linalg.eigvalsh` is available (PyTorch >= 1.9); fall back to `torch.symeig` if needed

## Training Command

```bash
cd target/
  --agent violet \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 4 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --grad-clip-norm 0.5 \
  --wandb-group violet-surface-curvature
```

Do not modify any other parameters beyond adding the curvature feature computation and input projection adjustment.

## Baseline Metrics to Beat

## Why This Should Work

- Curvature is the fundamental geometric quantity that drives boundary-layer behaviour: strong curvature → strong pressure gradients → high shear stress
- κ_H and κ_G directly encode whether a surface point is on a convex (wing-like), concave (wheel arch), or saddle (door edge) geometry — all important for aerodynamic prediction
- The tau_y and tau_z gaps are largest precisely in areas of high curvature where the flow reattaches or separates tangentially
- Low-complexity change (~100 LOC), no architectural changes beyond input projection dimension
- k-NN PCA curvature is a well-established method from point cloud processing (PCPNet, etc.) directly applicable here

## What to Report

1. Your rank-0 W&B run ID
4. Whether tau_y and tau_z specifically improved vs SOTA (primary diagnostic for curvature features)
5. Whether surface_pressure improved (curvature directly affects Cp distribution)
6. Histogram of kH and kG values in the dataset (to confirm curvature is being computed correctly and the distribution is reasonable)


morganmcg1:
ADVISOR: Kickoff — curvature features, verify numerical stability

Your surface curvature hypothesis is well-motivated. High-curvature regions (hood edge, wheel arches, windshield base) are exactly where tau_y and tau_z are hardest to predict. Adding κ_H and κ_G as explicit input features gives the model a geometric prior about where flow separation is likely.

**Key implementation notes:**

1. **Numerical stability of curvature**: The normal covariance eigenvalue approach can produce NaN/Inf at degenerate points (collinear neighbors, flat patches). Clip aggressively: `κ = torch.clamp(κ, -50, 50)`. Also add a small epsilon inside the eigenvalue computation.

2. **Precompute or online**: Curvature only depends on geometry (coordinates + normals), not the aerodynamic targets. You can precompute and cache it per mesh. However, since DrivAer has 400 configurations and geometry varies, compute it once per sample in the Dataset and cache to disk (numpy `.npy`) on first load. This avoids recomputing every epoch.

3. **Feature normalization**: After clipping to [-50, 50], apply z-score normalization using dataset statistics (mean and std of κ_H and κ_G across all samples). High curvature has a heavy tail — consider `arcsinh(κ)` as an alternative to clipping.

4. **Channel appending**: The current surface input features are 7 channels (coordinates, normals, area). Append κ_H and κ_G as channels 8 and 9. Verify the feature count matches the model's input projection layer — you may need to adjust `--model-surface-input-dim` or the relevant config flag.

5. **PCA vs divergence**: k-NN PCA on normals is a good approach. Alternative: use `div(n)` via finite differences (normal divergence approximates mean curvature). The PCA approach is more robust.

**Reproduce command**:
```bash
  --use-surface-curvature-features --curvature-k 16 \
  --wandb_group surface_curvature_features
```

Post W&B link at ep3 and report whether κ_H/κ_G learned features correlate with tau_y/tau_z error reduction vs baseline.

morganmcg1:
ADVISOR: Closing PR #405 because Closing as misassignment — student 'violet' does not have a senpai-drivaerml-ddp8 GPU pod allocated for this research target. The DrivAerML pod fleet only includes: alphonse, askeladd, edward, fern, frieren, nezuko, tanjiro, thorfinn (8 students total, not the 16 across all targets). This PR cannot run on this branch.

PR #420, opened 2026-05-02
feat: STRING-separable learnable positional encoding on yi branch

morganmcg1:
# STRING-Separable Learnable Positional Encoding on Yi Branch

## Hypothesis

The current yi SOTA (`val_abupt 7.546%`, PR #311, `gcwx9yaa`) was achieved by replacing
the fixed sincos position encoding with a **STRING-separable learnable PE** — per-axis
frequencies with learnable `log_freq` and `phase` parameters. However, PR #311 was
implemented and merged into the **`tay` branch**, not `yi`. The `yi` branch still uses
the original `ContinuousSincosEmbed` with a fixed `omega` buffer.

**Root cause of the gap:** The fixed sincos PE uses a single shared frequency sweep
across all three spatial axes. Aerodynamic point clouds have highly anisotropic structure:
the streamwise (x) dimension spans the full car length while the cross-section (y, z)
dimensions are much smaller. Fixed isotropic frequencies are a mismatch — the model
cannot independently tune how it resolves each axis.

**Fix:** Replace the fixed `omega` buffer with learnable `nn.Parameter` `log_freq` and
`phase` tensors — one scalar per frequency bin per axis. The model then auto-adjusts the
frequency spectrum for each axis during training, attending differently to streamwise vs.
cross-flow spatial scales. This is the known-best architectural improvement from the tay
research track.

**What to implement:** Modify `ContinuousSincosEmbed` in `train.py` to add an optional
`learnable=True` mode. When enabled, replace:
```python
self.register_buffer("omega", 1.0 / max_wavelength ** (arange / effective_dim_per_axis))
```
with:
```python
self.log_freq = nn.Parameter(
    torch.log(1.0 / max_wavelength ** (arange / effective_dim_per_axis))
    .unsqueeze(0).expand(input_dim, -1).clone()
)  # shape: [input_dim, num_freqs]
self.phase = nn.Parameter(torch.zeros(input_dim, len(arange)))  # shape: [input_dim, num_freqs]
```

And in `forward`, replace the fixed omega multiplication with per-axis learnable:
```python
def forward(self, coords: torch.Tensor) -> torch.Tensor:
    coords = coords.float()  # [..., input_dim]
    if self.learnable:
        omega = torch.exp(self.log_freq)  # [input_dim, num_freqs]
        # coords: [..., input_dim], omega: [input_dim, num_freqs]
        out = coords.unsqueeze(-1) * omega  # [..., input_dim, num_freqs]
[...]

PR #421, opened 2026-05-02
feat: Dual-stream Transformer with cross-attention bridge on yi

morganmcg1:
# Dual-Stream Transformer with Cross-Attention Bridge

## Hypothesis

The current `SurfaceTransolver` concatenates surface and volume tokens into a single
sequence and passes them through one shared Transformer backbone. With 4 layers and
128 slices, the slice assignments (physics tokens) must simultaneously serve two
very different physical regimes: boundary-layer surface aerodynamics and far-field
volume pressure. The surface branch's wall shear predictions (tau_y/z at 2.53× and
2.88× above target) are the dominant failure mode.

**Root cause:** When surface and volume tokens share a single Transformer, they
compete for the same physics-token (slice) budget. The `TransolverAttention` learns
a soft assignment of tokens to slices, but a slice cannot simultaneously represent
fine-grained near-wall boundary layer variation AND coarse far-field pressure
gradients. The surface tokens get "diluted" by the volume tokens, preventing the
model from developing specialized surface representations.

**Fix:** Replace the single shared backbone with two separate `Transformer` stacks
(surface stream and volume stream) with a **cross-attention bridge** inserted between
them every `--cross-attn-every` layers. The cross-attention allows each stream to
condition on the other's hidden states without contaminating its own slice
representations.

The architecture is:
```
surface_tokens → [SurfaceTransformer layers] ← cross-attn from volume
volume_tokens  → [VolumeTransformer layers]  ← cross-attn from surface
```

Each domain keeps its own slice budget. Cross-attention is lightweight (standard
multi-head attention, not Transolver-style slicing).

This is directly analogous to the dual-encoder architecture that won the SE(2)-
equivariant geometry experiment in the wandb/senpai noam/radford track — separate
encoders with cross-domain communication proved superior to a single shared encoder.

## Current Baseline

| Metric | Baseline (PR #311) | AB-UPT Target | Gap |
|--------|-------------------|---------------|-----|
| val_abupt (primary) | **7.546%** | — | merge bar |
| test_abupt | 8.771% | — | — |
| surface_pressure_rel_l2_pct | 4.485% | 3.82% | 1.17× |
| wall_shear_rel_l2_pct | 8.227% | 7.29% | 1.13× |
| wall_shear_y_rel_l2_pct | 9.233% | 3.65% | **2.53×** |
| wall_shear_z_rel_l2_pct | 10.449% | 3.63% | **2.88×** |
| volume_pressure_rel_l2_pct | 12.438% | 6.08% | 2.05× |

Baseline W&B run: `gcwx9yaa` (PR #311, edward, STRING-sep PE + yi compounding wins)

### Step 1: Add `CrossAttention` module to `train.py`

After the `TransformerBlock` class (around line 233), add a new module:

```python
class CrossAttentionBlock(nn.Module):
    """Lightweight cross-attention between surface and volume streams.
    
    Each stream attends to the other as key-value source.
    Uses standard multi-head attention (not Transolver slicing).
    """
    def __init__(self, hidden_dim: int, num_heads: int, dropout: float = 0.0):
        super().__init__()
        self.hidden_dim = hidden_dim
        self.num_heads = num_heads
        # Surface attends to volume
        self.norm_surf_q = nn.LayerNorm(hidden_dim, eps=1e-6)
        self.norm_vol_kv = nn.LayerNorm(hidden_dim, eps=1e-6)
        self.surf_to_vol = nn.MultiheadAttention(
            embed_dim=hidden_dim,
            num_heads=num_heads,
            dropout=dropout,
            batch_first=True,
        )
        # Volume attends to surface
        self.norm_vol_q = nn.LayerNorm(hidden_dim, eps=1e-6)
        self.norm_surf_kv = nn.LayerNorm(hidden_dim, eps=1e-6)
        self.vol_to_surf = nn.MultiheadAttention(
            embed_dim=hidden_dim,
            num_heads=num_heads,
            dropout=dropout,
            batch_first=True,
        )
        # Scale cross-attention contribution — residual gating
        self.surf_gate = nn.Parameter(torch.zeros(1))
        self.vol_gate = nn.Parameter(torch.zeros(1))

    def forward(
        self,
        surf: torch.Tensor,       # [B, Ns, D]
        vol: torch.Tensor,        # [B, Nv, D]
        surf_mask: torch.Tensor,  # [B, Ns] 1=valid
        vol_mask: torch.Tensor,   # [B, Nv] 1=valid
    ) -> tuple[torch.Tensor, torch.Tensor]:
        # Build key padding masks for nn.MultiheadAttention
        # (True means IGNORE — opposite of our convention)
        surf_key_pad = ~surf_mask.bool()  # [B, Ns]
        vol_key_pad = ~vol_mask.bool()    # [B, Nv]

        # Surface queries → volume keys/values
        surf_q = self.norm_surf_q(surf)
        vol_kv = self.norm_vol_kv(vol)
        surf_delta, _ = self.surf_to_vol(
            surf_q, vol_kv, vol_kv,
            key_padding_mask=vol_key_pad,
        )
        surf = surf + self.surf_gate.tanh() * surf_delta

        # Volume queries → surface keys/values
        vol_q = self.norm_vol_q(vol)
        surf_kv = self.norm_surf_kv(surf)
        vol_delta, _ = self.vol_to_surf(
            vol_q, surf_kv, surf_kv,
            key_padding_mask=surf_key_pad,
        )
        vol = vol + self.vol_gate.tanh() * vol_delta

        return surf, vol
```

**Key design choices:**
- Zero-initialized gates (`surf_gate`, `vol_gate`) → tanh(0)=0 → cross-attention is
  zero at init. Training gradually opens the gates, providing a stable residual path.
- Separate layer norms for query and key/value of each direction — prevents scale
  mismatch between streams.
- Standard `nn.MultiheadAttention` (not Transolver slicing) — cross-stream
  communication should be dense/global, not physics-token-based.

### Step 2: Add `DualStreamTransformer` class to `train.py`

After the `Transformer` class (around line 330), add:

```python
class DualStreamTransformer(nn.Module):
    """Two separate Transformer streams (surface + volume) with cross-attention bridges.
    
    Args:
        depth: number of Transformer blocks per stream
        hidden_dim: token dimension
        num_heads: attention heads (both self-attention and cross-attention)
        mlp_expansion_factor: MLP hidden = hidden_dim * factor
        num_slices: slice count for TransolverAttention
        cross_attn_every: insert cross-attention bridge every N layers (default 2)
        dropout: dropout probability
        stochastic_depth_prob: max stochastic depth drop rate
    """
    def __init__(
        self,
        depth: int,
        hidden_dim: int,
        num_heads: int,
        mlp_expansion_factor: int | float,
        num_slices: int,
        cross_attn_every: int = 2,
        dropout: float = 0.0,
        stochastic_depth_prob: float = 0.0,
    ):
        super().__init__()
        if depth <= 1:
            drop_path_rates = [stochastic_depth_prob] * depth
        else:
            drop_path_rates = [
                stochastic_depth_prob * i / (depth - 1) for i in range(depth)
            ]
        self.depth = depth
        self.cross_attn_every = cross_attn_every

        self.surface_blocks = nn.ModuleList([
            TransformerBlock(
                hidden_dim=hidden_dim,
                num_heads=num_heads,
                mlp_expansion_factor=mlp_expansion_factor,
                num_slices=num_slices,
                dropout=dropout,
                drop_path_prob=drop_path_rates[i],
            )
            for i in range(depth)
        ])
        self.volume_blocks = nn.ModuleList([
            TransformerBlock(
                hidden_dim=hidden_dim,
                num_heads=num_heads,
                mlp_expansion_factor=mlp_expansion_factor,
                num_slices=num_slices,
                dropout=dropout,
                drop_path_prob=drop_path_rates[i],
            )
            for i in range(depth)
        ])
        # Cross-attention bridges: one after every `cross_attn_every` layers
        num_bridges = depth // cross_attn_every
        self.cross_attn_bridges = nn.ModuleList([
            CrossAttentionBlock(hidden_dim=hidden_dim, num_heads=num_heads, dropout=dropout)
            for _ in range(num_bridges)
        ])

    def forward(
        self,
        surf: torch.Tensor,       # [B, Ns, D]
        vol: torch.Tensor,        # [B, Nv, D]
        surf_mask: torch.Tensor,  # [B, Ns]
        vol_mask: torch.Tensor,   # [B, Nv]
    ) -> tuple[torch.Tensor, torch.Tensor]:
        bridge_idx = 0
        for i in range(self.depth):
            surf = self.surface_blocks[i](surf, attn_mask=surf_mask)
            vol = self.volume_blocks[i](vol, attn_mask=vol_mask)
            # Apply cross-attention bridge every `cross_attn_every` layers
            if (i + 1) % self.cross_attn_every == 0 and bridge_idx < len(self.cross_attn_bridges):
                surf, vol = self.cross_attn_bridges[bridge_idx](surf, vol, surf_mask, vol_mask)
                bridge_idx += 1
        return surf, vol
```

### Step 3: Add `--use-dual-stream` and `--cross-attn-every` flags to `Config`

In the `Config` dataclass (around line 549), add:
```python
use_dual_stream: bool = False
cross_attn_every: int = 2
```

These auto-generate CLI flags `--use-dual-stream` and `--cross-attn-every N`.

### Step 4: Modify `SurfaceTransolver.__init__` to support dual-stream mode

In `SurfaceTransolver.__init__`, replace the single `self.backbone = Transformer(...)` with:

```python
if getattr(config, 'use_dual_stream', False):
    # This branch won't be used here — instead pass use_dual_stream as a param
    pass
```

Actually, add `use_dual_stream: bool = False` and `cross_attn_every: int = 2` to
`SurfaceTransolver.__init__`'s signature, then:

```python
if use_dual_stream:
    self.backbone = DualStreamTransformer(
        depth=n_layers,
        hidden_dim=n_hidden,
        num_heads=n_head,
        mlp_expansion_factor=mlp_ratio,
        num_slices=slice_num,
        cross_attn_every=cross_attn_every,
        dropout=dropout,
        stochastic_depth_prob=stochastic_depth_prob,
    )
else:
    self.backbone = Transformer(
        depth=n_layers,
        hidden_dim=n_hidden,
        num_heads=n_head,
        mlp_expansion_factor=mlp_ratio,
        num_slices=slice_num,
        dropout=dropout,
        stochastic_depth_prob=stochastic_depth_prob,
        film_geom_dim=film_encoder_dim if use_film else 0,
    )
self.use_dual_stream = use_dual_stream
```

Also add two separate `nn.LayerNorm` heads for dual stream (the shared `self.norm` won't work when processing streams separately):

```python
if use_dual_stream:
    self.surface_norm = nn.LayerNorm(n_hidden, eps=1e-6)
    self.volume_norm = nn.LayerNorm(n_hidden, eps=1e-6)
```

### Step 5: Modify `SurfaceTransolver.forward` for dual-stream path

In `SurfaceTransolver.forward`, after encoding surface and volume tokens separately
(already done in `_encode_group`), add a branch:

```python
if self.use_dual_stream:
    # Run dual-stream transformer
    # Handle cases where one modality is absent (rare in practice)
    if surface_x is not None and volume_x is not None:
        surf_hidden, vol_hidden = self.backbone(
            surf_h, vol_h, surface_mask, volume_mask
        )
        surf_hidden = _apply_token_mask(self.surface_norm(surf_hidden), surface_mask)
        vol_hidden = _apply_token_mask(self.volume_norm(vol_hidden), volume_mask)
    elif surface_x is not None:
        # Volume absent — fall back to surface-only single-stream
        surf_hidden = surf_h
        for blk, xblk in zip(self.backbone.surface_blocks, self.backbone.volume_blocks):
            surf_hidden = blk(surf_hidden, attn_mask=surface_mask)
        surf_hidden = _apply_token_mask(self.surface_norm(surf_hidden), surface_mask)
        vol_hidden = surf_hidden.new_zeros(surf_hidden.shape[0], 0, surf_hidden.shape[-1])
    else:
        # Surface absent — volume only
        vol_hidden = vol_h
        for blk in self.backbone.volume_blocks:
            vol_hidden = blk(vol_hidden, attn_mask=volume_mask)
        vol_hidden = _apply_token_mask(self.volume_norm(vol_hidden), volume_mask)
        surf_hidden = vol_hidden.new_zeros(vol_hidden.shape[0], 0, vol_hidden.shape[-1])
    # surface_preds and volume_preds output remain the same
    ...
else:
    # Original single-stream path (unchanged)
    ...
```

**Tip:** Restructure the forward to separate the encoding step from the backbone
step. Store `surf_h` and `vol_h` as local variables before the backbone call so
both modes can reuse them.

The key is: in the dual-stream mode, `surface_tokens` and `volume_tokens` are now
processed by `DualStreamTransformer`, which returns them as separate tensors rather
than a single concatenated tensor. The output heads (`surface_out`, `volume_out`)
remain unchanged.

### Step 6: Update `build_model` to pass new config flags

```python
def build_model(config: Config) -> SurfaceTransolver:
    return SurfaceTransolver(
        n_layers=config.model_layers,
        n_hidden=config.model_hidden_dim,
        dropout=config.model_dropout,
        n_head=config.model_heads,
        mlp_ratio=config.model_mlp_ratio,
        slice_num=config.model_slices,
        stochastic_depth_prob=config.stochastic_depth_prob,
        use_film=config.use_film,
        film_encoder_dim=config.film_encoder_dim,
        pos_max_wavelength=config.pos_max_wavelength,
        use_dual_stream=config.use_dual_stream,
        cross_attn_every=config.cross_attn_every,
    )
```

## Parameter Budget

With `--use-dual-stream`, the model has:
- 2× single-stream TransformerBlocks (one per domain) → ~2× backbone parameters
- Cross-attention bridges: each `CrossAttentionBlock` adds ~4 * D² parameters
  (2 MHA modules with D=512, 8 heads → ~2M params per bridge for 4 layers, 2 bridges)
- Total increase: ~1.5× baseline params (backbone dominates)

To maintain a comparable parameter count with baseline, reduce `--model-layers` from 4
to 3 when using `--use-dual-stream`. At 3 layers dual-stream, the total params should
be close to 4-layer single-stream.

Alternatively, keep 4 layers and let the model be larger — the VRAM budget (4×96GB)
is not the bottleneck at 512d.

## Training Configuration

**Arm A: Control (single-stream baseline, reproducibility check)**
```bash
  --epochs 10 \
  --lr 5e-4 \
  --weight-decay 1e-4 \
  --batch-size 4 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --pos-max-wavelength 1000 \
  --ema-decay 0.999 \
  --lr-warmup-steps 100 \
  --lr-warmup-start-lr 1e-6 \
  --clip-grad-norm 0.5 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
  --wandb-group dual-stream-cross-attn-yi \
  --wandb-name arm-a-single-stream-control \
  --seed 42
```

**Arm B: Dual-stream, 4 layers, cross-attn every 2 layers (2 bridges)**
```bash
  --epochs 10 \
  --lr 5e-4 \
  --weight-decay 1e-4 \
  --batch-size 4 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --use-dual-stream \
  --cross-attn-every 2 \
  --pos-max-wavelength 1000 \
  --ema-decay 0.999 \
  --lr-warmup-steps 100 \
  --lr-warmup-start-lr 1e-6 \
  --clip-grad-norm 0.5 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
  --wandb-group dual-stream-cross-attn-yi \
  --wandb-name arm-b-dual-stream-4l-xattn-every2 \
  --seed 42
```

**Arm C (optional, if time allows): Dual-stream with reduced depth to match params**
```bash
  --epochs 10 \
  --lr 5e-4 \
  --weight-decay 1e-4 \
  --batch-size 4 \
  --model-layers 3 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --use-dual-stream \
  --cross-attn-every 1 \
  --pos-max-wavelength 1000 \
  --ema-decay 0.999 \
  --lr-warmup-steps 100 \
  --lr-warmup-start-lr 1e-6 \
  --clip-grad-norm 0.5 \
  --wallshear-y-weight 2.0 \
  --wallshear-z-weight 2.0 \
  --wandb-group dual-stream-cross-attn-yi \
  --wandb-name arm-c-dual-stream-3l-xattn-every1 \
  --seed 42
```

Run Arm A and Arm B in parallel first. If both complete without divergence, add
Arm C to explore cross-attn frequency. Arm B is the primary test.

## Reporting

For each arm, per-epoch table:
- `val_abupt` (primary — must beat **7.546%** to merge)
- `wall_shear_y_rel_l2_pct` and `wall_shear_z_rel_l2_pct` (gap targets)
- `surface_pressure_rel_l2_pct`
- `volume_pressure_rel_l2_pct`
- W&B run ID

Also report:
- **Gate values**: After training, log the values of `surf_gate` and `vol_gate`
  for each cross-attention bridge. If they remain near zero, the gradients aren't
  flowing through the cross-stream path — check whether the `tanh()` activation
  provides enough gradient signal. If gates stay at 0, try initializing them at
  0.1 instead of 0.0.
- **Per-domain slice analysis**: If W&B gradient histograms show surface and volume
  blocks developing different gradient norms, that confirms domain specialization
  is happening.

If val_abupt is diverging or stuck above 20% by epoch 2, kill and report.

After finding the best val checkpoint, run `--eval-only` to get test metrics.

## What Success Looks Like

Arm B beats Arm A by ≥0.5pp on val_abupt, and specifically the
`wall_shear_y_rel_l2_pct` and `wall_shear_z_rel_l2_pct` metrics decrease more than
proportionately (relative to total val_abupt improvement). This would confirm that
domain separation + cross-attention communication is the right inductive bias for
this problem.

If the cross-attention gates open significantly (|tanh(gate)| > 0.1), it confirms
that cross-domain information is actively being used, which validates the hypothesis
that the shared-backbone architecture was creating a cross-contamination bottleneck.

## Suggested Follow-ups

If Arm B beats baseline:
1. Try a **directional cross-attention** variant: surface attends to volume but
   volume does NOT attend to surface (one-directional bridge). This tests whether
   the benefit comes from the surface branch getting global pressure context, or
   from both directions.
2. Combine with **learnable PE** (PR #420, `--learnable-pe`) to stack the gains.
3. Try deeper dual-stream with `--model-layers 6` — if the parameter count concern
   is resolved, deeper domain-specific stacks may improve further.

### DIFF
--- a/train.py
+++ b/train.py
@@ -235,41 +235,154 @@ class TransformerBlock(nn.Module):
         return x
 
 
-class CrossAttentionBlock(nn.Module):
-    """Lightweight cross-attention bridge between surface and volume streams.
+class CrossAttentionBridge(nn.Module):
+    """One-directional cross-stream attention via a register-token bottleneck.
 
-    Each stream uses standard multi-head attention to read from the other
-    stream's hidden states. Output is gated by a learnable `tanh(g)` scalar
-    initialised at 0 so cross-stream contribution starts at zero and the
-    gradient signal can gradually open the bridge.
+    Two-pass design (Perceiver / Set Transformer / DINOv3 registers):
+      Pass 1: K learned register tokens attend to source tokens
+              (registers=Q, source=K/V) -> registers summarise the source.
+      Pass 2: target tokens attend to the summarised registers
+              (target=Q, registers=K/V) -> per-token cross-stream delta.
+
+    Complexity is O((N_source + N_target) * K), avoiding O(N^2) over raw
+    65k-token streams. Output is gated by `tanh(gate)` initialised at 0, so
+    the bridge contributes nothing at init and gradient on the gate gradually
+    opens it.
     """
 
-    def __init__(self, hidden_dim: int, num_heads: int, dropout: float = 0.0):
+    def __init__(
+        self,
+        hidden_dim: int,
+        num_heads: int,
+        num_registers: int = 128,
+        dropout: float = 0.0,
+    ):
         super().__init__()
         self.hidden_dim = hidden_dim
         self.num_heads = num_heads
-        self.norm_surf_q = nn.LayerNorm(hidden_dim, eps=1e-6)
-        self.norm_vol_kv = nn.LayerNorm(hidden_dim, eps=1e-6)
-        self.surf_to_vol = nn.MultiheadAttention(
+        self.num_registers = int(num_registers)
+        # Learnable register tokens, shaped [1, K, D].
+        self.registers = nn.Parameter(
+            torch.randn(1, self.num_registers, hidden_dim) * 0.02
+        )
+        # Pass 1: registers (Q) attend to source (K/V).
+        self.norm_reg_q = nn.LayerNorm(hidden_dim, eps=1e-6)
+        self.norm_src_kv = nn.LayerNorm(hidden_dim, eps=1e-6)
+        self.attn_reg_from_src = nn.MultiheadAttention(
             embed_dim=hidden_dim,
             num_heads=num_heads,
             dropout=dropout,
             batch_first=True,
         )
-        self.norm_vol_q = nn.LayerNorm(hidden_dim, eps=1e-6)
-        self.norm_surf_kv = nn.LayerNorm(hidden_dim, eps=1e-6)
-        self.vol_to_surf = nn.MultiheadAttention(
+        # Pass 2: target (Q) attends to registers (K/V).
+        self.norm_tgt_q = nn.LayerNorm(hidden_dim, eps=1e-6)
+        self.norm_reg_kv = nn.LayerNorm(hidden_dim, eps=1e-6)
+        self.attn_tgt_from_reg = nn.MultiheadAttention(
             embed_dim=hidden_dim,
             num_heads=num_heads,
             dropout=dropout,
             batch_first=True,
         )
-        # Zero-init residual gates -> tanh(0)=0 -> cross-attn contributes 0 at init.
-        self.surf_gate = nn.Parameter(torch.zeros(1))
-        self.vol_gate = nn.Parameter(torch.zeros(1))
+        # Zero-init residual gate -> bridge contributes 0 at init.
+        self.gate = nn.Parameter(torch.zeros(1))
 
-    def _has_tokens(self, mask: torch.Tensor) -> bool:
-        return mask.shape[1] > 0 and bool(mask.any())
+    def forward(
+        self,
+        source: torch.Tensor,        # [B, N_src, D]
+        target: torch.Tensor,        # [B, N_tgt, D]
+        source_mask: torch.Tensor,   # [B, N_src] 1=valid
+        target_mask: torch.Tensor,   # [B, N_tgt] 1=valid
+    ) -> torch.Tensor:
+        batch_size = source.shape[0]
+        registers = self.registers.expand(batch_size, -1, -1).contiguous()
+
+        # Pass 1: registers attend to source tokens.
+        # nn.MultiheadAttention key_padding_mask: True = ignore.
+        src_key_pad = ~source_mask.bool()
+        # Guard rows where every source token is padding -- MHA returns NaN if
+        # a query attends to a fully-masked KV. Disable padding for those rows;
+        # the gate is zero-init so any noise is multiplied to ~0 anyway.
+        all_pad_rows = src_key_pad.all(dim=-1)
+        if bool(all_pad_rows.any()):
+            src_key_pad = src_key_pad.clone()
+            src_key_pad[all_pad_rows] = False
+
+        reg_q = self.norm_reg_q(registers)
+        src_kv = self.norm_src_kv(source)
+        reg_summary, _ = self.attn_reg_from_src(
+            reg_q, src_kv, src_kv,
+            key_padding_mask=src_key_pad,
+            need_weights=False,
+        )
+        registers = registers + reg_summary  # residual on registers
+
+        # Pass 2: target attends to registers (always valid -- no padding mask).
+        tgt_q = self.norm_tgt_q(target)
+        reg_kv = self.norm_reg_kv(registers)
+        tgt_delta, _ = self.attn_tgt_from_reg(
+            tgt_q, reg_kv, reg_kv,
+            need_weights=False,
+        )
+        target = target + self.gate.tanh() * tgt_delta
+        target = _apply_token_mask(target, target_mask)
+        return target
+
+
+class CrossAttentionBlock(nn.Module):
+    """Bi-directional register-token cross-attention bridge.
+
+    Holds two `CrossAttentionBridge` modules: one for surface<-volume and one
+    for volume<-surface. Each direction owns its own learned register tokens,
+    so the two streams' summaries do not have to share a representation space.
+    """
+
+    def __init__(
+        self,
+        hidden_dim: int,
+        num_heads: int,
+        num_registers: int = 128,
+        dropout: float = 0.0,
+    ):
+        super().__init__()
+        self.hidden_dim = hidden_dim
+        self.num_heads = num_heads
+        self.num_registers = int(num_registers)
+        # surface receives information from volume
+        self.vol_to_surf = CrossAttentionBridge(
+            hidden_dim=hidden_dim,
+            num_heads=num_heads,
+            num_registers=num_registers,
+            dropout=dropout,
+        )
+        # volume receives information from (post-update) surface
+        self.surf_to_vol = CrossAttentionBridge(
+            hidden_dim=hidden_dim,
+            num_heads=num_heads,
+            num_registers=num_registers,
+            dropout=dropout,
+        )
+
+    @property
+    def surf_gate(self) -> torch.Tensor:
+        return self.vol_to_surf.gate
+
+    @property
+    def vol_gate(self) -> torch.Tensor:
+        return self.surf_to_vol.gate
+
+    def _touch_all_params(self) -> torch.Tensor:
+        """DDP-safe no-op: 0-valued tensor with grad path through every parameter.
+
+        Needed because some DrivAerML batches have surface_views < volume_views
+        (or vice-versa), so a batch can arrive with N_surface=0 or N_volume=0.
+        In that case MHA cannot run; we skip the bridge but must still register
+        each parameter as "used" or DDP find_unused_parameters=False will error.
+        """
+        accum = None
+        for p in self.parameters():
+            term = p.sum()
+            accum = term if accum is None else accum + term
+        return accum * 0.0 if accum is not None else torch.zeros((), device=next(self.parameters()).device)
 
     def forward(
         self,
@@ -278,69 +391,26 @@ class CrossAttentionBlock(nn.Module):
         surf_mask: torch.Tensor,
         vol_mask: torch.Tensor,
     ) -> tuple[torch.Tensor, torch.Tensor]:
-        surf_has = self._has_tokens(surf_mask)
-        vol_has = self._has_tokens(vol_mask)
-
-        # Without keys/values on one side, attention is undefined — skip the
-        # bridge. We still touch all parameters via a 0-multiplied dummy pass
-        # so DDP / autograd see every gate even when this batch happens to be
-        # surface-only or volume-only.
-        if not (surf_has and vol_has):
-            zero = (
-                self.surf_gate.sum() * 0.0
-                + self.vol_gate.sum() * 0.0
-            )
+        # Empty-stream guard: if either modality has 0 tokens (e.g. a view-count
+        # mismatch in the DrivAerML loader), MHA cannot run on a 0-length axis.
+        if surf.shape[1] == 0 or vol.shape[1] == 0:
+            zero = self._touch_all_params()
             return surf + zero, vol + zero
 
-        # nn.MultiheadAttention key_padding_mask convention: True = ignore.
-        surf_key_pad = ~surf_mask.bool()
-        vol_key_pad = ~vol_mask.bool()
-        # Guard fully-padded query rows: nn.MultiheadAttention returns NaN when
-        # a query attends to a fully-masked key set. Disable padding for any
-        # row that has no valid keys (the corresponding query rows are
-        # themselves padding and will be re-zeroed by the post-call mask
-        # multiplication).
-        if bool(vol_key_pad.all(dim=-1).any()):
-            all_pad_rows = vol_key_pad.all(dim=-1)
-            vol_key_pad = vol_key_pad.clone()
-            vol_key_pad[all_pad_rows] = False
-        if bool(surf_key_pad.all(dim=-1).any()):
-            all_pad_rows = surf_key_pad.all(dim=-1)
-            surf_key_pad = surf_key_pad.clone()
-            surf_key_pad[all_pad_rows] = False
-
-        surf_q = self.norm_surf_q(surf)
-        vol_kv = self.norm_vol_kv(vol)
-        # nn.MultiheadAttention does not support bf16 autocast on all paths;
-        # cast to fp32 internally to avoid silent NaNs and let DropPath/EMA see
-        # finite values in the residual.
-        surf_q_f = surf_q.float()
-        vol_kv_f = vol_kv.float()
-        surf_delta, _ = self.surf_to_vol(
-            surf_q_f, vol_kv_f, vol_kv_f,
-            key_padding_mask=vol_key_pad,
-            need_weights=False,
+        # Surface pulls a summary of the volume stream.
+        surf = self.vol_to_surf(
+            source=vol,
+            target=surf,
+            source_mask=vol_mask,
+            target_mask=surf_mask,
         )
-        surf_delta = surf_delta.to(surf.dtype)
-        surf = surf + self.surf_gate.tanh() * surf_delta
-        surf = _apply_token_mask(surf, surf_mask)
-
-        vol_q = self.norm_vol_q(vol)
-        # use the *updated* surf as the KV source for vol_to_surf — this matches
-        # the deterministic ordering described in the PR (surface receives,
-        # then volume reads from the post-update surface).
-        surf_kv = self.norm_surf_kv(surf)
-        vol_q_f = vol_q.float()
-        surf_kv_f = surf_kv.float()
-        vol_delta, _ = self.vol_to_surf(
-            vol_q_f, surf_kv_f, surf_kv_f,
-            key_padding_mask=surf_key_pad,
-            need_weights=False,
+        # Volume pulls a summary of the (just-updated) surface stream.
+        vol = self.surf_to_vol(
+            source=surf,
+            target=vol,
+            source_mask=surf_mask,
+            target_mask=vol_mask,
         )
-        vol_delta = vol_delta.to(vol.dtype)
-        vol = vol + self.vol_gate.tanh() * vol_delta
-        vol = _apply_token_mask(vol, vol_mask)
-
         return surf, vol
 
 
@@ -458,6 +528,7 @@ class DualStreamTransformer(nn.Module):
         mlp_expansion_factor: int | float,
         num_slices: int,
         cross_attn_every: int = 2,
+        cross_attn_registers: int = 128,
         dropout: float = 0.0,
         stochastic_depth_prob: float = 0.0,
     ):
@@ -470,6 +541,7 @@ class DualStreamTransformer(nn.Module):
             ]
         self.depth = depth
         self.cross_attn_every = max(1, int(cross_attn_every))
+        self.cross_attn_registers = int(cross_attn_registers)
 
         self.surface_blocks = nn.ModuleList(
             [
@@ -503,6 +575,7 @@ class DualStreamTransformer(nn.Module):
                 CrossAttentionBlock(
                     hidden_dim=hidden_dim,
                     num_heads=num_heads,
+                    num_registers=self.cross_attn_registers,
                     dropout=dropout,
                 )
                 for _ in range(num_bridges)
@@ -551,6 +624,7 @@ class SurfaceTransolver(nn.Module):
         pos_max_wavelength: int = 1000,
         use_dual_stream: bool = False,
         cross_attn_every: int = 2,
+        cross_attn_registers: int = 128,
     ):
         super().__init__()
         self.space_dim = space_dim
@@ -565,6 +639,7 @@ class SurfaceTransolver(nn.Module):
         self.pos_max_wavelength = pos_max_wavelength
         self.use_dual_stream = bool(use_dual_stream)
         self.cross_attn_every = int(cross_attn_every)
+        self.cross_attn_registers = int(cross_attn_registers)
 
         self.pos_embed = ContinuousSincosEmbed(
             hidden_dim=n_hidden,
@@ -599,6 +674,7 @@ class SurfaceTransolver(nn.Module):
                 mlp_expansion_factor=mlp_ratio,
                 num_slices=slice_num,
                 cross_attn_every=self.cross_attn_every,
+                cross_attn_registers=self.cross_attn_registers,
                 dropout=dropout,
                 stochastic_depth_prob=stochastic_depth_prob,
             )
@@ -859,6 +935,7 @@ class Config:
     pos_max_wavelength: int = 1000
     use_dual_stream: bool = False
     cross_attn_every: int = 2
+    cross_attn_registers: int = 128
     amp_mode: str = "bf16"
     num_workers: int = -1
     pin_memory: bool = True
@@ -1044,6 +1121,7 @@ def build_model(config: Config) -> SurfaceTransolver:
         pos_max_wavelength=config.pos_max_wavelength,
         use_dual_stream=config.use_dual_stream,
         cross_attn_every=config.cross_attn_every,
+        cross_attn_registers=config.cross_attn_registers,
     )
 
 
@@ -1400,6 +1478,35 @@ def slope_source_metrics(metrics: dict[str, object]) -> dict[str, float]:
     )
     numeric = _numeric_metric_items(metrics)
     return {key: value for key, value in numeric.items() if any(word in key for word in keywords)}
+
+
+def collect_cross_attn_gate_metrics(model: nn.Module) -> dict[str, float]:
+    """Per-bridge tanh(gate) magnitudes — sanity check that bridges open over time."""
+
+    base_model = getattr(model, "_orig_mod", model)
+    backbone = getattr(base_model, "backbone", None)
+    bridges = getattr(backbone, "cross_attn_bridges", None)
+    if bridges is None:
+        return {}
+    metrics: dict[str, float] = {}
+    abs_values: list[float] = []
+    for idx, bridge in enumerate(bridges):
+        for name, mod in (
+            ("vol_to_surf", getattr(bridge, "vol_to_surf", None)),
+            ("surf_to_vol", getattr(bridge, "surf_to_vol", None)),
+        ):
+            if mod is None:
+                continue
+            gate = getattr(mod, "gate", None)
+            if gate is None:
+                continue
+            value = float(gate.detach().tanh().item())
+            metrics[f"train/cross_attn/bridge_{idx}/{name}/tanh_gate"] = value
+            abs_values.append(abs(value))
+    if abs_values:
+        metrics["train/cross_attn/tanh_gate_abs_mean"] = sum(abs_values) / len(abs_values)
+        metrics["train/cross_attn/tanh_gate_abs_max"] = max(abs_values)
+    return metrics
 
 
 class MetricSlopeTracker:
@@ -2324,6 +2431,11 @@ def main(argv: Iterable[str] | None = None) -> None:
             train_loss_sum += float(loss.detach().cpu().item())
             n_batches += 1
             global_step += 1
+            cross_attn_gate_metrics = (
+                collect_cross_attn_gate_metrics(model)
+                if config.use_dual_stream and should_log_weights
+                else {}
+            )
             train_log: dict[str, object] = {
                 "train/loss": float(loss.detach().cpu().item()),
                 "train/surface_loss": batch_loss_metrics["surface_loss"],
@@ -2337,6 +2449,7 @@ def main(argv: Iterable[str] | None = None) -> None:
                 "global_step": global_step,
                 **gradient_metrics,
                 **weight_metrics,
+                **cross_attn_gate_metrics,
             }
             if "wallshear_pred_normal_rms" in batch_loss_metrics:
                 train_log["train/wallshear_pred_normal_rms"] = batch_loss_metrics[
--- a/train.py
+++ b/train.py
@@ -806,11 +806,11 @@ class SurfaceTransolver(nn.Module):
             surface_tokens = surf_h.shape[1] if surf_h is not None else 0
             volume_tokens = vol_h.shape[1] if vol_h is not None else 0
             surface_hidden = (
-                hidden_norm[:, :surface_tokens] if surface_tokens > 0 else None
+                hidden_norm[:, :surface_tokens] if surf_h is not None else None
             )
             volume_hidden = (
                 hidden_norm[:, surface_tokens : surface_tokens + volume_tokens]
-                if volume_tokens > 0
+                if vol_h is not None
                 else None
             )
 
--- a/train.py
+++ b/train.py
@@ -233,6 +233,115 @@ class TransformerBlock(nn.Module):
         return x
 
 
+class CrossAttentionBlock(nn.Module):
+    """Lightweight cross-attention bridge between surface and volume streams.
+
+    Each stream uses standard multi-head attention to read from the other
+    stream's hidden states. Output is gated by a learnable `tanh(g)` scalar
+    initialised at 0 so cross-stream contribution starts at zero and the
+    gradient signal can gradually open the bridge.
+    """
+
+    def __init__(self, hidden_dim: int, num_heads: int, dropout: float = 0.0):
+        super().__init__()
+        self.hidden_dim = hidden_dim
+        self.num_heads = num_heads
+        self.norm_surf_q = nn.LayerNorm(hidden_dim, eps=1e-6)
+        self.norm_vol_kv = nn.LayerNorm(hidden_dim, eps=1e-6)
+        self.surf_to_vol = nn.MultiheadAttention(
+            embed_dim=hidden_dim,
+            num_heads=num_heads,
+            dropout=dropout,
+            batch_first=True,
+        )
+        self.norm_vol_q = nn.LayerNorm(hidden_dim, eps=1e-6)
+        self.norm_surf_kv = nn.LayerNorm(hidden_dim, eps=1e-6)
+        self.vol_to_surf = nn.MultiheadAttention(
+            embed_dim=hidden_dim,
+            num_heads=num_heads,
+            dropout=dropout,
+            batch_first=True,
+        )
+        # Zero-init residual gates -> tanh(0)=0 -> cross-attn contributes 0 at init.
+        self.surf_gate = nn.Parameter(torch.zeros(1))
+        self.vol_gate = nn.Parameter(torch.zeros(1))
+
+    def _has_tokens(self, mask: torch.Tensor) -> bool:
+        return mask.shape[1] > 0 and bool(mask.any())
+
+    def forward(
+        self,
+        surf: torch.Tensor,
+        vol: torch.Tensor,
+        surf_mask: torch.Tensor,
+        vol_mask: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        surf_has = self._has_tokens(surf_mask)
+        vol_has = self._has_tokens(vol_mask)
+
+        # Without keys/values on one side, attention is undefined — skip the
+        # bridge. We still touch all parameters via a 0-multiplied dummy pass
+        # so DDP / autograd see every gate even when this batch happens to be
+        # surface-only or volume-only.
+        if not (surf_has and vol_has):
+            zero = (
+                self.surf_gate.sum() * 0.0
+                + self.vol_gate.sum() * 0.0
+            )
+            return surf + zero, vol + zero
+
+        # nn.MultiheadAttention key_padding_mask convention: True = ignore.
+        surf_key_pad = ~surf_mask.bool()
+        vol_key_pad = ~vol_mask.bool()
+        # Guard fully-padded query rows: nn.MultiheadAttention returns NaN when
+        # a query attends to a fully-masked key set. Disable padding for any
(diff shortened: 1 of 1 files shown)


morganmcg1:
STUDENT kohaku:

## Status: Round 2 launch (r2)

Earlier session launched arms with `--train-surface-points 8192` which inflated batches/epoch 8x (173k vs 22k) and would have run ~4h per epoch — I killed those (W&B `odbtk7a8`, `907a6tt9`, `m9j7nzyz`) and relaunched.

**r2 launch (single-GPU per arm, in parallel):**
- Arm A (control, single-stream 4L): GPU 0, W&B `0olvujja`
- Arm B (dual-stream 4L, xattn-every=2): GPU 1, W&B `cbozsgmj`
- Arm C (dual-stream 3L, xattn-every=1): GPU 2, W&B `vh7m9qf3`

Group: `dual-stream-cross-attn-yi-r2`

**Settings:** `--epochs 10 --batch-size 4 --lr 5e-4 --weight-decay 1e-4 --train-surface-points 65536 --eval-surface-points 65536 --train-volume-points 65536 --eval-volume-points 65536 --model-hidden-dim 512 --model-heads 8 --model-slices 128 --pos-max-wavelength 1000 --ema-decay 0.999 --lr-warmup-steps 100 --lr-warmup-start-lr 1e-6 --clip-grad-norm 0.5 --wallshear-y-weight 2.0 --wallshear-z-weight 2.0 --validation-every 2 --no-compile-model --seed 42`

**Param counts (visible at startup):**
- Arm A: 12.70M (single-stream baseline)
- Arm B: 28.55M (dual-stream 4L)
- Arm C: 24.84M (dual-stream 3L)

Note: PR estimate was 1.5x; actual is 2.0–2.25x because TransolverAttention's per-block params dominate (the slice budget doubles in dual-stream).

**Caveat about merge bar:** This branch was forked from `yi` before PR #355 (DDP fix). Single-GPU bs=4 is a 4x effective-batch reduction vs the DDP-trained baseline behind val_abupt=7.546% (PR #311). Arm A vs Arm B comparison will be apples-to-apples (both single-GPU, same seed/config), but absolute val_abupt will not be directly comparable to the merge bar; if dual-stream wins clearly, an obvious follow-up is to rebase + re-run under DDP for the absolute claim.

Will report when arms complete or hit timeout (`SENPAI_TIMEOUT_MINUTES=320`).

morganmcg1:
STUDENT kohaku:

## Status update: Round 2 — arms B and C killed (compute infeasible at N=65536)

### What I observed

Looking at the logged iter speeds after ~7 minutes of training:

| Arm | Config | Iter speed | ETA per epoch | Epochs in 320min |
|-----|--------|-----------|---------------|------------------|
| A (single-stream control) | 4L, no x-attn | **1.50 it/s** | ~4 h | ~0.95 |
| B (dual-stream 4L, xattn-every=2) | 2 bridges | **0.20 it/s** (5.05 s/it) | ~30 h | **~0.18** |
| C (dual-stream 3L, xattn-every=1) | 3 bridges | **0.14 it/s** (7.09 s/it) | ~42 h | **~0.12** |

Arms B and C cannot reach a single validation step before the timeout, so they would produce zero usable signal. I killed PIDs 65110 (B) and 65172 (C). Arm A (PID 65049) is still running and will produce a single-GPU control reference (≈1 full epoch).

### Why dual-stream is so much slower per iter

The `CrossAttentionBlock` uses `nn.MultiheadAttention` over the **full token sequence** of N=65536 surface and N=65536 volume tokens. That's quadratic O(N²) in compute (≈268B flops per head per direction), and PyTorch's `MultiheadAttention` is being explicitly cast to FP32 on the SDPA path (silent-NaN guard from when the module was added), so we don't even get the BF16 / Flash speedup. Self-attention is Transolver-sliced (linear in N) but **cross-attention dominates** the iter time once you stack 2-3 bridges.

The PR spec said "Cross-attention is lightweight (standard multi-head attention, not Transolver-style slicing)." That premise holds at small N but breaks at 65536 — **compute scales linearly with view size for the whole epoch** (per-view cost grows N², per-epoch view count shrinks 1/N, so total work ∝ N), so we can't fix this by changing per-view points alone.

### Bug found while reading: branch is missing PR #355 (DDP) + #317 (Huber)

This branch was forked from `yi` at commit `776afd8`, before:
- `224e8b4` PR #355 (emma DDP infrastructure fix — merged 17:05 today)
- `85992b0` PR #317 (violet Huber loss — merged today, no metric bar change)

The earlier session's caveat about "single-GPU bs=4 vs DDP-32 baseline" was correct: this branch literally had no `init_process_group` / `DistributedSampler` wiring. I will merge `yi` into this branch to pick up DDP (low-risk, this is exactly what PR #355 added to `yi`) and re-launch dual-stream under 4-GPU DDP.

### Proposed fix to cross-attention (will deviate slightly from spec)

To make the dual-stream architecture computationally tractable at N=65536, I plan to:

1. **Drop the FP32 cast** in `CrossAttentionBlock` — let SDPA use BF16 / Flash when available. ~2× speedup, no semantic change.
2. **Pool the KV side through a small set of learned register tokens** before cross-attention (`K=128`, matching `--model-slices`). This makes each direction O(N·K) instead of O(N²) — strictly more efficient than `nn.MultiheadAttention` on raw tokens, and a well-known design (Perceiver / Set Transformer / DINOv3 registers). The intent ("dense, global cross-stream communication, not Transolver-slicing inside the bridge") is preserved.
3. Cherry-pick / merge DDP from `yi` so we can run on 4 GPUs.

This deviates slightly from the spec ("standard MHA"), but I think it's the only way to make the hypothesis testable in a 320-min budget on 4 H100s. If you'd prefer I don't make this change (e.g. you want to keep raw MHA at smaller N to test the original formulation), please let me know in a comment.

### Next steps

I'll proceed with: arm A continues as the single-GPU single-stream control reference; merge `yi`; rewrite `CrossAttentionBlock` with register-token bottleneck + drop the FP32 cast; re-launch a single dual-stream variant under 4-GPU DDP. Will report metrics when they land.

morganmcg1:
ADVISOR: The register-token bottleneck approach is the right call. Proceed with it.

**Verdict on your diagnosis:** Your root cause is correct — O(N²) cross-attention over 65536 tokens is computationally dead-on-arrival at N=65536. The Transolver's O(N·S) slice-based attention is specifically designed to avoid this, and you've identified that cross-attention is the bottleneck exactly as expected when you add global cross-stream communication naively.

**Approved modifications:**

1. **Register-token bottleneck (K=128):** Yes. Map both stream's token sequences to K=128 learned register tokens before cross-attention. This makes each cross-attention direction O(N·K) — equivalent to the Transolver's sliced complexity. Use learnable registers (nn.Parameter shaped [1, K, hidden_dim]) as the KV side of the cross-attention, and let the query be the source stream's tokens. This is standard and well-motivated (Perceiver IO, DINOv3 registers, etc.).

2. **Drop FP32 cast on SDPA path:** Yes. The silent-NaN guard was overly conservative. BF16 SDPA (Flash) is stable for this problem.

3. **Merge yi DDP + Huber into this branch before re-launch:** Yes. DDP is needed to make the comparison fair vs the CTRL arm.

**Implementation note:** For the register-token architecture, a minimal clean implementation:
```python
class CrossAttentionBridge(nn.Module):
    def __init__(self, hidden_dim, num_heads, num_registers=128):
        super().__init__()
        self.registers = nn.Parameter(torch.randn(1, num_registers, hidden_dim) * 0.02)
        self.to_kv = nn.MultiheadAttention(hidden_dim, num_heads, batch_first=True)  # source tokens → KV on registers
        self.to_q = nn.MultiheadAttention(hidden_dim, num_heads, batch_first=True)  # target tokens query registers
```
Two-pass:
- Pass 1: registers attend to source tokens (registers=Q, source=K/V) → updated registers
- Pass 2: target tokens attend to registers (target=Q, registers=K/V) → output for target stream

This is O(K·N) per direction — efficient and clean.

**Run spec for the revised dual-stream:** 4-GPU DDP, Lion optimizer (matching the yi SOTA config), `--epochs 10` under the 360-min budget. Run Arm A (single-stream control) and Arm B (dual-stream with register-token cross-attention) in parallel on 2+2 GPUs, or sequentially on 4+4 if the host has 8 GPUs. Report val_abupt and the per-channel decomposition when both arms complete.

morganmcg1:
STUDENT kohaku:

## Bug fix during launch

Hit a DDP unused-params crash on Arm A (single-stream control) at step 6:

```
RuntimeError: Expected to have finished reduction in the prior iteration before starting a new one.
Parameter indices which did not receive grad for rank 0: 88 89
```

Indices 88 and 89 are `surface_out.project.weight/bias`. Root cause: when DrivAerML view-count asymmetry produces a batch where every sample has 0 surface tokens (or 0 volume tokens), `pad_collate` makes a `(B, 0, F)` tensor — `surface_x is not None` but `surface_x.shape[1] == 0`. The single-stream forward path used `surface_tokens > 0` to decide whether to call `surface_out`, leaving `surface_out.{weight,bias}` unused for that step → DDP crash with `find_unused_parameters=False`.

**Fix** (commit `a0420b0`): condition on `surf_h is not None` (i.e. whether the stream input was provided), not on token count. `surface_out` is now always called when `surface_x` was passed, even with a 0-length input — `nn.Linear` on an empty tensor still registers the param in the autograd graph (gradient becomes a tensor of zeros), satisfying DDP's used-param check. Symmetric fix for `volume_out`.

Verified across 6 cases (`{single,dual}-stream × {both, empty-surface, empty-volume}` batches): all produce 0 unused params after the change.

The dual-stream path was already correct (it conditions on `surf_out is not None` from the backbone, which is set whenever `surface_x` was passed), so Arm B was unaffected and is still running from the original launch.

This bug appears to be present in `yi/train.py` HEAD as well — the same `surface_tokens > 0` guard exists there. It only manifests when DDP is enabled AND a batch happens to land where every sample has 0 surface (or 0 volume) tokens, which is rare but reproducible. Worth back-porting to yi if you agree.

morganmcg1:
ADVISOR: Excellent diagnostic work and clean fix.

**On the fix itself:** Good — conditioning on `surf_h is not None` (presence of the stream input) rather than `surface_tokens > 0` (post-pad token count) is exactly right. `nn.Linear` on a `(B, 0, F)` tensor produces a `(B, 0, out_F)` output with the params registered in the autograd graph, so DDP's `find_unused_parameters=False` path stays valid. The 6-case verification matrix is the right level of rigor for a DDP correctness fix.

**On the yi back-port:** Yes, please back-port to `yi/train.py` HEAD. This bug is latent in the baseline — it just hasn't fired because the DDP runs to date have been seed-lucky on view-count distribution. We've had enough DDP runs in flight (PR #355 onward) that the chance of a bad batch pulling a `surface_tokens == 0` collate at any step is non-trivial; eliminating the latent crash improves DDP robustness across all yi students. Cherry-pick `a0420b0` (or send the patch to me directly here) and I'll merge it to yi as a hotfix outside your experiment PR.

**On the experiment status:** It's been ~3.5h since your bug-fix comment. Please post an interim:
1. After commit `a0420b0`, did Arm A relaunch cleanly under DDP? If yes, what's the current epoch / val_abupt?
2. Arm B (dual-stream xattn-every=2 with register-token bottleneck K=128) — has that landed yet? If still on the original O(N²) MultiheadAttention path, it's still computationally infeasible at N=65536 and won't produce signal in budget.
3. Are you parallel single-GPU per arm, or DDP per arm? If single-GPU per arm, drop the `find_unused_parameters` concern — it only bites under DDP.

Two-line status comment is fine; we just need to know whether arms are alive and where they are in epoch terms.

morganmcg1:
STUDENT kohaku:

## Interim status — arms B (dual-stream xattn-r3) and A (control r4) are both alive under DDP

**Both arms running in parallel, 2-GPU DDP each, Lion optimizer, register-token bridge K=128.**

| Arm | W&B | Mode | Started | Epoch 1 val_abupt | Current pos | Iter speed |
|-----|-----|------|---------|-------------------|-------------|-----------|
| B (dual-stream, K=128 register bridge) | `cbozsgmj` (run cah98laj) | DDP×2, lion, bs=4 | 19:55 | **13.2561%** | epoch 2 / 12% | 1.07 s/it |
| A (single-stream control)              | `0olvujja` (run f6s84dp8) | DDP×2, lion, bs=4 | 20:07 | 13.8830% | epoch 2 / 72% | 1.49 it/s |

(W&B "name vs id" mapping: I tagged the runs with my name strings; the run IDs shown above are the full storage keys.)

### Key observations

- **Register-bottleneck made dual-stream tractable.** Per-iter cost ratio dropped from ~10× (raw MultiheadAttention at N=65536) to ~1.6× — exactly the O(N·K) win the bottleneck buys.
- **Arm B is ~0.6pp ahead of Arm A at epoch 1.** This is a single-epoch read with bs=8 effective, so noise floor is wide; need at least 2-3 epochs each for a real signal.
- **Both will hit the 360-min budget before 10 epochs.** Arm A pace ~7.4k s/epoch → ~3 epochs at timeout. Arm B pace ~11.9k s/epoch → ~2 epochs at timeout. The val_abupt curves are still descending steeply at epoch 1, so the comparison must be at matched-epoch.
- **No DDP crashes.** Bug fix `a0420b0` was committed at 20:06 — Arm A r4 launched at 20:07 includes it; Arm B r3 launched at 19:55 was lucky-batched until now.

### Re: yi back-port

Will cut a separate one-line patch comment with `a0420b0` for you to cherry-pick (next message after this one).

### Next

Arms will continue under DDP until the 360-min budget; will post final per-channel decomposition (surface_p / wall_shear_y / wall_shear_z / volume_p) at matched epoch when both finish. Best matched comparison will be at **epoch 2** (Arm A finishes epoch 2 in ~30 min; Arm B in ~3h, so likely the comparable cutoff for the report).

morganmcg1:
STUDENT kohaku:

## Correction on yi back-port — the bug does NOT exist in yi/train.py

When you said "yes, please back-port to yi" you were taking my word that the same `surface_tokens > 0` guard was on yi HEAD. I went to extract the patch and found I was wrong — the bug is **specific to my branch**, introduced by my dual-stream restructuring (commit `5391d47`).

### What I checked

`yi/train.py` HEAD (`origin/yi`):

```python
cursor = 0
surface_hidden = hidden_norm[:, cursor : cursor + surface_tokens]   # always slices, may be empty
cursor += surface_tokens
volume_hidden = hidden_norm[:, cursor : cursor + volume_tokens]

if surface_x is not None:
    surface_preds = self.surface_out(surface_hidden) * surface_mask.unsqueeze(-1)
```

The guard around `self.surface_out(...)` is `surface_x is not None`, NOT `surface_tokens > 0`. When `surface_x` is a `(B, 0, F)` tensor, `surface_x is not None` is True, so `surface_out` is called on a `(B, 0, D)` slice → params are still registered in the autograd graph. No DDP crash.

`grep "surface_tokens > 0" origin/yi:train.py` returns nothing.

### Why I introduced it

In my dual-stream restructuring I split the forward into a single-stream branch and a dual-stream branch, and along the way reformulated `surface_hidden = hidden_norm[:, :surface_tokens] if surface_tokens > 0 else None`. That `else None` is what propagates to the `if surface_hidden is not None:` guard on the next line, which is what stranded `surface_out` and `volume_out` from the autograd graph when token counts hit zero.

`a0420b0` essentially reverts to the yi pattern (use input-presence rather than token-count) for the single-stream code path on my branch, fixing the regression I introduced.

### Implication

**No back-port needed.** Sorry for the false alarm — I should have diffed against `origin/yi` before claiming the bug was latent there. The fix lives only on my branch and is needed to make the merged-yi single-stream control arm safe under DDP.

Will continue to monitor the running arms and report at matched-epoch when both finish.

morganmcg1:
STUDENT kohaku:

## Results — Round 3 (DDP×2 each, lion, register-token bridge K=128)

### TL;DR — negative result on val_abupt under matched wall-clock

Dual-stream + register-bridge (Arm B) is **slightly better at matched-epoch but worse at matched-wall-clock**, because the cross-attn bridge adds ~1.6× per-iter cost and the gates barely opened (|tanh|≤0.055, well below the 0.1 "in use" bar). Under the 265-min budget Arm A reached partial epoch 3 at val_abupt=9.37%, while Arm B only reached partial epoch 2 at 11.09%.

### Final metrics

| Metric | Arm A (single-stream, 4L) | Arm B (dual-stream + xattn-every=2, K=128) | Δ vs A |
|---|---|---|---|
| W&B run id | `f6s84dp8` | `cah98laj` | — |
| W&B run name | `kohaku/arm-a-single-stream-control-r4` | `kohaku/arm-b-dual-stream-register-bridge-r3` | — |
| Param count | 12.70M | 28.55M | +125% |
| Iter speed | ~1.49 it/s | ~1.07 s/it (≈1.6× slower) | — |
| Time/epoch | ~123 min | ~198 min | +61% |
| Total runtime | 270.6 min (3 partial epochs) | 272.9 min (2 partial epochs) | — |
| Peak memory/GPU | 67.6 GB | 77.8 GB | +10.2 GB |
| **Best val_abupt_axis_rel_l2_pct** | **9.3729%** (partial ep3) | 11.0890% (partial ep2) | **−1.72pp (A wins)** |
| **test_abupt_axis_rel_l2_pct** | **10.5223%** | 11.9667% | **−1.44pp (A wins)** |
| test_surface_pressure_rel_l2_pct | 5.7558% | 6.7145% | −0.96pp |
| test_wall_shear_rel_l2_pct | 10.3624% | 11.8528% | −1.49pp |
| test_wall_shear_y_rel_l2_pct | 12.6322% | 14.5024% | −1.87pp |
| test_wall_shear_z_rel_l2_pct | 12.5602% | 13.9215% | −1.36pp |
| test_volume_pressure_rel_l2_pct | 12.7569% | 14.3930% | −1.64pp |

**Reference: PR #311 STRING-sep PE merge bar val_abupt = 7.546% (test 8.771%)** — neither arm reaches this, but see Caveat below; this is expected since both arms only got 2-3 partial epochs of a planned 10.

### Per-epoch val curves

| Epoch | Arm A val_abupt | Arm A surf_p | Arm A ws | Arm A ws_y | Arm A ws_z | Arm A vol_p |
|---|---|---|---|---|---|---|
| 1 (full) | 13.883 | 9.253 | 15.425 | 18.338 | 19.444 | 9.050 |
| 2 (full) | 9.424 | 6.107 | 10.563 | 12.917 | 13.184 | 5.911 |
| 3 (partial) | **9.373** | 6.039 | 10.488 | **12.797** | **13.075** | 5.997 |

| Epoch | Arm B val_abupt | Arm B surf_p | Arm B ws | Arm B ws_y | Arm B ws_z | Arm B vol_p |
|---|---|---|---|---|---|---|
| 1 (full) | 13.256 | 8.739 | 14.355 | 17.173 | 17.609 | 10.255 |
| 2 (partial) | 11.089 | 7.069 | 12.072 | 14.791 | 14.700 | 8.497 |

**Matched-epoch read (only ep1 fully completed on both):** Arm B is 0.6pp ahead on val_abupt and ~1pp ahead on surface_p / wall_shear_y/z, but 1.2pp **behind** on volume_p. Promising direction at matched-epoch but compute infeasible at the configuration tested.

### Cross-attention gate values (Arm B, end of training)

| Bridge | direction | tanh(gate) |
|---|---|---|
| 0 | surf→vol | +0.0101 |
| 0 | vol→surf | +0.0472 |
| 1 | surf→vol | −0.0553 |
| 1 | vol→surf | +0.0306 |
| **all** | mean abs | **0.0358** |
| **all** | max abs | **0.0553** |

**All four gates stayed under the |tanh|=0.1 "in use" threshold the PR called out.** The cross-stream communication pathway was barely contributing — meaning the dual-stream architecture is paying the full ~2× param + ~1.6× per-iter compute cost while running essentially as two parallel single-stream models with a near-zero residual cross-channel.

### Exact commands

**Arm A (control, single-stream 4L):**
```bash
cd target/ && torchrun --standalone --nproc_per_node=2 train.py \
  --epochs 10 --optimizer lion --lr 1e-4 --weight-decay 5e-4 --no-compile-model \
  --pos-max-wavelength 1000 --ema-decay 0.999 \
  --lr-warmup-epochs 1 --clip-grad-norm 0.5 \
  --no-log-gradient-histograms --no-log-weight-histograms \
  --weight-log-every 50 --gradient-log-every 50 --agent kohaku \
  --wandb-group dual-stream-cross-attn-yi-r3 \
  --wandb-name kohaku/arm-a-single-stream-control-r4 --seed 42
```

**Arm B (dual-stream, register-bridge K=128):**
```bash
cd target/ && torchrun --standalone --nproc_per_node=2 train.py \
  --epochs 10 --optimizer lion --lr 1e-4 --weight-decay 5e-4 --no-compile-model \
  --use-dual-stream --cross-attn-every 2 --cross-attn-registers 128 \
  --pos-max-wavelength 1000 --ema-decay 0.999 \
  --lr-warmup-epochs 1 --clip-grad-norm 0.5 \
  --no-log-gradient-histograms --no-log-weight-histograms \
  --weight-log-every 50 --gradient-log-every 50 --agent kohaku \
  --wandb-group dual-stream-cross-attn-yi-r3 \
  --wandb-name kohaku/arm-b-dual-stream-register-bridge-r3 --seed 42
```

### Caveats

1. **Branch was forked from yi pre-PR #355 (DDP fix).** I merged origin/yi to pick up DDP infra (commit `6a632c4`) before launching r3, so both arms ran with the same DDP wiring, but the merge bar (val_abupt=7.546%) on PR #311 was achieved on a different DDP-32 effective batch (8 GPUs × bs=4) and a full 10 epochs. Both arms here ran DDP×2 (4 GPUs total across the two arms, not 8 per arm) at bs=8 effective on 2-3 partial epochs — so neither arm is directly comparable to the merge bar in absolute terms. The **A vs B comparison is apples-to-apples** (same hardware split, same hyperparameters, same seed).

2. **Compute budget cut both arms short.** Arm A finished epoch 3 only ~75% through (22443/30k steps). Arm B finished epoch 2 only ~64% through (14119/22k steps). Both val curves were still descending steeply at the cutoff. A 10-epoch run on 8-GPU DDP would test the dual-stream hypothesis at the merge-bar regime — but the iter-cost gap between A and B would persist proportionally.

3. **Cross-attn gates barely opened.** With learning rate 1e-4 (lion) and zero-init `surf_gate`/`vol_gate` (PR design), the gradient signal through `tanh(gate) * delta` is intrinsically bounded by the gate magnitude, which then stays small in a tight self-reinforcing loop. The PR explicitly anticipated this and suggested initializing gates at 0.1 if they stayed near zero — that's a clean follow-up.

### What happened

- The **inductive bias of dual-stream is not obviously wrong.** At matched epoch 1, Arm B is ahead on every surface-channel metric (surf_p, ws, ws_y, ws_z) by ~0.6-1.8pp. Volume_p is the only loser at ep1 (10.25 vs 9.05), consistent with the volume stream having less data per iter than in the single-stream setup. This is the expected directional signal the hypothesis predicts.
- But the **cost is too high in this configuration.** Doubling the backbone (12.7M → 28.6M params) and adding 2 register-bridge cross-attentions adds ~60% wall-clock per epoch — too steep when the cross-attn gates aren't even opening. The model is learning to ignore the cross-stream pathway, which means we're paying the full architectural cost for the slight matched-epoch advantage that comes from doubling the capacity, not from the cross-stream coupling itself.
- **Net: under matched compute, single-stream wins clearly** (9.37% vs 11.09% val, 10.52% vs 11.97% test).

### Suggested follow-ups

1. **Initialize cross-attn gates at 0.1 instead of 0.0.** PR #421 explicitly suggested this; the gate-stayed-near-zero outcome confirms it should be tried. With gates non-trivially open from step 0, the cross-stream channel actually contributes to gradients and the hypothesis gets a cleaner test. Cost: one parameter change, no architecture change.
2. **Test dual-stream at matched-params, not matched-depth.** Reduce `--model-layers 4 → 2` for the dual-stream arm so total params (and per-iter cost) match Arm A. This isolates the architectural contribution from the param-count contribution. If 2-layer dual-stream still beats 4-layer single-stream at matched epoch, the cross-stream inductive bias is the real signal.
3. **Skip dual-stream, keep the wallshear-y/z reweighting infrastructure that was already in baseline.** The "surface tokens get diluted by volume tokens" diagnosis from the PR may be wrong — the 2.5–2.9× wallshear_y/z gap vs AB-UPT may simply be a learning-rate / loss-balance issue, not an architectural one. Worth ablating `--wallshear-y/z-weight` at higher values (e.g. 4–8) on the single-stream baseline before reaching for architectural changes.
4. **If pursuing dual-stream, run on 8-GPU DDP so the per-iter cost can be amortized over more samples per step.** At 8-way DDP × bs=4 = 32 effective batch (matching the PR #311 merge bar regime), 10 full epochs of Arm B should fit under SENPAI_TIMEOUT_MINUTES, which gives a fair shot at the absolute merge bar.

### W&B links

- Arm A: https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/f6s84dp8
- Arm B: https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/cah98laj
- Group: `dual-stream-cross-attn-yi-r3`

morganmcg1:
ADVISOR:

Kohaku — rigorous experiment design and thorough analysis. Thank you for the detailed results.

## Verdict: Hypothesis falsified in this configuration — close PR

The dual-stream + register-bridge hypothesis is **falsified under the tested configuration**. The evidence is clear:

1. **Cross-attn gates never opened** (max |tanh|=0.055, mean 0.036). The model learned to treat the cross-stream pathway as dead weight. You're paying 125% more parameters and 61% more wall-clock time for a near-zero residual cross-channel.
2. **Under matched wall-clock, single-stream wins by 1.72pp val_abupt** (9.37% vs 11.09%). The ep1 matched-epoch advantage for Arm B (~0.6pp) is entirely attributable to the 2× parameter count, not the cross-stream inductive bias.
3. **The architecture cost is too high to justify** given the gates are not activating. Even with gate-init=0.1, you would need the gates to reach ~0.3–0.5 to recover the per-iter compute penalty.

The core diagnosis from the PR — that surface tokens get diluted by volume tokens — may be addressable with a much simpler lever. Your suggestion #3 is the right next step: **ablate  at higher values on the single-stream baseline.** The 2.5–2.9× wallshear gap vs AB-UPT is a loss-balance issue that doesn't require architectural surgery.

## What to pick up next

Your suggestion #3 is your next assignment: test  and  at values 4.0 and 8.0 on the current yi baseline (4L/512d, Lion, lr=1e-4, wd=1e-4, clip=0.5). This is a clean one-variable sweep with high expected impact. Watch for: whether boosted wallshear weights depress surf_p/vol_p (competing loss terms), and whether val_abupt improves as the model allocates capacity away from volume.

## DDP bug fix

The  fix in commit `a0420b0` is clean — good catch. Confirmed not needed on yi HEAD.

Closing this PR. Excellent experimental methodology throughout — the negative result is a real contribution and rules out a significant architectural direction efficiently.

morganmcg1:
ADVISOR: Closing PR #421 because Closing: dual-stream + register-bridge hypothesis falsified — cross-attn gates never opened (max |tanh|=0.055), single-stream wins by 1.72pp val_abupt under matched wall-clock. See advisor verdict in comments above. Next step for kohaku: wallshear reweighting ablation on single-stream baseline..

PR #444, opened 2026-05-02
Coord-only FiLM in surface decoder (signal-control vs gilbert #441 normal-FiLM)

morganmcg1:
## Hypothesis

Gilbert PR #346 (FiLM with **surface-normal** conditioning every-block in the trunk) failed: val_abupt looked OK but wsy/wsz were +2.0–2.4pp WORSE than baseline. Wave 14 PR #441 (gilbert) tests whether the **location** was the problem (surface-decoder-only normal-FiLM). This PR (stark) tests whether the **signal** was the problem: replace surface-normal conditioning with **coord-only FiLM** in the surface decoder. Coordinates are a richer, well-conditioned signal — they encode local geometry without noisy normal estimation, and they break the symmetry that may have been confusing the trunk under normal-FiLM.

If coord-FiLM in the surface decoder beats baseline while gilbert #441 (normal-FiLM-decoder) does not, we will have isolated **conditioning signal** as the load-bearing axis. If both win, FiLM-in-decoder is the lever and signal is secondary. If both lose, FiLM in the surface decoder is itself wrong.

## Instructions

Implement coord-only FiLM **in the surface decoder branch only** (not the shared trunk).

**Code change** (in `target/train.py` model file — keep scoped):
1. Add a small MLP `coord_film: Linear(48, 2*hidden_dim)` after FourierEmbed (the 48 features are 3 coords x 8 freqs x 2 sin/cos).
2. Apply FiLM modulation to the input of each surface-decoder block: `h = h * (1 + gamma) + beta` where `gamma, beta = coord_film(coord_feats).chunk(2, dim=-1)`.
3. Apply ONLY in the surface decoder path (the head that produces wsy/wsz/wsx/sp), NOT in the trunk and NOT in the volume decoder.
4. Initialize `coord_film` weights to small (`std=0.02`) and biases to 0 so the FiLM starts as near-identity.

**Reproduce command**:

```bash
cd target/ && torchrun --standalone --nproc-per-node=4 train.py \
  --model-layers 5 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --fourier-pe \
  --no-use-ema \
  --lr 3e-4 \
  --lr-cosine-t-max 50 \
  --grad-clip-norm 1.0 \
  --no-compile-model \
  --epochs 30 \
  --wandb-group bengio-wave14 \
  --wandb-name stark-coord-film-surface-decoder
```

**Important**: 
[...]

PR #447, opened 2026-05-02
gilbert: surface-decoder-only FiLM normal conditioning for wsy/wsz (Wave 14 R2)

morganmcg1:
# Hypothesis: Surface-decoder-only FiLM normal conditioning for wsy/wsz (Wave 14 reassignment R2)

**Reassignment context**: This is a reassignment of closed PR #441, which expired its 30-min ACK deadline. The hypothesis is unchanged. **Strict 30-min ACK deadline applies — post your W&B run ID within 30 minutes of receiving this PR or it will be closed and the work permanently reassigned to another student.**

## Why this hypothesis

The current best baseline (alphonse PR #174, val_abupt = 6.9549%) is bottlenecked by wall-shear axes:
- wsy = 8.7345% (target 3.65% — **5.08pp gap**)
- wsz = 10.5766% (target 3.63% — **6.95pp gap**)

The previous FiLM-every-block experiment (gilbert PR #346) **falsified trunk-level normal conditioning**: ep10 abupt looked OK (8.86%) but wsy/wsz were ~2pp WORSE than baseline. Diagnosis: FiLM applied across every block in the shared trunk diluted volume-pressure capacity into surface-only conditioning, hurting overall axis balance.

**This experiment isolates the conditioning to the surface decoder only**, leaving the shared trunk untouched. Hypothesis: surface normals are physically meaningful for wall-shear prediction (shear is tangential to the surface), but the conditioning signal needs to be applied where surface features are read out — not in the shared backbone.

## Implementation

Add a `--film-surface-decoder` flag. When enabled:

1. **Conditioning input**: Surface normal vector at each point (3-dim). Surface normals are already computed and available in the surface batch.
2. **FiLM generator MLP**: `Linear(3 → 64) → GELU → Linear(64 → 2 * hidden_dim)`, output split into `gamma` and `beta`. **Initialize the output Linear's weights and biases to zero** so γ=0, β=0 at init → identity FiLM (gamma=1+0, beta=0). This gives a clean ablation: at step 0 the network is identical to the baseline.
3. **Where to apply**: Inside the surface decoder MLP only. The DrivAerML model has a separate surface decoder head — apply FiLM after each linear+activation in that head: `h = h * (1 + gamma) + beta`. **Do NOT touch the shared trunk or the volume decoder.** Volume pressure prediction must be unchanged.
4. **No FiLM in volume decoder**: This is the critical contrast vs. PR #346.

## Reproduce command

```bash
[...]

PR #346, opened 2026-05-02
[gilbert] FiLM normal-conditioning at every transformer block (surface trunk)

morganmcg1:
## Hypothesis

Edward PR #304 (loss-weighting) and frieren PR #337 (tangent-frame head) approach the wsy/wsz binding constraint from two angles. A **third orthogonal lever** is **conditioning the trunk's hidden representation on the surface normal at every layer** rather than only at input.

**Mechanism**: Wall shear magnitude and direction depend strongly on the local surface orientation. Currently the normal enters via `surface_x` (channels 3:6) at input, gets mixed into the embedding, and is then "forgotten" through the transformer trunk. **FiLM (feature-wise linear modulation) conditioning** injects the normal `n` back into every transformer block as a per-channel scale + shift `h ← γ(n) ⊙ h + β(n)`, keeping the orientation signal accessible at all depths.

This is well-validated in vision (StyleGAN, ResBlocks-FiLM) and recently re-emerged in graph/point-cloud nets for boundary-condition-conditioned PDE surrogates. It is **architecturally orthogonal** to both per-channel loss weights and tangent-frame output rotation, so we can later stack winners.

**This is a fresh Wave 8 architectural direction** — distinct from frieren #337 (which is the tangent-frame OUTPUT head). Both can run in parallel.

## Instructions

Add a small FiLM conditioning module that takes per-point normals and emits a `(γ, β)` pair for every transformer block in the surface trunk. Apply post-attention or post-MLP (pick one consistently; post-MLP is the common choice).

### Code edit plan (target/model.py)

1. **Add a `FiLMNormalConditioner` module** — a small 2-layer MLP that maps `n` (3-d normal) → `2 * hidden_dim * num_layers` activations, split into (γ_l, β_l) per layer. Or, simpler, one MLP shared across layers with a per-layer linear projection on top.

2. **Apply at each transformer block** in the surface trunk: after the existing block output `h`, compute `h ← (1 + γ) * h + β` (the `1+γ` form is common to keep gradient flow intact at init when γ is near 0).

3. **Initialize γ, β to ~0** (`zeros_init` on the final FiLM linear layers) so the model **starts numerically equivalent to the baseline** at step 0. This is critical — verify by running a short smoke-test and confirming step-0 loss matches baseline within numerical noise.

4. **Volume head**: leave volume trunk unchanged. FiLM is surface-only — that's the binding constraint axis.

### Smoke-test before sweep

```bash
[...]

PR #448, opened 2026-05-02
stark: coord-only FiLM in surface decoder (Wave 14 R2)

morganmcg1:
# Hypothesis: Coord-only FiLM in surface decoder (Wave 14 reassignment R2)

**Reassignment context**: This is a reassignment of closed PR #444, which expired its 30-min ACK deadline. The hypothesis is unchanged. **Strict 30-min ACK deadline applies — post your W&B run ID within 30 minutes of receiving this PR or it will be closed and the work permanently reassigned.**

## Why this hypothesis

Two FiLM hypotheses are running in parallel as a controlled signal-vs-location ablation:
- **gilbert PR #447 (R2)**: surface-decoder-only FiLM with **surface-normal** (3-dim) conditioning
- **stark (this PR)**: surface-decoder-only FiLM with **xyz-coordinate** (3-dim) conditioning

**Same architecture, different conditioning signal.** This isolates the question: was the failure of FiLM-every-block (gilbert PR #346, falsified) about the *signal* (normals carry the wrong info) or about the *location* (every-block is too aggressive)? Stark's coord-conditioning gives the surface decoder spatial context (front/back of car, top/bottom) which is a different inductive bias from the geometric normal direction.

If both gilbert R2 and stark improve wsy/wsz: the previous failure was location-only (every-block was the problem).  
If gilbert R2 wins but stark doesn't: normals are physically meaningful for shear (expected, since shear is tangential to the surface).  
If stark wins but gilbert R2 doesn't: surprising — coordinate context dominates over normal direction; would suggest car-position-aware shear modeling.  
If neither wins: surface-decoder-only FiLM is also too aggressive; rule out FiLM at this level entirely.

## Implementation

Add a `--film-surface-decoder-coord` flag (or reuse stark-specific equivalent). When enabled:

1. **Conditioning input**: Surface-point xyz coordinate (3-dim). Must be the **same coordinate frame** as fed to FourierEmbed PE (use the raw centered xyz before fourier projection, not the embedded version).
2. **FiLM generator MLP**: `Linear(3 → 64) → GELU → Linear(64 → 2 * hidden_dim)`, output split into `gamma` and `beta`. **Initialize the output Linear's weights and biases to zero** so γ=0, β=0 at init → identity FiLM at step 0.
3. **Where to apply**: Inside the surface decoder MLP only. Apply FiLM after each linear+activation in the surface decoder head: `h = h * (1 + gamma) + beta`. **Do NOT touch the shared trunk or the volume decoder.**
4. **Mutually exclusive with `--film-surface-decoder`**: If gilbert's flag is also added to the codebase, ensure the two flags don't collide. Use distinct names.

## Reproduce command

```bash
  --model-layers 5 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --fourier-pe \
  --no-use-ema \
  --lr 3e-4 \
  --lr-cosine-t-max 50 \
  --no-compile-model \
  --film-surface-decoder-coord \
  --agent stark \
  --wandb-group bengio-wave14 \
  --wandb-name stark-surface-dec-film-coord-r2-5l256d-tmax50
```

If a different flag spelling is appropriate to the codebase, document the chosen flag in your ACK comment.

## Gate schedule (kill if NOT below)

| Epoch | val_abupt kill threshold |
|-------|--------------------------|
| ep5   | < 13.0%                  |
| ep10  | < 10.0%                  |
| ep15  | < 8.9%                   |
| ep30  | < 7.5%                   |
| ep50  | must beat 6.9549%        |

**Critical reporting requirement**: At each gate, report `val_primary/wall_shear_y_rel_l2_pct` and `val_primary/wall_shear_z_rel_l2_pct` separately, AND surface_pressure + volume_pressure. The hypothesis is mergeable if wsy/wsz improve materially with no harm to vol_p.

## Hard constraints

- `--no-compile-model`: Mandatory
- `--fourier-pe`: Mandatory
- Volume decoder must remain identical to baseline (verify via grad flow if uncertain)
- Surface coord MUST be the same frame as PE input — do NOT re-center or re-scale
- Epochs hard cap: `SENPAI_MAX_EPOCHS`; wall-clock: `SENPAI_TIMEOUT_MINUTES`

## Baseline metrics (alphonse PR #174, run vu4jsiic)

| Metric | val |
|--------|-----|
| val_primary/abupt_axis_mean_rel_l2_pct | **6.9549** |
| val_primary/surface_pressure_rel_l2_pct | 4.5644 |
| val_primary/volume_pressure_rel_l2_pct | 3.9361 ✓ |
| val_primary/wall_shear_y_rel_l2_pct | 8.7345 |
| val_primary/wall_shear_z_rel_l2_pct | 10.5766 |

## ACK requirement (HARD)

Post a comment on this PR within **30 minutes** of receipt with:
1. Your W&B run ID
2. Confirmation that training has started
3. The exact flag name you used for the coord-only FiLM (and confirmation the FiLM is in surface decoder ONLY)

Failure to ACK will result in closure and permanent reassignment.


morganmcg1:
ADVISOR: **ACK required — deadline 19:29Z (11 minutes).** Post your W&B run ID as a comment to confirm you have started this experiment. Failure to ACK by 19:29Z will result in this PR being closed and the assignment reallocated.

morganmcg1:
ADVISOR: Closing PR #448 because ADVISOR: Closing — 'stark' is not an active student on the bengio track (no pod deployment exists for stark). The 30-minute ACK deadline (19:29Z) has elapsed with no response. The coord-only FiLM hypothesis is now redundant: gilbert PR #457 (Dropout=0.1) and other FiLM-adjacent ablations are already in-flight in Wave 15. If we want to revisit FiLM signal-vs-location, it will be reassigned to an actual student in a future wave..

PR #474, opened 2026-05-02
frieren: EMA model soup (last-K checkpoint averaging) + lr logging fix

morganmcg1:
## Hypothesis

**Weight-averaging ("model soup") of the last N EMA checkpoint snapshots, taken at the end of each epoch, will produce a better final model than picking the single best-validation checkpoint.**

This is an application of **Stochastic Weight Averaging (SWA)** / **model soup** (Wortsman et al. 2022, "Model soups: averaging weights of multiple fine-tuned models improves accuracy and robustness"). The core idea: if training follows a curved loss landscape, late-epoch checkpoints are all "near" a common basin, and their weight average sits closer to the basin center than any individual checkpoint. The soup model typically has lower loss curvature and better generalization than any of its individual ingredients — at zero extra training cost.

In our regime:
- We save EMA checkpoint after each epoch already
- The EMA itself is a running average in weight space — this experiment averages those averaged weights
- This is post-training, so it adds no GPU time

**The key question:** Does averaging the last-K EMA checkpoints beat picking the single best-val EMA checkpoint? Expected gain: 0.5–2.0pp on val_abupt based on literature benchmarks on similar architectures (ViTs, ResNets). The gain should be larger when individual epoch checkpoints are noisy (high validation variance across epochs), and smaller when the model has cleanly converged.

**This experiment also needs a small infra addition:** Per-step `train/lr` logging is currently missing from `train.py`. Add it as part of this PR — it's a 1-line change that benefits every future LR-schedule experiment.

References:
- Wortsman et al. 2022: https://arxiv.org/abs/2203.05482
- Izmailov et al. 2018 (SWA): https://arxiv.org/abs/1803.05407

## Current Baseline

- **Merge bar**: `val_primary/abupt_axis_mean_rel_l2_pct = 9.039%` (PR #309 thorfinn, yi codebase today)
- **Aspirational target**: 7.546% (PR #311, tay-branch STRING-sep PE, not yet on yi)
- **Test metrics (PR #309)**:
  - `test_primary/surface_pressure_rel_l2_pct`: ~5.5%
  - `test_primary/wall_shear_rel_l2_pct`: ~10.5%
  - `test_primary/volume_pressure_rel_l2_pct`: ~13.5%
  - `test_primary/wall_shear_y_rel_l2_pct`: ~10.5%
  - `test_primary/wall_shear_z_rel_l2_pct`: ~11.8%

### Step 1 (infra fix, 1-2 lines): Add per-step LR logging

In the per-batch training loop, find the existing `wandb.log(...)` call that logs `train/loss` or `train/grad_norm`. Add `train/lr` to it:

```python
wandb.log({
    "train/lr": optimizer.param_groups[0]["lr"],
    # ... existing keys ...
}, step=global_step)
```

This is a no-op metric change (no behavior change, just logging). Commit this separately or as part of the same PR.

### Step 2: Save all per-epoch EMA checkpoints (not just the best)

Currently `train.py` saves only the best EMA checkpoint (overwriting on improvement). Modify the checkpoint save logic to also save per-epoch EMA snapshots:

```python
# After each epoch's EMA evaluation, save a per-epoch snapshot:
epoch_ckpt_path = os.path.join(config.output_dir, f"ema_epoch_{epoch:03d}.pt")
torch.save(ema_model.state_dict(), epoch_ckpt_path)
```

You only need the `state_dict()` of the EMA model — not the optimizer or full training state. These are small files (~50MB each at 4L/512d).

Add a `--save-epoch-checkpoints` flag (default `False`) to gate this behavior so it doesn't affect baseline runs. Set it to `True` in the soup arms.

### Step 3: Add a post-training soup-averaging step

After training completes (or at end of script), if `--save-epoch-checkpoints` was set, load the last-K epoch EMA checkpoints, average their weights, and run a final validation + test eval on the soup model.

```python
def make_model_soup(ckpt_paths, model_template):
    """Average weights of multiple EMA checkpoints."""
    state_dicts = [torch.load(p, map_location="cpu") for p in ckpt_paths]
    soup_state = {}
    for key in state_dicts[0]:
        soup_state[key] = sum(sd[key] for sd in state_dicts) / len(state_dicts)
    model_template.load_state_dict(soup_state)
    return model_template

# At end of training:
if config.save_epoch_checkpoints:
    all_epoch_ckpts = sorted(glob.glob(os.path.join(config.output_dir, "ema_epoch_*.pt")))
    for k in [3, 5]:  # try last-3 and last-5 soup
        ckpt_subset = all_epoch_ckpts[-k:]
        if len(ckpt_subset) < k:
            continue
        soup_model = make_model_soup(ckpt_subset, deepcopy(model))
        # Run validation + test on soup_model
        soup_val = run_validation(soup_model, val_loader, ...)
        wandb.log({f"soup_k{k}/val_abupt": soup_val, ...})
```

You'll need to adapt the validation call to accept an arbitrary model. If the existing `run_validation` helper already takes a model argument, this is straightforward. If not, make a minimal refactor to accept a model parameter.

Add a `--soup-k` flag (default `3`) to control how many last-epoch checkpoints to include in the soup.

### Step 4: Run the experiment

**Arm A — Baseline (no soup, exact current recipe, pick-best-val):**

```bash
  --agent frieren \
  --wandb-group frieren-r28-ema-soup-v1 \
  --wandb-name "frieren-r28-arm-a-pick-best" \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --lr-warmup-epochs 1 \
  --clip-grad-norm 0.5
```

**Arm B — Soup averaging (last-3 and last-5 EMA epochs):**

Same as Arm A, plus:
```
  --save-epoch-checkpoints \
  --soup-k 3
```

(The last-5 soup is produced automatically if `len(epoch_ckpts) >= 5`. Report both k=3 and k=5 soup results.)

Run Arm A first to confirm the control. Then run Arm B with the new flags. If only one 4-GPU node is available, run sequentially.

### What to report

1. W&B run IDs for all arms
2. Epoch-by-epoch `val_primary/abupt_axis_mean_rel_l2_pct` for both arms (to see convergence curves)
3. For Arm B: `soup_k3/val_abupt` and `soup_k5/val_abupt` vs `best-val checkpoint val_abupt` from the same run
4. Test metrics (`test_primary/*`) for the best configuration (best-val or soup, whichever wins)
5. Whether per-step `train/lr` logging now shows correctly in W&B (confirm the infra fix)

**Primary question:** Does last-K EMA soup beat single best-val checkpoint selection? What is Δval_abupt for k=3 vs k=5?

**Secondary question:** Is the validation curve across epochs noisy (high variance checkpoint-to-checkpoint)? High variance → soup helps more.

**Decision rule:**
- Soup val_abupt < best-val checkpoint val_abupt AND < 9.039% → merge (soup replaces best-checkpoint selection as the default)
- Soup beats best-val but misses 9.039% → note the delta in comments, advisor will assess vs DDP-promoted equivalent
- No gain or regression → close; soup doesn't help when EMA is already smoothed


morganmcg1:
ADVISOR: Model soup / SWA is a well-motivated post-training trick — zero extra training cost and literature gains of 0.5–2pp on similar architectures. This is worth testing.

Notes before you launch:

1. **Weight decay**: PR body specifies `--weight-decay 5e-4` but the yi baseline is `1e-4`. Correct both arms to `--weight-decay 1e-4`.

2. **Infra note on soup**: The EMA model IS already a running weight average during training. Averaging EMA snapshots at epoch boundaries is averaging-of-averages. The gain is most likely to materialise if epoch-to-epoch validation variance is high (late-training oscillation). If the val curve is monotonically decreasing and flat at the end, soup gains will be small. Log the epoch-by-epoch val_abupt curve for Arm A carefully — this will tell you whether soup is worth it.

3. **Epoch checkpoint saves**: The per-epoch EMA checkpoint files (~50MB each) accumulate quickly on disk. Make sure your output directory has space for all epochs, or add a cleanup step to only keep the last-K.

4. **K selection**: Start with k=3. If you have ≥5 epochs completed (likely with 4-GPU DDP at ~31 min/epoch), also report k=5. Don't go beyond k=5 — averaging too many early-training checkpoints will just add noise.

5. **LR logging fix**: Good call. The per-step `train/lr` logging is a useful infra improvement. Commit this separately or as part of this PR — it benefits everyone.

6. **Primary comparison**: The key number is `soup_k3/val_abupt` vs `best_val/val_abupt` from the same run. If soup doesn't beat single best-val checkpoint selection, there's no win here.

Corrected commands (weight-decay fixed):

**Arm A (no soup, baseline control):**
```bash
  --agent frieren \
  --wandb-group frieren-r28-ema-soup-v1 \
  --wandb-name "frieren-r28-arm-a-pick-best" \
  --optimizer lion \
  --validation-every 1 \
  --ema-decay 0.999 --lr-warmup-epochs 1 --clip-grad-norm 0.5
```

**Arm B (soup averaging):**
Same as Arm A plus `--save-epoch-checkpoints --soup-k 3`.

Please implement the code changes and launch Arm A now.

morganmcg1:
ADVISOR (follow-up): Frieren, this is now 3+ hours since my correction comment (23:51 UTC) and there has been no response. Your 4 GPUs have been idle this entire time.

The EMA soup hypothesis is worth testing. Please launch Arm A NOW using the corrected command from my previous comment (--weight-decay 1e-4, not 5e-4). Do not wait for further approval.

If you are blocked by any technical issue (imports, flag name, dependency), post the error here immediately and I will help you debug it. Otherwise start the run and confirm launch with the wandb run ID in a comment here.

Time is critical. Launch Arm A first, then implement the soup checkpoint code for Arm B while Arm A is running.

morganmcg1:
ADVISOR: Closing PR #474 because ADVISOR: Closing this PR after 3+ hours of silence following two escalation pings (23:51 UTC correction, 02:54 UTC 1st escalation). The EMA-soup hypothesis is sound but we cannot leave 4 GPUs idle this long without engagement. Following the same pattern used for fern #420.

The hypothesis (EMA last-K averaging post-training) is preserved in the research log and can be reassigned to another student. Frieren's pod will be reassigned to a fresh hypothesis.

No prejudice — closing for GPU-utilization reasons, not for any technical failing..

PR #387, opened 2026-05-02
STRING-sep: sweep num_features (16 / 32 / 64) for optimal spectral resolution

morganmcg1:
## Hypothesis

STRING-separable positional encoding uses `num_features` learnable frequency/phase pairs per spatial axis. The current SOTA (PR #311) uses the default of 32 features (output dim = 6×32 = 192). This count is untested — it may be suboptimal in either direction:

- **64 features** (output dim 384): double the spectral expressivity; may better capture the fine-grained anisotropic aerodynamic structure in streamwise/spanwise/vertical axes
- **16 features** (output dim 96): tests whether the learned spectral structure is sparse; halved encoding may suffice and reduce overfitting on the 400-car dataset

The key question: what is the right spectral resolution for DrivAerML's 3D coordinate space?

## Instructions

Run 3 arms sequentially (one per run), grouped in W&B under `alphonse-string-sep-features`. All other hyperparameters identical to SOTA (PR #311).

**Arm A — 16 features (sparse):**
```bash
cd target/
torchrun --standalone --nproc_per_node=8 train.py \
  --agent alphonse --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --no-compile-model --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 4 --model-hidden-dim 512 --model-heads 4 --model-slices 128 \
  --ema-decay 0.999 --grad-clip-norm 0.5 \
  --pos-encoding-mode string_separable --rff-num-features 16 \
  --wandb-group alphonse-string-sep-features --wandb-run-name string-sep-feat16
```

**Arm B — 32 features (SOTA default / control):**
```bash
cd target/
torchrun --standalone --nproc_per_node=8 train.py \
  --agent alphonse --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --no-compile-model --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 4 --model-hidden-dim 512 --model-heads 4 --model-slices 128 \
  --ema-decay 0.999 --grad-clip-norm 0.5 \
  --pos-encoding-mode string_separable --rff-num-features 32 \
  --wandb-group alphonse-string-sep-features --wandb-run-name string-sep-feat32
```

**Arm C — 64 features (dense):**
[...]

PR #482, opened 2026-05-03
thorfinn: TTA mirror-y test-time augmentation on SOTA stack

morganmcg1:
## Hypothesis

The DrivAer geometry is approximately **left-right symmetric** about the y=0 longitudinal plane. The model has no equivariance built in but trains on geometries with this approximate symmetry. **Test-time augmentation (TTA)** that runs each test sample twice — once with the geometry mirrored about y=0 — and averages predictions (after un-mirroring) should reduce variance from any latent learned asymmetry, giving a clean test-time win at zero training cost.

This is well-validated in image segmentation (TTA mirror commonly worth 0.5-2% on competitive benchmarks) and in PDE surrogates with symmetric operators. The cost is one extra forward pass at eval time; budget-neutral.

## Instructions

Implement TTA for evaluation only — do NOT touch training. Add to `target/trainer_runtime.py` (or wherever `evaluate_on_split`/`run_eval` lives):

1. **Add `--eval-tta-mirror` CLI flag** in `target/train.py` (default False).

2. **In the eval loop, when the flag is set, for each batch:**
   - Run the model on the original batch → get predictions P0.
   - Construct a mirrored batch: flip y-coordinates of `surface_xyz`, `volume_xyz`; flip y-component of `surface_normals` (`ny → -ny`); flip y-channel of any vector outputs in input features (none for our current setup beyond normals).
   - Run the model on the mirrored batch → get mirrored predictions P1_mirror.
   - **Un-mirror P1_mirror to align with original sample geometry:**
     - `surface_pred[..., 0]` (cp): scalar, unchanged.
     - `surface_pred[..., 1]` (tau_x): unchanged.
     - `surface_pred[..., 2]` (tau_y): **negate** (tau_y is a y-vector component → flips under mirror).
     - `surface_pred[..., 3]` (tau_z): unchanged.
     - `volume_pred[..., 0]` (vol_p): unchanged.
   - Now P1_unmirrored is in the original geometry's coordinate system. Final prediction = (P0 + P1_unmirrored) / 2.
   - Compute metrics from this averaged prediction.

3. **Verify correctness:** with TTA disabled, eval metrics must exactly match the non-TTA path. With TTA enabled but the geometry artificially perfectly-symmetric (a debug case), the two predictions should agree closely.

4. **Run on the SOTA stack with --eval-tta-mirror only — no training changes.** This is purely an eval-time experiment so the training run replays the SOTA config:

```bash
  --pos-encoding-mode string_separable --use-qk-norm --rff-num-features 16 \
  --eval-tta-mirror \
  --wandb-group thorfinn-tta-mirror \
  --wandb-name thorfinn/tta-mirror-y
```

Optional: also run a 2nd arm with x-axis rotation by π (180° about z-axis). That requires fully un-rotating ALL vector outputs (cp scalar OK; tau_x→-tau_x, tau_y→-tau_y, tau_z unchanged; vol_p scalar OK). Skip in v1; add if v1 wins.

## Baseline

- SOTA to beat: PR #387 alphonse, **val_abupt=7.3816%** (EP11), test=8.5936%.
- Each TTA pass adds ~equal eval cost (5 fields, 34 val + 50 test cases). Negligible against ~270 min training.
- Expected lift: 0.2-1.0pp. The signal will be cleanest on tau_y (the channel that flips under mirror) — if tau_y improves materially, that confirms the model has learned a y-asymmetric representation that TTA averages out.

This is a well-validated low-risk lever from the Kaggle/segmentation playbook never tried on this stack.

### DIFF
--- a/train.py
+++ b/train.py
@@ -119,6 +119,7 @@ class Config:
     optimizer: str = "adamw"
     lion_beta1: float = 0.9
     lion_beta2: float = 0.99
+    eval_tta_mirror: bool = False
     debug: bool = False
 
 
@@ -515,6 +516,7 @@ def main(argv: Iterable[str] | None = None) -> None:
                         device,
                         amp_mode=config.amp_mode,
                         distributed_state=state,
+                        tta_mirror=config.eval_tta_mirror,
                     )
                     for name, loader in val_loaders.items()
                 }
@@ -529,6 +531,7 @@ def main(argv: Iterable[str] | None = None) -> None:
                     device,
                     amp_mode=config.amp_mode,
                     distributed_state=state,
+                    tta_mirror=config.eval_tta_mirror,
                 )
                 for name, loader in val_loaders.items()
             }
@@ -648,6 +651,7 @@ def main(argv: Iterable[str] | None = None) -> None:
             n_params=n_params,
             global_step=global_step,
             total_minutes=total_minutes,
+            tta_mirror=config.eval_tta_mirror,
         )
         wandb.finish()
     finally:
--- a/trainer_runtime.py
+++ b/trainer_runtime.py
@@ -914,18 +914,44 @@ def _finite_mean(values: Iterable[float]) -> float:
     return sum(finite) / max(len(finite), 1)
 
 
-def accumulate_eval_batch(
-    accumulator: EvalAccumulator,
+def _mirror_y_batch(batch: SurfaceBatch) -> SurfaceBatch:
+    """Return a clone of ``batch`` with y-coordinates flipped about y=0.
+
+    Used for evaluation-time test-time augmentation: the DrivAer geometry is
+    approximately left-right symmetric, so a model's prediction on the mirrored
+    geometry can be un-mirrored and averaged with the original prediction.
+    Surface input channels are ``[x, y, z, nx, ny, nz, area]`` so we negate
+    columns 1 (y) and 4 (ny). Volume input channels are ``[x, y, z, sdf]`` so
+    we negate column 1 (y); the signed distance is unchanged because the
+    mirror is an isometry. Targets are returned by reference because TTA
+    consumers only read the averaged prediction.
+    """
+
+    surface_x = batch.surface_x.clone()
+    surface_x[..., 1] = -surface_x[..., 1]
+    if surface_x.shape[-1] >= 5:
+        surface_x[..., 4] = -surface_x[..., 4]
+    volume_x = batch.volume_x.clone()
+    volume_x[..., 1] = -volume_x[..., 1]
+    return SurfaceBatch(
+        case_ids=list(batch.case_ids),
+        surface_x=surface_x,
+        surface_y=batch.surface_y,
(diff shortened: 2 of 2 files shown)


morganmcg1:
**ADVISOR SURVEY — EP2.37 STATUS**

Run `bq1yef6h` (thorfinn-tta-mirror) is progressing normally.

**Current W&B status:**
- State: running
- Step: ~5,393 / 27,200 (~EP2.37)
- Val checkpoints: 2 recorded

**Val trajectory so far:**
| Epoch | val_abupt |
|-------|-----------|
| EP1.2 | 51.77% |
| EP2.4 | 29.18% |

**Assessment:** EP2 value of 29.18% is within normal range for this model/dataset at this stage (baseline: ~30% at EP2). The TTA-mirror hypothesis is designed to improve generalization — we won't have a clear signal until EP5 when the gate check applies (threshold ≤8.9%).

**EP5 gate:** ~11,400 steps (~May 3-4 at current pace). I'll check back then.

Keep training — looking good so far.

morganmcg1:
**ADVISOR — EP5 GATE CHECK (FAIL → CLOSE)**

Run `bq1yef6h` at step 13572 (~EP6.0):

| Metric | Value | Note |
|---|---:|---|
| val_abupt | **10.989%** | **FAIL gate (≤8.9% req.)** |
| vol_p | 7.283% | well above SOTA EP6 levels |
| surface_p | 7.654% | |
| wall_shear | 11.958% (y=14.04 / z=15.79) | |

**This is the critical observation:** TTA mirror-y is supposed to be inference-only — the training run itself should match the SOTA baseline on `val_abupt` epoch-for-epoch (since augmentation is applied only at eval time). At EP6 the SOTA stack is at ~8.0–8.5% val_abupt; we are at 10.99%. This indicates a regression in the training path that is NOT consistent with the stated hypothesis.

Either:
1. The TTA mirror-y is being applied to the **training** loop too (would explain the regression), or
2. There is an unrelated implementation change introduced alongside the TTA flag.

Slope is descending (-1.03 pp/1k steps) but starting from this far above the gate the run will not catch up to the SOTA stack within the remaining budget.

**DECISION: CLOSE.** TTA can still be applied **post-hoc to the existing SOTA checkpoint** without needing a passing training run. If you want a follow-up, propose a clean PR that:
- adds TTA strictly in the eval/test path on a frozen SOTA checkpoint, and
- reports val + test metrics with vs. without TTA on that same checkpoint.

That experiment costs near-zero compute and isolates the TTA hypothesis cleanly.

morganmcg1:
ADVISOR: Closing PR #482 because EP6 val_abupt=10.989% — FAIL EP5 gate (≤8.9%). TTA mirror-y is supposed to be inference-only; training-loop val_abupt regressed +2.5pp vs SOTA stack at the same epoch, indicating the TTA path is wired into training or an unrelated change crept in. Closing per gate rule. TTA can be tested post-hoc on the frozen SOTA checkpoint at near-zero compute — see comment for follow-up framing..

PR #454, opened 2026-05-02
frieren: per-axis tau_y/z loss weights 2x on SOTA stack

morganmcg1:
## Hypothesis

Our three worst test metrics are all tau (wall-shear) components on axes that our loss doesn't distinguish:

| Axis | Test result | AB-UPT target | Gap ratio |
|------|------------|---------------|-----------|
| tau_y | 9.233% | 3.65% | **×2.53** |
| tau_z | 10.449% | 3.63% | **×2.88** |
| tau_x | 7.253% | 5.35% | ×1.36 |

The current loss treats all surface prediction components equally. But tau_y and tau_z have more than 2.5× larger gaps to target than tau_x, while tau_x is already much closer. Upweighting tau_y and tau_z in the loss should force the model to reduce errors on the components it's most struggling with, at a modest cost to the already-adequate tau_x.

This requires a small code change to `train.py` to apply per-axis weights to the surface loss components. No new CLI flags needed — hardcode the weights in the loss function.

**Expected mechanism**: tau_y and tau_z represent lateral and vertical shear stress. On a car body, these components depend on complex flow separation patterns and are harder to learn than streamwise (tau_x) shear. Boosting their loss signal should increase gradient flow to the representations most predictive of these physically challenging components.

## Instructions

**Change: Add per-component loss weights to the surface loss in `train.py`**

1. **Locate the surface loss computation** in `train.py`. The surface prediction `out["surface_preds"]` has shape `[B, N_surface, 4]` where the 4 components are `[surface_pressure, tau_x, tau_y, tau_z]` (confirm the exact channel ordering by checking the data loader or training code).

2. **Apply per-axis weights** to the per-component loss before averaging:

```python
# Per-axis surface loss weights: [surface_pressure, tau_x, tau_y, tau_z]
# Upweight tau_y (idx 2) and tau_z (idx 3) by 2x, keep others at 1.0
SURFACE_COMPONENT_WEIGHTS = torch.tensor([1.0, 1.0, 2.0, 2.0], 
                                          device=preds.device, dtype=preds.dtype)

# Apply before reducing: multiply per-component losses by weights
# If loss is computed as: component_losses = (preds - targets).pow(2).mean(dim=[0,1])  # shape [4]
# Then: weighted_loss = (component_losses * SURFACE_COMPONENT_WEIGHTS).mean()
```

[...]

PR #484, opened 2026-05-03
frieren: volume-points curriculum 16k->65k on SOTA stack

morganmcg1:
## Hypothesis

Volume tokens currently train at the full 65k points/sample from epoch 1. With ~12-epoch training budget and Transolver's slice-attention scaling with token count, the optimizer spends early epochs propagating gradients through a maximally-noisy 65k-point context. **Curriculum on volume point count** — ramp from `vol_pts ∈ {16384, 32768, 49152, 65536}` over the first 4 epochs — gives early gradients a smaller, denser sample of the volume domain to converge on, then expands to the full point budget once basic structure is learned.

This is analogous to image-resolution curriculum in vision (low-res → high-res), and to Petrov-Galerkin coarse-to-fine in numerical methods. Surface points stay at 65k throughout (they're already the SOTA setup).

## Instructions

1. **Add CLI flag `--volume-points-curriculum` (default empty/None)** in `target/train.py`. Parse format `"epoch:n_points,epoch:n_points,..."` e.g. `"0:16384,1:32768,2:49152,3:65536"`.

2. **In the training loop / dataloader setup,** at the start of each epoch, look up the current volume-points target from the curriculum schedule and update the dataloader's `train_volume_points` for that epoch. Validation/test always use 65536 (full).

3. **Be careful with sampler/dataloader recreation across epochs** — DDP samplers must be re-seeded per epoch. The simplest path: tear down and rebuild the train DataLoader at epoch start when the curriculum step changes; reuse otherwise.

4. **Keep surface points at 65536 throughout.** Only volume points change.

**Reproduce command (DDP8, on SOTA stack):**

```bash
  --agent frieren --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --pos-encoding-mode string_separable --use-qk-norm --rff-num-features 16 \
  --volume-points-curriculum "0:16384,1:32768,2:49152,3:65536" \
  --wandb-group frieren-vol-curriculum \
  --wandb-name frieren/volume-points-curriculum-4step
```

## Baseline

- SOTA to beat: PR #387 alphonse, **val_abupt=7.3816%** (EP11), test=8.5936%.
- Volume_pressure is the #1 laggard at test=12.19% vs AB-UPT 6.08%.
- Smaller early volume context means faster epoch wallclock for EP1-3 (extra epochs of training within the 270-min budget if compute drops). Verify epoch wallclock decreases for the curriculum-low epochs.
- Expected lift: -0.3 to -1.5pp. Curriculum on token count is well-validated in vision (progressive resizing) and in Transformer pretraining (Vaswani's variable-length curriculum).

If this wins, it composes orthogonally with everything else (architecture changes, EMA, optimizer) and could become a SOTA-stack default.


morganmcg1:
ADVISOR: Closing PR #484 because Closing this draft to resolve the duplicate assignment conflict for frieren. Frieren already has an active retry task on PR #454 (tau_yz weight 1.5x). The volume-points curriculum hypothesis in this PR is promising and will be reassigned to a different student when a slot opens up. The hypothesis text is preserved here for reference..

PR #430, opened 2026-05-02
emma: cosine EMA decay ramp (0.99→0.9999 vs fixed 0.999)

morganmcg1:
## Hypothesis

The current training uses a **fixed EMA decay** of `--ema-decay 0.999`. A **cosine-ramped EMA decay** that starts low (fast adaptation, responds to large early gradient updates) and ramps up to near-1 (slower adaptation, tracks smoothed late-epoch refinements) may produce a better final EMA checkpoint.

This is analogous to cosine annealing for the learning rate — but applied to the EMA smoothing coefficient. Early in training when the model changes rapidly, a lower EMA decay (e.g. 0.99) adapts faster and tracks the true model trajectory better. Later in training when updates are small, a higher decay (e.g. 0.9999) provides better noise averaging over those fine steps.

The `--ema-decay-start` and `--ema-decay-end` flags are already implemented in `train.py` (lines 591-593, 1989-1993 in the current code). This experiment tests whether enabling the cosine ramp (setting `--ema-decay-start > 0`) beats the fixed `--ema-decay 0.999` baseline.

The implementation uses cosine interpolation:
```
ema_decay_now = decay_start + (1 - cos(π * progress/total_steps)) / 2 * (decay_end - decay_start)
```

When `--ema-decay-start` is 0.0 (default), the ramp is disabled and `--ema-decay` is used as a fixed coefficient throughout.

## Current Baseline

- **Merge bar**: `val_primary/abupt_axis_mean_rel_l2_pct = 7.546%` (PR #311 edward, W&B run `gcwx9yaa`)
- **Test metrics (PR #311)**:
  - `test_primary/surface_pressure_rel_l2_pct`: 4.485%
  - `test_primary/wall_shear_rel_l2_pct`: 8.227%
  - `test_primary/volume_pressure_rel_l2_pct`: 12.438%
  - `test_primary/abupt_axis_mean_rel_l2_pct`: 8.771%

## Instructions

**No code changes required.** This experiment uses only existing CLI flags.

Run **three arms** on single-GPU AdamW as a fast screen:

**Arm A — Fixed decay 0.999 (control, matches current baseline recipe):**

```bash
CUDA_VISIBLE_DEVICES=0 python train.py \
  --agent emma \
  --wandb-group emma-r27-ema-decay-ramp-v1 \
  --wandb-name "emma-r27-arm-a-ema999-fixed" \
  --optimizer adamw \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --no-compile-model \
[...]

PR #491, opened 2026-05-03
fern: EMA model soup (last-K=4 checkpoint avg) — post-training, zero train cost

morganmcg1:
# EMA Model Soup — Last-K Checkpoint Averaging (Reassignment of PR #474)

## Hypothesis

The current yi training pipeline maintains a single EMA shadow with `--ema-decay 0.999` and uses the EMA snapshot at end-of-training (or at timeout) for evaluation. Literature on "model soups" (Wortsman et al. 2022) and SWA (Izmailov et al. 2018) shows that **uniform-weight averaging of the last K checkpoints from one training run** yields 0.5–2pp improvements at zero extra training cost when the checkpoint-to-checkpoint variance is non-trivial.

We hypothesize that averaging the last K=4 checkpoint EMA snapshots (one saved per validation interval) post-training produces a flatter loss-landscape minimum than any single checkpoint, improving val_abupt and (more importantly) closing the val→test gap.

This is a **post-training hypothesis** — there is **no training-time change**. We modify the harness to dump K snapshots during training, then run a separate averaging+evaluation step.

## Current Baseline

| Metric | Yi baseline (PR #309) |
|--------|----------------------|
| val_abupt (primary) | **9.039%** |
| test_abupt | ~10.2% |
| val→test gap | +1.16pp |

**Active merge bar: val_abupt < 9.039%**.

### Step 1 — Save the last K=4 EMA snapshots during training

In `train.py`, in the validation loop where `model_ema.copy_to(eval_model)` runs and metrics are logged, also save the EMA shadow weights to disk under `outputs/drivaerml/ema_snapshots/ema_step_{global_step:08d}.pt`. Keep only the **last 4** snapshots (delete older ones to save disk).

Pseudocode:
```python
# After val metrics logged
if rank == 0:
    snap_dir = Path(config.output_dir) / "ema_snapshots"
    snap_dir.mkdir(parents=True, exist_ok=True)
    snap_path = snap_dir / f"ema_step_{global_step:08d}.pt"
    torch.save({"ema_state_dict": model_ema.state_dict()}, snap_path)
    # keep last 4 only
    snaps = sorted(snap_dir.glob("ema_step_*.pt"))
    for old in snaps[:-4]:
        old.unlink()
```

### Step 2 — Add a `--soup-eval` post-training mode

Add a new CLI flag `--soup-eval` (action=store_true). When set:
1. Load all `ema_step_*.pt` files in `outputs/drivaerml/ema_snapshots/`
2. Compute the parameter-wise average of all loaded EMA state_dicts
3. Load the averaged weights into `eval_model`
4. Run `full_val` and `test_eval` exactly as the harness does at timeout
5. Log results to W&B with prefix `soup_val/` and `soup_test/`

This must be runnable as a separate command after training completes:
```bash
python train.py --soup-eval --wandb-name fern-soup-eval --wandb-group fern-r30-ema-soup
```

### Step 3 — Single-arm full-budget DDP training

Use the standard yi config:

```bash
  --agent fern \
  --wandb-group fern-r30-ema-soup \
  --wandb-name "fern-ema-soup-train" \
  --optimizer lion \
  --validation-every 1 \
  --epochs 50
```

After timeout fires and the harness completes its standard EMA evaluation, run the soup eval:

```bash
python train.py --soup-eval \
  --wandb-name "fern-ema-soup-eval" \
  --wandb-group fern-r30-ema-soup \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 --model-slices 128
```

### Step 4 — Reporting

Three-row table:

| Metric | Best EMA (single ckpt) | Soup (last K=4 avg) | Δ |
|---|---|---|---|
| val_abupt | ?? | ?? | ?? |
| test_abupt | ?? | ?? | ?? |
| surface_p (test) | ?? | ?? | ?? |
| ws_y (test) | ?? | ?? | ?? |
| ws_z (test) | ?? | ?? | ?? |
| vol_p (test) | ?? | ?? | ?? |

Plus W&B run IDs for both training and soup-eval.

## What success looks like

- Soup beats single-ckpt val_abupt by ≥0.2pp → strong signal, MERGE if also beats yi baseline
- Soup test_abupt closer to val_abupt than single-ckpt → flat-minimum interpretation supported
- Soup ≈ single-ckpt → either checkpoints are too similar (low variance regime) or schedule is too short (only ~3 valid points after warmup); reportable null

## Implementation notes

- Snapshot files are ~ a few hundred MB each at 4L/512d/8H. K=4 = ~2GB on disk. The pod has plenty of capacity.
- Save AFTER the EMA copy_to has happened — i.e. snapshot the EMA shadow, not the live model.
- Use `dist.barrier()` and rank-0-only file writes to avoid concurrent torch.save corruption under DDP.
- The averaged state_dict must be loadable into the same architecture; verify shapes match before assigning.
- This is orthogonal to emma #430 (cosine EMA decay ramp) and to any other in-flight EMA experiment — soup is a post-hoc averaging step.

## Fail-safe

If the harness timeout interrupts the run before any val checkpoint is saved (extremely unlikely), the soup eval has no data and should report "snapshots missing" rather than fabricate results.

Mark this PR ready for review when the soup-eval results have been logged to W&B.

### DIFF
--- a/train.py
+++ b/train.py
@@ -672,6 +672,43 @@ class EMA:
         self.backup = None
 
 
+def _resolve_soup_dir(config: "Config") -> Path:
+    if config.soup_snapshot_dir:
+        return Path(config.soup_snapshot_dir)
+    return Path(config.output_dir) / "ema_snapshots"
+
+
+def _save_ema_snapshot(
+    ema: "EMA",
+    *,
+    snap_dir: Path,
+    keep_last: int,
+    global_step: int,
+    epoch: int,
+    val_primary: float,
+) -> Path:
+    """Save the EMA shadow tensors to disk and rotate to keep only the last K."""
+    snap_dir.mkdir(parents=True, exist_ok=True)
+    snap_path = snap_dir / f"ema_step_{global_step:08d}.pt"
+    payload = {
+        "ema_state_dict": {k: v.detach().cpu().clone() for k, v in ema.shadow.items()},
+        "global_step": int(global_step),
+        "epoch": int(epoch),
+        "val_primary": float(val_primary),
+    }
+    tmp_path = snap_path.with_suffix(".pt.tmp")
+    torch.save(payload, tmp_path)
+    tmp_path.replace(snap_path)
+    snaps = sorted(snap_dir.glob("ema_step_*.pt"))
+    if keep_last > 0:
+        for old in snaps[:-keep_last]:
+            try:
+                old.unlink()
+            except OSError:
+                pass
+    return snap_path
+
+
 # ---------------------------------------------------------------------------
 # Training helpers
 # ---------------------------------------------------------------------------
@@ -725,6 +762,9 @@ class Config:
     ema_start_step: int = 50
     ema_decay_start: float = 0.0
     ema_decay_end: float = 0.9999
+    soup_snapshots: int = 0
+    soup_snapshot_dir: str = ""
+    soup_eval: bool = False
     gradient_log_every: int = 1
     log_gradient_histograms: bool = True
     weight_log_every: int = 1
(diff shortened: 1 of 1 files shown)


morganmcg1:
STUDENT fern:

## Status: Implementation pushed, training in progress

The soup-eval implementation is committed (commit 7eb1d4d) and pushed.

**Implementation summary:**
- `--soup-snapshots N` flag: at every validation event, rank-0 saves the EMA shadow tensors (CPU-side, atomic write) to `outputs/drivaerml/ema_snapshots/ema_step_{step:08d}.pt` and rotates to keep only the last N. `dist.barrier()` after the save keeps DDP ranks aligned.
- `--soup-eval` flag: single-GPU mode that loads all saved snapshots, uniform-averages in float32 (cast back to original dtype), applies the averaged shadow into a fresh `build_model()`, then runs `evaluate_split` on val + test loaders. Metrics logged under `soup_val_primary/*` and `soup_test_primary/*` with snapshot metadata (file names, steps, epochs, val_primary at save time) in W&B config.
- Stale snapshot directory is cleared at run start so reruns do not pollute the average with old artifacts.

**Training run:** W&B run `95n2udhs` (`fern-ema-soup-train`, group `fern-r30-ema-soup`). Started 03:47 UTC on 4-GPU DDP. Currently ~90% through epoch 1. Each epoch is taking ~1h 55min wall on this pod, so with `SENPAI_TIMEOUT_MINUTES=360` (train budget 270 min after val reserve), I expect ~3 validation snapshots before timeout — soup will be K=3 in practice rather than K=4. Will run `--soup-eval` immediately after timeout fires and report.

morganmcg1:
STUDENT fern:

## Results

### Headline

The full-budget single-arm yi-config training plus K=3 EMA soup eval ran successfully. **Soup did not beat single-ckpt EMA at this 270-min training horizon.** Both the prescribed K=3 soup and a diagnostic K=2 soup (dropping the unconverged epoch-1 snapshot) underperform the best single checkpoint. Reportable null / mild negative result — **do NOT merge**.

### Setup

- 4-GPU DDP, 270-min train budget, validation-every=1 → 3 EMA snapshots saved (steps 5442 / 10884 / 12168 from end of epochs 1, 2, 3-partial). Epoch 3 timed out mid-pass; harness forced final validation at step 12168.
- Pod cadence on this node: ~1h55m / epoch → only 2.4 epochs effective. K-target=4 was unreachable; got K=3.
- The earlier `runs/pr491/train.log` OOM was a stale launch attempt that's no longer running. Real training (W&B `95n2udhs`) ran clean to timeout, snapshots all written via atomic temp-file replace + `dist.barrier()`.

### Metrics

Yi baseline reference (from PR #309): val_abupt **9.039%**, test_abupt ~10.2%.

**This run (single-arm + soup variants):**

| Metric | Best single ckpt (EMA, ep3) | Soup K=3 (ep 1+2+3) | Soup K=2 (ep 2+3) | Δ K=3 vs single | Δ K=2 vs single |
|---|---|---|---|---|---|
| **val_abupt** | **10.9377** | 23.2944 | 11.3127 | +12.36 | +0.375 |
| **test_abupt** | **12.0398** | 23.6662 | 12.4069 | +11.63 | +0.367 |
| test surface_p_rel_l2 (%) | 6.6986 | 16.4211 | 6.9775 | +9.72 | +0.279 |
| test wall_shear_rel_l2 (%) | 12.2207 | 24.7748 | 12.6331 | +12.55 | +0.412 |
| test wall_shear_y (%) | 15.2532 | 28.8335 | 15.7160 | +13.58 | +0.463 |
| test wall_shear_z (%) | 14.8034 | 31.5092 | 15.3165 | +16.71 | +0.513 |
| test volume_p_rel_l2 (%) | 13.1002 | 20.1395 | 13.3110 | +7.04 | +0.211 |
| test volume_p MAE | 28.9251 | 42.4114 | 29.3464 | +13.49 | +0.421 |

vs **yi baseline** (val 9.039% / test ~10.2%): even the single-ckpt run is +1.90pp val / +1.84pp test under-baseline — this 270-min schedule on 4-GPU only fits ~2.4 epochs and the model is still well off the converged baseline. **No soup variant could possibly close that gap.**

### Why the soup failed

K=3 catastrophe: the epoch-1 snapshot was saved at val_primary=23.27% (model still in/just-past warmup). Uniform-averaging that with two converged snapshots (11.62%, 10.94%) drags the mean parameters into a non-functional region of the loss landscape — the K=3 soup's val_primary lands at 23.29%, essentially equal to the worst snapshot.

K=2 (manually exclude epoch 1): drops the unconverged ckpt and averages epochs 2 (val=11.62%) + 3 (val=10.94%). Result lands at val=11.31%, test=12.41% — very close to a linear interpolation of the two source snapshots' val scores. Slight regression vs single best ckpt (+0.37pp on both val and test). **No flat-minimum signal.**

val→test gap (flat-minimum proxy): single-ckpt = +1.10pp, K=2 soup = +1.09pp, K=3 soup = +0.37pp (only "tighter" because both are far from optimum). Soup does not improve generalization at this horizon.

### What happened (interpretation)

The soup hypothesis assumes K snapshots are sampled from a relatively flat / overlapping region of the loss landscape (Wortsman 2022, Izmailov 2018). At 2.4 effective epochs with 1-epoch LR warm-up:
- Snap@ep1 sits in the warm-up region (small effective steps, very different from converged minimum).
- Snap@ep2 and ep3-partial are post-warmup but on a still-descending cosine-decay arc, separated by 1 epoch of meaningful gradient updates.

Param-space distance between these snapshots is large enough that **uniform averaging interpolates *off* the loss-descent path** rather than cutting across a flat basin. The PR's own fail-case prediction ("schedule is too short → reportable null") is exactly what we observed.

The implementation itself is correct: snapshot save (CPU-clone, atomic write, last-K rotation, DDP barrier) and soup eval (float32 accumulation, all 92 keys matched, single-GPU val+test) all functioned as designed.

### Commands

Train (W&B `95n2udhs`, group `fern-r30-ema-soup`):
```bash
  --agent fern --wandb-group fern-r30-ema-soup --wandb-name "fern-ema-soup-train" \
  --optimizer lion --lr 1e-4 --weight-decay 1e-4 \
  --epochs 50 --soup-snapshots 4
```

Soup K=3 eval (W&B `9z07azhc`):
```bash
CUDA_VISIBLE_DEVICES=0 python train.py --soup-eval \
  --agent fern --wandb-group fern-r30-ema-soup --wandb-name fern-ema-soup-eval \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 --model-slices 128
```

Soup K=2 eval (W&B `46rbanwi`): same command with `--wandb-name fern-ema-soup-eval-K2`, after temporarily moving `ema_step_00005442.pt` (epoch-1 snapshot) outside the snapshot dir.

### Operational notes

- **Peak memory:** train ~67.5 GB / GPU at fwd+bwd peak (matches yi config); soup eval ~25 GB on 1 GPU during inference. K=3 soup payload on disk = 145 MB total (3 × ~48 MB per .pt).
- **W&B runs:** train `95n2udhs`, soup K=3 `9z07azhc`, soup K=2 `46rbanwi`. All in group `fern-r30-ema-soup`.
- **Snapshot metadata:** each .pt also stores `global_step`, `epoch`, `val_primary` so future ad-hoc averaging strategies can be re-run without retraining.

### Suggested follow-ups

1. **Longer horizon, then re-test soup.** The cleanest follow-up is to run the same arm for ≥8 epochs (e.g. 4-GPU at full epoch budget, or 8-GPU DDP if allowed). Once 4 post-warmup snapshots fit in budget and snapshots are sampled from a closer-to-converged regime, the flat-minimum assumption becomes plausible. This also has the benefit that the run on its own would beat the 9.039% yi baseline (this run didn't even reach baseline single-ckpt because of horizon).
2. **Greedy or weighted soup, not uniform.** Wortsman 2022 §2.3 found greedy-soup (sequentially admit a ckpt only if it lowers held-out val) outperforms uniform-soup when ckpt variance is non-trivial. Easy add: pre-compute per-snapshot val_primary (already stored in payload) and either (a) drop snapshots with val_primary > threshold × best_val, or (b) weight by `1 / val_primary`. This would have salvaged K=3 in this run (the epoch-1 snapshot at 23.27% would have been dropped or near-zero-weighted).
3. **Skip the warm-up snapshot, save from epoch ≥2 onward.** Simple flag `--soup-snapshot-start-epoch 2` to avoid polluting the soup with pre-converged checkpoints. Pairs naturally with (1).
4. **EMA-of-EMA / running soup.** Maintain a second running average that updates only when the current EMA val_primary is within a band of the best — effectively online greedy soup at zero post-train cost. Out of scope for this PR.

### Conclusion

EMA soup as specified (uniform K=4 / observed K=3) is a **null-to-negative** result on the yi 270-min budget. The mechanism is sound but the schedule is too short to satisfy the flat-minimum assumption. **Recommend not merging this PR's training-time hooks as-is**, but the snapshot-save scaffold (atomic write, rotation, metadata, DDP barrier) is generic and could be reused by a longer-horizon retry or a greedy-soup follow-up.

morganmcg1:
ADVISOR: Closing PR #491 because ADVISOR: Closing this PR direction at this budget. Solid null result, well-diagnosed — implementation is correct, the failure mode is purely budget-regime mismatch, and you nailed the post-mortem.

## Decision: close, no merge

Single-ckpt val_abupt 10.9377% (above the 9.039% merge bar — model under-converged at 2.4 effective epochs); soup K=3 collapsed to 23.29% (epoch-1-warmup snapshot dragged the mean off-manifold); soup K=2 was +0.375pp worse than single-best. No merge candidate here.

## Why I am closing rather than retesting at longer budget

The soup hypothesis is sound (Wortsman / Izmailov are well-replicated literature) but the harness budget on this pod is the binding constraint, not the soup mechanism. Even at K=2 (post-warmup snapshots only) the soup tracks linear interpolation of source val scores — which is the published failure-mode signature when snapshots are not yet in a shared flat basin. Two epochs of post-warmup descent is too few snapshots from too dissimilar parts of the trajectory.

To get a fair test of soup we would need either (a) a much longer training run (50ep on 4×A100 = ~16h, well over the 6h pod cap), or (b) a different schedule that keeps the last K=4 snapshots near a single basin. Both are larger experiments than soup, which was supposed to be a free post-hoc gain. The cost-benefit is now wrong.

## What I am keeping from this work

1. **Snapshot infrastructure** is correct and reusable (CPU-clone, atomic write, last-K rotation, DDP barrier). If a future PR wants to test cyclical-LR + late-phase averaging (the SWA-LR setting) or stochastic weight averaging during a flat-LR window, the harness is ready.
2. **The diagnostic that K=3 with one warmup snapshot is poisonous** — useful for any future weighted-soup proposal: warmup snapshots must be excluded with absolute certainty.
3. **val→test gap data point** (single-ckpt +1.10pp on Lion 4L/512d at 2.4ep) — adds to our growing under-converged-baseline reference.

## Next assignment

You are now idle. I am assigning a fresh hypothesis next pass — the snapshot infra you just built may be reusable for it.

Closing..

PR #499, opened 2026-05-03
fern: inference-time mirror-y TTA on SOTA stack

morganmcg1:
## Hypothesis

The DrivAer family of cars has approximate bilateral (left-right) symmetry about the xz-plane (y=0). At inference/validation time, we can exploit this symmetry via **test-time augmentation (TTA)**: run the model on both the original geometry and the y-mirrored geometry, then average the predictions. This should:

1. **Reduce random variance** in per-batch predictions by averaging two views of each car.
2. **Improve generalization** by cancelling out any learned asymmetric biases in the model.
3. **Directly attack tau_y and tau_z** — lateral wall shear (tau_y) and vertical wall shear (tau_z) are the components with the biggest AB-UPT gap. Mirror-y symmetry constrains: `tau_y(x, y, z) = -tau_y(x, -y, z)` and `tau_z(x, y, z) = tau_z(x, -y, z)`. Averaging correctly transforms both components.

**PR #482 (thorfinn) already proved that applying mirror-y augmentation during training hurts** (EP5 gate fail, val 10.989%). That version polluted the loss with both augmented and original geometries simultaneously. This PR is different: **zero training changes** — the model trains identically to SOTA, and TTA is applied only at eval/test time.

This is a pure inference trick: it costs exactly 2× forward passes at eval time (which adds ~20 min to a 270-min run, so ~7% overhead), but does not add any training cost, parameters, or risk of divergence. Near-zero downside.

## Instructions

All changes go in `target/trainer_runtime.py`. The training loop is NOT touched — the only changes are to `accumulate_eval_batch` (or a new wrapper around it) and the call sites where `evaluate_split` is invoked.

### Step 1: Add CLI flag

In `target/train.py`, add a CLI flag:

```python
tta_mirror_y: bool = False  # --tta-mirror-y flag
```

Pass it through to `evaluate_split` via config.

### Step 2: Create a mirror-y batch transform helper

Add a helper function in `trainer_runtime.py` (near `accumulate_eval_batch`):

```python
def mirror_y_batch(batch: SurfaceBatch) -> SurfaceBatch:
    """
    Return a new SurfaceBatch with y-axis mirrored geometry.
    
    Coordinate layout (from data/loader.py):
      surface_x: [xyz(0:3), normals(3:6), area(6)]
      surface_y: [surface_pressure(0), tau_x(1), tau_y(2), tau_z(3)]
      volume_x:  [xyz(0:3), sdf(3)]
      volume_y:  [volume_pressure(0)]
    
[...]

PR #496, opened 2026-05-03
tanjiro: homoscedastic uncertainty-weighted multitask loss

morganmcg1:
## Hypothesis

The SOTA stack optimizes a flat equally-weighted sum of surface and volume MSE. With 5 output channels (surface_pressure, tau_x, tau_y, tau_z, volume_pressure) that have widely different loss magnitudes and gradient scales, the optimizer must spread capacity across tasks that compete. tau_y and tau_z trail AB-UPT by ×2.5–2.8×, suggesting insufficient optimization pressure on those channels. Static loss weights (already shown ineffective in PRs #142, #454, #467) fail because they require hand-tuning and cannot adapt to where the model's uncertainty actually lies.

**Homoscedastic uncertainty weighting** (Kendall et al., CVPR 2018: "Multi-Task Learning Using Uncertainty to Weigh Losses in Deep Neural Networks") provides a principled adaptive solution: each task gets a learnable log-variance parameter `log_sigma_i`, and the loss becomes:

```
L = sum_i [ L_i / (2 * sigma_i^2) + log(sigma_i) ]
```

This is equivalent to maximum likelihood under a Gaussian with per-task variance `sigma_i^2`. Tasks with inherently high uncertainty (like tau_y/tau_z, which are dominated by complex separated-flow regions) get learnable attenuation. The `log(sigma_i)` regularizer prevents the trivial solution of setting all `sigma_i → ∞`.

Key advantages over static weighting:
1. **Adaptive** — `sigma_i` learns the right effective weight per task during training, not via grid search.
2. **Theoretically grounded** — directly maximizes joint log-likelihood.
3. **Low complexity** — adds only 5 learnable scalars (`log_sigma` for [surface_pressure, tau_x, tau_y, tau_z, volume_pressure]) with near-zero compute overhead.
4. **Orthogonal to all prior attempts** — doesn't change target distribution (unlike PR #481 log1p), doesn't add parameters to the model backbone (unlike PR #467), and addresses the loss formulation rather than the optimizer state.

Reference: Kendall, A., Gal, Y., & Cipolla, R. (2018). Multi-Task Learning Using Uncertainty to Weigh Losses in Deep Neural Networks. CVPR. https://arxiv.org/abs/1705.07115

## Instructions

All changes go in `target/train.py`. The existing `compute_loss_for_batch` function computes surface_loss and volume_loss separately — you need to decompose surface_loss into its 4 channel contributions and apply per-channel uncertainty weighting.

### Step 1: Add CLI flag

[...]

PR #489, opened 2026-05-03
thorfinn: volume-points curriculum 16k→65k ramp (4-stage)

morganmcg1:
## Hypothesis

Volume pressure (`vol_p`) is our primary laggard: test 12.19% vs AB-UPT 6.08% (×2× gap). We currently fix `--train-volume-points 65536` throughout all 12 epochs. This means the model is always learning from views that cover ~63% unique volume points per epoch per case (replacement sampling).

**Hypothesis:** Starting with a coarser view (fewer volume points per view) and progressively increasing it acts as an implicit curriculum: early epochs learn the low-frequency volumetric structure from a broader but sparser view; later epochs refine on dense, high-fidelity views. This mirrors resolution curricula in NeRF / 3D implicit function literature and has shown consistent benefits on noisy-point-cloud regression tasks.

**Prediction:** A 4-stage ramp `16k → 32k → 48k → 65k` (3 epochs each) should improve vol_p by exposing the model to more diverse spatial views early (wider coverage) while finishing on the dense 65k config that the current SOTA already converges on.

Note: fern (PR #480) is showing val_vol_p=5.30% at EP6 via the cosine EMA ramp (an orthogonal mechanism). If this curriculum variant also improves vol_p, they are composable.

## Instructions

Modify `train.py` to support a step-based volume-points curriculum. Add a `--vol-points-schedule` argument that accepts a colon-separated list of `epochs:points` pairs defining the volume point count to use from that epoch onward (inclusive). Example: `0:16384:3:32768:6:49152:9:65536` means:
- Epochs 0–2: 16384 vol points
- Epochs 3–5: 32768 vol points
- Epochs 6–8: 49152 vol points
- Epochs 9–11: 65536 vol points

Implementation guidance:
1. Parse the schedule in `main()` into a sorted list of `(start_epoch, n_points)` tuples.
2. At the start of each epoch in the training loop, look up the current epoch in the schedule and call `datamodule.set_train_volume_points(n_points)` (or equivalent — update the DataLoader's sampler parameters so the next epoch uses the new count). If the DataLoader cannot be reconstructed mid-run, recreate it from scratch at epoch boundaries — this is fine since the cost is one-time per epoch transition.
3. Only `train-volume-points` changes; `eval-volume-points` always stays fixed at 65536 for fair comparison.
4. Log `train/vol_points` as a W&B scalar each step so the schedule is traceable in the run.
5. `--train-volume-points` continues to work as before (constant schedule = no change for existing runs).

[...]

PR #488, opened 2026-05-03
alphonse: multi-sigma log_freq init for STRING-sep encoding

morganmcg1:
## Hypothesis

The current SOTA stack uses `StringSeparableEncoding` (PR #311) with a single
sigma initialization for all log_freq parameters: `log_freq = log(sigma)` (default sigma=1.0).
This warm-starts all basis frequencies at the same scale, meaning the learnable
per-axis frequencies all start near 1.0 Hz/unit and gradient descent must learn
spectral coverage from scratch.

**Automotive aerodynamics spans a wide range of spatial frequencies simultaneously:**
- Volume pressure (smooth wake, body scale): low frequency, large wavelength
- Surface pressure (A-pillar, front diffuser): medium frequency
- Tangential wall shear tau_y / tau_z (mirror wake, wheel arch, A-pillar vortex): HIGH frequency, sharp geometric features

The tau_y/tau_z gap to AB-UPT is 2.5× / 2.8× — the worst per-axis lag on the board.
PR #467 just proved this gap is NOT in the output head calibration (all scale parameters
converged to near-uniform 0.87 attenuation). The gap is upstream in the spectral representation.

**Hypothesis:** Initializing `log_freq` with a spread across multiple frequency octaves
([0.25, 0.5, 1.0, 2.0, 4.0]) will give the learnable encoding a diverse spectral
warm-start, reducing the gradient cost of learning high-frequency basis functions
(tau_y, tau_z, surface pressure near sharp edges) while retaining low-frequency
coverage for the volume pressure field.

This is related to the Neural Tangent Kernel + Fourier Features theory (Tancik et al.
2020): models with multi-scale RFF initialization converge faster and generalize better
on tasks with mixed spatial frequencies. The STRING-separable encoding is already a
learnable generalization of RFF — the initialization of `log_freq` controls the
frequency coverage before gradient descent.

**Implementation:** Add a new CLI flag `--rff-init-sigmas` that accepts a
comma-separated list of sigma values and distributes the per-axis `num_features`
uniformly across those frequency scales during `StringSeparableEncoding.__init__`.

## Instructions

This is a surgical change to `StringSeparableEncoding.__init__` in `target/model.py`
and a new CLI flag in `target/train.py`. The running experiment uses the EXACT
SOTA stack — the only change is `log_freq` initialization.

### Step 1: Add CLI flag to `train.py`

In the `Config` dataclass, add after `rff_sigma: float = 1.0`:

```python
[...]

PR #500, opened 2026-05-03
edward: surface curvature (mean H + Gaussian K) input features for tau_y/tau_z gap

morganmcg1:
## Hypothesis

**Surface curvature input features will close the tau_y / tau_z gap.**

The current SOTA (PR #387) reaches val_abupt=7.3816% with strong surface_pressure (4.44%) and increasingly strong volume_pressure (~4.7-5.2% across the fleet, all below AB-UPT 6.08%). The remaining laggards are **tau_y (9.11% vs AB-UPT 3.65%, ×2.5)** and **tau_z (10.27% vs AB-UPT 3.63%, ×2.8)**.

Wall shear stresses tau_y / tau_z are dominated by **boundary-layer separation, ridge edges, and high-curvature regions** (A-pillar, mirror, wheel arch). Three independent NEG results — per-axis output scaling (#467), tau_yz loss-weight at w=2.0 (#142) and w=1.5 (#454) — confirm the gap is **upstream / representational**, not a loss-balance issue. The model is missing geometric information.

xyz coordinates alone cannot reconstruct local geometric shape (convex/concave, principal directions, edge sharpness). The STRING-separable RFF encodes spatial position frequency content but is rotationally / curvature-blind. Adding **scalar curvature features** (mean curvature H, Gaussian curvature K) gives the model an explicit, rotation-invariant prior on where high-frequency tau components live.

This is **fundamentally different** from the negative tangent-frame INPUT features (#423): tangent frames are vector orientation features (n, t1, t2 basis); curvature H, K are **scalar geometric invariants** of the surface. They are also orthogonal to every active in-flight experiment (cosine EMA #480, log1p tau v2 #481, multi-sigma STRING #488, vol curriculum #489, signed-log #471, mlp_ratio=8 #458, uncertainty weighting #496, TTA mirror-y #499).

References:
- Meyer et al. "Discrete Differential-Geometry Operators for Triangulated 2-Manifolds" (Visualization & Math 2003) — robust per-vertex H and K from triangle one-rings.
- Bobenko & Schroeder, "Discrete Curvature Flow" — practical numerical recipes.
- For point clouds without explicit triangulation: PCA over k-NN normals → smaller eigenvalue ratios approximate principal curvatures.

### Goal
Add per-surface-point curvature features (mean H, Gaussian K, and optionally |H|, max(K,0), min(K,0)) as **additional input channels** to the surface branch only. The surface branch already supports `surface_extra_dim = max(0, surface_input_dim - space_dim)` in `model.py`, so the architecture machinery is in place — you only need to (a) compute curvature in the data pipeline, (b) thread a CLI flag, (c) bump `SURFACE_X_DIM`, (d) verify shapes.

### Step 1: Compute mean & Gaussian curvature from the source mesh
DrivAerML provides per-case STL/triangle meshes. Curvature is most reliable computed once **on the full triangle mesh** (offline or as a one-time per-case preprocessing step at dataset load), then **gathered to the sampled surface point indices**.

Use the discrete Laplace-Beltrami / cotangent formula (Meyer 2003) per vertex:
- Mean curvature normal:  K_H(v) = (1 / 2 A_mixed) * sum_{j in N(v)} (cot α_ij + cot β_ij) * (v_j - v) → H(v) = ||K_H(v)|| / 2
- Gaussian curvature:     K(v) = (2π - sum_{j} θ_j) / A_mixed
- A_mixed = mixed Voronoi/barycentric area as in Meyer §3.3 (clip obtuse triangles).

If meshes are not available at dataset time, fall back to **point-cloud PCA**: for each sampled point, take k=32 nearest neighbors, compute the local covariance, the smallest eigenvalue's eigenvector approximates the normal, and the ratio λ1/λ2 of the two larger eigenvalues approximates anisotropy (a curvature proxy). Prefer the mesh-based method when triangles are available — it is exact and well-conditioned.

Cache curvature arrays per-case to disk (e.g. `<case>/curvature.npy` shape `[n_surface_vertices, 2]` for `[H, K]`). Recompute is expensive (one-time) but loading is essentially free.

### Step 2: Wire it into `data/loader.py`
- After sampling surface points, gather `H, K` at the sampled vertex indices (or interpolate from triangle barycentric weights if you sample over faces).
- Apply **robust normalization**: `H_norm = sign(H) * log1p(|H|)`, same for K. This compresses the heavy-tailed curvature distribution at sharp edges.
- Append `[H_norm, K_norm]` as columns to `surface_x` so the per-point feature dim becomes `xyz (3) + H (1) + K (1) = 5` (or 6 if you also add `|H|`).

### Step 3: Add a CLI flag in `train.py`
Add a dataclass field on the config:
```python
use_surface_curvature_features: bool = False
surface_curvature_neighbors: int = 32   # for PCA fallback only
```
Plumb `use_surface_curvature_features` to the dataloader so disabling it is zero-cost.

### Step 4: Update `SURFACE_X_DIM` and model wiring
- Where `SURFACE_X_DIM` is defined (search for it in `data/loader.py` or `train.py`), make it dependent on the flag: `SURFACE_X_DIM = 3 + (2 if use_surface_curvature_features else 0)`.
- The model already handles `surface_extra_dim` via a separate input-projection path (see `model.py` around the `surface_extra_dim = max(0, surface_input_dim - space_dim)` line). **Do not modify the STRING-sep encoding** — it operates only on xyz (`space_dim=3`). The extra channels feed through the linear input projection unchanged.
- Verify with a tiny debug pass that surface_x has the expected last-dim = 5 when the flag is on.

### Step 5: Run the experiment
Single run with curvature ON, otherwise SOTA config:
```bash
  --agent edward --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --rff-num-features 16 \
  --use-surface-curvature-features \
  --wandb-group edward-curvature \
  --wandb-name edward/curvature-HK-pr-NEW
```

### Reporting
- After EP12, comment back with: full val_abupt curve, val per-axis (tau_x, tau_y, tau_z, surface_p, vol_p), test_abupt + test per-axis evaluated **at the best-val checkpoint**, the W&B run id, and any hiccups in the curvature precomputation.
- If first-val (EP1) is more than 5pp above the SOTA arm at the equivalent step, ping me — the input-projection might need re-init.

### Gates
- **EP5 gate:** val_abupt < 8.9% OR slope < -0.4 pp/1k steps (slope-waiver auto-granted if curvature precompute slowed first epoch).
- **EP8 gate:** val_abupt < 8.0%.
- **Win condition:** val_abupt < **7.3816%** (current SOTA, PR #387).
- **Stretch win:** test tau_y < 8.5% AND test tau_z < 9.5% even if headline misses — that confirms the geometric prior helps the right axes and we'd compose it.

## Baseline

Current SOTA — **PR #387 alphonse (feat16 RFF + QK-norm + STRING-sep)**, val_abupt **7.3816%** (EP11), W&B run `wj6mn6ve`, group `alphonse-rff-sweep`.

| Metric | val EP11 | test (best-val ckpt) | AB-UPT |
|---|---:|---:|---:|
| `abupt` | **7.3816%** | 8.5936% | — |
| `surface_pressure` | — | 4.4377% | 3.82% |
| `wall_shear` | — | 7.9989% | 7.29% |
| `volume_pressure` | — | 12.1885% | 6.08% |
| `tau_x` | — | 6.9622% | 5.35% |
| **`tau_y`** | — | **9.1058%** | **3.65%** |
| **`tau_z`** | — | **10.2736%** | **3.63%** |

**Reproduce SOTA (without curvature flag):**
```bash
  --agent alphonse --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --rff-num-features 16 \
  --wandb-group alphonse-rff-sweep \
  --wandb-name alphonse/feat16-qknorm-string-sep-pr387
```

### DIFF
--- a/research/CURRENT_RESEARCH_STATE.md
+++ b/research/CURRENT_RESEARCH_STATE.md
@@ -1,6 +1,6 @@
 # SENPAI Research State — `tay` (DrivAerML / DDP8)
 
-- **Date:** 2026-05-03 (post W&B snapshot, Round 25 active — PR #489 assigned; 8 students active)
+- **Date:** 2026-05-03 (post W&B snapshot, late EP9 region — 8 students active, 0 idle)
 - **Branch:** `tay`
 - **Target repo:** `morganmcg1/DrivAerML`
 - **W&B project:** `wandb-applied-ai-team/senpai-v1-drivaerml-ddp8`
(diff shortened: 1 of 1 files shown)


morganmcg1:
ADVISOR: Closing PR #500 because Closing PR #500 — surf-vol-xattn experiment finished with val_abupt=7.6695%, a regression of +0.2879pp vs SOTA (7.3816%). The cross-attention between surface and volume representations did not improve performance. The added complexity of cross-attention (more parameters, slower training) was not justified by the results. Closing as a dead end. Thank you edward for running this experiment — the negative result helps us understand that the surface/volume interaction mechanism we have is already sufficient at this scale..

morganmcg1:
STUDENT: Hi advisor — PR #500 has been closed but the closure comment text references **surf-vol-xattn** (`val_abupt=7.6695%`, surface/volume cross-attention) while PR #500 is actually the **surface curvature (mean H + Gaussian K) input features** experiment for the tau_y/tau_z gap. Looks like the comment may have been copy-pasted from a different PR.

Status of this PR's work when the close came in:
- All implementation done on `edward/surface-curvature-input-features` and pushed locally (not committed to remote yet): `data/curvature.py` (cotangent-style point-cloud Weingarten LSQ), wiring through `data/loader.py`, `train.py` CLI flag `--use-surface-curvature-features`, `trainer_runtime.py` plumbing, and a one-time `scripts_precompute_curvature.py` driver.
- Curvature precompute had reached **414 / 484** cases (~85%) on GPU 0 when I killed it after seeing the close.
- Smoke tests on synthetic shapes (sphere / cylinder / saddle) pass; cached HK arrays are finite with a reasonable distribution; full forward pass with `surface_input_dim=9` works end-to-end.

Could you confirm whether you intended to close PR #500 (curvature) or whether the close was meant for a different PR? Happy to resume — the precompute can finish in ~6 min and the training run is ready to launch on 8 GPUs. If the close was intentional, no further action needed and I'll discard the local changes.

PR #517, opened 2026-05-03
askeladd: Lion (lr, clip) joint sweep at fixed lr·clip product

morganmcg1:
## Hypothesis

Your previous PR #431 produced a clear structural finding: with Lion @ clip=0.5, **97.5% of optimizer steps are clipped** (pre-clip p50 grad-norm = 1.1, threshold = 0.5). In this regime gradient clipping is **not** a "rare-spike defender" — it acts as a **constant multiplicative scaler on the effective learning rate**.

That suggests a sharper hypothesis: the truly relevant knob in our regime is the product `lr·clip` (the effective per-step magnitude when the clip is biting), and (lr, clip) are **redundant degrees of freedom** along iso-product curves. The current SOTA-recipe `lr=1e-4, clip=0.5` may not be the optimum in this product-space.

This is testable cheaply with a 4-arm sweep that walks an iso-product line and an off-iso reference, all at the same total per-step magnitude budget:

| Arm | lr     | clip | lr·clip | rationale |
|---|---|---|---|---|
| A | 1e-4   | 0.5  | 5e-5    | control = current SOTA recipe |
| B | 5e-5   | 1.0  | 5e-5    | iso-product, looser clip → fewer clipped steps, more "raw" Lion sign-update info preserved |
| C | 2e-4   | 0.25 | 5e-5    | iso-product, tighter clip → more aggressive clip, smaller effective update at peak LR |
| D | 1e-4   | 1.0  | 1e-4    | off-iso (2× product) — does loosening clip while holding lr at SOTA win or lose? |

If the "clip is just an LR scaler" finding is correct, A/B/C should be near-identical (within noise) and D should be ~2× the effective LR. If A wins clearly we have evidence Lion's sign-update has structure that benefits from a *specific* `lr·clip` ratio, not just the product. Either outcome is publishable inside this programme.

## Instructions

**Run all 4 arms in parallel on your single 4-GPU pod via 4 sequential DDP-4 launches** — each arm gets the full 6h `SENPAI_TIMEOUT_MINUTES`, and we do the arms one at a time on the same 4 GPUs. (You ran arm-by-arm last time; keep that pattern.)

**No code changes — pure CLI sweep.** The `--lr` and `--clip-grad-norm` flags are both already wired.

**Common config (all arms):**

```bash
cd target/
torchrun --standalone --nproc_per_node=4 train.py \
  --agent askeladd \
  --optimizer lion \
  --weight-decay 1e-4 \
  --no-compile-model \
  --batch-size 4 \
[...]

PR #529, opened 2026-05-03
violet: stream-normal per-point ws loss weight for ws_y/z gap attack

morganmcg1:
## Hypothesis

**Per-point stream-normal-conditioned wall-shear loss weighting reduces the dominant ws_y/z gap (currently 2.5–2.9× above AB-UPT targets).**

The physics motivation: wall shear stress in the cross-flow directions (τ_y, τ_z) is largest on surface panels that are **near-perpendicular to the freestream** (e_x = [1,0,0] in car coordinates). These are underbody panels, wheel arches, and side panels. Points where the surface normal is perpendicular to the freestream carry the full cross-flow separation signal; points where the normal is parallel to freestream (front/rear face) carry the streamwise signal which is better-predicted.

**Proposed per-point weighting:**

At each surface point i, compute:
```
n_cross_i = sqrt(n_y_i^2 + n_z_i^2)  # cross-flow component of unit surface normal
w_i = 1 + gamma * n_cross_i            # weight: 1 at streamwise-facing points, 1+gamma at cross-flow panels
```

Then in the wall-shear MSE/L2 loss, multiply the per-point squared residual by `w_i`:
```
loss_ws = mean(w_i * |tau_pred_i - tau_true_i|^2, over all points i)
```

This is physics-informed because:
1. Cross-flow-facing panels (high `n_cross`) are where the model struggles most (τ_y/z errors 2.5–2.9×)
2. The weight is derived from the surface geometry (normals are available as `batch.surface_x[..., 3:6]`)
3. No architecture change — purely a loss modification

This is orthogonal to the channel-level `--wallshear-y-weight` / `--wallshear-z-weight` flags (which apply uniform scaling across all points for a given axis). This approach applies per-point differential weighting within each axis.

## Baseline

Current yi SOTA: **val_abupt 9.039%** (PR #309 thorfinn). Active merge bar: val_abupt < 9.039%.

| Metric | Yi baseline (PR #309) | AB-UPT target | Ratio |
|---|---:|---:|---:|
| val_abupt | **9.039%** | — | — |
| wall_shear_y (test) | ~10.5% | 3.65% | **2.53×** |
| wall_shear_z (test) | ~11.8% | 3.63% | **2.88×** |
| volume_pressure (test) | ~13.5% | 6.08% | **2.05×** |

### Step 1 — Add `--stream-normal-ws-gamma` flag

Add one new config field and CLI flag:
```python
stream_normal_ws_gamma: float = 0.0   # disabled if 0.0
```
CLI: `--stream-normal-ws-gamma GAMMA`

### Step 2 — Implement per-point stream-normal weights in the wall-shear loss

Find the wall-shear loss computation in `train.py` (search for `wall_shear` near the surface loss block). Surface normals are available as `batch.surface_x[..., 3:6]` (shape `[B, N_surface, 3]`, unit normals in car-frame xyz).

Replace the existing wall-shear per-point MSE with a weighted version:
```python
if config.stream_normal_ws_gamma > 0.0:
    normals = batch.surface_x[..., 3:6].float()  # [B, N, 3]
    # cross-flow magnitude (y-z components of normal)
    n_cross = (normals[..., 1:3] ** 2).sum(dim=-1, keepdim=True).sqrt()  # [B, N, 1]
    # per-point weight: 1.0 at streamwise-facing, 1+gamma at cross-flow-facing
    ws_point_weight = 1.0 + config.stream_normal_ws_gamma * n_cross  # [B, N, 1]
    # normalize so mean weight = 1.0 (preserves loss scale)
    ws_point_weight = ws_point_weight / ws_point_weight.mean()
    # apply to per-point squared residual before mean
    ws_pred = outputs.surface_tau  # [B, N, 3]
    ws_true = batch.surface_tau    # [B, N, 3]
    ws_residual_sq = (ws_pred - ws_true) ** 2  # [B, N, 3]
    ws_loss = (ws_point_weight * ws_residual_sq).mean()
else:
    # existing implementation unchanged
    ...
```

**IMPORTANT**: Check the actual variable names in `train.py` — `outputs.surface_tau` / `batch.surface_tau` may differ from the above placeholder names. Use the existing unweighted wall-shear loss block as a template and wrap it with the weighting.

Also check: if tangential loss is separately computed (`use_tangential_wallshear_loss`), apply the same `ws_point_weight` weighting to the tangential loss component too. Keep the weight normalization (`/ weight.mean()`) to avoid altering the overall loss scale.

### Step 3 — Add W&B diagnostic logging

When gamma > 0, log the mean and std of `ws_point_weight` every `gradient_log_every` steps:
```python
wandb.log({"train/stream_normal_ws_weight_mean": ws_point_weight.mean().item(),
           "train/stream_normal_ws_weight_std": ws_point_weight.std().item()}, step=global_step)
```
This verifies the weighting is active and checks for degenerate distributions.

### Step 4 — Two-arm DDP-4 run

**Arm A** — gamma=1.0 (moderate cross-flow emphasis):
```bash
  --agent violet \
  --wandb-group violet-r24-stream-normal-ws \
  --wandb-name "violet-r24-arm-a-gamma1.0" \
  --optimizer lion \
  --validation-every 1 \
  --epochs 50 \
  --stream-normal-ws-gamma 1.0
```

**Arm B** — gamma=3.0 (stronger cross-flow emphasis):
```bash
  --agent violet \
  --wandb-group violet-r24-stream-normal-ws \
  --wandb-name "violet-r24-arm-b-gamma3.0" \
  [... all same flags ...] \
  --stream-normal-ws-gamma 3.0
```

Run both arms sequentially (or on a fresh pod for Arm B if Arm A completes and pod budget is exhausted). If only one arm fits in budget, run Arm A (gamma=1.0) first.

### Step 5 — Reporting

For each arm, report:

| Metric | Control (est. from PR #309) | Arm A (γ=1.0) | Arm B (γ=3.0) |
|---|---:|---:|---:|
| val_abupt | 9.039% | | |
| wall_shear (val) | | | |
| ws_y (val) | | | |
| ws_z (val) | | | |
| surface_p (val) | | | |
| vol_p (val) | | | |
| test_abupt | | | |

Plus: confirm `stream_normal_ws_weight_mean ≈ 1.0` and `_std` magnitude (should be O(0.1–0.5) for gamma=1).

## What success looks like

- Either arm reduces ws_y or ws_z by ≥0.5pp relative to the matched no-weight baseline → promising, push to full budget
- val_abupt improvement ≥0.2pp and below 9.039% merge bar → MERGE
- No significant regression on surface_p or vol_p (collateral damage check)

## Notes

- Normalize `ws_point_weight` so its mean = 1.0. Without normalization, gamma=3 would raise overall loss scale by ~(1 + 3·0.5) = 2.5× (if mean n_cross ≈ 0.5), which would need a compensating LR reduction. Normalization avoids this.
- The existing `--wallshear-y-weight` / `--wallshear-z-weight` flags operate at the channel level (scalar per axis). Keep them at 2.0/2.0 as specified above — this is orthogonal (axis-level scaling × point-level weighting).
- If the surface_tau batch attribute name differs from `batch.surface_tau`, trace back from the existing wall-shear loss computation (search `loss_ws` or the Huber loss block from PR #317's code).


morganmcg1:
ADVISOR: approved. Stream-normal-conditioned per-point ws weighting is a clean physics-informed loss reformulation, and the gap analysis is right — wall_shear_y (2.53×) and wall_shear_z (2.88×) are the largest channel gaps to AB-UPT.

**Updated baseline:** the active merge bar has moved. True yi SOTA is now **val_abupt 7.1792%** (PR #489 thorfinn, vol-points curriculum, run `r5rw40rn`), not 9.039%. Adjust your reporting table — your control row should be the matched standard-config baseline, not a stale figure.

**Implementation note — surface_x layout:** before launching, verify the normal channels are at `surface_x[..., 3:6]` by reading the dataset loader / batch dataclass. Some recent PRs add channels (cross-flow-index, etc.); the slice may have shifted. Trace it from the dataset class definition rather than assuming.

**On normalization:** good call to normalize `ws_point_weight` so its mean = 1.0 — otherwise gamma=3.0 inflates total loss scale and would couple to LR. Log both `weight_mean` (should sit at 1.0) and `weight_std` (should be O(0.15) for γ=1, O(0.45) for γ=3 if normals are roughly isotropic).

**Tangential loss interaction:** if `--use-tangential-wallshear-loss` is enabled in your run (it isn't in your launch command, good), the same point-weight should apply to that branch. Since you're not setting it, this is moot — but document the intended composition for the next round.

**Two arms is correct:** γ=1.0 (Arm A) is the safe baseline, γ=3.0 (Arm B) is the bold variant. If A wins and B loses, the right next step is a γ=2.0 mid-point. If both lose modestly, this is null and we move on.

**Reporting:** include test_abupt_axis_mean (single number, primary metric) at the best-val EMA checkpoint, plus the per-axis test breakdown so we can see whether ws_y/ws_z actually moved.

Ready when you are. Mark ready-for-review after both arms complete.

morganmcg1:
ADVISOR CORRECTION: my previous comment cited val_abupt 7.1792% from PR #489 as the active merge bar. That was an error — **PR #489 (thorfinn, vol-points curriculum) was merged into the `tay` branch, not `yi`**. The active yi merge bar is unchanged: **val_abupt 9.039%** (PR #309 thorfinn grad-clip=0.5).

7.1792% (PR #489) and 7.546% (PR #311) are aspirational tay-branch numbers, not currently reproducible on the yi codebase. Beat 9.039% to merge. Apologies for the confusion.

morganmcg1:
ADVISOR: Closing PR #529 because ADVISOR: Closing — null result. Arm A gamma=1.0 val=11.468%, Arm B gamma=3.0 val=10.900%; both well above the active yi merge bar of 9.032% (PR #517). Stream-normal per-point ws loss weighting did not target the τ_y/τ_z gap as hypothesized — neither arm shifts the primary metric in the right direction. Hypothesis falsified at this loss-weight formulation. The τ_y/τ_z structural gap remains: PR #541 (askeladd, streamline-aligned target frame) and PR #540 (emma, GradNorm adaptive task weighting) are the next attacks on this gap. Thanks for the clean γ sweep..

PR #511, opened 2026-05-03
extended epoch training EP15 cosine on current SOTA stack

morganmcg1:
## Hypothesis

The validation curve in PR #488 (current SOTA, run `ki2q9ko9`) was **still descending at EP11** — the terminal epoch. This means the model had not converged and was leaving performance on the table. Extending the cosine LR schedule with `--lr-cosine-t-max 15` and training to EP15 should allow the model to find a lower minimum before LR decays to near-zero. This is a clean single-variable test: same architecture and all hyperparameters identical to SOTA, only the cosine decay length and epoch count change.

Historical context: PR #393 (edward) tried extended epochs on an older STRING-sep stack (no QK-norm, no multi-sigma init) and was closed. The current SOTA stack is substantially stronger — multi-sigma init dramatically changes the spectral coverage of the encoding — so the optimal epoch budget may have shifted.

Expected gain: if the val curve was still dropping at EP11 at the same rate as it descended from EP8→EP11, extending to EP15 could yield another 0.03–0.10pp improvement, which is meaningful at this precision level.

## Instructions

Run a single arm with the current SOTA config extended to 15 epochs, using a cosine schedule that extends proportionally:

```bash
torchrun --standalone --nproc_per_node=8 train.py \
  --agent edward --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --no-compile-model --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 4 --model-hidden-dim 512 --model-heads 4 --model-slices 128 \
  --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 \
  --pos-encoding-mode string_separable --use-qk-norm \
  --rff-num-features 16 \
  --lr-cosine-t-max 15 \
  --epochs 15 \
  --wandb-group edward-extended-cosine --wandb-name edward-extended-ep15
```

**Important:** The multi-sigma STRING-sep init from PR #488 is not exposed as a CLI flag — it was implemented directly in model.py by alphonse. **You must obtain and apply the same multi-sigma init patch from alphonse's PR #488 student branch (branch `alphonse/multi-sigma-rff-init`) to your model.py before running.** This is critical — without the multi-sigma init you will not be building on the current SOTA.

[...]

PR #278, opened 2026-05-01
[gilbert] Train-time mirror-symmetry augmentation for wsy/wsz

morganmcg1:
## Hypothesis

DrivAerML vehicles are approximately symmetric about the longitudinal (xz) plane. Wall-shear y (`wsy`) is the hardest metric by far (val=9.10% vs AB-UPT target 3.65%) and is the primary bottleneck to beating the 7.21% abupt baseline. Under a y-mirror transform, the wall-shear vector transforms as:

```
(wsx, wsy, wsz) → (wsx, -wsy, wsz)
```

If we randomly flip each training sample about the y-axis (negating y-coords of points and negating the wsy component of the wall-shear target) we give the model **2× more diverse antisymmetric shear signal** at no extra data collection cost. This is the train-time version of frieren's TTA idea (PR #256) — training with the augmentation should improve the model's intrinsic wsy representation, not just average-out at eval time. Training augmentation and TTA are orthogonal — this experiment tests the train-time improvement.

The mechanism: the model currently sees each geometry once per epoch; augmented training sees both the original and its mirror, doubling the effective variety of anti-symmetric shear patterns. The surface-pressure and volume-pressure channels are scalars and are sign-invariant — no correction needed. Wall-shear is a 3-vector; only wsy flips.

## Instructions

Implement a `--mirror-aug` flag in `target/train.py`. This is a pure data augmentation in the training collation loop — no model changes.

### Step 1: CLI flag

Add `--mirror-aug` / `--no-mirror-aug` (default `False`) to the argparser. Log the flag value to W&B config.

### Step 2: In the training DataLoader / batch collation

For each training sample independently, with probability 0.5:
1. Negate the y-coordinate of all point cloud coordinates (surface + volume).
2. Negate the wsy channel (channel index 1 of wall-shear, i.e. `wall_shear[:, 1]`).
3. Leave all other channels (wsx, wsz, surface_pressure, volume_pressure) unchanged.

Implement this as a per-sample transform in the dataset `__getitem__` or as a batched collate step. The easiest approach is to add a `RandomMirrorY` transform class that wraps the existing dataset and applies the flip independently per sample with p=0.5.

Verify the flip is correct by checking that before/after values satisfy: y_coords_flipped = -y_coords_orig; wsy_flipped = -wsy_orig; all other coords/channels unchanged.

### Step 3: Eval path

**Do NOT apply augmentation during eval.** The eval path must run the original (un-flipped) inputs only.

### Step 4: Training run

[...]

PR #510, opened 2026-05-03
surface-loss-weight sweep 0.5/2.0/4.0 on current SOTA stack

morganmcg1:
## Hypothesis

Surface pressure (`surface_pressure`) is currently val=4.805% against an AB-UPT reference of 3.82% — a 0.985pp gap that is the second-largest remaining gap to reference after wall_shear. The loss weighting between surface and volume terms has not been tuned on the current SOTA stack (STRING-sep + QK-norm + multi-sigma init + feat16 RFF). The last surface-loss-weight experiment (PR #322, `--surface-loss-weight 2.0`) was run in an early round on a substantially weaker stack and was closed. We expect that tuning `--surface-loss-weight` on the current stack may push surface_pressure down toward 4.0-4.4% range, and that the optimal value may differ from what was tried before due to the improved model capacity.

## Instructions

Run **three arms** of the surface-loss-weight sweep on the current SOTA stack. Use `--wandb-group alphonse-slw-sweep` to group them in W&B.

**Arm A** (`--surface-loss-weight 0.5`):
```bash
torchrun --standalone --nproc_per_node=8 train.py \
  --agent alphonse --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --no-compile-model --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 4 --model-hidden-dim 512 --model-heads 4 --model-slices 128 \
  --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 \
  --pos-encoding-mode string_separable --use-qk-norm \
  --rff-num-features 16 \
  --surface-loss-weight 0.5 \
  --wandb-group alphonse-slw-sweep --wandb-name alphonse-slw-0.5
```

**Arm B** (`--surface-loss-weight 2.0`):
```bash
torchrun --standalone --nproc_per_node=8 train.py \
  --agent alphonse --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --no-compile-model --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 4 --model-hidden-dim 512 --model-heads 4 --model-slices 128 \
  --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 \
[...]

PR #536, opened 2026-05-03
nezuko: y-mirror training augmentation for tau_y/tau_z gap

morganmcg1:
## Hypothesis

The DrivAerML cars are geometrically symmetric about the xz-plane (y=0). A y-mirror (flip about xz-plane) is an exact physics symmetry: reflected surface pressure `cp` is unchanged, reflected wall shear `tau_y` negates (antisymmetric), reflected `tau_x` and `tau_z` are unchanged. Training with y-mirror augmentation therefore effectively doubles the training dataset from 400 to 800 cars at zero data-collection cost.

Our current tau_y gap is 8.77% vs AB-UPT 3.65% — a 2.4× shortfall. tau_y is the antisymmetric channel under y-flip; a sign-correct mirror augmentation provides the model with balanced positive- and negative-tau_y training signal, which should reduce the model's bias on this channel. tau_z (orthogonal to the flip axis) and surface_pressure should be unaffected or slightly improved from the doubled dataset.

**Key insight:** This is NOT the same as the inference-time TTA that failed in PR #499. That experiment trained *without* augmentation but applied the average at inference, which picked the wrong checkpoint. Here we augment during *training* so the model learns invariance directly from data. Training augmentation and inference TTA are fundamentally different interventions.

**Precedent:** On the `bengio` track, y-mirror training augmentation (PR #278) improved wall shear accuracy. We are re-implementing it from scratch on `tay` as the code was never ported. The commit `b1dc4ca` on the `bengio` track shows the correct sign-flip channels.

## Instructions

Add `--mirror-aug-prob P` CLI flag (float in [0, 1], default 0.0 = off) and apply per-sample independent y-mirror augmentation in the `train_loss` function **during training only** (never applied in eval/test).

### Implementation

Add the following to `train.py`:

**1. Add CLI argument** (in `parse_args`):
```python
parser.add_argument("--mirror-aug-prob", type=float, default=0.0,
    help="Per-sample y-mirror augmentation probability during training (0.0 = disabled)")
```

**2. Add to Config dataclass:**
```python
mirror_aug_prob: float = 0.0
```

**3. Add augmentation function** (near the top of train.py, after imports):
```python
def apply_mirror_y_aug(batch, prob: float) -> object:
    """Randomly flip each sample about the xz-plane (y -> -y) with given probability.
    
    Sign conventions (from data/loader.py):
      surface_x dim layout: [x(0), y(1), z(2), nx(3), ny(4), nz(5), area(6)]
      surface_y dim layout: [cp(0), tau_x(1), tau_y(2), tau_z(3)]
      volume_x dim layout:  [x(0), y(1), z(2), sdf(3)] (4-dim, y at index 1)
    
    Under y-reflection (xz-plane):
      surface_x[:, :, 1] (y pos)     -> negate
      surface_x[:, :, 4] (ny normal) -> negate
      surface_y[:, :, 2] (tau_y)     -> negate   (antisymmetric)
      volume_x[:, :, 1]  (y pos)     -> negate
      all other channels are sign-invariant (cp, tau_x, tau_z, area, nx, nz, sdf unchanged)
    """
    if prob <= 0.0:
        return batch
    
    B = batch.surface_x.shape[0]
    flip_mask = torch.rand(B, device=batch.surface_x.device) < prob  # [B]
    
    if not flip_mask.any():
        return batch
    
    # Clone only the tensors we'll modify (avoid in-place on original batch)
    surface_x = batch.surface_x.clone()
    surface_y = batch.surface_y.clone()
    volume_x = batch.volume_x.clone()
    
    # Apply flip to selected samples: expand mask to [B, N, 1] for broadcasting
    flip_3d = flip_mask[:, None, None]  # [B, 1, 1]
    
    # surface_x: negate y (col 1) and ny (col 4)
    surface_x[:, :, 1] = torch.where(flip_3d.squeeze(-1), -surface_x[:, :, 1], surface_x[:, :, 1])
    surface_x[:, :, 4] = torch.where(flip_3d.squeeze(-1), -surface_x[:, :, 4], surface_x[:, :, 4])
    
    # surface_y: negate tau_y (col 2 only — antisymmetric under y-flip)
    surface_y[:, :, 2] = torch.where(flip_3d.squeeze(-1), -surface_y[:, :, 2], surface_y[:, :, 2])
    
    # volume_x: negate y (col 1)
    volume_x[:, :, 1] = torch.where(flip_3d.squeeze(-1), -volume_x[:, :, 1], volume_x[:, :, 1])
    
    # Return a new batch-like object with augmented tensors
    # Replace the relevant fields using dataclasses.replace or manual construction
    import dataclasses
    return dataclasses.replace(batch, surface_x=surface_x, surface_y=surface_y, volume_x=volume_x)
```

**4. Modify `train_loss` to accept and apply augmentation:**

Change the signature to add `mirror_aug_prob: float = 0.0`:
```python
def train_loss(
    model: nn.Module,
    batch,
    transform: TargetTransform,
    device: torch.device,
    amp_mode: str,
    *,
    surface_loss_weight: float = 1.0,
    volume_loss_weight: float = 1.0,
    mirror_aug_prob: float = 0.0,
    batch = batch.to(device)
    # Apply y-mirror augmentation BEFORE computing targets (so flip applies to both x and y)
    if mirror_aug_prob > 0.0:
        batch = apply_mirror_y_aug(batch, mirror_aug_prob)
    surface_target = transform.apply_surface(batch.surface_y)
    ...
```

**5. Pass the flag through in the training call** (find where `train_loss` is called in the training loop and add `mirror_aug_prob=config.mirror_aug_prob`).

**6. Log the augmentation fraction** for monitoring:
```python
# Inside train_loss, after flip_mask computation:
logs["train/mirror_aug_frac"] = float(flip_mask.float().mean().item())
```

### Run command

```bash
  --agent nezuko --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --mirror-aug-prob 0.5 \
  --wandb-group nezuko-mirror-y-aug --wandb-name nezuko-mirror-aug-p50
```

**Note:** Use `--train-volume-points 65536` (constant, not `--vol-points-schedule`) to match the SOTA #511 stack. The `--eval-volume-points` applies to validation — **do NOT apply mirror-aug during eval/test**, the augmentation is training-only.

## Gate structure

- **EP3 gate**: val_abupt ≤ 14.0%
- **EP5 gate**: val_abupt ≤ 9.2% (sanity: augmentation should not hurt)
- **EP8 gate**: val_abupt ≤ 7.8%
- **Win condition**: val_abupt < **7.0134%** (PR #511 SOTA)

## Baseline

**Current SOTA: edward PR #511, W&B run `5o7jc7wi`, val_abupt=7.0134%, test_abupt=8.3130% (EP13)**

| Metric | PR #511 (SOTA) | AB-UPT ref |
|---|---:|---:|
| val_abupt | 7.0134% | — |
| val_surface_pressure | 4.5104% | 3.82% |
| val_wall_shear | 7.9650% | 7.29% |
| val_volume_pressure | 4.2168% | 6.08% (BEATEN) |
| val_tau_x | 7.0053% | 5.35% |
| **val_tau_y** | **8.7717%** | **3.65%** |
| **val_tau_z** | **10.5629%** | **3.63%** |
| test_abupt | 8.3130% | — |

tau_y (8.77%) and tau_z (10.56%) are 2.4× and 2.9× above the AB-UPT reference. These are the primary open problem.

## Monitoring

During training, monitor `train/mirror_aug_frac` to confirm augmentation is firing (~0.5 expected). At EP3, report:
1. val_abupt vs SOTA EP3 trajectory (SOTA was ~14% at EP3)
2. Per-axis breakdown: specifically val_tau_y and val_tau_z
3. `train/mirror_aug_frac` average

The critical test is whether tau_y closes faster (per-epoch improvement rate) than the SOTA baseline's trajectory.

### DIFF
--- a/train.py
+++ b/train.py
@@ -82,6 +82,7 @@ class Config:
     validation_every: int = 1
     surface_loss_weight: float = 1.0
     volume_loss_weight: float = 1.0
+    mirror_aug_prob: float = 0.0
     manifest: str = "data/split_manifest.json"
     data_root: str = ""
     output_dir: str = "outputs/drivaerml"
@@ -223,6 +224,45 @@ def build_model(config: Config) -> SurfaceTransolver:
     )
 
 
+def apply_mirror_y_aug(batch, prob: float) -> tuple[object, float]:
+    """Per-sample y-mirror (xz-plane) augmentation.
+
+    Sign conventions (from data/loader.py):
+      surface_x layout: [x(0), y(1), z(2), nx(3), ny(4), nz(5), area(6)]
+      surface_y layout: [cp(0), tau_x(1), tau_y(2), tau_z(3)]
+      volume_x  layout: [x(0), y(1), z(2), sdf(3)]
+
+    Under y-reflection only the y position, y normal, and tau_y target flip sign;
+    cp / tau_x / tau_z / area / nx / nz / sdf and volume_y (scalar pressure) are
+    invariant. Returns the (possibly augmented) batch and the realized fraction
+    of flipped samples for logging.
+    """
+    if prob <= 0.0:
+        return batch, 0.0
+
+    B = batch.surface_x.shape[0]
+    flip_mask = torch.rand(B, device=batch.surface_x.device) < prob  # [B]
+    flip_frac = float(flip_mask.float().mean().item())
+    if not bool(flip_mask.any().item()):
+        return batch, flip_frac
+
+    sign = torch.where(flip_mask, -1.0, 1.0).to(batch.surface_x.dtype)  # [B]
+    surface_sign = sign[:, None]  # [B, 1]
+    volume_sign = sign[:, None]   # [B, 1]
+
+    surface_x = batch.surface_x.clone()
+    surface_y = batch.surface_y.clone()
+    volume_x = batch.volume_x.clone()
+
+    surface_x[:, :, 1] = surface_x[:, :, 1] * surface_sign
+    surface_x[:, :, 4] = surface_x[:, :, 4] * surface_sign
+    surface_y[:, :, 2] = surface_y[:, :, 2] * surface_sign
+    volume_x[:, :, 1] = volume_x[:, :, 1] * volume_sign
+
+    import dataclasses
+    return dataclasses.replace(batch, surface_x=surface_x, surface_y=surface_y, volume_x=volume_x), flip_frac
+
+
 def train_loss(
     model: nn.Module,
     batch,
@@ -232,8 +272,12 @@ def train_loss(
     *,
     surface_loss_weight: float = 1.0,
     volume_loss_weight: float = 1.0,
+    mirror_aug_prob: float = 0.0,
 ) -> tuple[torch.Tensor, dict[str, float]]:
     batch = batch.to(device)
+    mirror_aug_frac = 0.0
+    if mirror_aug_prob > 0.0:
+        batch, mirror_aug_frac = apply_mirror_y_aug(batch, mirror_aug_prob)
     surface_target = transform.apply_surface(batch.surface_y)
     volume_target = transform.apply_volume(batch.volume_y)
     with autocast_context(device, amp_mode):
@@ -255,6 +299,7 @@ def train_loss(
         "volume_loss": float(volume_loss.detach().cpu().item()),
         "surface_loss_weighted": float(weighted_surface_loss.detach().cpu().item()),
         "volume_loss_weighted": float(weighted_volume_loss.detach().cpu().item()),
+        "mirror_aug_frac": mirror_aug_frac,
     }
 
 
@@ -501,6 +546,7 @@ def main(argv: Iterable[str] | None = None) -> None:
                     config.amp_mode,
                     surface_loss_weight=config.surface_loss_weight,
                     volume_loss_weight=config.volume_loss_weight,
+                    mirror_aug_prob=config.mirror_aug_prob,
                 )
                 optimizer.zero_grad(set_to_none=True)
                 global_step += 1
@@ -537,6 +583,7 @@ def main(argv: Iterable[str] | None = None) -> None:
                             "train/volume_loss": batch_loss_metrics["volume_loss"],
                             "train/surface_loss_weighted": batch_loss_metrics["surface_loss_weighted"],
                             "train/volume_loss_weighted": batch_loss_metrics["volume_loss_weighted"],
+                            "train/mirror_aug_frac": batch_loss_metrics["mirror_aug_frac"],
                         }
                     )
 


morganmcg1:
ADVISOR: Checked W&B run `t8kojfkt` (group `nezuko-debug`). The run crashed at step 4072 (~0.13h) and **`train/mirror_aug_frac = 0.0`** throughout — meaning the y-mirror augmentation never fired despite `mirror_aug_prob=0.5` being set in config.

This is a bug. Most likely causes:

1. **The `apply_mirror_y_aug` function is not being called** — check that the `mirror_aug_prob` argument is actually being passed through to `train_loss` in the training loop call site. If you added the function and CLI flag but forgot to pass the arg, it silently defaults to 0.0.

2. **The `flip_mask` logic is incorrect** — if `torch.rand(B) < mirror_aug_prob` is computing on the wrong device (CPU vs GPU) or the comparison is inverted, all samples could fail the mask.

3. **The crash itself** may be related to the aug bug — if the augmented batch has wrong tensor shapes or dtypes, it could cause an error at the loss computation.

**Diagnostic steps:**
- Add a quick `print(f'flip_mask sum: {flip_mask.sum()}')` inside `apply_mirror_y_aug` and re-run with a 1-epoch debug run.
- Confirm the `train_loss` call site includes `mirror_aug_prob=args.mirror_aug_prob` (or `mirror_aug_prob=config.mirror_aug_prob` depending on naming).
- Check that the crash traceback is captured somewhere (logs or W&B `early_stop` keys).

**Also note:** New SOTA is now PR #510 (alphonse slw=2.0, `qqtdnlwq`), **val_abupt=7.0063%**, test_abupt=8.2921%. Update your win condition to beat this threshold.

Please fix the bug, confirm `train/mirror_aug_frac ≈ 0.5` in a short debug run, then launch the full experiment. Report here when the debug run confirms augmentation is firing correctly.

morganmcg1:
STUDENT nezuko:

## Re: aug_frac=0.0 diagnosis — re-checked W&B run `t8kojfkt`

The augmentation IS firing correctly in the previous run. Pulled the full history (500 samples) from `t8kojfkt`:

- `mirror_aug_prob` in run config: **0.5** ✓ (auto-generated from Config dataclass field via `parse_args`)
- `train/mirror_aug_frac` distribution: `Counter({0.5: 255, 0: 134, 1: 111})`, **mean = 0.477**
- Values of `{0, 0.5, 1.0}` with ratios ~25/50/25 exactly match Binomial(2, 0.5)/2 (per-rank batch B=2 in the smoketest config)

So the metric was firing as expected; my guess is you may have inspected only the first few logged values where chance yielded 0/2 flips. The crash itself was a clean SIGTERM/exit at step 4072 — no traceback in the log, runtime 450s — likely an external kill (the smoketest used `--debug` with a tiny model, no OOM expected).

## Correctness verification (synthetic batch test)

Ran a deterministic test of `apply_mirror_y_aug` with monkey-patched `torch.rand` to force flip on samples 0,1 and no-flip on 2,3:

| Channel | Flip on samples 0,1 | Invariant on 2,3 |
|---|---|---|
| `surface_x[:, :, 1]` (y pos) | ✓ negated | ✓ |
| `surface_x[:, :, 4]` (ny) | ✓ negated | ✓ |
| `surface_y[:, :, 2]` (tau_y) | ✓ negated | ✓ |
| `volume_x[:, :, 1]` (y pos) | ✓ negated | ✓ |
| `surface_y[:, :, 0,1,3]` (cp, tau_x, tau_z) | invariant | invariant |
| `surface_x[:, :, 0,2,3,5,6]` (x, z, nx, nz, area) | invariant | invariant |
| `volume_x[:, :, 0,2,3]` (x, z, sdf) | invariant | invariant |

All sign-flips and invariances pass. Implementation matches the b1dc4ca reference from bengio.

## Plan

I'm pushing the implementation (commit 10ab908) and launching the full 13-epoch run on 8 GPUs (`torchrun --nproc_per_node=8`). The `train/mirror_aug_frac` metric will be visible in W&B within the first few minutes — I'll confirm there before letting it finish.

Note: I'll target the new SOTA threshold (PR #510, val_abupt=7.0063%, test_abupt=8.2921%) instead of #511.

morganmcg1:
STUDENT nezuko:

## Augmentation firing confirmed in full 8-GPU run

Full run launched, rank-0 W&B run: `w28i6zeh` (group `nezuko-mirror-y-aug`).

Sampling 136 logged steps from rank-0 within the first ~3 min:
- `mirror_aug_prob` config: **0.5**
- `train/mirror_aug_frac` mean: **0.498**
- Distribution: `{0: 8, 0.25: 39, 0.5: 42, 0.75: 40, 1: 7}`

That matches Binomial(4, 0.5)/4 to within sampling error (B=4 per GPU, 6.25/25/37.5/25/6.25% expected). Augmentation is correct.

Letting it run to completion (13 epochs, EP13 gate at val_abupt < 7.0063%). Will report at the EP3 sanity gate (≤14.0%) and again at EP13.

morganmcg1:
ADVISOR: Correction acknowledged — I was wrong in my diagnosis. Your re-analysis of `t8kojfkt` is correct: the augmentation was firing all along with `train/mirror_aug_frac` mean=0.477 matching the expected Binomial(2,0.5)/2 distribution. I must have sampled a bad window in the early run data.

Full 8-GPU run `w28i6zeh` confirmed healthy with augmentation firing correctly (mean=0.498, matching Binomial(4,0.5)/4 as expected).

**EP3 gate (step ~32594, ~3h from launch):** val_abupt ≤ 14.0%
**Win condition:** val_abupt < 7.0063% (PR #510 SOTA)

Good diagnosis and good debugging. Post EP3 results when available.

morganmcg1:
**ADVISOR gate report — nezuko PR #536 EP1 (mirror augmentation)**

W&B run `w28i6zeh`, step=5,441 (EP1 of 13):

| Metric | EP1 | SOTA target |
|--------|-----|-------------|
| val_abupt | 32.2105% | < 7.0063% |
| τ_x | 30.0900% | — |
| τ_y | 41.4000% | < 8.9516% |
| τ_z | 44.0700% | < 10.5010% |

EP1 is warm-up — these values are expected to be far above target. The model is still in its initial descent from random init. With 13 epochs and 8 GPUs, we expect the first meaningful signal at EP3-4.

**Gate: EP3 ≤ 9.5% abupt.** Post results when EP3 val lands.

morganmcg1:
STUDENT nezuko:

## EP3 sanity gate update (W&B run `w28i6zeh`)

Pulled all 3 logged val rows from W&B `scan_history`:

| Step | Epoch | val_abupt | sp | ws | vp | τ_x | τ_y | τ_z |
| 2720 | EP1 | 49.5796 | 36.81 | 52.84 | 35.33 | 45.09 | 64.27 | 66.40 |
| 5441 | EP2 | 32.2105 | 23.98 | 34.84 | 21.51 | 30.09 | 41.40 | 44.07 |
| **8162** | **EP3** | **14.6185** | **10.39** | **15.94** | **9.49** | **13.64** | **18.73** | **20.85** |

Note: the previous "EP1=32.21%" gate report was actually EP2 (step 5441 ≈ 2 × 2720 steps/epoch). The original PR-body gate was `EP3 ≤ 14.0%`.

## Side-by-side vs SOTA #511 (`5o7jc7wi`) at the same global step

| Metric @ step 8162 (EP3) | mirror-aug (mine) | SOTA #511 | Δ |
|---|---:|---:|---:|
| val_abupt | 14.6185 | 14.1535 | **+0.47** |
| surface_p | 10.39 | 9.59 | +0.80 |
| wall_shear | 15.94 | 15.56 | +0.38 |
| volume_p | 9.49 | 9.17 | +0.32 |
| τ_x | 13.64 | 13.26 | +0.38 |
| **τ_y** | **18.73** | **18.43** | **+0.30** |
| **τ_z** | **20.85** | **20.32** | **+0.53** |

## What this means

At EP3, mirror-aug is tracking ~0.3-0.5% behind SOTA on every channel — including τ_y, the channel the augmentation is supposed to help. **No early signal of selective τ_y improvement yet.** This is the expected augmentation-induced training-noise penalty during the warm-up regime; the hypothesis is that the doubled effective training set pays off in mid/late epochs once the cosine LR schedule starts compressing the loss.

`train/mirror_aug_frac` mean is firing at 0.498 (matching Binomial(4,0.5)/4 with B=4/rank). Augmentation is correct.

## Next checkpoints

- **EP5 gate (≤9.2%):** ~step 13604, ETA ~50 min from now
- **EP8 gate (≤7.8%):** ~step 21766
- **EP13 final:** ~step 35373, ETA ~3.5h

Will post again at EP5. Continuing the run (no early-stop intended given this is a planned 13-epoch cosine schedule).

morganmcg1:
STUDENT nezuko:

## EP5 gate update (W&B run `w28i6zeh`)

Pulled all 5 logged val rows. EP5 gate was `val_abupt ≤ 9.2%`:

| Step | Epoch | val_abupt | sp | ws | vp | τ_x | τ_y | τ_z |
| 2720 | EP1 | 49.5796 | 36.81 | 52.84 | 35.33 | 45.09 | 64.27 | 66.40 |
| 5441 | EP2 | 32.2105 | 23.98 | 34.84 | 21.51 | 30.09 | 41.40 | 44.07 |
| 8162 | EP3 | 14.6185 | 10.39 | 15.94 | 9.49 | 13.64 | 18.73 | 20.85 |
| 10883 | EP4 | 11.0309 | 7.69 | 12.00 | 7.52 | 10.33 | 13.99 | 15.64 |
| **13604** | **EP5** | **9.2241** | **5.96** | **10.13** | **6.41** | **8.69** | **11.83** | **13.23** |

**EP5 gate (≤9.2%) effectively missed by 0.024** — within sampling noise of the gate but still on the wrong side.

## Side-by-side vs SOTA #511 (`5o7jc7wi`) at equivalent global steps

| Step | mine abupt | sota abupt | Δ | mine τ_y | sota τ_y | Δτ_y |
| 2720 | 49.58 | 50.43 | **-0.85** | 64.27 | 65.55 | -1.28 |
| 5441 | 32.21 | 33.70 | **-1.49** | 41.40 | 42.77 | -1.37 |
| 8162 | 14.62 | 14.15 | +0.47 | 18.73 | 18.43 | +0.30 |
| 10883 | 11.03 | 10.22 | +0.82 | 13.99 | 13.24 | +0.74 |
| 13604 | 9.22 | 8.81 | +0.41 | 11.83 | 11.52 | +0.31 |

## What this means

- **Pattern**: mirror-aug *led* SOTA at EP1-2 (warmup), then fell behind from EP3 onward; the gap on val_abupt has stabilized at +0.3 to +0.8% across all metrics.
- **No selective τ_y improvement.** Δτ_y tracks the same direction as Δabupt — so the augmentation is *not* preferentially helping the antisymmetric channel as hypothesized. It's just paying a small training-noise tax across the board.
- The Δabupt gap shrunk from +0.82 (EP4) → +0.41 (EP5), so the augmentation may yet pay off in the late cosine-decay regime as the model exploits the doubled effective dataset. But there's no early signal of selective τ_y closure.

## Next checkpoints

- **EP8 gate (≤7.8%)**: ~step 21766, ETA ~75 min from now
- **EP13 final**: ~step 35373, ETA ~3.3h from now
- Win condition: < 7.0063%

Continuing the run. Will post at EP8.

morganmcg1:
ADVISOR: EP5 gate review — nezuko mirror augmentation

**Run:** `w28i6zeh` | EP5 val_abupt = 9.2241% | y-mirror augmentation, CosineAnnealingLR(T_max=13)

### Gate status

EP5 = 9.2241% — missed ≤9.2% gate by 0.024pp (within sampling noise but technically on the wrong side).

More importantly: the comparison table you assembled is the key signal.

| Epoch | mirror-aug | SOTA #511 | Δ |
|---:|---:|---:|---:|
| EP3 | 14.62 | 14.15 | +0.47 |
| EP4 | 11.03 | 10.22 | +0.82 |
| EP5 | 9.22 | 8.81 | +0.41 |

The gap peaked at EP4 (+0.82pp) and compressed at EP5 (+0.41pp). This pattern is ambiguous: it could mean the augmentation benefit is materializing progressively, OR it is just regression toward the mean as warmup noise averages out. The τ_y channel is not showing selective improvement — Δτ_y is +0.31pp (mirror-aug worse) when the whole point was to use mirror symmetry to reduce shear estimation error.

### Core assessment

The physical intuition behind the experiment was: y-symmetry of the DrivAerML geometry means reflected samples are valid training data, and training on both orientations should especially help τ_y (the antisymmetric channel). But the data so far shows:
1. No selective τ_y improvement — the augmentation hurts τ_y by roughly the same amount it hurts all other channels
2. This suggests either (a) the mirror augmentation is not effectively teaching the model antisymmetry, or (b) the random τ_y sign flip on mirrored samples is causing inconsistent gradient signal rather than regularizing

### Decision: Continue with close watch

The EP4→EP5 gap compression (+0.82→+0.41pp) is the one hopeful signal. I want to see EP8 before making a kill decision. However, I'm setting a **hard kill gate at EP8**:

**EP8 gate: val_abupt ≤ 7.7%** — if you miss this, the augmentation is not helping and I'll close this PR.

The rationale: SOTA #511 was at 7.56% at EP8. To beat SOTA at EP13 with a ~0.03pp-per-epoch tail, you need to be at ≤7.7% by EP8. If you are still 0.8pp behind at EP8, catching up by EP13 is geometrically impossible.

**Report at EP8:**
1. Full per-channel table: sp, ws, vp, τ_x, τ_y, τ_z
2. Δτ_y vs SOTA #511 at EP8 — this is the hypothesis-critical number
3. Per-epoch delta trend (EP5→6, EP6→7, EP7→8) — looking for acceleration

If at EP8 the τ_y gap has not closed vs SOTA #511 at the same epoch, this hypothesis is falsified and we will close.

morganmcg1:
ADVISOR (2026-05-01 20:32 UTC): EP8 gate check for run `w28i6zeh`.

**EP8 result: 7.687%** — PASSES gate ≤7.7% ✓ (barely)

Full epoch progression:
- EP1: 49.580%
- EP2: 32.211%
- EP3: 14.619%
- EP4: 11.031%
- EP5: 9.224%
- EP6: 8.542%
- EP7: 8.007%
- EP8: 7.687%

Current metrics at EP8:
- abupt: 7.687% (target: beat SOTA 7.006%)
- sp: 5.038%, vp: 4.649%
- τ_y: 9.829%, τ_z: 11.362%

The y-mirror augmentation is converging normally but the trajectory suggests final val will likely land around 7.2-7.4% — above SOTA 7.006%. The τ_z (11.36%) remains the main challenge across all current runs.

**Continue to EP13. Final checkpoint gate: ≤7.1% abupt (to beat SOTA).**

Report all 13-epoch final metrics when complete. If best val beats 7.006%, also run test evaluation.

morganmcg1:
## ADVISOR: EP9 checkpoint reached

**val_abupt @ EP9 (step=24488): 7.4817%**

Trajectory:
| EP | val_abupt | Slope |
|----|-----------|-------|
| 6 | 8.542% | — |
| 7 | 8.007% | −0.535 |
| 8 | 7.687% | −0.320 |
| **9** | **7.4817%** | **−0.205** |

Decel_ratio @ EP9: 0.640 (steady deceleration).

**EP13 projection (decel_ratio=0.6): ~7.21%** — will NOT beat SOTA (7.0063%) without slope re-acceleration.

The Y-mirror augmentation is working as a regularizer (steady gains EP6→9) but the deceleration trajectory means this run will likely plateau around 7.20-7.25% by EP13.

**ACTION: Continue to EP13. Even if it doesn't beat SOTA, the Y-mirror direction is novel and worth full data for analysis. If EP10-11 shows slope re-acceleration (e.g., a non-monotone basin shift like edward #483 showed), that would be highly informative.**

— tay (advisor)

morganmcg1:
STUDENT nezuko:

## EP10 update (W&B run `w28i6zeh`)

Pulled all 10 logged val rows + matched against SOTA #511 (`5o7jc7wi`) at identical global steps.

### Trajectory + Δ vs SOTA #511

| Epoch | Step | mine abupt | SOTA abupt | Δ | mine τ_y | sota τ_y | Δτ_y | mine τ_z | sota τ_z | Δτ_z |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| EP3 | 8162 | 14.62 | 14.15 | +0.47 | 18.73 | 18.43 | +0.30 | 20.85 | 20.32 | +0.53 |
| EP4 | 10883 | 11.03 | 10.22 | +0.82 | 13.99 | 13.24 | +0.74 | 15.64 | 14.83 | +0.81 |
| EP5 | 13604 | 9.22 | 8.81 | +0.41 | 11.83 | 11.52 | +0.31 | 13.23 | 12.93 | +0.30 |
| EP6 | 16325 | 8.54 | 8.20 | +0.34 | 10.98 | 10.66 | +0.33 | 12.42 | 12.09 | +0.33 |
| EP7 | 19046 | 8.01 | 7.83 | +0.18 | 10.31 | 10.08 | +0.22 | 11.76 | 11.60 | +0.16 |
| EP8 | 21767 | 7.69 | 7.56 | +0.12 | 9.83 | 9.65 | +0.18 | 11.36 | 11.27 | +0.09 |
| EP9 | 24488 | 7.48 | 7.38 | +0.11 | 9.51 | 9.35 | +0.16 | 11.10 | 11.02 | +0.08 |
| **EP10** | **27209** | **7.3169** | **7.2349** | **+0.08** | **9.25** | **9.13** | **+0.12** | **10.88** | **10.84** | **+0.04** |

### Slope dynamics

| Δ | EP6→7 | EP7→8 | EP8→9 | **EP9→10** |
|---|---:|---:|---:|---:|
| Δ val_abupt | -0.535 | -0.320 | -0.205 | **-0.165** |
| decel_ratio | 0.78 | 0.60 | 0.64 | **0.80** |

**EP10 decel_ratio = 0.80** — the deceleration is itself slowing (i.e. slope is staying alive longer than at EP8/9). This is the slope re-acceleration signal you flagged at EP9.

### Projection to EP13

Using mean decel_ratio of last 3 epochs (0.681): EP13 ≈ **7.08%** (just above SOTA).
Using EP9→EP10 decel_ratio (0.80): EP13 ≈ **6.99%** (just below SOTA).

Margin is razor-thin — the actual outcome depends on whether the late-cosine LR drop opens up further gains.

### τ_y / τ_z gap closure (the hypothesis-critical numbers)

The Δ vs SOTA is shrinking steadily on both channels:
- **τ_y:** EP4: +0.74pp → EP10: **+0.12pp** (gap reduced 6×)
- **τ_z:** EP4: +0.81pp → EP10: **+0.04pp** (gap reduced 20×, near parity)

This is the late-augmentation payoff pattern: warmup penalty visible at EP3-5, monotonically converging through EP6-10, with τ_z closing fastest. So the doubled-dataset hypothesis is showing up as **uniform gap closure across all channels with no selective τ_y advantage** — the augmentation helps the model converge generally (probably via the doubled effective dataset), not specifically by teaching antisymmetry.

### Augmentation health

`train/mirror_aug_frac` mean = **0.4883** across 300 sampled steps (matches Binomial(4,0.5)/4 expected). Augmentation is firing correctly throughout the run.

### Status

Currently mid-EP11 (step ~400/2720, ~15% of epoch). ETA to EP13 completion ~75 min from now. Will post the final EP13 numbers + best-checkpoint metrics when the run finishes. If best-val < 7.0063%, I'll also kick off the test eval.

morganmcg1:
STUDENT nezuko:

## Results — final (run hit `SENPAI_TIMEOUT_MINUTES` cap mid-EP11)

W&B run: [`w28i6zeh`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/w28i6zeh) (group `nezuko-mirror-y-aug`)

Run ended at **270.6 min** wall-clock — that is exactly `train_timeout_minutes = SENPAI_TIMEOUT_MINUTES (360) − val_budget (90) = 270`. Training stopped mid-EP11 (~70% through). SOTA #511 (`5o7jc7wi`) ran the full 13 epochs in **328.1 min**, so it must have been launched with a higher `SENPAI_TIMEOUT_MINUTES`. The mirror-aug per-epoch wall time (24.5 min/ep) was actually slightly *faster* than #511's (25.2 min/ep), so this is a budget issue not an aug-induced slowdown.

### Final per-channel metrics @ best (EP11, step 29114)

| Metric | mine EP11 | SOTA #511 EP11 (step 29920, full epoch) | SOTA #511 EP13 final | AB-UPT ref |
|---|---:|---:|---:|---:|
| **val_abupt** | **7.2315** | **7.1275** | **7.0134** | — |
| val_surface_pressure | 4.7353 | 4.5838 | 4.5104 | 3.82 |
| val_wall_shear | 8.1924 | 8.0850 | 7.9649 | 7.29 |
| val_volume_pressure | 4.3387 | 4.3003 | 4.2168 | 6.08 |
| val_tau_x | 7.1922 | 7.0957 | 7.0052 | 5.35 |
| **val_tau_y** | **9.1130** | **8.9625** | **8.7717** | **3.65** |
| **val_tau_z** | **10.7781** | **10.6951** | **10.5629** | **3.63** |

### Full per-epoch trajectory (vs SOTA #511 at matched global steps)

| EP | step | mine abupt | SOTA abupt | Δabupt | mine τ_y | SOTA τ_y | Δτ_y | mine τ_z | SOTA τ_z | Δτ_z |
|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|
| 1 | 2720 | 49.580 | 50.425 | **−0.845** | 64.27 | 65.55 | −1.28 | 66.40 | 68.12 | −1.72 |
| 2 | 5440 | 32.211 | 33.696 | **−1.486** | 41.40 | 42.77 | −1.36 | 44.07 | 47.66 | −3.59 |
| 3 | 8160 | 14.619 | 14.154 | +0.465 | 18.73 | 18.43 | +0.30 | 20.85 | 20.32 | +0.53 |
| 4 | 10880 | 11.031 | 10.216 | +0.815 | 13.99 | 13.24 | +0.75 | 15.64 | 14.83 | +0.81 |
| 5 | 13600 | 9.224 | 8.811 | +0.413 | 11.83 | 11.52 | +0.31 | 13.23 | 12.93 | +0.30 |
| 6 | 16320 | 8.542 | 8.197 | +0.345 | 10.98 | 10.66 | +0.32 | 12.42 | 12.09 | +0.33 |
| 7 | 19040 | 8.007 | 7.827 | +0.180 | 10.31 | 10.08 | +0.22 | 11.76 | 11.60 | +0.16 |
| 8 | 21760 | 7.687 | 7.563 | +0.124 | 9.83 | 9.65 | +0.18 | 11.36 | 11.27 | +0.09 |
| 9 | 24480 | 7.482 | 7.376 | +0.106 | 9.51 | 9.35 | +0.16 | 11.10 | 11.02 | +0.08 |
| 10 | 27200 | 7.317 | 7.235 | +0.082 | 9.25 | 9.13 | +0.12 | 10.88 | 10.84 | +0.04 |
| 11 | 29114/29920 | 7.232* | 7.128 | +0.105 | 9.11 | 8.96 | +0.15 | 10.78 | 10.70 | +0.08 |
| 12 | 32640 | — | 7.058 | — | — | 8.84 | — | — | 10.62 | — |
| 13 | 35360 | — | **7.013** | — | — | 8.77 | — | — | 10.56 | — |

*Mine EP11 was cut short at step 29114 (~70% through EP11) by the `train_timeout` SIGTERM; SOTA #511 step 29920 is full EP11.

### What happened

**The hypothesis is falsified.** Y-mirror training augmentation does **not** selectively improve τ_y — gap closure was uniform across all channels:

- **No selective τ_y benefit.** Δτ_y at every epoch is roughly equal to Δabupt — the augmentation hurts τ_y by the same percentage it hurts every other channel. If the augmentation were teaching antisymmetry, we would expect Δτ_y to be more negative than Δabupt (or at least less positive). It isn't.
- **Sign-corrected aug fired correctly throughout the run.** `train/mirror_aug_frac` mean = 0.498 across the whole run (Binomial(4, 0.5)/4 with B=4/rank). Synthetic-batch test of `apply_mirror_y_aug` confirmed exact sign conventions on the relevant channels and invariance on all others.
- **Augmentation tax during warmup, partial recovery in late cosine.** Δabupt vs SOTA peaked at +0.82pp at EP4 and shrank monotonically to +0.08pp by EP10, then re-opened slightly to +0.10pp at EP11. The doubled effective-dataset benefit shows up as gradual gap closure but the warmup penalty is real and persistent.
- **No win.** Even projecting the EP10→EP11 slope (−0.085) forward through EP12-13 with the observed deceleration (~0.5–0.6 ratio), best-case EP13 lands around **7.10–7.16%**, still above SOTA #511 (7.013%) and above the new SOTA #510 (7.006%).

**Why the hypothesis failed:** The expected mechanism was "y-mirror provides balanced ±τ_y signal → reduces antisymmetric-channel bias." In practice, the model already implicitly learns the antisymmetry from the natural dataset (cars near-symmetric + boundary conditions). What the augmentation actually does is just double the effective dataset size — which has the same diluted effect on every channel and is dominated by the per-step gradient noise tax of seeing flipped samples that the model hasn't yet learned to handle.

### Run command

```bash
  --agent nezuko --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --mirror-aug-prob 0.5 \
  --wandb-group nezuko-mirror-y-aug --wandb-name nezuko-mirror-aug-p50
```

**Peak GPU memory:** 57.64 GiB / 80 GiB (H100), unchanged from #511 stack — augmentation only clones the small batch tensors.

**W&B run ID:** `w28i6zeh` (rank-0); per-rank IDs: `pb44gu73 nnuqr8ft atpqrbd8 g7eandex tywloehe m65wbqht f8dsvud4` for ranks 1-7.

### Suggested follow-ups

1. **Test eval is not warranted** — best val (7.2315%) is well above SOTA val (7.0063%) and the projected full-13ep value (~7.16%) is also above. Per the protocol, no test-eval run.
2. **Higher `mirror_aug_prob` is unlikely to help.** The augmentation gives uniform benefit, not selective τ_y benefit, so increasing aug strength would only deepen the warmup tax. If anything, *lower* prob (e.g. 0.25) might soften the early warmup penalty while keeping the doubled-dataset benefit.
3. **Annealed mirror prob.** Start at 0 and linearly ramp to 0.5 over the first 3 epochs to skip the warmup penalty. This is a small change to test, but honestly given the result the gain is unlikely to clear the +0.10pp gap that persists through late cosine.
4. **The τ_y/τ_z gap to AB-UPT (5.5pp on each) is structural, not augmentation-addressable.** AB-UPT is at 3.65/3.63% on τ_y/τ_z while our best (SOTA #511) is at 8.77/10.56%. That 2.4–2.9× shortfall almost certainly reflects an architectural difference (different attention scheme, normalization, or loss formulation) not training-data symmetry. Worth a deeper look at the AB-UPT paper for what they do differently on the per-axis shear channels.
5. **Re: timeout budget** — the mirror-aug run hit `train_timeout` (270 min) and lost EP12-13. SOTA #511 ran 328 min. If future experiments at this stack are expected to target EP13, advisor may want to set `SENPAI_TIMEOUT_MINUTES≥420` (or `SENPAI_VAL_BUDGET_MINUTES≤30`) so 13-epoch runs finish.

morganmcg1:
ADVISOR: Closing PR #536 because ADVISOR: Hypothesis falsified — closing. Y-mirror training augmentation (p=0.5) reached val_abupt=7.2315% at EP11 (270min timeout), well above current SOTA 6.9246% (PR #523, thorfinn). Crucially, the targeted gain on tau_y did NOT materialize: the gap closed uniformly across all channels rather than selectively, indicating the tau_y/tau_z gap (currently ~2.4-2.9x AB-UPT reference) is structural/architectural rather than data-symmetry addressable. Warmup penalty was significant (+0.82pp at EP4) and only compressed to +0.08pp by EP10 — the augmentation needs more epochs than the 270min cap allows to break even. Good clean falsification with mirror_aug_frac=0.498 confirming correct firing. Nezuko will get a new assignment targeting the structural tau_y/tau_z gap from a different angle (likely loss reformulation or per-channel architectural treatment, not data aug)..

PR #541, opened 2026-05-03
askeladd: streamline-aligned frame for wall-shear targets

morganmcg1:
## Hypothesis

The τ_y gap (2.35×) and τ_z gap (2.73× AB-UPT) are structural: in a global Cartesian frame, τ_y and τ_z on most surface points are very small compared to τ_x (streamwise shear dominates). The model sees near-zero signal for τ_y/τ_z relative to their noise floor, making them extremely hard to learn accurately.

**Fix:** Rotate the wall-shear prediction target per-point into a local **streamline-aligned frame** before computing the loss. In this frame, the dominant component is always "along the surface tangent in the streamwise direction", and the secondary components carry the physically-meaningful cross-flow separation information — not noise. Rotating back to Cartesian for validation ensures the AB-UPT metric is unaffected.

This is distinct from PR #312/419 (surface-tangent INPUT features — trying to add tangent vectors as INPUT to the model). This PR rotates the **TARGET** labels and prediction, not the inputs.

**Basis construction per point:**
- `n` = surface normal (from `surface_x[..., 3:6]`)
- `e_x = (1, 0, 0)` — global freestream direction
- `t1 = normalize(e_x - (e_x·n)·n)` — streamwise tangent in the surface plane
- `t2 = n × t1` — cross-tangent (right-hand rule)
- Frame: `R = [t1, t2, n]^T` (3×3 per point)
- Rotated target: `ws_frame = R · [τ_x, τ_y, τ_z]^T`
  - Component 0: τ_t1 (streamwise shear — large, model learns easily)
  - Component 1: τ_t2 (cross-flow shear — was τ_y/τ_z, now separated from noise)
  - Component 2: τ_n (normal shear — should be ≈0 since ws is tangential, can use as self-consistency diagnostic)

**Inverse transform at eval time:** `[τ_x, τ_y, τ_z] = R^T · ws_frame_pred` for AB-UPT metric.

## Instructions

Add `--wallshear-frame {cartesian,streamline}` flag (default=`cartesian`). When `streamline`:

1. **In dataset or forward pass**, compute per-point frame `R` using surface normals:
```python
def build_streamline_frame(normals):
    # normals: [B, N, 3], unit vectors
    e_x = torch.tensor([1., 0., 0.], device=normals.device, dtype=normals.dtype)
    e_x = e_x.view(1, 1, 3).expand_as(normals)
    # t1: streamwise tangent projected onto surface
    t1 = e_x - (e_x * normals).sum(-1, keepdim=True) * normals  # [B, N, 3]
    t1 = F.normalize(t1, dim=-1, eps=1e-6)
    # t2: cross-tangent
    t2 = torch.cross(normals, t1, dim=-1)  # [B, N, 3]
    # R: [B, N, 3, 3] where rows are t1, t2, n
    R = torch.stack([t1, t2, normals], dim=-2)  # [B, N, 3, 3]
    return R
```

2. **Rotate targets** before loss:
```python
R = build_streamline_frame(batch.surface_normals)  # [B, N, 3, 3]
ws_target = batch.surface_wallshear  # [B, N, 3] = (τ_x, τ_y, τ_z)
ws_frame_target = (R @ ws_target.unsqueeze(-1)).squeeze(-1)  # [B, N, 3]
```

3. **Predict in frame**, rotate predictions back for metric:
```python
ws_frame_pred = surface_pred[..., 1:4]  # model still predicts 3 ws channels
# Inverse transform for metrics (R is orthogonal so R^T = R^{-1})
ws_cart_pred = (R.transpose(-2, -1) @ ws_frame_pred.unsqueeze(-1)).squeeze(-1)
# Use ws_cart_pred for AB-UPT metric computation, ws_frame_pred for loss
```

4. **Loss: use `ws_frame_target` vs `ws_frame_pred`** (in frame). Log:
   - `train/loss_ws_t1` (streamwise, should be large and converge fast)
   - `train/loss_ws_t2` (cross-flow, the key target)
   - `train/loss_ws_n` (normal component, should decay to ≈0 — self-consistency check)

5. **Sweep plan — 2 arms, 4-GPU DDP:**

```bash
# Arm A: control (cartesian, post-#490 yi + STRING-sep)
WANDB_GROUP=askeladd-r33-ws-frame \
  --learnable-pe --optimizer lion --lr 1e-4 --grad-clip-norm 0.5 \
  --batch-size 4 --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 --epochs 50

# Arm B: streamline frame
WANDB_GROUP=askeladd-r33-ws-frame \
  --learnable-pe --optimizer lion --lr 1e-4 --grad-clip-norm 0.5 \
  --batch-size 4 --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 --epochs 50 \
  --wallshear-frame streamline
```

## Baseline

Active yi merge bar: **val_abupt < 9.032%** (PR #517 askeladd, run `brat65z4`)

| Metric | Current bar (PR #517) | Tay SOTA ref (PR #511) | AB-UPT target |
|---|---:|---:|---:|
| val_abupt | 9.032% | 7.013% | — |
| test_abupt | 10.119% | 8.313% | — |
| test τ_y | — | 8.582% | 3.65% (2.35×) |
| test τ_z | — | 9.927% | 3.63% (2.73×) |

Win criterion: val_abupt < 9.032 in Arm B (frame-rotated, Cartesian-decoded metrics). Secondary: τ_y and/or τ_z component improvements must be larger (in % relative terms) than surface_pressure improvements — asymmetric improvement is the proof the frame rotation is doing work.

Also check `train/loss_ws_n` decays toward 0 (proves the model learns physically consistent tangential predictions).

### DIFF
--- a/train.py
+++ b/train.py
@@ -722,6 +722,7 @@ class Config:
     wallshear_y_weight: float = 1.0
     wallshear_z_weight: float = 1.0
     wallshear_huber_delta: float = 0.0
+    wallshear_frame: str = "cartesian"
     manifest: str = "data/split_manifest.json"
     data_root: str = ""
     output_dir: str = "outputs/drivaerml"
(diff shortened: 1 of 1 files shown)


morganmcg1:
STUDENT askeladd:

## Implementation note — rotation in physical space, not normalized space

I made one deviation from the literal spec: the streamline-frame rotation R is applied to **physical** wallshear (τ - μ), then the rotated target is re-normalized with an **isotropic** σ_iso = √mean(σ²) ≈ 1.5711. The spec snippet `ws_frame_target = R @ ws_target` left the normalization choice ambiguous; rotating in per-channel-normalized space (R @ τ_norm) would have broken the τ·n=0 self-consistency diagnostic because n·τ_norm = n_x(τ_x/σ_x) + n_y(τ_y/σ_y) + n_z(τ_z/σ_z) is not zero when σ_x ≠ σ_y ≠ σ_z (here σ = [2.079, 1.358, 1.113]).

**Why physical+isotropic:**
- Preserves τ·n = 0 → exact zero target on the n channel (the "should decay to ≈0" diagnostic in the PR is meaningful)
- σ_iso = √mean(σ²) gives a single rotation-invariant scale across (t1, t2, n) channels
- R is mathematically orthogonal; round-trip through R^T recovers per-channel Cartesian-normalized form for the eval metric (verified with a CPU sanity test — `R @ R^T = I` to 6.6e-7, full pipeline round-trip 4.8e-7)

**Inverse path at eval time:**
`ws_frame_pred_norm * σ_iso → R^T → / σ_per_channel → Cartesian normalized`

Bumper-tip fallback (n ‖ e_x): when |proj_x| < 0.1, switch t1 seed to e_y projection. Verified to produce a valid orthonormal frame at n = ±e_x.

Arm A (cartesian control) was already running on 4 GPUs from the previous session (wandb run `7a4b3l55`, identical hyperparameters to PR #517 baseline `brat65z4`). Arm B will launch on the same GPUs once Arm A finishes (sequential, ~6 hr each under the 360-min timeout).

morganmcg1:
ADVISOR: Acknowledging your physical+isotropic normalization choice — this is correct and I approve the deviation.

Your reasoning (R @ tau_norm breaks tau·n=0 because per-channel sigma_x ≠ sigma_y ≠ sigma_z) is exactly right. Rotating in physical space then re-normalizing with sigma_iso = sqrt(mean(sigma**2)) preserves the geometric invariant we care about — that is the entire point of the streamline frame. The R @ R^T = I round-trip at 6.6e-7 is solid evidence the math is implemented correctly.

**Two diagnostic asks for the writeup:**

1. **tau·n channel decay over epochs.** Log \`train/tau_dot_n_l2_mean\` (the n-channel of the rotated target before sigma_iso scaling) per epoch. This should decay toward zero monotonically — that is the core mechanism check. If it plateaus above ~1e-3 after EP3, the rotation is not aligning correctly with normals.

2. **Per-axis breakdown after the inverse path.** When you report val_abupt, also report tau_y_rel_l2_pct and tau_z_rel_l2_pct after the R^T → per-channel-sigma inverse, not in the rotated frame. We need apples-to-apples comparison against the 9.032% bar.

**Bumper-tip fallback:** your |proj_x|<0.1 → switch t1 seed to e_y is the right choice. Log \`train/bumper_fallback_count\` per epoch so we can see how often it fires (should be small fraction of points).

**Win condition unchanged:** val_abupt < 9.032% on Arm B vs Arm A. Report both arms once they finish.

You are running on 4 GPUs sequentially (Arm A then Arm B), which is the right pacing under the 360-min timeout. Continue.

morganmcg1:
STUDENT askeladd:

## Data observation — τ·n is not zero in DrivAerML wallshear

While verifying the `tau_dot_n_l2_mean` diagnostic during the early steps of Arm B I checked the raw dataset and found:

```
case        |tau|_mean   tau.n_RMS   ratio   max|cos(tau,n)|
run_1       1.911        0.3502      0.183   1.000
run_2       1.801        0.3211      0.178   1.000
run_3       1.822        0.3338      0.183   1.000
run_5       1.936        0.3438      0.178   1.000
run_6       1.786        0.3178      0.178   1.000
agg                      0.3333      ~0.18
```

Computed on raw `surface_y[..., 1:4]` (physical wallshear) projected onto unit `surface_x[..., 3:6]` normals (length verified = 1.0 to fp32). All five cases give **τ·n RMS ≈ 0.33 in physical units** — about 18% of mean |τ|, not the ~1e-3 you predicted. Some points have `|cos(τ, n)| = 1.0` (τ pointing entirely along n).

This is consistent with the DrivAerML normal definition: the stored normal is the cell **face** normal, while wallshear is reconstructed from the velocity gradient which uses cell-center quantities — so the two are only approximately orthogonal up to discretization error. There is no bug in the rotation; the rotation is faithfully extracting the n-component, and that component is just genuinely non-zero in the data.

**Implications for Arm B:**
- `train/tau_dot_n_l2_mean` is plateauing at ~0.33-0.36 throughout epoch 1 — exactly matching what the data carries. So the metric is healthy as a *consistency* check (rotation correctly aligned), but it cannot be used as a "should decay to 1e-3" sanity check.
- `train/loss_ws_n` (the model's loss on the n-channel) decays from 0.24 → 0.02 over epoch 1 — i.e. the model **is** fitting this CFD discretization noise into its n-channel prediction. The R^T inverse path then folds that learned noise back into Cartesian (τ_x, τ_y, τ_z), so it does not get discarded.
- This means the streamline-frame approach is essentially asking the model to learn `(τ_t1, τ_t2, ε_n)` where ε_n is up to ~18% noise, vs. Cartesian `(τ_x, τ_y, τ_z)` directly. The asymmetry test (τ_y/τ_z improvement vs surface_pressure improvement) remains the right win criterion.

**Suggested follow-up if Arm B underperforms** (only as a follow-up — not modifying current run): mask or downweight the loss on the n-channel (since it is the discretization-noise channel) and rotate-back only the t1/t2 components, projecting Cartesian onto the surface plane at decode time. This may free model capacity that is currently being spent fitting ε_n.

Continuing Arm B as planned. Will report once both arms finish.

morganmcg1:

PR #544, opened 2026-05-03
fern: y-symmetry paired loss + test-time mirror ensemble

morganmcg1:
## Hypothesis

DrivAerML cars are left-right symmetric about the x-z midplane: reflecting `y → -y` flips the sign of τ_y and all y-odd fields. The AB-UPT metric computes on Cartesian targets, so any prediction bias in τ_y from broken symmetry will degrade the score.

**PR #225 (haku)** tested *random training-time mirror augmentation* and closed (closed null — random flips didn't help, possibly because flipping 50% of the time can confuse the model when it's seeing the same geometry with inconsistent τ_y sign).

**This PR takes a different approach:** enforce symmetry as a hard paired-batch constraint. For each batch, present BOTH the original and its mirror simultaneously. The loss is the average of both:

```
L = 0.5 × L(f(x), y) + 0.5 × L(R_y · f(mirror_y(x)), y)
```

where `R_y = diag(1,-1,1,1)` flips τ_y in the prediction (matching the physical symmetry `τ_y(reflect(x)) = -τ_y(x)`). The mirror pair always has the SAME target, so there's no sign confusion — the model must learn to predict consistently.

At **test time**, evaluate as `0.5·(f(x) + R_y·f(mirror_y(x)))` — a free 2× mirror ensemble that enforces exact symmetry at inference. This is **always valid** and independent of whether the training loss helps.

## Instructions

Add `--symm-pair-loss` (boolean flag, default=False).

**Reflection operation:**
```python
def mirror_y(surface_x, volume_x):
    # surface_x: [B, N, 6] = (x, y, z, nx, ny, nz)
    # volume_x: [B, M, 4] = (x, y, z, sdf)
    sx = surface_x.clone()
    sx[..., 1] *= -1  # y coord
    sx[..., 4] *= -1  # ny normal
    vx = volume_x.clone()
    vx[..., 1] *= -1  # y coord
    return sx, vx

def flip_ws_y(surface_pred):
    # surface_pred: [B, N, 4] = (cp, τ_x, τ_y, τ_z)
    pred = surface_pred.clone()
    pred[..., 2] *= -1  # τ_y flips sign under y-reflection
    return pred
```

**Training step (when `--symm-pair-loss`):**
```python
# Forward original batch
pred_orig = model(batch)  # [B, N, 4] surface + [B, M, 1] volume
loss_orig = compute_loss(pred_orig, batch.targets)

# Forward mirrored batch
sx_mirror, vx_mirror = mirror_y(batch.surface_x, batch.volume_x)
batch_mirror = batch_with_new_inputs(batch, sx_mirror, vx_mirror)
pred_mirror = model(batch_mirror)
# Apply symmetry to predictions and compare with original targets
pred_mirror_corrected = flip_ws_y(pred_mirror)  # flip τ_y to match original orientation
loss_mirror = compute_loss(pred_mirror_corrected, batch.targets)

# Averaged loss
loss = 0.5 * loss_orig + 0.5 * loss_mirror
```

Also log `train/symm_consistency = mean(|pred_orig - pred_mirror_corrected|^2)` — should decrease toward 0 as model converges to symmetric predictions.

**Test-time evaluation:** Always run test as ensemble of original + mirror (regardless of `--symm-pair-loss`):
```python
pred_ens = 0.5 * (model(x) + flip_ws_y(model(mirror_y(x))))
```
Report BOTH standard val_abupt (no ensemble) and `val_abupt_mirror_ens` (with ensemble) for both arms so we can see whether the ensemble itself helps even without the paired loss.

**Sweep plan — 2 arms, 4-GPU DDP:**

```bash
# Arm A: control (no symm-pair-loss, NO ensemble at eval)
WANDB_GROUP=fern-r33-symm-pair \
  --learnable-pe --optimizer lion --lr 1e-4 --grad-clip-norm 0.5 \
  --batch-size 4 --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 --epochs 50

# Arm B: paired symmetry loss + mirror ensemble at eval
WANDB_GROUP=fern-r33-symm-pair \
  --learnable-pe --optimizer lion --lr 1e-4 --grad-clip-norm 0.5 \
  --batch-size 4 --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 --epochs 50 \
  --symm-pair-loss
```

Note: `--symm-pair-loss` doubles the per-step forward passes (doubles GPU memory and time per step). If the batch effectively doubles, you may need to halve `--batch-size` to 2 to stay within VRAM. Test memory usage on step 0 and adjust.

Also add: Even in Arm A control run, evaluate with mirror ensemble at the end (`val_abupt_mirror_ens` metric) — this quantifies the free TTA gain independent of training.

## Baseline

Active yi merge bar: **val_abupt < 9.032%** (PR #517 askeladd, run `brat65z4`)

| Metric | Current bar (PR #517) | Tay SOTA ref (PR #511) | AB-UPT target |
|---|---:|---:|---:|
| val_abupt | 9.032% | 7.013% | — |
| test_abupt | 10.119% | 8.313% | — |
| test τ_y | — | 8.582% | 3.65% (2.35×) |

Win criterion: val_abupt < 9.032 in Arm B (using either standard or mirror-ensemble metric). If mirror ensemble alone (no paired loss) drops val_abupt below bar in Arm A, that's a free win worth logging separately.


morganmcg1:
ADVISOR: Approved — proceed with the y-symmetry paired-batch loss.

**Important context — this is NOT PR #225.** PR #225 (random y-flip augmentation) was closed as null. Your design is fundamentally different: hard paired-batch constraint (both x and mirror_y(x) in the same forward, joint loss including a sign-flip on tau_y). This is a much stronger inductive bias than random augmentation because the model sees the violation as a per-batch differentiable signal, not a stochastic regularizer. So the prior null does NOT predict failure here.

**Required implementation checks:**
1. \`mirror_y(surface_x, volume_x)\` must flip:
   - y coordinate (index 1 of xyz)
   - n_y normal (index 1 of normal vector)
   - leave area, x, z, n_x, n_z unchanged
2. \`flip_ws_y(surface_pred)\` must flip the sign of the tau_y output channel (index 2 of [cp, tau_x, tau_y, tau_z]) and ALSO flip the sign of mu_y if you have a coupled head — be careful about this.
3. Loss = 0.5*loss(f(x), y) + 0.5*loss(flip_ws_y(f(mirror_y(x))), y) — the symmetric expectation. Both halves use the SAME ground truth y; only the input and the output sign-flip differ.
4. Test-time ensemble must be applied unconditionally: predict = 0.5*(f(x) + flip_ws_y(f(mirror_y(x)))). Report val_abupt under both \`val_abupt\` (single-pass) and \`val_abupt_mirror_ens\` (ensemble) — the ensemble version is the merge target.

**Diagnostics:**
- \`train/symm_consistency\` = mean L2 between f(x) and flip_ws_y(f(mirror_y(x))) per epoch — should decrease monotonically if the constraint is being learned.
- \`val/tau_y_rel_l2_pct\` and \`val/tau_z_rel_l2_pct\` per epoch — the symmetry constraint should help tau_y/tau_z disproportionately.

**Memory note:** paired batch doubles forward memory. With bs=4 you may need bs=2 or to checkpoint the encoder. Report your final batch size when launching.

**Win condition:** val_abupt_mirror_ens < 9.032% (active merge bar). If the single-pass val_abupt is also below 9.032%, that's a stronger result. We'll merge on the ensemble metric — report both.

W&B group: \`yi-r34-fern-symm-pair\`. Use \`--clip-grad-norm 0.5\`.

morganmcg1:
STUDENT fern: Status update

Implementation in place (already on branch from prior session):
- `mirror_y_surface_x` flips y-coord (idx 1) and ny-normal (idx 4); area (idx 6), nx (idx 3), nz (idx 5), x (idx 0), z (idx 2) unchanged. **Note**: surface_x has 7 channels (xyz+normals+area), not 6 as the PR body spec says — indices verified.
- `mirror_y_volume_x` flips y-coord (idx 1); sdf invariant.
- `flip_y_surface_pred` flips τ_y (idx 2) of [cp, τ_x, τ_y, τ_z]; cp/τ_x/τ_z unchanged.
- Loss: 0.5·L(f(x), y) + 0.5·L(flip_y(f(mirror_y(x))), y) — both halves use SAME ground-truth y; only input and output sign-flip differ.
- Test-time mirror ensemble runs unconditionally (both arms get `val_abupt` and `val_abupt_mirror_ens`), so the free-TTA gain is isolated.
- `train/symm_consistency`, `symm_pair_loss_orig/mirror`, per-axis mirror losses all logged.

Research pass (Frame Averaging Puny et al. ICLR 2022, REMUL ICLR 2025, Wang/Walters/Yu ICLR 2021): paired symmetrization is theoretically stronger than random mirror augmentation (per-sample vs per-distribution constraint). Watch for "lazy symmetric collapse" where τ_y → 0 satisfies the paired loss trivially — `train/symm_consistency` is the right diagnostic; if it drops faster than τ_y/τ_z metrics improve, that's degenerate. LayerNorm is safe with the doubled-batch concat (per-sample stats; no BN contamination).

**Run plan (sequential due to single-pod 4-GPU constraint):**
- **Arm A (running now):** v2 launched 20:57 UTC, W&B run `1akvkfh7`, group `fern-r33-symm-pair`. ~31% through epoch 1 at t+30min. ETA ~85 min/epoch on 4 GPUs → expect ~3-4 epochs in 360-min timeout. Will record val_abupt at epoch 1 + timeout-hit epoch + final-eval.
- **Arm B:** queued, will launch on the same 4 GPUs after Arm A finishes.

Both arms use exactly the advisor's command (--clip-grad-norm 0.5, no --validation-every override → default 10). Note: with the 360-min budget on 4 GPUs we expect to complete roughly 4 epochs vs the historic 9-epoch runs that hit val_abupt 9.032% on 8 GPUs. The merge bar may be unreachable in this single-arm budget; the comparative claim will be Arm B vs Arm A at matched compute, plus the standalone test-time mirror-ensemble gain in Arm A.

morganmcg1:
ADVISOR: 7-channel surface_x (xyz + nx,ny,nz + area) clarification accepted — flipping y-coord (idx 1) and ny-normal (idx 4) is the correct mirror operation for y-symmetry. Pred y-flip on channels τ_y (idx 2) is correct — surface_pressure (idx 0) and τ_x (idx 1), τ_z (idx 3) are y-symmetric (do NOT flip), τ_y is y-antisymmetric (sign flip).

Quick sanity check before you commit GPU: please verify the mirror leaves volume_pressure unchanged (volume scalar field is y-symmetric, no flip needed) and that the mirror+forward path produces the same prediction up to floating-point on a single sample (i.e. `model(mirror(x)) ≈ mirror_pred(model(x))` should be exactly the loss → 0 floor your aux term is pulling toward).

Headline bar is 9.032% val_abupt. Continue with the planned arms.

morganmcg1:
STUDENT fern: Pre-GPU sanity checks (responding to advisor's check request)

Ran the requested verification on the mirror/flip helpers (`/workspace/senpai/target/train.py:1518-1534`). All four physics conditions hold:

**1. `mirror_y_surface_x` (7-channel `[xyz, nx, ny, nz, area]`)**
- y (idx 1): sign-flipped ✓
- n_y (idx 4): sign-flipped ✓
- x (0), z (2), n_x (3), n_z (5), area (6): unchanged ✓
- involutive: `mirror(mirror(x)) == x` ✓

**2. `mirror_y_volume_x` (4-channel `[xyz, sdf]`)**
- y (idx 1): sign-flipped ✓
- x (0), z (2), sdf (3): unchanged ✓ (sdf is reflection-invariant; signed distance to a y-symmetric surface is the same at (x,y,z) and (x,-y,z))
- involutive ✓

**3. `flip_y_surface_pred` (4-channel `[cp, τ_x, τ_y, τ_z]`)**
- τ_y (idx 2): sign-flipped ✓
- cp (0), τ_x (1), τ_z (3): unchanged ✓
- involutive ✓

**4. Volume prediction is correctly NOT sign-flipped**
- volume pressure is a y-symmetric scalar field: `p(x,-y,z) = p(x,y,z)` ⇒ no `flip_y_volume` function needed (and none defined)
- training: `volume_pred_mirror` is compared against the unflipped `volume_target` (line 1659) ✓
- eval ensemble: `0.5 * (volume_pred_orig + volume_pred_mirror_norm)` (line 1802), no sign flip ✓

**Aux-term floor verification**
The `train/symm_consistency` metric is computed (line 1690) as
`mean((f(x) - flip_pred(f(mirror(x))))² · mask)`
which is exactly the equivariance gap. It is 0 iff `f(x) = flip_pred(f(mirror(x)))` ⟺ `f(mirror(x)) = flip_pred(f(x))` (using involutivity of `flip_pred`). So this is the right floor the aux term pulls toward — matches your specification exactly. (Will be logged only in Arm B since the metric is gated on `symm_pair_loss=True`.)

**Status**: Arm A (control, run `1akvkfh7`) just finished epoch 1 (5441 steps; epoch-1 validation in flight at the time of this comment). Arm B will queue on the same 4-GPU pod after Arm A completes.

morganmcg1:
STUDENT fern: Status update — torch.compile inductor bug at validation, relaunching without compile

**Arm A v2 (run `1akvkfh7`, control)**: Trained the full epoch 1 (5441 steps, ~87 min) but **crashed at validation entry** with:
```
torch._inductor.exc.InductorError: AssertionError: For s24 + 65536, expected [s24] to have been codegen-ed.
```
This is a torch.compile dynamic-shape bug triggered when eval-time tensor shapes differ from training-time shapes. Specifically, eval batches use deterministic strided point chunks (per program.md sampling contract) which don't have the exact `points_per_view=65536` count that training uses with replacement sampling. The padded second dim of `surface_x` becomes a symbolic dim `s24` that wasn't pre-codegen'd, so inductor bails. **No checkpoint or validation metrics produced** because the save-best-checkpoint logic runs *after* eval.

**Arm B (paired) bs=2**: Started cleanly at 22:28 UTC (run `mh6wrw9i`), reached step ~600 (~10 min in), `train/symm_consistency` already low at ~2e-3. Training was healthy. **But this run was guaranteed to hit the same eval crash at epoch 1 boundary (~165 min from start)**. Killed at 22:39 UTC to save compute (~155 min that would have been wasted).

**Relaunch plan (sequential, single 4-GPU pod):**
1. **Arm B v3** (just launched, pid=811510): same hyperparameters + `--no-compile-model`. Without compile, expect ~30–50% slower steps (~0.65 it/s), but eval will not crash. With `SENPAI_TIMEOUT_MINUTES=360` (270 min training budget), Arm B will complete ~1 epoch + timeout-triggered eval + final full_val/test on best checkpoint.
2. **Arm A v3** (control, queued): same launch but with `--symm-pair-loss` removed and bs=4 + `--no-compile-model`. Will follow Arm B.

W&B: `fern-r33-symm-pair` group, run name `fern/symm-pair-armB-paired-bs2-noc`. Will report results once both arms finish.

morganmcg1:
ADVISOR: Good diagnosis. The Inductor tiling assertion is a known issue with dynamic shapes at validation boundaries — `--no-compile-model` is the right workaround here.

For both Arm A v3 and Arm B v3 with `--no-compile-model`: expect ~30% slower step throughput, so you'll see fewer epochs in the 360-min budget. That's fine — the hypothesis test is still valid.

Key metrics to watch during the run:

1. **`train/symm_consistency`** — should decrease monotonically and stabilize by epoch 2–3. If it plateaus high (>0.05), the mirror augmentation may not be coupling well with the loss.
2. **`val_abupt`** — compare Arm B (with symm pair loss) vs Arm A v3 (control, no symm loss) directly. The delta tells us whether the y-equivariance constraint is helping.
3. **`val_abupt_mirror_ens`** — the test-time ensemble metric; if the model has learned equivariance, this should match or beat `val_abupt`.

Target: beat 9.032% (current bar from askeladd PR #517). Even a small improvement like 8.9% would be a meaningful signal that the symmetry prior is useful.

Good luck with the relaunched runs.

morganmcg1:

PR #539, opened 2026-05-03
frieren: STRING-sep PE + Lion lr=1e-4 clip=0.5 from scratch

morganmcg1:
## Hypothesis

PR #490 (frieren STRING-sep learnable PE) and PR #517 (askeladd Lion lr=1e-4 clip=0.5) have both been merged into `yi`. The combination of `--learnable-pe` (per-axis learnable `log_freq + phase` in `ContinuousSincosEmbed`) with the Lion lr=1e-4/clip=0.5 tuning was proven in your resumed run `zwh9qzjw` (best_val=8.087% vs bar 9.032%). However that run **resumed from a checkpoint** (run `0p3rjqzz` at epoch 3) rather than training from scratch — making it a non-canonical merge candidate.

This PR tests the canonical from-scratch combination on the now-merged yi codebase: Lion lr=1e-4 clip=0.5 + `--learnable-pe`. Expected: val_abupt ≤ 8.5%, locking in a clean new baseline ~1pp below the current 9.032% bar.

**Reference runs:**
- Frieren resume `zwh9qzjw`: val=8.087% (STRING-sep + Lion, BUT resumed from checkpoint — non-canonical)
- Askeladd `brat65z4` (PR #517 merged bar): val=9.032% (Lion lr=1e-4 clip=0.5, NO STRING-sep)
- Frieren initial `0p3rjqzz`: val=10.318% (STRING-sep, compute-starved at 2 full epochs)

The from-scratch full-budget run closes this gap.

## Instructions

No code changes needed — the yi codebase already has both `--learnable-pe` (PR #490) and Lion lr=1e-4 clip=0.5 (PR #517). Your task is to run from scratch with the combined config.

**Arm A (canonical from-scratch):**
```bash
SENPAI_VAL_BUDGET_MINUTES=30 \
WANDB_GROUP=frieren-r33-string-sep-lion-scratch \
torchrun --standalone --nproc_per_node=4 target/train.py \
  --learnable-pe \
  --optimizer lion \
  --lr 1e-4 \
  --grad-clip-norm 0.5 \
  --weight-decay 5e-4 \
  --lr-warmup-epochs 1 \
  --ema-decay 0.999 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --batch-size 4 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --epochs 50
```

[...]

PR #545, opened 2026-05-03
senku: RoPE on Transolver slice tokens keyed on 3D centroids

morganmcg1:
## Hypothesis

The Transolver architecture uses slice tokens as compressed latent representations of point-cloud subsets. Currently, slice tokens have no explicit positional encoding — they're permutation-invariant with respect to their identity. Adding Rotary Position Embeddings (RoPE) to the slice tokens, keyed on the **centroid coordinates** of each slice (weighted average of points assigned to that slice), should give the model an inductive bias that nearby-in-space slices attend to each other preferentially and distant slices are discounted. This is particularly important for DrivAerML's 3D geometry where spatial locality of slice assignments strongly correlates with physical coherence.

RoPE on slice tokens (rather than on point tokens) is cheap: there are only `--model-slices 128` slice tokens vs 65536 point tokens, so the overhead is negligible. The key insight from RoPE (Su et al. 2021, https://arxiv.org/abs/2104.09864) is that it encodes relative position implicitly via rotation of Q/K vectors — exactly what we want for slice-token attention. Modded-NanoGPT (https://github.com/KellerJordan/modded-nanogpt) demonstrated that RoPE + modern attention is a reliable win even on small models.

This is **different from PR #490 (STRING-sep learnable PE)** which adds PE to the input point tokens. PR #490 is already merged on yi. This PR adds PE to the *slice tokens* inside the Transolver attention layers.

### 1. Add a `SliceRoPE` module

In `target/train.py` (or the model file it imports), add a `SliceRoPE` class:

```python
import torch, math

class SliceRoPE(torch.nn.Module):
    """RoPE for slice tokens keyed on 3D centroid coordinates."""
    def __init__(self, head_dim: int, max_freq: float = 10.0):
        super().__init__()
        # head_dim must be divisible by 6 (2 per spatial axis)
        assert head_dim % 6 == 0, f"head_dim {head_dim} must be divisible by 6 for 3D RoPE"
        d = head_dim // 6  # frequencies per axis
        inv_freq = 1.0 / (max_freq ** (torch.arange(0, d, dtype=torch.float32) / d))
        self.register_buffer("inv_freq", inv_freq)

    def _rotate_half(self, x):
        x1, x2 = x[..., ::2], x[..., 1::2]
        return torch.stack([-x2, x1], dim=-1).flatten(-2)

    def forward(self, q, k, centroids):
        """
        q, k: (B, heads, S, head_dim) — slice-token Q/K
        centroids: (B, S, 3) — xyz centroid of each slice
        Returns rotated (q, k)
        """
        B, H, S, D = q.shape
        # Build per-axis frequency embeddings
        freqs = []
        for axis in range(3):
            c = centroids[..., axis]  # (B, S)
            f = torch.einsum("bs,d->bsd", c, self.inv_freq)  # (B, S, d)
            freqs.append(torch.cat([f, f], dim=-1))  # (B, S, 2d)
        theta = torch.cat(freqs, dim=-1)  # (B, S, head_dim)
        theta = theta.unsqueeze(1)  # (B, 1, S, head_dim)
        cos_t, sin_t = theta.cos(), theta.sin()
        q_rot = q * cos_t + self._rotate_half(q) * sin_t
        k_rot = k * cos_t + self._rotate_half(k) * sin_t
        return q_rot, k_rot
```

### 2. Compute slice centroids in the Transolver attention forward

In the Transolver slice-attention forward pass (wherever `slice_tokens` are assembled from point assignments), compute centroids **after** the soft assignment:

```python
# soft_assign: (B, N, S) — softmax weights over slices for each point
# point_coords: (B, N, 3) — xyz of each input point
# centroid[b, s] = sum_n(soft_assign[b,n,s] * point_coords[b,n]) / sum_n(soft_assign[b,n,s])
centroid_num = torch.einsum("bns,bnd->bsd", soft_assign, point_coords)  # (B, S, 3)
centroid_den = soft_assign.sum(dim=1, keepdim=True).transpose(1,2) + 1e-8  # (B, S, 1)
slice_centroids = centroid_num / centroid_den  # (B, S, 3)
```

Then apply `SliceRoPE` to Q/K before the dot-product attention:
```python
q_rot, k_rot = self.slice_rope(q, k, slice_centroids)
attn = (q_rot @ k_rot.transpose(-2, -1)) * scale
```

### 3. Add a `--slice-rope` flag

```python
parser.add_argument("--slice-rope", action="store_true",
    help="Add RoPE to Transolver slice tokens keyed on slice centroids.")
```

Only instantiate `SliceRoPE` when `--slice-rope` is set. Use `--wandb-group yi-r34-senku-slice-rope` so both arms are grouped.

### 4. Sweep plan (2 arms, 4-GPU DDP each)

**Arm A — control** (no RoPE, post-#490 baseline):
```bash
  --learnable-pe \
  --batch-size 4 \
  --epochs 50 \
  --wandb-group yi-r34-senku-slice-rope
```

**Arm B — slice RoPE**:
Same as Arm A plus `--slice-rope`.

### 5. Logging

Add to W&B every epoch:
- `train/slice_centroid_spread_mean` — mean pairwise L2 distance between slice centroids (should be non-trivial for RoPE to matter)
- `train/slice_rope_q_cos_sim` — mean cosine similarity of q before/after rotation (sanity check; should differ from 1.0)

### Decision rule
- RoPE arm beats val_abupt < 9.032% (active merge bar) → open merge request.
- RoPE arm worse than control by >1pp → close as null.

## Baseline

**Current yi merge bar: val_abupt 9.032%** (PR #517 askeladd, W&B run `brat65z4`)
- **Expected next bar: ~8.1–8.5%** once PR #539 (frieren STRING-sep + Lion from-scratch) completes

| Metric | yi bar (PR #517) |
|---|---:|
| `abupt_axis_mean_rel_l2_pct` | 9.032 |
| `surface_pressure_rel_l2_pct` | ~5.3 (est) |
| `wall_shear_y_rel_l2_pct` | ~12.9 (est) |
| `wall_shear_z_rel_l2_pct` | ~13.0 (est) |
| `volume_pressure_rel_l2_pct` | ~5.5 (est) |

W&B: https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/brat65z4

**Reproduce baseline:**
```bash
  --learnable-pe \
  --batch-size 4 \
  --epochs 50
```

### DIFF
--- a/train.py
+++ b/train.py
@@ -301,8 +301,85 @@ class UpActDownMlp(nn.Module):
         return self.fc2(self.act(self.fc1(x)))
 
 
+class SliceRoPE(nn.Module):
+    """3D Rotary Position Embedding for Transolver slice tokens.
+
+    Keyed on per-head slice centroids (xyz). Splits ``head_dim`` into 3 axis groups,
+    each contributing ``2 * (head_dim // 6)`` channels. If ``head_dim`` is not divisible
+    by 6 the residual ``head_dim - 6 * (head_dim // 6)`` channels pass through unrotated
+    (e.g. head_dim=64 → rotate first 60 channels, pass 4 untouched).
+
+    Convention: interleaved frequency packing matched with interleaved rotate_half so
+    consecutive (even, odd) channels form a complex pair sharing one rotation angle.
+    """
+
+    def __init__(self, head_dim: int, max_freq: float = 10.0):
+        super().__init__()
+        d = head_dim // 6
+        if d == 0:
+            raise ValueError(
+                f"head_dim {head_dim} too small for 3D RoPE (need at least 6, got {head_dim}/6={d})"
+            )
+        self.head_dim = head_dim
+        self.max_freq = float(max_freq)
+        self.d_per_axis = d
+        self.rot_dim = 6 * d
+        inv_freq = 1.0 / (self.max_freq ** (torch.arange(0, d, dtype=torch.float32) / d))
+        self.register_buffer("inv_freq", inv_freq, persistent=False)
+
+    @staticmethod
+    def _rotate_half_interleaved(x: torch.Tensor) -> torch.Tensor:
+        x1 = x[..., 0::2]
+        x2 = x[..., 1::2]
+        return torch.stack([-x2, x1], dim=-1).flatten(-2)
+
+    def forward(
+        self,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        centroids: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        # q, k: (B, H, S, head_dim); centroids: (B, H, S, 3)
+        rot_dim = self.rot_dim
+        inv_freq = self.inv_freq
+        # Build per-axis frequency embeddings, then concat across axes.
+        # Pack each axis as [f0, f0, f1, f1, ..., f_{d-1}, f_{d-1}] to match the
+        # interleaved rotate_half convention.
+        axis_blocks = []
+        for axis in range(3):
+            c = centroids[..., axis]  # (B, H, S)
+            f = torch.einsum("bhs,d->bhsd", c.float(), inv_freq)  # (B, H, S, d)
+            f_pair = torch.repeat_interleave(f, 2, dim=-1)  # (B, H, S, 2d)
+            axis_blocks.append(f_pair)
+        theta = torch.cat(axis_blocks, dim=-1)  # (B, H, S, 6d)
+        cos_t = theta.cos().to(q.dtype)
+        sin_t = theta.sin().to(q.dtype)
+
+        q_rot = q[..., :rot_dim]
+        k_rot = k[..., :rot_dim]
+        q_rotated = q_rot * cos_t + self._rotate_half_interleaved(q_rot) * sin_t
+        k_rotated = k_rot * cos_t + self._rotate_half_interleaved(k_rot) * sin_t
+
+        if rot_dim < q.shape[-1]:
+            q_pass = q[..., rot_dim:]
+            k_pass = k[..., rot_dim:]
+            q_out = torch.cat([q_rotated, q_pass], dim=-1)
+            k_out = torch.cat([k_rotated, k_pass], dim=-1)
+        else:
+            q_out = q_rotated
+            k_out = k_rotated
+        return q_out, k_out
+
+
 class TransolverAttention(nn.Module):
-    def __init__(self, hidden_dim: int, num_heads: int, num_slices: int, dropout: float = 0.0):
+    def __init__(
+        self,
+        hidden_dim: int,
+        num_heads: int,
+        num_slices: int,
+        dropout: float = 0.0,
+        slice_rope: SliceRoPE | None = None,
+    ):
         super().__init__()
         if hidden_dim % num_heads != 0:
             raise ValueError("hidden_dim must be divisible by num_heads")
@@ -319,6 +396,10 @@ class TransolverAttention(nn.Module):
         self.qkv = LinearProjection(self.dim_head, self.dim_head * 3, bias=False)
         self.proj = LinearProjection(hidden_dim, hidden_dim)
         self.proj_dropout = nn.Dropout(dropout)
+        self.slice_rope = slice_rope
+        # Telemetry — populated per forward when slice_rope is enabled.
+        self.last_centroid_spread: float = 0.0
+        self.last_rope_q_cos_sim: float = 1.0
 
     def create_slices(self, x: torch.Tensor, attn_mask: torch.Tensor | None = None) -> tuple[torch.Tensor, torch.Tensor]:
         batch_size, num_tokens, _ = x.shape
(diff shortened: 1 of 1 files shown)


morganmcg1:
ADVISOR: Approved — proceed with the SliceRoPE 2-arm sweep.

**Compatibility with PR #490 (already merged on yi):** PR #490 added STRING-sep learnable PE on **input point tokens**. Your PR adds RoPE on **slice tokens** inside the Transolver attention layers. These are orthogonal mechanisms — the point-PE survives slice projection only as far as the soft-assign aggregation carries it, so adding RoPE on slice centroids is non-redundant. Both arms in your sweep already include \`--learnable-pe\`, which is correct.

**Architectural sanity checks before launching:**
1. **head_dim divisibility:** with model-hidden-dim=512 and heads=8, head_dim=64. 64 is NOT divisible by 6. Your assert will fire. You have two options:
   - (a) Use heads=4 (head_dim=128, divisible by 4 but not 6 either) — wrong, 128 also fails.
   - (b) **Recommended:** drop the strict head_dim%6==0 requirement to head_dim>=6 and pad/truncate the per-axis frequency dim. Use d = head_dim // 6 frequencies per axis, leaving (head_dim - 6d) channels unrotated. For head_dim=64: d=10, rotation occupies 60 channels, 4 unrotated. This is the standard NanoGPT/HF approach and it works fine in practice.
   - (c) Or change to model-heads=12 to make head_dim ~42 — but this changes the model and breaks the controlled comparison.
   Pick (b). Update the assert to \`assert head_dim >= 6\`.

2. **Centroid computation:** soft_assign is (B, N, S). Make sure you are using the SAME soft_assign that produces the slice tokens, and that gradients flow through it (do not detach centroids — RoPE through differentiable centroids is what makes this learnable).

3. **Numerical stability:** slice centroids may collapse near a single point early in training. Log \`train/slice_centroid_spread_mean\` from epoch 1 — if the spread is <0.1m, RoPE has no signal to encode and we'll see a null. Add \`train/slice_centroid_spread_std\` too for diagnostics.

**Win condition:** Arm B (RoPE) val_abupt < 9.032% AND ≥0.3pp improvement over Arm A (control). If <0.3pp delta we close as null.

**CLI:** use \`--clip-grad-norm 0.5\` (not \`--grad-clip-norm\`).

W&B group: \`yi-r34-senku-slice-rope\`.

morganmcg1:
ADVISOR: Checking in — your approval was posted ~12+ hours ago. Please confirm you've launched both arms of the SliceRoPE sweep (Arm A control + Arm B RoPE), and share your W&B run IDs. Remember the head_dim fix: use d = head_dim // 6 = 10 freqs per axis, 4 unrotated channels, and assert head_dim >= 6. Share early centroid-spread logs (train/slice_centroid_spread_mean) once you have a few steps.

morganmcg1:
ADVISOR check-in 2: It's been ~5 hours since the 20:17 UTC approval for the 2-arm RoPE sweep, and ~2.5 hours since my first check-in at 23:05 UTC, with no launch confirmation or W&B run id posted. Your pod is healthy.

If you're stuck on the head_dim configuration or anything else in the build, post a question now and I'll unblock you. If you don't post a launch confirmation in the next 60 minutes, I'll close this PR and reassign the RoPE-on-slice-tokens hypothesis to a different student to keep the GPU productive — the idea is good and someone should be testing it.

To unblock fast: paste your full launch command + W&B run id in a comment, or paste the error/blocker you hit.

morganmcg1:
STUDENT senku:

## Launch confirmation — Arm A (control) running

Sorry for the delay; my earlier (now-discarded) launch hit two foot-guns I had to fix before relaunching. Documenting both so it's clear what changed.

### Fixes applied before launch (commit `2a12a06`)

1. **Stopped detaching slice centroids before SliceRoPE.** Per your review note ("RoPE through differentiable centroids is what makes this learnable") — the v1 code had `.detach()` on the centroid tensor, breaking gradient flow from the rotation angles back through `slice_weights → in_project_slice → backbone params`. Now centroids carry grads into RoPE; the telemetry block still uses local `detach()`/`torch.no_grad()` for spread + cosine-sim computation so the autograd graph isn't extended.
2. **Added `train/slice_centroid_spread_std`** alongside `_mean` (mean of per-(B,H) std of pairwise centroid distance), per your "add spread_std for diagnostics" ask.

### Architectural sanity check (your review pt. 1)
Took option (b): partial rotation, `assert head_dim >= 6`. With `--model-hidden-dim 512 --model-heads 8` → `head_dim=64`, `d=10` per axis, `rot_dim=60` rotated channels and 4 untouched. Smoke-verified: gradient flows through `in_project_slice.{weight,bias}` end-to-end with non-zero magnitudes; spread/q-cos-sim populate per step.

### Earlier-attempt diagnostics (no longer relevant, just for context)
- v1 arm-a crashed end of epoch 1 with `torch._inductor.exc.InductorError`. Root cause: ran with `--compile-model` (default True) where the matching baseline brat65z4 used `compile_model: False`. Now using `--no-compile-model`.
- The first relaunch this session was 2x slower than baseline (1.25 s/iter vs brat65z4's ~0.68 s/iter). Cause was the per-step gradient/weight metric collection (defaults are `gradient_log_every: 1`, `weight_log_every: 1`); brat65z4 used 100. Killed and matched the baseline. Now at ~1.4 it/s steady state — projects to ~5 epochs inside the 360-min `SENPAI_TIMEOUT_MINUTES`, same as brat65z4's actual `_step=21986` at termination.

### Arm A — control, running now

- **W&B run:** `9hwww7x9` — https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/9hwww7x9
- **Group:** `yi-r34-senku-slice-rope`
- **Command:**
```bash
  --learnable-pe \
  --optimizer lion --lr 1e-4 --clip-grad-norm 0.5 \
  --batch-size 4 \
  --epochs 50 \
  --no-compile-model \
  --gradient-log-every 100 --weight-log-every 100 \
  --agent senku \
  --wandb-group yi-r34-senku-slice-rope \
  --wandb-name "senku/slice-rope-arm-a-control"
```

### Arm B — slice RoPE, queued

I have only 4 GPUs on this pod (not 8), so I'm running Arm A and Arm B sequentially on the full 4-GPU set rather than splitting 2+2 (which would slow each arm enough that neither would finish 4–5 epochs before the per-run timeout). Will launch Arm B with the same flags + `--slice-rope --slice-rope-max-freq 10.0` immediately after Arm A finishes; same group, name `senku/slice-rope-arm-b-slice-rope`. If you'd rather I split now in parallel, say so and I'll relaunch.

### Early centroid-spread sanity (Arm B will report on real data)
Synthetic-CUDA debug run with `--slice-rope` and randomly-scaled point coords: `slice_centroid_spread_mean ≈ 0.03`, `q_cos_sim ≈ 0.9999` (synthetic centroids are sub-meter so rotation angles are tiny). On real DrivAerML xyz the spread should be O(meters) and cos-sim well below 1.0; I'll post the actual value once Arm B is past warmup.

morganmcg1:

PR #3, opened 2026-04-28
[askeladd] Round-1: codex/optimized-lineage config (4L/256d, 65k pts, lr=2e-4)

morganmcg1:
## Hypothesis

The `codex/optimized-lineage` branch in `morganmcg1/DrivAerML` ships a config with
a larger model (4L/256d/4h/128sl), higher point counts (65k), lower lr (2e-4), higher
weight decay (5e-4), and stronger EMA (0.9995). These changes were included as a
"proven stronger baseline" — this PR runs that exact config so we can measure its
absolute benefit over the stock defaults on the fresh `yi` W&B project.

No code changes needed — only CLI flags.

## Instructions

No modifications to `train.py` are needed.

```bash
cd target/
python train.py \
  --lr 2e-4 \
  --weight-decay 5e-4 \
  --train-surface-points 65536 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --model-slices 128 \
  --ema-decay 0.9995 \
  --wandb-group round1-codex-lineage \
  --wandb-name askeladd-codex-lineage
```

All other flags stay at defaults (`batch_size=2`, `epochs` governed by `SENPAI_MAX_EPOCHS`,
`amp=bf16`, `validation_every=10`, etc.).

This config is ~3× the parameter count of the default baseline:
- Default: 3L/192d = small
- This run: 4L/256d/4h/128sl = ~3× params

## Baseline

No prior `yi` runs exist. Targets to beat (AB-UPT public reference, lower is better):

| Metric | AB-UPT target |
|---|---:|
| `test_primary/surface_pressure_rel_l2_pct` | **3.82** |
| `test_primary/wall_shear_rel_l2_pct` | **7.29** |
| `test_primary/volume_pressure_rel_l2_pct` | **6.08** |
| `test_primary/wall_shear_x_rel_l2_pct` | **5.35** |
| `test_primary/wall_shear_y_rel_l2_pct` | **3.65** |
| `test_primary/wall_shear_z_rel_l2_pct` | **3.63** |

Compare your results to PR #2 (alphonse, stock defaults) once both finish.

## Results (fill in after run)

Add a PR comment with:
1. `test_primary/abupt_axis_mean_rel_l2_pct` and all six per-target `test_primary/*_rel_l2_pct`.
2. `full_val_primary/*` numbers.
3. Wall-clock run time and epoch reached.
[...]

PR #2, opened 2026-04-28
[alphonse] Round-1 reference: stock train.py defaults baseline

morganmcg1:
## Hypothesis

Run `train.py` with its stock defaults to establish the calibration floor on the
`yi` branch. W&B project `wandb-applied-ai-team/senpai-v1-drivaerml` currently has
**zero** completed runs, so this PR pins every `test_primary/*` metric for the first
time and gives all other Round-1 experiments a concrete comparison point.

No code changes needed — this is a pure configuration reference run.

## Instructions

No modifications to `train.py` are needed.

Run the following on your 4-GPU pod:

```bash
cd target/
python train.py \
  --wandb-group round1-default-baseline \
  --wandb-name alphonse-default-baseline
```

Active defaults (do **not** change unless `SENPAI_MAX_EPOCHS` / `SENPAI_TIMEOUT_MINUTES` forces it):

| Flag | Value |
|---|---|
| `--lr` | 3e-4 |
| `--weight-decay` | 1e-4 |
| `--batch-size` | 2 |
| `--epochs` | 50 |
| `--train-surface-points` | 40000 |
| `--eval-surface-points` | 40000 |
| `--train-volume-points` | 40000 |
| `--eval-volume-points` | 40000 |
| `--model-layers` | 3 |
| `--model-hidden-dim` | 192 |
| `--model-heads` | 3 |
| `--model-slices` | 96 |
| `--ema-decay` | 0.999 |
| `--amp-mode` | bf16 |
| `--validation-every` | 10 |

## Baseline

No prior `yi` runs exist. Targets to beat (AB-UPT public reference, lower is better):

| Metric | AB-UPT target |
|---|---:|
| `test_primary/abupt_axis_mean_rel_l2_pct` | — |
| `test_primary/surface_pressure_rel_l2_pct` | **3.82** |
| `test_primary/wall_shear_rel_l2_pct` | **7.29** |
| `test_primary/volume_pressure_rel_l2_pct` | **6.08** |
| `test_primary/wall_shear_x_rel_l2_pct` | **5.35** |
| `test_primary/wall_shear_y_rel_l2_pct` | **3.65** |
| `test_primary/wall_shear_z_rel_l2_pct` | **3.63** |

## Results (fill in after run)

Add a PR comment with:
1. `test_primary/abupt_axis_mean_rel_l2_pct` and all six per-target `test_primary/*_rel_l2_pct`.
2. `full_val_primary/*` numbers (best-checkpoint validation).
[...]

PR #547, opened 2026-05-03
tanjiro: joint 2D Cholesky NLL head for τ_y/τ_z wall-shear

morganmcg1:
## Hypothesis

PR #530 (tangent-frame head) failed because hairy-ball discontinuities (basis_jump_p99 = 89.7°) corrupted the frame assignment. The **root failure** was that the basis construction used a single global reference vector → unavoidable topological discontinuity on a closed surface.

The τ_y and τ_z channels are not independent: on a car, τ_y (lateral shear) and τ_z (roof/underbody shear) are the two "cross-flow" components. They are physically correlated — where τ_y is large (door mirror wake, A-pillar), τ_z is often also non-zero (3D vortex has both y and z components). Predicting them **jointly** with a shared 2D output head that captures their cross-channel covariance should improve both.

**Approach:** Add a low-rank 2×2 covariance-aware output head for (τ_y, τ_z) only. Instead of predicting (μ_y, μ_z) independently, predict (μ_y, μ_z, L_11, L_12, L_22) where L is the Cholesky factor of a 2×2 covariance matrix. Train with a 2D NLL loss: `0.5 * (log det Σ + (τ - μ)^T Σ^{-1} (τ - μ))`. This is a principled joint prediction that:
1. Encourages the model to learn the τ_y/τ_z correlation structure
2. Weights predictions by their own uncertainty (high-separation regions with large Σ get appropriate gradient scaling)
3. Requires no coordinate-frame construction — no hairy-ball problem

The key difference from PR #496 (tanjiro homoscedastic uncertainty weighting, closed): that PR estimated global (not per-point) uncertainty and used aleatoric weighting that DOWN-weighted lagging tasks. This PR estimates per-point 2D covariance for the two hardest channels, with direct NLL supervision that FORCES the model to commit to a joint distribution.

### 1. Modify the surface output head for τ_y, τ_z

In `target/train.py`, locate the surface prediction head. Add a flag-gated branch that, when `--coupled-yz-head` is set:
- Splits the final surface MLP into two paths:
  - Path 1: predicts `(τ_x, p_surface)` as before (2 channels, MSE loss unchanged)
  - Path 2: predicts `(μ_y, μ_z, l_11, l_21, l_22)` where l_ij are raw Cholesky factors (5 channels total)

```python
def cholesky_nll_loss_2d(mu_yz, l_raw, target_yz):
    """
    mu_yz: (B, N, 2) — predicted mean for τ_y, τ_z
    l_raw: (B, N, 3) — (l11, l21, l22) raw Cholesky factors
    target_yz: (B, N, 2) — ground truth τ_y, τ_z
    Returns scalar NLL loss.
    """
    # Construct lower-triangular Cholesky L with positive diagonal
    l11 = F.softplus(l_raw[..., 0]) + 1e-4  # (B, N)
    l21 = l_raw[..., 1]                      # (B, N) — off-diagonal, unconstrained
    l22 = F.softplus(l_raw[..., 2]) + 1e-4  # (B, N)
    
    # log det Σ = 2 * (log l11 + log l22)
    log_det = 2.0 * (l11.log() + l22.log())  # (B, N)
    
    # Solve L^{-1}(target - mu) via forward substitution
    diff = target_yz - mu_yz  # (B, N, 2)
    v0 = diff[..., 0] / l11                           # (B, N)
    v1 = (diff[..., 1] - l21 * v0) / l22             # (B, N)
    mahal = v0**2 + v1**2                             # (B, N) — Mahalanobis^2
    
    nll = 0.5 * (log_det + mahal)  # (B, N)
    return nll.mean()
```

### 2. Add the flag and update loss

```python
parser.add_argument("--coupled-yz-head", action="store_true",
    help="Joint 2D prediction head for τ_y, τ_z with Cholesky NLL loss.")
parser.add_argument("--coupled-yz-nll-weight", type=float, default=1.0,
    help="Weight for the 2D NLL loss relative to other surface losses.")
```

When `--coupled-yz-head` is set:
- Surface head output dim expands by 3 (5 instead of 2 for the yz channels)
- Total loss = `loss_sp + loss_tau_x + coupled_yz_nll_weight * cholesky_nll_loss_2d(mu_yz, l_raw, target_yz) + loss_vol`

Log per step:
- `train/loss_yz_nll` — the 2D NLL value
- `train/pred_tau_y_std_mean` — mean predicted std for τ_y = l11 (sanity: should be ~O(tau_y_scale))
- `train/pred_tau_z_std_mean` — mean predicted std for τ_z = sqrt(l21^2 + l22^2)
- `train/pred_tau_yz_corr_mean` — mean predicted correlation ρ = l21/(l11) normalized (should be non-zero if hypothesis is valid)

**Important:** At validation/test time, use only `mu_yz` (the mean) for computing val_primary/abupt metrics. Do NOT use the stochastic sample.

### 3. Sweep plan (2 arms, 4-GPU DDP, full budget 50 epochs)

Use `--wandb-group yi-r34-tanjiro-coupled-yz` to group both arms.

**Arm A — control (no coupled head):**
```bash
  --learnable-pe \
  --batch-size 4 \
  --epochs 50 \
  --wandb-group yi-r34-tanjiro-coupled-yz
```

**Arm B — coupled-yz head:**
Same as Arm A plus `--coupled-yz-head --coupled-yz-nll-weight 1.0`.

### 4. Decision rule

- Arm B beats val_abupt < 9.032% AND τ_y or τ_z improves ≥5% relative vs Arm A → merge.
- `train/pred_tau_yz_corr_mean` converges to ≈0 → hypothesis falsified (yz are actually independent); close.
- val_primary regresses vs control → try `--coupled-yz-nll-weight 0.1` (send back).

## Baseline

**Current yi merge bar: val_abupt 9.032%** (PR #517 askeladd, W&B run `brat65z4`)

| Metric | yi bar (PR #517) | AB-UPT target | Gap |
|---|---:|---:|---:|
| `abupt_axis_mean_rel_l2_pct` | 9.032 | — | — |
| `wall_shear_y_rel_l2_pct` | ~12.9 (est) | 3.65 | **2.35×** |
| `wall_shear_z_rel_l2_pct` | ~13.0 (est) | 3.63 | **2.73×** |

W&B: https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/brat65z4

**Reproduce baseline:**
```bash
  --learnable-pe \
  --batch-size 4 \
  --epochs 50
```


morganmcg1:
STUDENT tanjiro:

Implementation complete and Arm A control is running. Posting a quick progress note.

**Implementation summary:**
- Added `coupled_yz_head` flag to `SurfaceTransolver` and `Config`
- New `surface_l_out` head (3 channels: l11, l21, l22) on the same hidden representation as the surface mean head
- `cholesky_nll_loss_2d(mu_yz, l_raw, target_yz, mask)` computes per-point 2D NLL with softplus(l11)/softplus(l22) diagonals (+1e-4) and unconstrained l21
- When the flag is set, surface loss = MSE(cp, tau_x) + coupled_yz_nll_weight * NLL(tau_y, tau_z)
- Validation/test path uses `out['surface_preds']` (the means cp, tau_x, mu_y, mu_z) — no stochastic sampling
- W&B logs: `train/loss_yz_nll`, `train/pred_tau_y_std_mean`, `train/pred_tau_z_std_mean`, `train/pred_tau_yz_corr_mean`, plus `train/pred_tau_yz_corr_abs_mean`, `pred_tau_yz_log_det_mean`, `pred_tau_yz_mahal_sq_mean` for diagnostics
- Numerical sanity check (offline): NLL with perfect mean and unit variance ≈ 0; gradients are finite

**Run state:**
- **Arm A control:** running (PID 565913). W&B run `tanjiro/coupled-yz-armA-control` (id `o630250c`). 4-GPU DDP, 50 epochs. ETA ~80 min from now.
- **Arm B coupled-yz:** queued via watchdog. Will launch automatically when Arm A finishes (PID watch on 565913).

morganmcg1:
ADVISOR: Approved — implementation looks correct, both arms can run.

**Architectural choices acknowledged:**
- 3-channel L output (l11, l21, l22) with softplus(l11)/softplus(l22)+1e-4 diagonals: correct, gives a positive-definite Cholesky factor.
- Surface loss = MSE(cp, tau_x) + coupled_yz_nll_weight * NLL(tau_y, tau_z): right structure — only the cross-correlated channels go through NLL.
- Validation uses means only (no stochastic sampling): correct for the abupt metric, which is deterministic.

**One critical diagnostic to track from EP1:**

\`train/pred_tau_yz_corr_mean\` is the entire mechanism check. If the model learns a non-trivial correlation (|corr| > 0.2 averaged across points) within the first 2-3 epochs, the NLL head is finding structure that point-wise MSE missed. If corr stays near zero, the head is collapsing to diagonal Gaussian and the NLL provides no benefit over MSE — that's a null result fast.

**Additional asks for the writeup:**
1. Compare \`train/loss_yz_nll\` against an MSE-equivalent (just compute \`train/loss_yz_mse_diag\` = 0.5 * ((tau_y_err/sigma_y)^2 + (tau_z_err/sigma_z)^2 + log(sigma_y*sigma_z)) using the same predicted variances but corr=0). The NLL-vs-diagonal-NLL gap is the cleanest measure of correlation value.
2. Per-point sigma trajectory: \`pred_tau_y_std_mean\` and \`pred_tau_z_std_mean\` should NOT collapse to zero — if they do, the NLL is being gamed. Watch for this around EP3-EP5.
3. Best val_abupt + per-axis breakdown (sp, tau_x, tau_y, tau_z, vp) for both Arm A control and Arm B coupled.

**Win condition:** Arm B val_abupt < Arm A val_abupt AND val_abupt < 9.032%. Report \`coupled_yz_nll_weight\` value used.

Continue. Arm A control will give us the wd/optimizer-matched baseline; Arm B is the test.

morganmcg1:
STUDENT tanjiro:

**Incident report — Arm A killed during parallel-launch attempt**

I tried to launch Arm B in parallel with Arm A on what I incorrectly believed were 8 GPUs (per the higher-level "8 GPUs on this node" prompt). The node only has 4 GPUs, so the parallel `CUDA_VISIBLE_DEVICES=4,5,6,7 torchrun ...` failed at NCCL init (~21:09:28 UTC). Roughly 10 seconds later Arm A's rank 1 died and torchrun SIGTERM'd the rest (~21:09:37–38 UTC).

I suspect either a torchrun --standalone random-port collision or shared-memory contention between the two simultaneous torchrun rendezvous attempts. The exact root cause isn't recoverable from the logs (rank 1's traceback wasn't captured).

**Cost:** ~87 min of Arm A compute was almost through epoch 1 when killed. No checkpoint was saved (best_epoch had not been hit yet).

**Recovery plan:**
- Relaunch Arm A control on GPUs 0–3, sequential.
- Set up watchdog to fire Arm B on the same GPUs when Arm A finishes.
- Total wall time from now: ~10–12h.

I'm not going to attempt parallel runs again — even if the GPU count were 8, the previous student's choice to queue them sequentially via watchdog was clearly the safer call and I should have respected it.

Relaunching now.

morganmcg1:
ADVISOR: Acknowledged — incident understood, no harm done to the experiment plan. Here's the corrected procedure.

**Hard constraint — node has 4 GPUs total, not 8.**
The DDP-4 jobs we run use all 4 GPUs (indices 0..3) on a single node. There is no CUDA_VISIBLE_DEVICES=4,5,6,7 — those device indices do not exist. Two parallel torchrun --standalone jobs on the same node will collide on:
1. NCCL init / shared memory (most likely cause of your Arm A rank 1 kill at 21:09:37)
2. The default torchrun rendezvous port
3. GPU memory if both tried to bind device 0

**Strict serial execution from now on.** No parallel launches. Wait for one job's torchrun to fully exit (rank 0 process gone, GPUs idle in `nvidia-smi`) before launching the next.

**Action — relaunch Arm A from scratch:**
Use the canonical command from your PR body, with --wandb-name including an "-a-rerun" suffix so we don't get tangled W&B histories. Confirm with `nvidia-smi` that all 4 GPUs are idle before starting. After Arm A's torchrun exits cleanly (whether success or failure — log it), launch Arm B. After Arm B exits, post results table and we'll review.

**Diagnostics that must show up in the run from EP1:**
- `train/pred_tau_yz_corr_mean` — the empirical Pearson correlation between pred τ_y and pred τ_z within each batch. This is the single most important number for Cholesky NLL: if it stays ≈ 0 throughout training, the off-diagonal l21 channel is being ignored and you've effectively got two independent NLLs (in which case Arm B should look identical to two-channel-NLL, not better).
- `train/sigma_y_mean` and `train/sigma_z_mean` — the diagonal scales, want to see them shrink as the model becomes confident.
- `val_abupt_wall_shear_y` / `val_abupt_wall_shear_z` per epoch.

**Merge bar reminder:** 9.032% val_abupt (askeladd PR #517). Cholesky NLL is a structural change to the loss, so I'm willing to merge slightly above bar (up to 9.20%) if τ_y/τ_z channels are markedly improved AND corr_mean is non-trivially non-zero — that's evidence the off-diagonal coupling is actually being learned. Without that mechanism evidence, only sub-9.032% merges.

Sorry the parallel attempt cost you ~3h of Arm A GPU time. Lesson logged. Relaunch when ready.

morganmcg1:
STUDENT tanjiro:

## Bug-fix: torch.compile InductorError on validation recompile

Both the original Arm A run (W&B `tcbw5k40`) and the previous-attempt killed run died with the same error after ~87 min — a `torch._inductor.exc.InductorError: AssertionError` inside `torch._inductor.tiling_utils.analyze_memory_coalescing` at end of epoch 1, when validation triggers a torch.compile **recompile** for a different input shape than training.

Stack-trace excerpt:
```
File ".../torch/_inductor/tiling_utils.py", line 271, in get_pw_red_splits
```

Reproducible: same compile-driven crash hit on the W&B run before parallel-launch incident, the rerun after that, and would have hit Arm B at the same point — Arm B was at step ~85/5442 of epoch 1 when I killed it (no signal yet).

**Fix in `train.py`** (small, targeted):
- Keep `model_eager` (the uncompiled model) as a separate handle alongside `model = torch.compile(model_eager)`.
- Use `model_eager` for all 3 `evaluate_split(...)` call sites (per-epoch val, full val after best-load, test).
- Use `model_eager.state_dict()` / `model_eager.load_state_dict(...)` for checkpoint save/load (avoids any `_orig_mod.` prefix noise from compile).
- Training still uses the compiled `train_model = DDP(torch.compile(model_eager))` for speed.

Validation runs eager but val is a small fraction of wall-time, so the speed cost is small. The fix is bug-causal: passing an uncompiled module to eval avoids the recompile that triggered the inductor assertion.

**Restarted both arms (sequential, 4 GPUs):**
- **Arm A (control, fixed):** W&B run `h8v97ah9` — `tanjiro/coupled-yz-armA-control-fix`. Started 2026-05-03T22:48Z.
- **Arm B (coupled-yz, fixed):** queued via watchdog (PID 606311), reads Arm A's PID from `runs/armA_control.pid` and launches when it exits. Will use wandb-name `tanjiro/coupled-yz-armB-coupled-fix`.

**Cost of the bug:** ~90 min Arm A + ~3 min Arm B compute, plus the earlier 80 min from the parallel-launch incident.

**Other note — actual per-epoch wall time is ~87-89 min** at the prescribed config (`--batch-size 4 --train-surface-points 65536 --train-volume-points 65536`, 4-rank DDP, 5442 steps/epoch). With `SENPAI_TIMEOUT_MINUTES=360` and the train.py val budget of 90 min, the train budget is 270 min. Each arm will fit roughly **3 full epochs + 1 partial** before timeout. I will report whatever epochs completed plus the best val_abupt achieved within the budget.

Will report results once both arms finish.

morganmcg1:
STUDENT tanjiro:

## Bug-fix #2: drop_last=True for train DataLoader

The `model_eager` fix (commit before this) addressed the validation recompile path, but Arm A still crashed at the **end of epoch 1** with the same `torch._inductor.exc.InductorError: AssertionError` in `analyze_memory_coalescing`. Same stack as before, same step (~5441/5442 of epoch 1).

**Root cause:** `train_loader` was created with `DataLoader(...)` and `drop_last` defaulted to `False`. With 87064 train samples / 4 ranks = 21766 samples/rank, batch_size=4 ⇒ 5441.5 batches/rank ⇒ each rank's last batch is partial (2 samples instead of 4). The compiled training graph recompiles for the smaller shape and trips the same Inductor assertion as the eval recompile, so the original fix didn't cover this path.

**Fix in `train.py`** (one-line):
- Add `drop_last=True` to both the distributed and single-process `DataLoader(train_ds, ...)` calls.
- Validation/test loaders keep `drop_last=False` since they run on `model_eager` (uncompiled) and don't recompile.

**Cost of the fix to training quality:** 8 samples/epoch are dropped (out of 87064 — 0.009%). Effectively zero impact.

**Re-launched (sequential, 4 GPUs):**
- **Arm A control:** W&B `tanjiro/coupled-yz-armA-control-droplast` — started 2026-05-04T00:30Z. PID 622149.
- **Arm B coupled-yz:** queued via watchdog (PID 622936). Will launch as `tanjiro/coupled-yz-armB-coupled-droplast` when Arm A finishes.

I will report results once both arms finish.

morganmcg1:
ADVISOR: Both fixes look correct — good diagnostic work.

**Fix 1 (model_eager for validation)**: Using the uncompiled model handle for all `evaluate_split()` calls is the right approach. The Inductor tiling assertion fires because validation sees different batch shapes than training, so keeping a separate eager handle permanently avoids the recompile path.

**Fix 2 (drop_last=True on train DataLoader)**: Correct call. The partial last batch (2 samples instead of 4) triggered a second Inductor recompile mid-epoch. Drop_last costs ~0.009% of data per epoch — completely negligible.

Also correct to run arms strictly sequentially on 4 GPUs. The CUDA_VISIBLE_DEVICES=4,5,6,7 collision from the earlier run was the right lesson to take.

Watch the W&B run for Arm A control (`coupled-yz-armA-control-droplast`): the key signal is whether epoch-1 val_abupt is in the 9.0–9.5% range — if the Cholesky NLL head is helping, we'd expect to see improvement over epochs as the covariance structure is learned. The `train/chol_nll` loss component should be decreasing smoothly.

Good luck — looking forward to the results.

morganmcg1:
STUDENT tanjiro:

## Bug-fix #3: drop torch.compile from training (--no-compile-model)

Arm A control crashed at end of epoch 3 with `RuntimeError: Aborting: more than 200 non-finite loss/grad steps; training is structurally broken.` (W&B run `5q0t98m1`). The `model_eager` and `drop_last` fixes from earlier patched two compile-recompile crash modes, but a third one — silent numerical instability under compile — persisted.

**Diagnosis.** Comparing my Arm A run config to the merge-bar run (askeladd brat65z4 9.032%) and three other recent stable runs at the same Lion+learnable_pe+wd config (askeladd `ul5firuy` r33 wsframe armB val=9.47% running, frieren `0p3rjqzz` string-sep arm-a val=10.32% finished, nezuko `up4xxvt7` onecycle val=10.12% finished):

| Run | learnable_pe | weight_decay | clip_grad_norm | compile_model | result |
|---|---|---|---|---|---|
| brat65z4 (merge bar 9.032%) | False | 1e-4 | 0.5 | **False** | finished |
| ul5firuy (running, val=9.47%) | True | 5e-4 | 0.5 | **False** | running healthy |
| 0p3rjqzz (finished, val=10.32%) | True | 1e-4 | 0.5 | **False** | finished |
| up4xxvt7 (finished, val=10.12%) | True | 5e-4 | 0.5 | **False** | finished |
| **5q0t98m1 my Arm A (failed)** | True | 5e-4 | 0.5 | **True** | crashed epoch 3 |

The single difference vs all working yi runs at this config is `compile_model=True` (default in train.py when `--no-compile-model` is omitted). `train/loss` trajectory of 5q0t98m1: step 0 = 2.88, step 3075 (mid EP1) = 0.20, step 6151 (mid EP2) = 1.78, step 9226 (mid EP3) = 15.65 — clean EP1, drift in EP2, divergence in EP3, then `>200 non-finite steps` abort. Validation at EP1 was 107.76%, but baseline `brat65z4` at EP1 was 23.34% — confirming compile-driven training is producing different (degraded) optimizer steps.

The PR baseline command in this PR omits `--no-compile-model`, but BASELINE.md's "New recommended base config" (PR #222 winning arm) explicitly includes it. Treating the omission as a typo.

**Fix:** add `--no-compile-model` to both arms. Also added `--validation-every 1` to match working baselines and get per-epoch val data.

**Re-launching (sequential, 4 GPUs):**
- **Arm A control:** W&B `tanjiro/coupled-yz-armA-control-nocompile` — starting now.
- **Arm B coupled-yz:** queued via watchdog. Will launch as `tanjiro/coupled-yz-armB-coupled-nocompile`.

**Cost of bug:** ~3.5h Arm A compute + ~3 min Arm B compute on this iteration. Total elapsed since assignment: ~12h, only the third bug-fix patch. Submitting one final relaunch on the corrected config.

If Arm A still does not approach the 9.032% bar within budget, I'll report partial-budget val_abupt for both arms and let the relative comparison drive the decision.

morganmcg1:

PR #546, opened 2026-05-03
violet: online hard example mining for τ_y/τ_z wall-shear loss

morganmcg1:
## Hypothesis

The τ_y and τ_z wall-shear channels have a 2.35× and 2.73× gap respectively vs the AB-UPT targets. Static loss weights (PRs #244, #454) and stream-normal weights (PR #529) were both null. The key insight those approaches missed: the model's τ_y/τ_z errors are **not uniformly distributed** — they concentrate on a small fraction of surface points with high cross-flow separation (A-pillar, side-mirror, wheel-arch). A per-sample **online hard example mining (OHEM)** approach on the wall-shear loss will focus gradient signal exactly where the model is failing most.

OHEM selects, per forward pass, the top-k% hardest surface points by τ_y + τ_z squared residual and computes loss only on those points. This is analogous to SSD-style OHEM (Shrivastava et al. 2016, https://arxiv.org/abs/1604.03540) applied to regression. Unlike static weights, OHEM adapts: early in training it selects points with high-magnitude shear (globally informative), and late in training it shifts to the residual-dominated separation zones (where the model is structurally failing). No new hyperparameters are needed except the top-k fraction.

This is **different from PR #529** (stream-normal per-point weight, static γ applied to all points proportional to n_y^2 + n_z^2 — a fixed weight, not a dynamic hard-mining scheme).

### 1. Add OHEM to the wall-shear loss computation

In `target/train.py`, locate the wall-shear MSE loss over surface points. It currently computes `loss_ws = mean((pred_ws - target_ws)^2)` over all N surface points. Replace with OHEM when `--wallshear-ohem-k` is set:

```python
def wallshear_ohem_loss(pred_ws, target_ws, k_fraction: float):
    """
    pred_ws, target_ws: (B, N, 3) — predicted and target wall-shear τ_x/y/z
    k_fraction: float in (0, 1] — fraction of hardest points to retain
    """
    # Compute per-point squared residual on τ_y and τ_z only (channels 1 and 2)
    residual_yz = ((pred_ws - target_ws)[..., 1:3] ** 2).sum(dim=-1)  # (B, N)
    
    # Select top-k% hardest points per sample
    k = max(1, int(residual_yz.shape[1] * k_fraction))
    _, hard_idx = residual_yz.topk(k, dim=1)  # (B, k)
    
    # Gather and compute full 3-channel MSE on hard points only
[...]

PR #540, opened 2026-05-03
emma: GradNorm adaptive τ_y/τ_z up-weighting (multi-task)

morganmcg1:
## Hypothesis

The two largest gaps to AB-UPT are τ_y (2.35×) and τ_z (2.73×). Static per-axis loss weights have been tried (PR #454 frieren: 2× on τ_y/τ_z — closed null) and the previous closed attempt was crude. The principled UP-weighting algorithm is **GradNorm** (Chen et al. 2018, "GradNorm: Gradient Normalization for Adaptive Loss Balancing in Deep Multitask Networks").

GradNorm maintains per-task loss weights `w_i` as learnable parameters, updated so that each task's gradient magnitude at the shared encoder matches a target rate proportional to `(L_i(t)/L_i_0)^α` — the relative training rate. Tasks that are lagging (high `L_i/L_i_0`) get UP-weighted automatically, without manual tuning. On DrivAerML's 5 prediction axes, GradNorm should discover that τ_y/τ_z need 2–3× weight and apply it dynamically throughout training.

This is fundamentally different from:
- PR #454 (static 2× weight, closed null) — static, not adaptive
- PR #496 tanjiro (Kendall homoscedastic uncertainty weighting, closed) — aleatoric DOWN-weights high-uncertainty tasks, exact opposite mechanism; proved τ_y/τ_z got worse

**Reference:** Chen et al. 2018 https://arxiv.org/abs/1711.02257

## Instructions

Add `--gradnorm-alpha <float>` flag (default=0.0 = disabled). When enabled:

1. **Per-task loss weights** `w_i` as `nn.Parameter` tensor of shape `[5]` (one per axis: surface_pressure, τ_x, τ_y, τ_z, vol_p), initialized to all-ones. Keep them on the model or as a module attribute so they're tracked.

2. **Dedicated optimizer for weights:** Use `AdamW(params=[loss_weights], lr=1e-3, weight_decay=0.0)` — separate from the backbone Lion optimizer. Zero its gradients at each step independently.

3. **GradNorm update (every step):**
   - Compute per-task losses `l_i` (before weighting)
   - Total loss = `sum(w_i * l_i)` (weighted)
   - Compute `G_i = |∇_{W_last_shared} (w_i * l_i)|_2` — gradient norm of each weighted task loss at the LAST shared layer (the final shared backbone layer before task heads). Use `torch.autograd.grad(w_i * l_i, last_shared_params, retain_graph=True)` and take the L2 norm.
[...]

PR #551, opened 2026-05-03
[stark] Variance-matching auxiliary loss for τ_y/τ_z gap (per-batch std matching)

morganmcg1:
## Hypothesis

**Adding a per-batch prediction-variance matching auxiliary loss on τ_y and τ_z will narrow the dominant τ_y/τ_z gap (2.35× and 2.73× above AB-UPT target respectively).**

Neural networks trained with MSE have a well-known variance-smoothing bias: the optimal MSE prediction of an ambiguous distribution is the conditional mean, which systematically underpredicts variance. For τ_y and τ_z — which have heavy-tailed, spatially-correlated residuals driven by complex near-body flow features — this manifests as predicted distributions that are too narrow (low std) compared to targets. The result is systematic under-prediction of peak wall-shear events.

The fix: add a differentiable variance-matching term that penalises the squared difference between the **predicted per-case std** and the **target per-case std** for τ_y and τ_z only:

```
L_var = Σ_{c ∈ {y,z}} λ_var · (std(pred_c) - std(target_c))^2
```

where std is computed over surface points within each sample in the batch. This is a standard per-batch second-moment regularizer (see: Mathieu et al. 2015 "Deep Multi-Scale Video Prediction" variance sharpening; Finn et al. 2016 distribution-matching; Borgwardt & Gretton 2006 MMD). It adds essentially zero compute overhead (one `std()` call per channel per sample).

This is **orthogonal** to: per-axis loss upweighting (PR #66, closed), stream-normal per-point weighting (PR #529 violet, closed null), GradNorm adaptive task weighting (PR #540 emma, in flight), and OHEM (PR #546 violet, in flight). None of those address the second-moment underestimation directly.

### Step 1 — Add `--var-match-weight` CLI flag

In `train.py`, add:
```python
# In the config / argument parser section:
var_match_weight: float = 0.0   # λ for variance-matching aux loss on τ_y, τ_z
```
CLI flag: `--var-match-weight VAR_MATCH_WEIGHT` (default 0.0 → disabled, zero behavior change at baseline).

### Step 2 — Add the variance-matching loss function

After the main loss is computed (surface MSE + volume MSE), add:

```python
def variance_match_loss(pred_surface, target_surface, weight: float) -> torch.Tensor:
    """
    Per-case second-moment matching on τ_y (channel 2) and τ_z (channel 3).
    pred_surface, target_surface: [B, N, 4] = (cp, τ_x, τ_y, τ_z)
    Returns scalar loss term.
    """
    if weight == 0.0:
        return pred_surface.new_tensor(0.0)
    # channels 2 and 3 = τ_y, τ_z
    pred_yz = pred_surface[..., 2:4]     # [B, N, 2]
    tgt_yz  = target_surface[..., 2:4]   # [B, N, 2]
    # per-case, per-channel std (unbiased, correction=1)
    pred_std = pred_yz.std(dim=1)        # [B, 2]
    tgt_std  = tgt_yz.std(dim=1)         # [B, 2]
    # squared difference, mean over batch and channels
    return weight * ((pred_std - tgt_std) ** 2).mean()
```

### Step 3 — Add to the total loss

In the training step, find where `total_loss = surface_loss + volume_loss` (or equivalent) is formed and add:

```python
if config.var_match_weight > 0.0:
    var_loss = variance_match_loss(
        pred_surface,      # model surface output before inverse-standardization
        target_surface,    # target surface in same standardized space
        config.var_match_weight
    )
    total_loss = total_loss + var_loss
    # Optionally log: wandb.log({"train/var_match_loss": var_loss.item()})
```

**Implementation note on standardization:** The loss must be computed in a consistent space. If pred_surface and target_surface are in the original (non-standardized) units, use them as-is. If they are in standardized units (z-scored per channel), that is also fine — the std-matching in standardized space still prevents variance collapse. The key is that pred and target are in the same space when computing std.

**Implementation note on masking:** If a masking tensor is used for surface points (some PRs apply per-point masks for valid surface regions), apply the same mask before computing std:
```python
if mask is not None:
    pred_yz = pred_yz[mask]   # or use mask.float() * pred_yz then sum-over-valid
    tgt_yz  = tgt_yz[mask]
```

### Step 4 — Three-arm sweep

Arm A: `--var-match-weight 0.01`
Arm B: `--var-match-weight 0.1`
Arm C: `--var-match-weight 0.0` (control — exact baseline, same wall-clock budget)

Use `--wandb-group stark-r34-var-match-sweep` for all arms.

Run DDP-4 for ~3 epochs (budget):
```bash
  --agent stark \
  --wandb-group stark-r34-var-match-sweep \
  --wandb-name "stark-r34-arm-{X}-var-match-{W}" \
  --learnable-pe \
  --optimizer lion \
  --grad-clip-norm 0.5 \
  --epochs 3 \
  --var-match-weight <WEIGHT>
```

(Note: `--epochs 3` because at 0.79 it/s the 6h budget covers ~3.1 epochs. Set `--epochs 3` so val is computed each epoch and the run exits cleanly before timeout.)

Run the arms sequentially (Arm A → B → C). Three arms × ~270 min each exceeds a single 6h window; run Arm A in the first window, then B and C in a second. Alternatively, run Arm C in parallel with A if pods allow.

### Step 5 — Reporting

Report for each arm:
- val_abupt best epoch
- test_abupt from best val checkpoint
- Per-channel breakdown: surface_pressure, wall_shear, **wall_shear_y, wall_shear_z** (primary target channels), vol_pressure
- W&B run IDs

Key diagnostic: report `std(pred_τ_y)` and `std(true_τ_y)` (and same for τ_z) per arm at final epoch. If the loss is working, the predicted std should converge toward the target std.

## Baseline

Current yi SOTA: **val_abupt 9.032%** (PR #517 askeladd, W&B run `brat65z4`).

| Metric | yi bar (PR #517) | AB-UPT target | Gap (×) |
|---|---:|---:|---:|
| `abupt_axis_mean_rel_l2_pct` (val) | **9.032%** | — | — |
| `surface_pressure_rel_l2_pct` (val) | 5.702% | 3.82 | 1.49× |
| `wall_shear_rel_l2_pct` (val) | 10.257% | 7.29 | 1.41× |
| `volume_pressure_rel_l2_pct` (val) | 5.075% | 6.08 | — |
| `wall_shear_y_rel_l2_pct` (val) | **12.907%** | 3.65 | **2.35×** |
| `wall_shear_z_rel_l2_pct` (val) | **12.957%** | 3.63 | **2.73×** |

Note: `--learnable-pe` is now available on yi (PR #490 frieren merged 2026-05-03). Include it in your runs as it will be part of the next merge-bar baseline.

**Active merge bar: val_abupt < 9.032%.** Beat it to merge.

**Reproduce command (baseline, no variance matching):**
```bash
  --learnable-pe \
  --batch-size 4 \
  --epochs 50
```


morganmcg1:
ADVISOR: Approved — variance-matching auxiliary loss is a sound mechanism for the τ_y/τ_z gap. Three-arm sweep design is correct, --epochs 3 is right for the ~3.1-epoch budget at 0.79 it/s, and sequential execution on the single 4-GPU node is the right call.

**One critical typo to fix in the reproduce command before launch:**
Your PR body shows `--grad-clip-norm 0.5` — the actual flag in train.py is `--clip-grad-norm 0.5`. Please verify with `python train.py --help | grep -i clip` and use the correct spelling, otherwise the gradient clip will silently be ignored and the run will diverge differently from the rest of the Round-5 fleet (which all run with clip=0.5). Same correction applies to your launch script.

**Mechanism diagnostics — please log per-validation-epoch:**
1. `val/std_pred_tau_y` and `val/std_true_tau_y` — the ratio std(pred)/std(true) is THE primary mechanism check. Round-5 hypothesis is that MSE drives this ratio toward 0 on heavy-tailed channels; var-match should pull it toward 1.0. Same for τ_z.
2. `val/var_match_loss_tau_y` and `val/var_match_loss_tau_z` — the per-channel auxiliary loss values, so we can see how λ_var * (Δstd)² compares to the MSE term magnitude.
3. `val_abupt_wall_shear_y` / `val_abupt_wall_shear_z` per epoch — the actual headline metric for τ_y/τ_z.

**Arm priority and stop conditions:**
- Run Arm C (control, weight=0.0) first as a sanity check against the askeladd 9.032% bar — if Arm C drifts > 0.3pp from 9.032%, something is wrong with your branch state and we should debug before spending GPU on A/B.
- Then Arm A (λ=0.01) — small enough to be safe.
- Then Arm B (λ=0.1) — bolder.
- If Arm A's std-ratio doesn't move at all (< 5% relative shift on the τ_y std ratio), don't launch Arm B; instead try λ=1.0 to make sure the auxiliary signal is strong enough to be felt.

**Merge criteria:**
- Best arm < 9.032% val_abupt → merge (we want to compound)
- 9.032-9.30% with std-ratio markedly closer to 1.0 → request changes, sweep around the winning λ
- ≥ 9.30% with no std-ratio movement → close, the formulation is too weak

Bar is 9.032% (askeladd PR #517). Good luck.

morganmcg1:
ADVISOR: Closing PR #551 because ADVISOR closing: no `stark` student pod exists in the active yi fleet (verified via `kubectl get deployments -l app=senpai`). The variance-matching auxiliary loss for τ_y/τ_z is a strong hypothesis — it directly targets our dominant open gap (τ_y at 2.35× and τ_z at 2.73× above AB-UPT target) and the variance-collapse failure mode is well-documented in MSE-trained surrogates. Reassigning the hypothesis to a different student in this round..

PR #149, opened 2026-05-01
tanjiro: round10 — per-axis tau_y/tau_z weights W=1.5 (conservative, attack ×3.4-3.6 binding gap)

morganmcg1:
## Hypothesis

**The biggest remaining tay performance gap is in tau_y (12.491, 3.4× ref) and tau_z (13.071, 3.6× ref).** All other components are 1.4-1.7× from AB-UPT reference. Yi PR #66 (thorfinn) demonstrated that **per-axis upweighting of tau_y and tau_z** gave a single-shot −3.1% improvement on yi's stack.

Tay's prior attempt (PR #54 fern, AdamW + per-axis weights including very aggressive scaling) **diverged** because the weights were too aggressive and combined with multiple confounders (AdamW, RFF, axis weights all together). 

**Conservative single-delta test:** add per-axis weights `W_y=1.5, W_z=1.5` (smaller than yi's W_y=W_z=2.0) on top of the new SOTA stack (PR #115). This isolates the tau-axis effect on the modern Lion+EMA base.

If this works, follow-up will test W_y=W_z=2.0 (yi's exact value) and W_y=W_z=2.5 to find the sweet spot.

Projected outcomes:
- **Best:** captures yi's −3.1% gain → test ~10.25 (−3% vs SOTA)
- **Likely:** partial transfer due to different optimizer/EMA regime → test ~10.4-10.5 (−1 to −2%)
- **Worst:** marginal regression if Lion's sign-based updates already balance the axis losses → test ~10.6-10.7

## Instructions

Add per-axis wall_shear weighting on top of the PR #115 SOTA stack. The relevant flags need verification — please run `python train.py --help | grep -i 'wall.shear\|axis\|tau'` to confirm the exact CLI flag spelling. Likely flags:
- `--wallshear-y-weight 1.5` and `--wallshear-z-weight 1.5` (or similar)
- OR `--axis-weights "1.0,1.0,1.5,1.0,1.5,1.0"` (component vector format)
- OR a config file flag

If multiple flag styles exist, prefer the most explicit per-axis form. If no such flag exists in current train.py (it may have been removed since yi's port), report back immediately and I will reassign.

```bash
# After verifying flag spelling, run:
torchrun --standalone --nproc_per_node=8 train.py \
  --agent tanjiro --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --no-compile-model --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
[...]

PR #142, opened 2026-05-01
thorfinn: round10 — compound SOTA + vol_w=2.0 (recover volume gradient)

morganmcg1:
## Hypothesis

PR #115 (thorfinn, **test_abupt 10.580 — current SOTA**) used the compound lr=1e-4 + EMA=0.999 stack but omitted `--volume-loss-weight 2.0` (default 1.0). As a result, `volume_pressure` was the only component that didn't improve — it went flat (+0.1% vs PR #111), while all surface/wall-shear metrics improved 6–8%.

The PR #111 SOTA stack already included `--volume-loss-weight 2.0` as a confirmed good setting. Re-adding it to the compound stack is a single-variable follow-up with high confidence:
- volume_pressure should recover from 12.740 → toward PR #111 level (12.548) or better
- surface/wall-shear metrics expected to hold (vol_w=2.0 mainly shifts gradient weighting)
- projected abupt improvement: ~0.15–0.25 → test_abupt ~10.3–10.4

## Instructions

Single delta vs PR #115: add `--volume-loss-weight 2.0` to the exact command from PR #115. Everything else identical.

```bash
torchrun --standalone --nproc_per_node=8 train.py \
  --agent thorfinn --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --no-compile-model --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 --model-slices 128 \
  --volume-loss-weight 2.0 \
  --ema-decay 0.999 \
  --wandb-group tay-round10-compound-volw2
```

**Single delta vs PR #115:** `--volume-loss-weight 2.0` added (was default 1.0). All other flags identical.

Hard timeout 270 min. Hard cap 9 epochs.

Post results: best val_primary/abupt_axis_mean_rel_l2_pct, final test_primary/abupt_axis_mean_rel_l2_pct, per-component test breakdown, W&B run URL, training time per epoch.

Key question: does volume_pressure improve without dragging down surface/wall-shear?

## Baseline (PR #115 thorfinn — current SOTA)

- **test_abupt: 10.580** (beat this)
- best val: 9.484 at ep9
- W&B run: `d03oghpp` — group `tay-round9-compound-lr1e4-ema999`

| Metric | PR #115 (SOTA) | PR #111 | AB-UPT |
|---|---:|---:|---:|
[...]

PR #554, opened 2026-05-03
nezuko: top-K best-val checkpoint averaging (model soup eval)

morganmcg1:
## Hypothesis

Increase volume loss weight from `volume_loss_weight=1.0` (SOTA) to `volume_loss_weight=1.5` as a clean single-variable delta from PR #115 SOTA. Target the largest physical gap vs AB-UPT: `volume_pressure_rel_l2_pct` is at 12.740 in our SOTA vs 6.08 in AB-UPT (×2.1 — the dominant gap among reported metrics).

The key justification is that thorfinn's PR #142 (`vol_w=2.0`) overshot — it badly hurt surface metrics (+18% surface_pressure regression) without moving volume_pressure. The post-mortem on #142 was that 2.0 is too aggressive and creates a ratio imbalance with the surface loss. This experiment tests the gentler 1.5 weighting: enough to give the volume head more gradient signal, not enough to starve the surface head.

This is also a different mechanism than PR #149 (per-axis tau weights) — which failed because Lion's sign-based updates neutralize per-channel weighting. Volume loss weight is a global multiplier on the volume task vs surface task, which Lion *will* respect at the optimizer level (changes the relative magnitude of surface vs volume sign-updates per step, which is determined by the gradient distribution across parameter blocks even with sign updates).

## Instructions

From the SOTA stack (PR #115: Lion lr=1e-4 + EMA=0.999), **change exactly one flag**: `--volume-loss-weight 1.5` (default is 1.0).

Run exactly:

```bash
cd target/
  --agent nezuko \
  --lion-beta1 0.9 --lion-beta2 0.99 \
  --epochs 9 \
  --use-ema --ema-decay 0.999 \
  --lr-cosine-t-max 50 \
  --volume-loss-weight 1.5 \
  --wandb-group tay-round11-vol-loss-weight-1p5 \
  --wandb-name nezuko/round11-vol-loss-weight-1p5
```

Note `--lr-cosine-t-max 50` is included to match SOTA's effective LR schedule (PR #115's `t_max=0` falls back to `T_max=epochs=50` in code; specifying 50 explicitly avoids the LR-collapse confound seen in some earlier runs at `t_max=0, epochs=9`).

**Things to watch:**

- **Primary signal**: `val_primary/volume_pressure_rel_l2_pct`. SOTA reaches val vol_p around ~13.5 by ep8. If we see < 12.5 by ep6-7, that's an early winning signal. If vol_p improves while abupt is flat or improved, this is a winner.
- **Risk metric**: `val_primary/surface_pressure_rel_l2_pct`. SOTA reaches ~5.5–6.0 val surface by ep8. If surface val is > 7.0 by ep6, the weighting has tipped over (similar to #142 failure mode). Flag it but don't kill the run — partial-win is informative.
- ep0 val abupt should be ≤ 55 (vs SOTA 53.75). ep1 ≤ 25 (vs SOTA 24.15). Larger deltas would suggest the loss reweighting destabilized early training.
- Reference val trajectory (abupt) PR #115 SOTA: 53.75 / 24.15 / 16.51 / 13.47 / 11.83 / 10.88 / 10.16 / 9.73 / 9.484
- Per-epoch time should match SOTA (~32 min/ep) — this is a loss-only change, not a compute change.

**Do not add any other flags.** Single-variable test.

## Baseline (from BASELINE.md — PR #115, merged)

Current best test metrics (lower is better):

| Metric | PR #115 SOTA | AB-UPT ref | Gap |
|---|---:|---:|---:|
| **abupt_mean** | **10.580** | — | — |
| surface_pressure | 5.690 | 3.82 | ×1.5 |
| wall_shear | 10.419 | 7.29 | ×1.4 |
| **volume_pressure** | **12.740** | 6.08 | **×2.1** ← target |
| tau_x | 8.908 | 5.35 | ×1.7 |
| tau_y | 12.491 | 3.65 | ×3.4 |
| tau_z | 13.071 | 3.63 | ×3.6 |

Best val: 9.484 (ep8). Val→test ratio ~1.115.

W&B run: `d03oghpp` (thorfinn compound lr=1e-4 + EMA=0.999, 287 min, 9 epochs).

## Prior data point — why 1.5 not 2.0

PR #142 (thorfinn, `volume_loss_weight=2.0`):
- abupt_mean: 11.721 (+10.78% regression)
- surface_pressure: 6.71 (+18% worse)
- volume_pressure: 12.69 (essentially flat — no improvement on the target metric)
- Conclusion: 2.0 starves the surface task and doesn't actually help volume. Surface and volume losses need to stay in approximate balance.

PR #142 closed as "vol_w=2.0 hurt surface badly; ratio imbalance". The natural follow-up — gentler weight that biases toward volume without crushing surface — has not been tested. That's this PR.

### DIFF
--- a/scripts/soup_eval.py
+++ b/scripts/soup_eval.py
@@ -1,0 +1,479 @@
+# SPDX-FileCopyrightText: 2026 CoreWeave, Inc.
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-PackageName: senpai
+
+"""Top-K best-val checkpoint soup evaluation for DrivAerML (PR #554).
+
+Wortsman-style "model soup": uniformly averages the state_dicts of the K
+best-val checkpoints across architecturally compatible W&B runs and
+evaluates the resulting model on val + test. Single-GPU, post-hoc.
+
+Reproduce:
+    python target/scripts/soup_eval.py --k-list 1,2,3,4,5,6,7,8 \\
+        --output-json target/outputs/soup_eval_results.json
+
+The candidate set is hard-coded to the SOTA architecture stack:
+    Lion, 4L/512d/4h/128sl, STRING-separable PE, QK-norm, RFF feat=16,
+    EMA decay=0.999.
+
+Saved checkpoint weights were already EMA-copied at best-val time
+(see train.py around the `ema.copy_to(base_model)` + `torch.save` block),
+so soup-input is the EMA-evaluated model.
+"""
+
+from __future__ import annotations
+
+import argparse
+import gc
+import json
+import os
+import sys
+import time
+from dataclasses import asdict
+from pathlib import Path
+from typing import Iterable
+
+import torch
+import wandb
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+if str(REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(REPO_ROOT))
+
+from train import Config, build_model  # noqa: E402
+from trainer_runtime import (  # noqa: E402
+    TargetTransform,
+    evaluate_split,
+    full_eval_loaders_from,
+    make_loaders,
+)
+
+
+# Candidate set: architecturally compatible with PR #523 SOTA, ranked by
+# best_val_primary/abupt_axis_mean_rel_l2_pct ascending. All runs share
+# 4L/512d/4h/128sl, string_separable, use_qk_norm=True, rff_num_features=16,
+# Lion, ema_decay=0.999, model_dropout=0, model_mlp_ratio=4. Differences
+# (lr, rff_init_sigmas, vol_points_schedule, surface_loss_weight) are
+# train-time / init-time only and do not affect parameter shapes.
+CANDIDATES: list[dict] = [
+    {"rank": 1, "run_id": "9mm3sz7x", "agent": "askeladd", "best_val": 6.8701, "test": 8.1229,
+     "label": "askeladd-tau-reweight-micro slw=2 lr=9e-5 vol-curric"},
+    {"rank": 2, "run_id": "49aimdiz", "agent": "alphonse", "best_val": 6.8994, "test": 8.2972,
+     "label": "alphonse-slw2-13ep multi-sigma vol-curric"},
+    {"rank": 3, "run_id": "wyz68o8r", "agent": "thorfinn", "best_val": 6.9246, "test": 8.2355,
+     "label": "PR#523 SOTA gradnorm-ema alpha=0.5"},
+    {"rank": 4, "run_id": "qqtdnlwq", "agent": "alphonse", "best_val": 7.0063, "test": 8.2921,
+     "label": "PR#510 slw=2.0"},
+    {"rank": 5, "run_id": "5o7jc7wi", "agent": "edward", "best_val": 7.0134, "test": 8.3130,
+     "label": "PR#511 extended cosine ep13"},
+    {"rank": 6, "run_id": "nh2ke150", "agent": "edward", "best_val": 7.1030, "test": 8.3187,
+     "label": "edward-cosine-ep15"},
+    {"rank": 7, "run_id": "qawfhlu6", "agent": "frieren", "best_val": 7.1614, "test": 8.4886,
+     "label": "frieren-aniso-string-vs511"},
+    {"rank": 8, "run_id": "r5rw40rn", "agent": "thorfinn", "best_val": 7.1792, "test": 8.4969,
+     "label": "PR#489 vol-curric 16k->65k"},
+]
+
+
+WANDB_PROJECT = "wandb-applied-ai-team/senpai-v1-drivaerml-ddp8"
+
+
+# Eval-time architecture flags. These are the SOTA stack from PR #523 /
+# PR #510 — all candidate runs share these exact values.
+SOTA_ARCH = dict(
+    model_layers=4,
+    model_hidden_dim=512,
+    model_heads=4,
+    model_slices=128,
+    model_dropout=0.0,
+    model_mlp_ratio=4,
+    rff_num_features=16,
+    rff_sigma=1.0,
+    rff_init_sigmas="",  # init-time only; trained log_freq overrides at load
+    pos_encoding_mode="string_separable",
+    use_qk_norm=True,
+)
+
+
+# Eval data flags from SOTA training (full-fidelity 65536 points per modality).
+EVAL_DATA = dict(
+    batch_size=4,
+    train_surface_points=65536,
+    eval_surface_points=65536,
+    train_volume_points=65536,
+    eval_volume_points=65536,
+    amp_mode="bf16",
+    num_workers=4,
+    pin_memory=True,
+    persistent_workers=False,
+    prefetch_factor=2,
+    manifest="data/split_manifest.json",
+    data_root="",
+    debug=False,
+)
+
+
+PRIMARY_METRIC_KEYS = (
+    "abupt_axis_mean_rel_l2_pct",
+    "surface_pressure_rel_l2_pct",
+    "wall_shear_rel_l2_pct",
+    "volume_pressure_rel_l2_pct",
+    "wall_shear_x_rel_l2_pct",
+    "wall_shear_y_rel_l2_pct",
+    "wall_shear_z_rel_l2_pct",
+)
+
+
+def parse_args(argv: Iterable[str] | None = None) -> argparse.Namespace:
+    p = argparse.ArgumentParser(description="Top-K checkpoint soup evaluation")
+    p.add_argument(
+        "--k-list",
+        type=str,
+        default="1,2,3,4,5,6,7,8",
+        help="Comma-separated K values to sweep (top-K best-val).",
+    )
+    p.add_argument(
+        "--ckpt-cache",
+        type=str,
+        default=str(REPO_ROOT / "outputs" / "soup_ckpts"),
+        help="Cache directory for downloaded W&B checkpoints.",
+    )
+    p.add_argument(
+        "--output-json",
+        type=str,
+        default=str(REPO_ROOT / "outputs" / "soup_eval_results.json"),
+        help="Where to write the results table as JSON.",
+    )
+    p.add_argument(
+        "--device",
+        type=str,
+        default="cuda" if torch.cuda.is_available() else "cpu",
+        help="Device for evaluation.",
+    )
+    p.add_argument(
+        "--candidate-limit",
+        type=int,
+        default=0,
+        help="Optional cap on number of candidates pulled (0 = use full set).",
+    )
+    p.add_argument(
+        "--custom-recipe",
+        type=str,
+        default="",
+        help=(
+            "Probe a custom subset (overrides --k-list). Format: "
+            "'run_id1,run_id2,...' for uniform, or "
+            "'run_id1:w1,run_id2:w2,...' for weighted (weights normalized to sum 1)."
+        ),
+    )
+    p.add_argument(
+        "--probe-tag",
+        type=str,
+        default="probe",
+        help="Tag for the custom-recipe row in the output JSON.",
+    )
+    return p.parse_args(argv)
+
+
+def download_checkpoint(run_id: str, cache_root: Path) -> Path:
+    """Download W&B checkpoint (cached) and return path to checkpoint.pt."""
+    cache_root.mkdir(parents=True, exist_ok=True)
+    run_cache = cache_root / run_id
+    ckpt_path = run_cache / "checkpoint.pt"
+    if ckpt_path.exists():
+        return ckpt_path
+    api = wandb.Api()
+    run = api.run(f"{WANDB_PROJECT}/{run_id}")
+    art = next(a for a in run.logged_artifacts() if a.type == "model")
+    art.download(root=str(run_cache))
+    if not ckpt_path.exists():
+        raise FileNotFoundError(f"Expected {ckpt_path} after artifact download")
+    return ckpt_path
+
+
+def load_state_dict(ckpt_path: Path) -> dict[str, torch.Tensor]:
+    ckpt = torch.load(ckpt_path, map_location="cpu", weights_only=True)
+    sd = ckpt["model"]
+    # Strip torch.compile prefix if present (none of our candidates have it,
+    # but defensive).
+    return {k.removeprefix("_orig_mod."): v for k, v in sd.items()}
+
+
+def soup_state_dicts(
+    state_dicts: list[dict[str, torch.Tensor]],
+    weights: list[float] | None = None,
+) -> dict[str, torch.Tensor]:
+    """Weighted average of K state_dicts. All sds must share keys/shapes/dtypes.
+
+    If `weights` is None, uses uniform mean. Otherwise weights are normalized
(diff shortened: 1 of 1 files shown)


morganmcg1:
STUDENT: The PR has a title/body mismatch — title and branch name are 'nezuko/checkpoint-soup-eval' (top-K best-val checkpoint averaging / model soup eval), but the body describes a `volume_loss_weight=1.5` experiment from PR #115's stack with stale baseline numbers (val 9.484% — current SOTA is PR #523 at val 6.9246%).

Could you confirm which experiment to run?

1. **Checkpoint soup eval** — average top-K best-val checkpoints (model soup) and evaluate on test, building from current SOTA (PR #523 EMA-proxy GradNorm, run `wyz68o8r`)?
2. **volume_loss_weight=1.5** — re-run with the body's exact command but updated to current SOTA stack?

Holding off on implementation until you confirm.

morganmcg1:
ADVISOR: Confirmed — run **option 1: checkpoint soup eval (model soup)**. Ignore the stale body (it was a copy-paste mismatch from an old assignment template). The branch/title is correct: top-K best-val checkpoint averaging.

## Goal

Test whether averaging the weights of the top-K best-val checkpoints across recent strong runs yields a better-generalizing model than any single checkpoint. This is a Wortsman-style "model soup" — uniform averaging of state_dicts from independently-converged runs that share the same architecture. Free parameter: K. Free question: does it transfer the val gains to test? Test_abupt has been the open weakness — soup is well-known to close train→test generalization gaps.

## Concrete protocol

This is a **post-hoc evaluation experiment**, not a new training run. There is no native `train.py` flag for this — write a small script under `target/scripts/soup_eval.py` (or similar). The full DDP training infra is not needed; load on a single GPU, evaluate.

### Step 1 — pick the K source checkpoints

Query the W&B project `wandb-applied-ai-team/senpai-v1-drivaerml-ddp8` for runs on the `tay` branch with the **same architecture as PR #523 SOTA** (Lion lr=1e-4, 4L/512d/8h/128sl, STRING-sep, QK-norm, feat16 multi-sigma RFF, ema=0.999). Rank by `summary["best_val_primary/abupt_axis_mean_rel_l2_pct"]` ascending and take the top-K.

Recommended starting set (the strongest mutually-architecture-compatible runs):

| Rank | PR | Run ID | best val_abupt |
|---|---|---|---|
| 1 | #523 thorfinn | `wyz68o8r` | 6.9246 (current SOTA) |
| 2 | #510 alphonse | `qqtdnlwq` | 7.0063 |
| 3 | #511 edward   | `5o7jc7wi` | 7.0134 |
| 4 | #489 (prior tier) | (look up) | ~7.13 |
| 5 | #488 (prior tier) | (look up) | ~7.18 |

Verify the architecture flags match across all candidates before you average — different `--rff-num-features`, `--model-heads`, or pos-encoding modes will produce shape mismatches or scrambled weights. Reject any candidate that doesn't match the SOTA arch exactly.

### Step 2 — download checkpoints from W&B

Each run has a W&B artifact (the file is `checkpoint.pt` inside the model artifact). Pseudocode:

```python
import wandb
api = wandb.Api()
def load_ckpt(run_path):
    run = api.run(run_path)  # e.g. "wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/wyz68o8r"
    art = next(a for a in run.logged_artifacts() if a.type == "model")
    ckpt_dir = art.download()
    return torch.load(f"{ckpt_dir}/checkpoint.pt", map_location="cpu", weights_only=True)
```

The checkpoint dict has key `"model"` containing the state_dict (see `trainer_runtime.py:1397-1403`).

### Step 3 — uniform soup the state_dicts

```python
sds = [load_ckpt(p)["model"] for p in run_paths]
keys = sds[0].keys()
soup = {}
for k in keys:
    # confirm dtype/shape match across all sds for this key
    soup[k] = torch.stack([sd[k].float() for sd in sds]).mean(dim=0).to(sds[0][k].dtype)
```

Sanity: assert all sds have identical key sets; if they differ, abort (architecture mismatch). Buffers (e.g. running stats if any) — average them too unless they are integer counters; in that case take the max from rank-1.

### Step 4 — evaluate the soup on val + test

Reuse the val/test evaluation path from `train.py` / `trainer_runtime.py`. Build the model from the SOTA hyperparameters, `model.load_state_dict(soup)`, then run val + test once each. Report all 7 channels (abupt, surface_pressure, wall_shear, volume_pressure, tau_x, tau_y, tau_z) on both val and test.

Important: use the **non-EMA** weights from each checkpoint by default (the saved `"model"` key is the rank-0 model that was selected as best-val — confirm in trainer_runtime which weights get serialized). If you discover the saved weights are already the EMA copy, that's fine — just be consistent.

### Step 5 — sweep K

Run K = 2, 3, 4, 5 (and optionally 6+). The soup curve usually has a knee — report the best K and a small table.

## Reporting

In the PR results comment, table format:

| K | runs included | val_abupt | test_abupt | val all 7 | test all 7 |
|---|---|---|---|---|---|

Decision rule: if any K beats the SOTA single-checkpoint **test_abupt = 8.2355%** (PR #523), this is a winner regardless of val. Beating val without test is informative but not a merge.

## Compute budget

This should be cheap — only K full-eval passes on 1 GPU (no training). Budget: <30 min total. If your worker is already idle, run it now.

## Caveats / things to watch

- If two checkpoints have very different optimizer trajectories (different LR schedules, different epoch counts), the soup may underperform each individual member. That is itself a useful negative result — log it.
- If shapes don't match, surface the offending key+shape rather than silently skipping.
- Keep the script self-contained under `target/scripts/` and add a one-line README in the script docstring so it's reproducible.

Ping me with the results table when ready.

morganmcg1:
STUDENT nezuko:

## Results

**Verdict: NEGATIVE — uniform soup of independently-trained checkpoints catastrophically fails. K=1 (single best checkpoint) wins by ~13×; K≥2 cannot beat the SOTA single-checkpoint test_abupt = 8.2355% (PR #523).** The math is verified correct; the failure is structural.

### TL;DR

| K | val_abupt | test_abupt | members (run IDs) |
|---|---:|---:|---|
| 1 | **6.8701** | **8.1229** | `9mm3sz7x` (askeladd PR #516 WIP, top val) |
| 2 | 93.4366 | 94.1326 | `9mm3sz7x` + `49aimdiz` |
| 3 | 94.1786 | 94.3403 | + `wyz68o8r` |
| 4 | 92.1904 | 92.3563 | + `qqtdnlwq` |
| 5 | 100.4215 | 100.3591 | + `5o7jc7wi` |
| 6 | 98.8454 | 98.7661 | + `nh2ke150` |
| 7 | 98.2378 | 98.2049 | + `qawfhlu6` |
| 8 | 91.1808 | 91.3007 | + `r5rw40rn` |

K=1 reproduces the W&B-recorded `test_primary` exactly (8.1229 ↔ summary 8.122860), confirming the eval pipeline is sound. K=2..8 collapse to ~90-100% rel-L2 across every channel — i.e., effectively random predictions. Greedy soup (Wortsman) would reject every candidate after the first and degenerate to K=1.

### K=1 (single best checkpoint) — full 7 channels

| Metric | val | test | AB-UPT |
|---|---:|---:|---:|
| abupt_axis_mean | 6.8701 | 8.1229 | — |
| surface_pressure | 4.5153 | 4.1275 | 3.82 |
| wall_shear | 7.7572 | 7.4535 | 7.29 |
| volume_pressure | 4.1441 | 12.0512 | 6.08 |
| wall_shear_x | 6.7626 | 6.5662 | 5.35 |
| wall_shear_y | 8.6629 | 8.3261 | 3.65 |
| wall_shear_z | 10.2656 | 9.5432 | 3.63 |

### Side observation worth flagging

**The K=1 single-best (`9mm3sz7x`, val 6.870, test 8.123) already beats the BASELINE.md SOTA `wyz68o8r` (PR #523, val 6.925, test 8.236) by −0.113pp on test (−1.36% relative).** This run is from `askeladd-tau-reweight-micro` (PR #516, currently `status:wip`, slw=2.0, lr=9e-5, vol-curric, multi-sigma init). It is not currently flagged as SOTA in BASELINE.md. The soup eval is a no-op vs picking the best individual.

### Diagnostic probes (root cause)

Math verified: `soup({sd, sd}) == sd` exactly, and `soup({a, b}) == (a+b)/2` exactly elementwise (no quantization round-trip — checkpoints are stored in fp32). The negative result is real, not a bug.

| Probe | Setup | val_abupt | test_abupt |
|---|---|---:|---:|
| Same-lineage uniform K=2 | `qqtdnlwq` + `49aimdiz` (both alphonse, slw=2.0, multi-sigma init, lr=1e-4 — nearly identical training configs) | 92.85 | 93.16 |
| Small-perturbation 0.95/0.05 | `9mm3sz7x` (×0.95) + `49aimdiz` (×0.05) | 22.30 | 22.37 |

Two structural facts:

1. **Hyperparameter similarity does not save the soup.** Two alphonse runs trained with identical loss weights, lr, init scheme, and vol-curric, but different seeds, still produce a catastrophic soup (val 92.85). The dominant factor is the random init / data-order seed, not the training recipe. This is consistent with the loss-basin permutation problem in independently-trained MLPs (Ainsworth et al. 2022, "Git Re-Basin"; Entezari et al. 2022).

2. **Basin walls are very steep.** A 5% mix of a second model degrades val from 6.87 → 22.30 (3.2× regression). A 50% mix gives 93.4 (13.6× regression). The single-checkpoint solution sits in a narrow well; the second model pulls the parameters off the manifold quickly.

These match Wortsman's "model soup" pre-condition: uniform weight averaging works only when the constituent models are fine-tuned from a shared pretrained checkpoint (so they live in connected basins). Our SOTA stack trains from random init each run; there is no shared anchor, so the basins are not connected.

### Compute used

- 8 checkpoints downloaded from W&B (~14 GB total, cached at `target/outputs/soup_ckpts/`).
- K=1 sanity smoke (1 GPU) — 716s.
- K=2..8 sweep parallelized across 7 GPUs (1 K each) — ~715s wall clock.
- 2 diagnostic probes parallelized on 2 GPUs — ~715s wall clock.
- Total wall clock: ~36 min (≈80 GPU-min). All inference; no training.

### Exact commands used

```bash
# K=1 smoke
CUDA_VISIBLE_DEVICES=0 python target/scripts/soup_eval.py \
  --k-list 1 --output-json target/outputs/soup_eval_k1_smoke.json

# K=2..8 parallel sweep (one K per GPU)
for K in 2 3 4 5 6 7 8; do
  GPU=$((K-2))
  CUDA_VISIBLE_DEVICES=$GPU python target/scripts/soup_eval.py \
    --k-list $K --output-json target/outputs/soup_eval_k${K}.json &
done

# Probe A: same-lineage uniform pair
CUDA_VISIBLE_DEVICES=0 python target/scripts/soup_eval.py \
  --custom-recipe "qqtdnlwq,49aimdiz" \
  --output-json target/outputs/soup_probe_same_lineage.json

# Probe B: 0.95/0.05 small perturbation
CUDA_VISIBLE_DEVICES=1 python target/scripts/soup_eval.py \
  --custom-recipe "9mm3sz7x:0.95,49aimdiz:0.05" \
  --output-json target/outputs/soup_probe_weighted95.json
```

Peak GPU memory: ~16-20 GB per K (single-rank fp32 model + bf16 activations on 65k+65k points, batch 4).

### What happened

- **The advisor's protocol was implemented exactly**: top-K best-val ranking, architecture compatibility verified across all 8 candidates (4L/512d/4h/128sl, string_separable, qk-norm, rff_num=16, Lion, ema=0.999), uniform state_dict average, single-GPU full-fidelity val+test eval.
- **K=1 reproduces W&B summaries to 6 sig figs**, so the eval pipeline is sound.
- **K≥2 fails because the runs trained from independent random inits.** Even the most architecturally and hyperparam-similar pair (two alphonse slw=2.0 multi-sigma runs) cannot be soup-merged; even a 5% perturbation degrades the well-tuned single. The negative result is structural, not numerical.

This is exactly the scenario the advisor flagged as "very different optimizer trajectories → soup may underperform each individual member" — but the deeper finding is that *even very SIMILAR* trajectories fail, because the common anchor missing here is a shared pretrained init.

### Suggested follow-ups

1. **Prediction (logit) ensemble** — average `surface_preds` and `volume_preds` from K models *at inference time*. This sidesteps the basin/permutation issue entirely. Cost: K× inference per batch, no extra training. Expected: small but positive gain over K=1 (typical ensemble bonus 1-3% on held-out targets). Easy to bolt on to `soup_eval.py` — load K models, ensemble in normalised space, compute metrics. Would recommend K=2 (top-2: `9mm3sz7x` + `49aimdiz`) and K=3 first.
2. **Within-run mid-training checkpoint averaging** — average checkpoints from epochs 9, 10, 11, 12, 13 of a single SOTA run (e.g. `5o7jc7wi` ep13, `49aimdiz` ep13). These live on the same trajectory, so connected by construction. Requires adding mid-epoch checkpoint logging to `train.py`. This is also the mechanism by which EMA already wins (and EMA is already enabled).
3. **Permutation-aware merging (git-rebasin / TIES / DARE)** — solve a per-layer LAP to align permutations of the K models before averaging (Ainsworth 2022). Substantially more complex (per-block matching cost is O(d² log d) per layer × K-1 alignments). Likely overkill given the prediction-ensemble alternative is essentially free.
4. **Promote `9mm3sz7x` to BASELINE.md** as the current SOTA on test_abupt (8.123 < 8.236) **independent of the soup question**, and revisit closing PR #516 once askeladd has captured the result. *(Mentioning here so it doesn't get lost in this PR, since I noticed it while building the candidate list.)*

### Artefacts

- Script: `target/scripts/soup_eval.py` (CLI; supports `--k-list`, `--custom-recipe`, weighted recipes)
- Per-K results: `target/outputs/soup_eval_k{1..8}.json`
- Probes: `target/outputs/soup_probe_same_lineage.json`, `target/outputs/soup_probe_weighted95.json`
- Consolidated: `target/outputs/soup_eval_FINAL.json`
- Cached checkpoints: `target/outputs/soup_ckpts/{run_id}/checkpoint.pt` (kept around in case the advisor wants to run prediction ensembles or re-evaluate without re-downloading).

morganmcg1:
ADVISOR: Closing PR #554 because ADVISOR: Excellent negative result — and a very well-run investigation. Closing as a clear dead end for model soup specifically; no merge needed.

**What we learned:** Uniform weight averaging of independently-initialized runs is structurally broken for our setup. The root cause is exactly what Ainsworth 2022 (Git Re-Basin) describes: models trained from different random seeds live in disconnected loss basins. Even two runs with essentially identical hyperparameters (same-lineage alphonse slw=2.0 runs) fail catastrophically with K=2 (val 6.87 → 92.85). A 5% mix of a second model degrades val 3.2× — the basin walls are extremely steep. The pre-condition for Wortsman-style uniform soup (shared pretrained checkpoint anchor) is absent in our setup.

**K=1 is already the correct operation** — your eval confirmed run 9mm3sz7x (val=6.8701, test=8.1229) as the best single checkpoint, which I have already committed to BASELINE.md as the new SOTA.

**Next steps from your suggested follow-ups:** I'll be assigning you a prediction ensemble experiment next — this is the correct variant that sidesteps the basin issue entirely and has the best risk/reward ratio. Stay tuned for a new PR.

Great work on the diagnostic probes — the same-lineage pair test (two alphonse near-identical-config runs still fail) is particularly valuable evidence that the problem is fundamental, not recipe-dependent..

PR #531, opened 2026-05-03
fern: unit-vector direction loss on wall shear (tau_y/tau_z)

morganmcg1:
## Hypothesis

The wall shear vector (tau_x, tau_y, tau_z) is a 3D vector defined by both magnitude and direction. Our current loss (`masked_mse` over all 4 surface channels: p_s + tau_xyz) treats each component independently with isotropic squared error — it does not encode the constraint that tau is a physical vector. Tau_y and tau_z are 2.5–2.9× above the AB-UPT reference (9.19% and 10.70% vs 3.65%/3.63%), while tau_x is much closer (7.02% vs 5.35%). This suggests the model has trouble learning the correct angular orientation of wall shear, particularly for the cross-stream (y) and vertical (z) components.

We hypothesize that adding an **auxiliary unit-vector cosine similarity loss on the wall shear direction** — independent of magnitude — will guide the model to learn directional fidelity for the tau vector, which the component-wise MSE does not explicitly target. This is analogous to adding a direction supervision signal, as used in normal estimation and optical flow tasks.

Reference: Cosine embedding loss / cosine similarity loss for vector-output regression is well-established in surface normal estimation (e.g., EigenNet, GeoNet, DORN). The wall shear vector plays the same geometric role as a surface normal — it lies in the tangent plane of the body surface.

## Instructions

You will modify `train.py` and/or `trainer_runtime.py` to add a unit-vector direction loss on the wall shear channels (indices 1, 2, 3 of `surface_preds` / `surface_target`).

### Step 1: Add `--tau-unit-loss-weight` CLI flag

In `train.py`, add a new argument:

```python
parser.add_argument("--tau-unit-loss-weight", type=float, default=0.0,
                    help="Weight for auxiliary cosine-similarity direction loss on wall shear unit vector. "
                         "0.0 disables it (default). Try 0.1 as a starting point.")
```

Also add it to the `TrainConfig` dataclass (or equivalent config object):

```python
tau_unit_loss_weight: float = 0.0
```

### Step 2: Implement the direction loss

Add a helper function (e.g. in `trainer_runtime.py` or inline in the loss step):

```python
def masked_cosine_direction_loss(
    preds: torch.Tensor,   # shape [B, N, C] — predictions (denormalized or normalized scale — see note)
    targets: torch.Tensor, # shape [B, N, C] — targets
    mask: torch.BoolTensor # shape [B, N]
) -> torch.Tensor:
    """
    Cosine similarity loss on the direction of the 3D wall shear vector.
[...]

PR #559, opened 2026-05-04
norman: flow-aligned tangent-frame tau prediction (tau_y/tau_z gap attack)

morganmcg1:
## Hypothesis

The persistent tau_y/tau_z gap (val 8.72% / 10.30% vs AB-UPT 3.65% / 3.63%) is structural — it's been resistant to data augmentation (#536), static reweighting (#142, #454, #467, #531 all NEG), output head scaling (#467 NEG), and direction loss (#531 NEG). All these levers attack the symptom; the underlying issue may be **the prediction frame itself**.

Wall shear stress is a vector field that lives intrinsically in the **tangent plane to the surface**. Predicting it in global (x,y,z) coordinates forces the model to (a) implicitly learn the surface tangent space and (b) decompose tau into Cartesian components that have **no axis-aligned physical meaning**. tau_x is "easier" than tau_y/tau_z mainly because the car is x-aligned (flow direction = +x) — but for general points the local tangent geometry varies arbitrarily.

**Hypothesis: predict tau in a flow-aligned tangent frame (t̂, ŝ) — t̂ = freestream direction projected onto local tangent plane, ŝ = orthogonal in-plane vector. The model output becomes (tau_t, tau_s) with the wall-normal component ≡ 0 by construction. Convert back to global (x,y,z) for loss computation. The model now expresses tau in the natural physical basis and may close the channel gap from the prediction-side rather than fighting it via reweighting.**

This is a *preprocessing/postprocessing* change to the surface output — model architecture is untouched.

## Instructions

This requires a focused code change in `target/train.py` (or wherever the surface forward/loss is constructed). **Read the existing surface output path carefully before modifying.**

1. **Add a new flag** `--tau-output-frame {global, flow-tangent}` (default `global` for SOTA reproduction).

2. When `tau-output-frame=flow-tangent`:
   - **Inputs to model**: unchanged (still 3D coords).
   - **Model output**: still 3D vector per surface point — but interpret it as `(pred_t, pred_s, pred_n)` in a local frame where:
     - n̂ = surface normal at the query point (use the same surface-normal field already computed for the validation pipeline — DrivAerML provides per-vertex normals; query-point normals come from nearest-vertex lookup or KNN average)
     - t̂ = (freestream − (freestream · n̂) n̂) / ||·||  — freestream = `(1, 0, 0)` in DrivAerML
     - ŝ = n̂ × t̂
   - **Loss computation path**:
     - Build the per-point rotation matrix `R = [t̂ | ŝ | n̂]` (3×3, rows are world-axis projections of the local frame)
     - `pred_global = R @ pred_local`
     - Compute MSE between `pred_global` and `target_tau` (which is in global Cartesian as before)
     - **Penalize pred_n separately or keep it free**: start by allowing pred_n to be free (model can learn it ≈ 0 from data); a stricter ablation could project it out, but start permissive.

3. **Validation/test metrics**: must still report tau_x, tau_y, tau_z in **global frame** (the rotation applied above is sufficient — metrics computed downstream of `pred_global` are unchanged).

4. **Edge cases**: surface points where freestream is parallel to n̂ (rare, e.g. directly upstream stagnation point). Add a safe fallback: if `||freestream − (freestream · n̂) n̂|| < 1e-6`, set `t̂ = arbitrary tangent` (e.g. project (0,1,0) onto the tangent plane). This prevents numerical blow-up at <1% of points.

5. **Reproduce command** — same SOTA stack with new flag enabled:

```bash
  --agent norman --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --gradnorm-alpha 0.5 --gradnorm-ema-beta 0.9 --gradnorm-min-weight 0.7 \
  --tau-output-frame flow-tangent \
  --wandb-group norman-tau-flow-frame
```

6. **Sanity gate**: with `--tau-output-frame global` (default), training must reproduce SOTA val_abupt within ±0.1pp on a quick 1-epoch run before launching the full experiment with `flow-tangent`.

## Baseline

Current SOTA (PR #523 thorfinn EMA-proxy GradNorm, EP6, run `wyz68o8r`):
- val_abupt = **6.9246%**
- test_abupt = **8.2355%**
- val tau_y = **8.7197%**, val tau_z = **10.2960%** ← these are the targets to attack
- val tau_x = 6.7193% (much easier — confirms axis-alignment is part of the issue)

**Win condition**: val_abupt < 6.9246% AND test_abupt < 8.2355%.
**Mechanistic gate**: even if val_abupt ties, a meaningful tau_y/tau_z reduction (>0.5pp on either) validates the hypothesis and unlocks composability with GradNorm-EMA.


morganmcg1:
ADVISOR: Closing PR #559 because ADVISOR retracting: misrouted assignment. The student does not have a deployed pod on the tay (drivaerml-ddp8) branch. Reopening on a branch with the student's deployment if relevant..

PR #560, opened 2026-05-04
stark: loss-aware tau importance sampling (boosting-style hard-region focus)

morganmcg1:
## Hypothesis

Training currently uses **uniform random sampling** of 65536 surface query points per case per step. The model sees 99.9% of these points repeatedly across epochs, but the regions where it does worst — high-tau-magnitude wakes, vortex separation lines, A/C-pillar attachments — are under-sampled relative to their loss contribution. **Hypothesis: replacing uniform sampling with importance sampling proportional to per-point historical tau-error gives the model proportionally more gradient signal where it most needs improvement, closing the tau_y/tau_z gap by raising the gradient SNR on hard regions.**

This is classic boosting-style data weighting: Adaboost-style or "hard example mining" applied to query-point selection. The mechanism is orthogonal to GradNorm-EMA (which reweights per-channel losses) and to coord-jitter (which regularizes inputs).

References: Hard Example Mining (Shrivastava et al. 2016, OHEM), Focal Loss (Lin et al. 2017), and the standard Kaggle practice of "boost on residuals" in ensemble training.

## Instructions

This requires careful changes to the train-time sampling path. **Do not touch validation sampling** — val must remain uniform-random for fair comparison.

1. **Add new flags**:
   - `--tau-importance-sampling` (boolean, default off)
   - `--tau-importance-warmup-epochs` (default 2 — uniform until then to avoid bootstrapping from random init)
   - `--tau-importance-temperature` (default 1.0 — softmax temp; 1.0 = pure prob proportional to error, infinite = uniform, 0 = greedy max-error)

2. **Implementation sketch**:
   - Maintain a **per-case per-vertex error buffer** (size = N_cases × max_surface_vertices, typically 34 × ~150k = ~5M floats per buffer; one buffer for tau_y, one for tau_z, one combined). Init to uniform 1.0.
   - During each training step, when sampling 65536 query points from a case's surface mesh: replace `np.random.choice(N, 65536)` with `np.random.choice(N, 65536, p=softmax(error_buffer / T))`.
   - On each backward, **update the error buffer** for the points actually sampled this step using EMA: `buf[idx] = 0.95 * buf[idx] + 0.05 * |pred - target|^2` (combined tau_y/tau_z magnitude). Use `tensor.detach().cpu().numpy()` to update the buffer.
   - Buffer lives in CPU memory (4-byte float, ~20MB total — trivial).
   - **First N (`tau-importance-warmup-epochs`) epochs**: keep uniform sampling. Buffer is being populated but probabilities aren't used yet.

3. **Numerical stability**: clip error buffer to `[1e-6, 1e3]` per entry to prevent NaN softmax.

4. **Sanity gate**: with `--tau-importance-sampling=False`, training must reproduce SOTA. Add a pre-flight 1-epoch test before the full run.

5. **Reproduce command** for the experiment arm:

```bash
  --agent stark --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --gradnorm-alpha 0.5 --gradnorm-ema-beta 0.9 --gradnorm-min-weight 0.7 \
  --tau-importance-sampling --tau-importance-warmup-epochs 2 --tau-importance-temperature 1.0 \
  --wandb-group stark-tau-importance-sampling
```

6. Report per-epoch table including: val_abupt, val_tau_y, val_tau_z, and `frac_sampled_top10_error` (the fraction of sampled points that fall in the top-10% of the error buffer at sampling time) — log this as a sanity-check that importance sampling is actually firing.

## Baseline

Current SOTA (PR #523 thorfinn, run `wyz68o8r`):
- val_abupt = **6.9246%**, test_abupt = **8.2355%**
- val tau_y = 8.7197%, val tau_z = 10.2960%

**Win condition**: val_abupt < 6.9246% AND test_abupt < 8.2355%.
**Mechanistic gate**: tau_y/tau_z reduction >0.3pp validates the importance-sampling mechanism even if abupt mean ties.


morganmcg1:
ADVISOR: Closing PR #560 because ADVISOR retracting: misrouted assignment. The student does not have a deployed pod on the tay (drivaerml-ddp8) branch. Reopening on a branch with the student's deployment if relevant..

PR #543, opened 2026-05-03
haku: principal curvatures (H,K) as surface input features

morganmcg1:
## Hypothesis

Your diagnostic work (PR #363) showed that τ_y/τ_z errors concentrate at high-curvature surface zones — the A-pillars, side-mirror wake foot, wheel-arch lip, rear spoiler edge. These are exactly the regions where boundary-layer separation strongly modulates cross-flow shear.

The current surface input has 6 channels: (x, y, z, n_x, n_y, n_z) — no second-order geometry. Adding **per-vertex principal curvatures** (κ_1, κ_2 or mean curvature H = (κ_1+κ_2)/2 and Gaussian curvature K = κ_1·κ_2) gives the model an explicit signal for **where** separation-driven cross-flow shear is physically expected, without changing the architecture.

This is distinct from:
- PR #225 (haku mirror aug — train-time augmentation)
- PR #314 (norman coord jitter — noise augmentation)
- PR #363 (haku diagnostic — read-only, no model change)

## Instructions

**Step 1: Compute curvatures in the dataset loader.** For each surface point, fit a local quadric patch using its k=8 nearest neighbors in surface_xyz space:

```python
import numpy as np
from scipy.spatial import KDTree

def compute_surface_curvatures(xyz, normals, k=8):
    """Compute per-point mean H and Gaussian K curvature.
    xyz: [N, 3], normals: [N, 3]
    Returns: H [N], K [N]
    """
    tree = KDTree(xyz)
    _, nn_idx = tree.query(xyz, k=k+1)  # includes self
    nn_idx = nn_idx[:, 1:]  # exclude self -> [N, k]
    
    H_all, K_all = np.zeros(len(xyz)), np.zeros(len(xyz))
    for i in range(len(xyz)):
        n = normals[i]  # local normal
        neighbors = xyz[nn_idx[i]] - xyz[i]  # [k, 3], centered
        # Project neighbors onto tangent plane
        proj = neighbors - (neighbors @ n[:, None]) * n[None, :]  # [k, 3]
        # Build local tangent basis
        u = np.cross(n, [0, 1, 0]); u = u / (np.linalg.norm(u) + 1e-8)
        v = np.cross(n, u)
        # UV coords
        uv = np.stack([proj @ u, proj @ v], axis=1)  # [k, 2]
        # Height in normal direction
        h = neighbors @ n  # [k]
        # Fit: h ≈ a*u² + b*u*v + c*v²
[...]

PR #565, opened 2026-05-04
thorfinn: principal curvatures (κ₁,κ₂) as surface input features

morganmcg1:
## Hypothesis

**Adding per-vertex principal surface curvatures (κ₁, κ₂) as extra input features will help the model identify high-curvature surface regions (A-pillar, mirror wake foot, wheel-arch lip) where τ_y/τ_z are largest, reducing the dominant 2.35×/2.73× gap above AB-UPT targets.**

The current surface input is 6 channels: `(x, y, z, nx, ny, nz)`. This gives the model point position and local normal — first-order geometric information. There is no second-order information: the model cannot distinguish a flat region from a curved region with the same normal at a given point.

**Why curvatures matter for τ_y/τ_z:** Wall-shear cross-components τ_y, τ_z are overwhelmingly driven by:
1. Separation-line crossings (where boundary layer detaches) — these occur at lines of high negative curvature (concave into the flow)
2. Horseshoe vortex footprints (A-pillar, roof-side junction) — these occur at convex-concave transitions
3. Wake-reattachment zones (rear of car) — again high curvature

Principal curvatures κ₁, κ₂ (maximum and minimum normal curvature at each surface point) directly encode this second-order geometry. Mean curvature H = (κ₁+κ₂)/2 identifies separation-prone regions; Gaussian curvature K = κ₁·κ₂ distinguishes saddle points (K<0, likely separation) from dome-like (K>0, stable attached flow) regions.

**Expected mechanism:** with κ₁, κ₂ as input features, the model can learn `f(geom) → τ_y high if κ₁ < threshold` without having to infer curvature from neighboring point positions (which requires global reasoning across the point cloud). This is a classic geometric feature engineering win in point cloud learning — see: PointNet++ (Qi et al. 2017), CurveNet (Xiang et al. 2021), and standard discrete differential geometry tools.

**This is orthogonal to all existing PRs:** it's an input feature change (not loss, not architecture, not LR). Orthogonal to STRING-sep PE (which is about positional encoding), GradNorm (loss weighting), OHEM (mining), and streamline-frame (target rotation).

### Step 1 — Implement per-point curvature computation

Add a new function to `target/train.py` (or a separate `curvature_utils.py` imported from there):

```python
import numpy as np

def compute_surface_curvatures(surface_pts_np, k_neighbors=8):
    """
    Compute per-point principal curvatures (kappa1, kappa2) for a surface point cloud.
    
    Args:
        surface_pts_np: [N, 3] numpy array of surface point xyz coordinates
        k_neighbors: number of neighbors for local quadratic fit (default 8)
    
    Returns:
        kappa1: [N] maximum principal curvature (signed)
        kappa2: [N] minimum principal curvature (signed)
    """
    N = surface_pts_np.shape[0]
    kappa1 = np.zeros(N, dtype=np.float32)
    kappa2 = np.zeros(N, dtype=np.float32)
    
    # Build kNN index
    nbrs = NearestNeighbors(n_neighbors=k_neighbors+1, algorithm='kd_tree').fit(surface_pts_np)
    distances, indices = nbrs.kneighbors(surface_pts_np)
    
    for i in range(N):
        neighbors = surface_pts_np[indices[i, 1:], :]  # exclude self
        center = surface_pts_np[i]
        local = neighbors - center  # [k, 3]
        
        # PCA to find local tangent plane
        cov = local.T @ local
        eigenvalues, eigenvectors = np.linalg.eigh(cov)
        # smallest eigenvalue = normal direction, two largest = tangent plane
        t1 = eigenvectors[:, 2]  # largest eigenvec
        t2 = eigenvectors[:, 1]  # second largest
        
        # Project neighbors onto tangent plane (u, v) coords
        u = local @ t1  # [k]
        v = local @ t2  # [k]
        w = local @ eigenvectors[:, 0]  # height above tangent plane
        
        # Fit quadratic: w = a*u^2 + b*u*v + c*v^2
        A = np.stack([u**2, u*v, v**2], axis=1)  # [k, 3]
        try:
            coeffs, _, _, _ = np.linalg.lstsq(A, w, rcond=None)
            a, b, c = coeffs
            # Shape operator: [[2a, b], [b, 2c]]
            # Eigenvalues of shape operator = principal curvatures
            H_matrix = np.array([[2*a, b], [b, 2*c]])
            kvals = np.linalg.eigvalsh(H_matrix)
            kappa1[i] = kvals[1]  # max
            kappa2[i] = kvals[0]  # min
        except np.linalg.LinAlgError:
            kappa1[i] = 0.0
            kappa2[i] = 0.0
    
    return kappa1, kappa2
```

### Step 2 — Add CLI flag

```python
parser.add_argument('--surface-curvature', type=str, default='none',
    choices=['none', 'hk', 'k1k2'],
    help='Add surface curvature as extra input features. "hk" adds mean+Gauss (H, K), "k1k2" adds principal curvatures (kappa1, kappa2).')
```

### Step 3 — Precompute curvatures at dataset load time

In the dataset/data loading section, add curvature computation per case (once, before training starts, cached in memory):

```python
if args.surface_curvature != 'none':
    print(f"Precomputing surface curvatures ({args.surface_curvature}) for all cases...")
    import time
    t0 = time.time()
    curvature_cache = {}
    for case_id, pts in case_surface_pts.items():  # adapt to actual data structure
        k1, k2 = compute_surface_curvatures(pts[:, :3], k_neighbors=8)
        if args.surface_curvature == 'k1k2':
            curvature_cache[case_id] = np.stack([k1, k2], axis=1)  # [N, 2]
        else:  # 'hk'
            H = (k1 + k2) / 2.0
            K = k1 * k2
            curvature_cache[case_id] = np.stack([H, K], axis=1)  # [N, 2]
    print(f"Curvatures computed in {time.time()-t0:.1f}s")
```

Adapt the case loading to your actual data structure (look for where `surface_x` is assembled from per-case files).

### Step 4 — Append curvature features to surface input

When assembling `surface_x` for each batch, concatenate the precomputed curvatures:

```python
if args.surface_curvature != 'none':
    curv = torch.from_numpy(curvature_cache[case_id]).float().to(device)  # [N, 2]
    # Normalize: curvatures can be large; z-score per-feature across the dataset
    # Quick approach: clamp to [-5, 5] and scale by 0.1
    curv = torch.clamp(curv * 0.1, -5.0, 5.0)
    surface_x_case = torch.cat([surface_x_case, curv], dim=-1)  # [N, 6+2] = [N, 8]
```

### Step 5 — Update model input projection

The Transolver surface input projection currently expects 6 channels. With curvature features it needs 8. Add dynamic input dimension:

```python
# Where the surface input projection (first linear layer) is defined:
surface_in_dim = 6 + (2 if args.surface_curvature != 'none' else 0)
# Pass surface_in_dim to the model constructor so it can set the first layer correctly
```

Verify the correct location by checking `model.surface_input_proj` or equivalent in the model code. Log `train/surface_input_dim` at startup.

### Step 6 — Sweep (3 arms, sequential on 4×H100 DDP)

**Arm A (control):** `--surface-curvature none` (standard 6-channel input)  
**Arm B:** `--surface-curvature hk` (mean curvature H + Gaussian curvature K, 8 channels)  
**Arm C:** `--surface-curvature k1k2` (principal curvatures κ₁, κ₂, 8 channels)  

Use `--wandb-group thorfinn-r34-curvature-features` for all 3 arms.

**Full launch command for Arm A (control):**
```bash
cd target/
nohup torchrun --standalone --nproc_per_node=4 train.py \
  --agent thorfinn \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --clip-grad-norm 0.5 \
  --batch-size 4 \
  --validation-every 1 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --lr-warmup-epochs 1 \
  --learnable-pe \
  --no-compile-model \
  --surface-curvature none \
  --epochs 3 \
  --wandb-group thorfinn-r34-curvature-features \
  --wandb-name thorfinn-r34-arm-a-ctrl \
  > _runs/arm_a_ctrl.log 2>&1 &
echo "PID: $!"
```

Replace `--surface-curvature none` with `hk` (Arm B) and `k1k2` (Arm C) for the other arms. Run **sequentially** — each arm ~5.5–6h on 4×H100.

**Precomputation timing:** after implementing, run a quick timing check:
```python
import time, numpy as np
pts = np.random.randn(65536, 3).astype(np.float32)
t0 = time.time()
compute_surface_curvatures(pts, k_neighbors=8)
print(f"65536 pts: {time.time()-t0:.1f}s")
```
If >60s per case and there are many cases, switch to a faster implementation (ball-tree instead of kd-tree, or use Open3D's `estimate_normals` + `compute_point_cloud_curvatures`). Report timing in your first comment.

### IMPORTANT flag check

Use `--clip-grad-norm 0.5` (NOT `--grad-clip-norm`). Verify:
```bash
python train.py --help | grep -i clip
```

### Step 7 — Required diagnostics

Log at startup (step 0):
- `train/surface_input_dim` — should be 6 (Arm A), 8 (Arms B, C)
- `train/curvature_kappa1_mean`, `train/curvature_kappa1_std` — sanity that values are non-trivial
- `train/curvature_kappa2_mean`, `train/curvature_kappa2_std`

At each epoch end, log per-decile curvature error if possible (split val surface points into 10 buckets by |κ₁|, report τ_y rel_l2 in each bucket for Arm B vs Arm A). This is the key mechanistic check: improvement should be concentrated in the high-curvature deciles.

## Results to report

For each arm:
1. W&B run id + link
2. Curvature precomputation time per case (seconds)
3. Per-epoch `val/abupt_axis_mean_rel_l2_pct` table
4. Per-channel at best epoch: `surface_pressure`, `tau_y`, `tau_z`, `volume_pressure`
5. `test/abupt_axis_mean_rel_l2_pct` from best-val checkpoint
6. High-curvature vs low-curvature error breakdown (if you implement per-decile logging)

Win criterion: **val_abupt < 9.032%** AND τ_y or τ_z shows ≥5% relative improvement vs Arm A. The per-decile breakdown is optional but highly encouraged — if improvement is flat across curvature deciles, the feature isn't doing what we think.

## Baseline

Current best on yi:
- **val_abupt: 9.032%** (PR #517 askeladd, W&B run `brat65z4`, Lion lr=1e-4 clip=0.5)
- STRING-sep learnable PE now on yi via PR #490; frieren resumed run reached **8.087%** with `--learnable-pe` (awaiting clean from-scratch confirmation from PR #539)
- Expected next bar: ~8.1–8.5% once PR #539 lands

AB-UPT reference targets:
- τ_y (ws_y): 3.65% — **current yi: ~8.58%, gap 2.35×**
- τ_z (ws_z): 3.63% — **current yi: ~9.93%, gap 2.73×**
- surface_pressure: 3.82%
- volume_pressure: 6.08%
- wall_shear (overall): 7.29%

Reproduce baseline (6-channel, no curvature):
```bash
cd target/
  --agent thorfinn \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --clip-grad-norm 0.5 \
  --learnable-pe \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --ema-decay 0.999 \
  --lr-warmup-epochs 1 \
  --epochs 3
```


morganmcg1:
ADVISOR: Closing PR #565 because ADVISOR closing this PR immediately — duplicate hypothesis. Haku is already running the surface curvature features experiment in PR #543 (Arm A complete at 10.39% val_abupt, Arm B currently running). I made this assignment without fully checking PR #543's status first — apologies for the confusion. A new distinct experiment will be assigned to you shortly..

PR #595, opened 2026-05-04
asinh target-space normalization for wall-shear loss (τ_y/τ_z gap attack)

morganmcg1:
## Hypothesis

Wall-shear targets (τ_x, τ_y, τ_z) span multiple decades of magnitude across a vehicle surface — stagnation regions (τ ≈ 0) vs. leading edges (τ >> 0). Training the MSE/Huber loss directly on these raw values causes the loss to be dominated by the high-magnitude regions, leaving the near-zero regions (which often coincide with τ_y and τ_z over flat panels) effectively untrained.

**Applying asinh normalization to the wall-shear targets before loss computation** compresses the dynamic range from ~4 decades to ~1.5 decades, giving near-zero and large-magnitude regions equal gradient weight. This is fundamentally different from the closed PR #485 (asinh loss *scale sweep* on the input) — here we transform the **target space** so the loss landscape sees uniformly calibrated residuals.

Expected improvement: 10–25% reduction in wall_shear_y and wall_shear_z (dominant 2.35–2.73× gap vs AB-UPT), with neutral impact on surface_pressure and volume_pressure.

Reference: asinh normalization is standard in seismology and signal processing for heavy-tailed distributions. In ML: Yeo-Johnson transformation, Box-Cox.

## Instructions

Modify `target/train.py` to apply **asinh normalization** to the wall-shear target channels (channels 1, 2, 3 of the surface target tensor) before loss computation, and inverse-asinh the predictions before metric evaluation.

### Specific changes

**Step 1: Add a global scale constant near the top of `train.py`** (after imports):
```python
# asinh normalization scale for wall-shear targets (tau_x, tau_y, tau_z)
# Set to the ~75th percentile of |wall_shear| across training set to normalize the "typical" value to ~0.88
# Use 0.1 as a conservative starting point (1 Pa ≈ 1.0 in scaled units)
WS_ASINH_SCALE = 0.1  # Arm A: 0.1, Arm B: 0.01
```

**Step 2: In the surface loss computation** (find where `surface_loss` is computed), apply the transform to wall-shear channels only:
```python
# Apply asinh normalization to wall-shear target channels (1,2,3)
# surface_targets shape: [B, N, 4] — ch0=surface_pressure, ch1-3=tau_{x,y,z}
surface_targets_ws = surface_targets[..., 1:4]
surface_preds_ws   = surface_preds[..., 1:4]

# Normalize
surface_targets_ws_norm = torch.asinh(surface_targets_ws / WS_ASINH_SCALE)
surface_preds_ws_norm   = torch.asinh(surface_preds_ws   / WS_ASINH_SCALE)

# Replace wall-shear channels in loss computation only (not in metric eval)
surface_targets_for_loss = torch.cat([surface_targets[..., :1], surface_targets_ws_norm], dim=-1)
surface_preds_for_loss   = torch.cat([surface_preds[..., :1],   surface_preds_ws_norm],   dim=-1)
# Use surface_targets_for_loss and surface_preds_for_loss in the loss, but
# use the ORIGINAL surface_targets and surface_preds for metric computation.
```

**Step 3: Ensure metrics are computed on raw (un-normalized) predictions** — verify that the `compute_metrics()` or equivalent call receives the original `surface_preds` (not the asinh-transformed version).

**Step 4: Add CLI flags** so the scale is tunable without code changes:
```python
parser.add_argument("--ws-asinh-scale", type=float, default=0.0,
                    help="If > 0, apply asinh(x/scale) normalization to wall-shear loss targets. 0 = disabled.")
```
When `--ws-asinh-scale 0` (default), the behavior is identical to the current baseline (no regression).

### Arms to run

Run **2 arms sequentially** (use watchdog / background launch pattern for Arm B):

**Arm A** — scale=0.1:
```bash
  --agent alphonse \
  --learnable-pe \
  --optimizer lion \
  --lr 1e-4 \
  --clip-grad-norm 0.5 \
  --weight-decay 5e-4 \
  --lr-warmup-epochs 1 \
  --ema-decay 0.999 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --batch-size 4 \
  --epochs 50 \
  --ws-asinh-scale 0.1 \
  --wandb-group yi-round34-asinh-ws-target-norm
```

**Arm B** — scale=0.01 (finer compression, more aggressive for near-zero regions):
```bash
  --agent alphonse \
  --learnable-pe \
  --optimizer lion \
  --lr 1e-4 \
  --clip-grad-norm 0.5 \
  --weight-decay 5e-4 \
  --lr-warmup-epochs 1 \
  --ema-decay 0.999 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --batch-size 4 \
  --epochs 50 \
  --ws-asinh-scale 0.01 \
  --wandb-group yi-round34-asinh-ws-target-norm
```

### Important implementation notes
- `torch.asinh` is differentiable and available in PyTorch ≥ 1.7.
- The asinh transform is applied **only to loss computation**, not to metric logging. Metrics must be computed on raw predictions in physical units.
- Do NOT apply asinh to `surface_pressure` (channel 0) — that channel is already well-conditioned (range ~[−1, 1] Cp).
- Do NOT apply asinh to volume targets.
- Check that `WS_ASINH_SCALE` does not cause the normalized residuals to saturate (asinh(x/scale) should remain < 3 for typical values; if scale=0.01 produces large values, report it).

## Baseline (yi — must beat)

| Metric | val | W&B run |
|---|---:|---|
| `val_abupt` | **9.032%** | `brat65z4` |
| `surface_pressure_rel_l2_pct` | 5.702% | `brat65z4` |
| `wall_shear_rel_l2_pct` | 10.257% | `brat65z4` |
| `volume_pressure_rel_l2_pct` | 5.075% | `brat65z4` |
| `wall_shear_x_rel_l2_pct` | 8.520% | `brat65z4` |
| `wall_shear_y_rel_l2_pct` | 12.907% | `brat65z4` |
| `wall_shear_z_rel_l2_pct` | 12.957% | `brat65z4` |

**Win gate: val_abupt < 9.032% on either arm.**

Focus: expected τ_y/τ_z improvement. Surface pressure should be neutral (transform not applied to ch0).

## W&B group
`yi-round34-asinh-ws-target-norm`

### DIFF
--- a/run_asinh_arms.sh
+++ b/run_asinh_arms.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+# Sequential 2-arm sweep for PR #595 asinh wall-shear target-norm.
+set -u
+cd /workspace/senpai/target
+mkdir -p train_logs
+
+export SENPAI_TIMEOUT_MINUTES=${SENPAI_TIMEOUT_MINUTES:-360}
+export CUDA_VISIBLE_DEVICES=${CUDA_VISIBLE_DEVICES:-0,1,2,3}
+
+COMMON=(
+  --learnable-pe
+  --optimizer lion
+  --lr 1e-4
+  --clip-grad-norm 0.5
+  --weight-decay 5e-4
+  --lr-warmup-epochs 1
+  --ema-decay 0.999
+  --model-layers 4
+  --model-hidden-dim 512
+  --model-heads 8
+  --model-slices 128
+  --batch-size 4
+  --train-surface-points 65536 --eval-surface-points 65536
+  --train-volume-points 65536 --eval-volume-points 65536
+  --epochs 50
+  --validation-every 1
+  --no-compile-model
+  --agent alphonse
+  --wandb-group yi-round34-asinh-ws-target-norm
+)
+
+run_arm() {
+  local arm_name="$1"
+  local scale="$2"
+  local logfile="train_logs/${arm_name}.log"
+  local donefile="train_logs/${arm_name}.done"
+  echo "[$(date -Is)] launching ${arm_name} scale=${scale}"
+  torchrun --standalone --nproc_per_node=4 train.py \
+    "${COMMON[@]}" \
+    --ws-asinh-scale "${scale}" \
+    --wandb-name "alphonse/asinh-target-${arm_name}" \
+    > "${logfile}" 2>&1
+  local rc=$?
+  echo "[$(date -Is)] ${arm_name} exited rc=${rc}" | tee -a "${donefile}"
+  return $rc
+}
+
+run_arm "armA-scale0p1" 0.1
+echo "[$(date -Is)] arm A done — Arm B cancelled per advisor decision (saturation kill-gate hit, see PR #595 comment 16:16 UTC)"
+# run_arm "armB-scale0p01" 0.01  # cancelled
+echo "[$(date -Is)] sweep complete (Arm A only)" | tee -a train_logs/asinh_sweep_done
--- a/train.py
+++ b/train.py
@@ -722,6 +722,7 @@ class Config:
     wallshear_y_weight: float = 1.0
     wallshear_z_weight: float = 1.0
     wallshear_huber_delta: float = 0.0
+    ws_asinh_scale: float = 0.0
     manifest: str = "data/split_manifest.json"
     data_root: str = ""
     output_dir: str = "outputs/drivaerml"
@@ -1528,6 +1529,7 @@ def train_loss(
     wallshear_y_weight: float = 1.0,
     wallshear_z_weight: float = 1.0,
     wallshear_huber_delta: float = 0.0,
+    ws_asinh_scale: float = 0.0,
 ) -> tuple[torch.Tensor, dict[str, float]]:
     batch = batch.to(device)
     surface_target = transform.apply_surface(batch.surface_y)
@@ -1567,6 +1569,38 @@ def train_loss(
         else:
             surface_pred_used = surface_pred_norm
             surface_target_used = surface_target
+        if ws_asinh_scale > 0.0:
+            # asinh(x/scale) wall-shear target-norm. Denormalize ch1:4 to
+            # physical Pa, transform in fp32 (bf16 underflows the asinh
+            # derivative 1/sqrt(1+(x/s)^2) at large x/s), then keep the asinh
+            # result in fp32 alongside the cp channel cast to fp32. The
+            # downstream MSE auto-promotes (target is fp32) so this matches the
+            # baseline's effective fp32 MSE compute. cp (ch0) is untouched
+            # numerically. Eval/metrics see raw surface_pred_norm.
+            ws_std_t = transform.surface_y_std[1:4].to(
+                device=surface_pred_used.device, dtype=torch.float32
+            )
+            ws_mean_t = transform.surface_y_mean[1:4].to(
+                device=surface_pred_used.device, dtype=torch.float32
+            )
+            ws_pred_phys = surface_pred_used[..., 1:4].float() * ws_std_t + ws_mean_t
+            ws_true_phys = surface_target_used[..., 1:4].float() * ws_std_t + ws_mean_t
+            scale_t = float(ws_asinh_scale)
+            ws_pred_asinh = torch.asinh(ws_pred_phys / scale_t)
+            ws_true_asinh = torch.asinh(ws_true_phys / scale_t)
+            surface_pred_used = torch.cat(
+                [surface_pred_used[..., :1].float(), ws_pred_asinh], dim=-1
+            )
+            surface_target_used = torch.cat(
+                [surface_target_used[..., :1].float(), ws_true_asinh], dim=-1
+            )
+            if ws_pred_asinh.numel() > 0:
+                ws_asinh_max_pred_t = ws_pred_asinh.detach().abs().amax()
+                ws_asinh_max_target_t = ws_true_asinh.detach().abs().amax()
+            else:
+                _zero = torch.zeros((), device=ws_pred_asinh.device)
+                ws_asinh_max_pred_t = _zero
+                ws_asinh_max_target_t = _zero
         surface_loss, per_axis_unweighted = weighted_masked_mse_per_channel(
             surface_pred_used,
             surface_target_used,
@@ -1598,6 +1632,10 @@ def train_loss(
         metrics["aux_rel_l2_loss"] = aux_rel_l2_value
     if use_tangential_wallshear_loss:
         metrics["wallshear_pred_normal_rms"] = normal_rms
+    if ws_asinh_scale > 0.0:
+        metrics["ws_asinh_scale"] = float(ws_asinh_scale)
+        metrics["ws_asinh_max_pred"] = float(ws_asinh_max_pred_t.cpu().item())
+        metrics["ws_asinh_max_target"] = float(ws_asinh_max_target_t.cpu().item())
     if "geom_token" in out:
         geom_token = out["geom_token"].detach().float()
         metrics["film/geom_token_norm_mean"] = float(
@@ -2172,6 +2210,7 @@ def main(argv: Iterable[str] | None = None) -> None:
                 wallshear_y_weight=config.wallshear_y_weight,
                 wallshear_z_weight=config.wallshear_z_weight,
                 wallshear_huber_delta=config.wallshear_huber_delta,
+                ws_asinh_scale=config.ws_asinh_scale,
             )
             optimizer.zero_grad(set_to_none=True)
             loss_is_finite = bool(torch.isfinite(loss).item())
@@ -2290,6 +2329,9 @@ def main(argv: Iterable[str] | None = None) -> None:
                 train_log["train/aux_rel_l2_loss"] = batch_loss_metrics[
                     "aux_rel_l2_loss"
                 ]
+            for asinh_key in ("ws_asinh_scale", "ws_asinh_max_pred", "ws_asinh_max_target"):
+                if asinh_key in batch_loss_metrics:
+                    train_log[f"train/{asinh_key}"] = batch_loss_metrics[asinh_key]
             if ema_decay_now is not None:
                 train_log["train/ema_decay"] = ema_decay_now
             for key, value in batch_loss_metrics.items():


morganmcg1:
STUDENT alphonse:

## Picked up — implementation in progress

Implemented asinh wall-shear target-norm in \`train.py\` per PR spec:

- New CLI flag: \`--ws-asinh-scale\` (default 0.0 = baseline path).
- Inside \`train_loss\`, after \`surface_pred_used\`/\`surface_target_used\` are formed (works alongside the optional tangential-projection branch and the Huber path), the wall-shear channels (1:4) are denormalized to physical Pa, transformed by \`asinh(x/scale)\`, then concatenated back with the standardized \`cp\` channel for the loss. **cp (ch0) is untouched, volume targets untouched, eval/metrics path untouched** — metrics still come from raw predictions in physical units.
- Computation forced to fp32 inside autocast for stability (per researcher pass: \`d/dx asinh(x/s) = 1/sqrt(1+(x/s)^2)\` can underflow in bf16 for scale=0.01 + large \`x\`).
- Logged to wandb per-step: \`train/ws_asinh_scale\`, \`train/ws_asinh_max_pred\`, \`train/ws_asinh_max_target\` to monitor saturation (PR sanity-check requested |asinh| < 3 for typical values).

### Numerics check (synthetic τ ∈ [-15, 15] Pa)
| scale | max \|asinh\| |
|---|---:|
| 1.0 | 3.46 |
| 0.1 | 5.77 |
| 0.01 | 8.07 |

Confirms the PR's caveat that scale=0.01 saturates beyond 3 — will report observed \`ws_asinh_max_target\` per arm.

### Runs launched
4 GPUs allocated to this pod, so arms run sequentially via chain script (\`run_asinh_arms.sh\`):
- **Arm A**: \`--ws-asinh-scale 0.1\` (50 epochs, Lion lr=1e-4, learnable-pe, bs=4, 65k pts)
- **Arm B**: \`--ws-asinh-scale 0.01\` (same, sequential after A)

W&B group: \`yi-round34-asinh-ws-target-norm\`. Will report results once both arms finish.

morganmcg1:
ADVISOR: Implementation looks solid — fp32 cast inside autocast for the gradient stability concern is the right call for scale=0.01.

**Saturation concern — scale=0.01 is likely too aggressive:**
Your numerics table shows scale=0.01 gives |asinh|=8.07 for τ≈15 Pa. The gradient of asinh(x/s) = 1/sqrt(1+(x/s)²) approaches zero as |x/s|→∞ — at scale=0.01, x=15 Pa gives x/s=1500, gradient≈0.0007. The loss signal for large wall-shear values is essentially killed. Watch `train/ws_asinh_max_target` in W&B carefully: if it's consistently above 50×scale (i.e., >5.0 for scale=0.01), we're in flat territory for a significant fraction of training examples.

**Arm A win gates (scale=0.1):**
- EP1 val_abupt < 20% → on track (EP1 is still dominated by the untrained model, so 20% is a loose sanity gate)
- EP2 val_abupt < 11.73% → on track to beat baseline
- EP4 val_abupt < 9.032% → **merge bar**

**Arm B (scale=0.01):**
- If Arm A EP1 shows `ws_asinh_max_target` >> 3×scale, Arm B will be worse due to heavier saturation — flag this early so we can kill Arm B early if Arm A is already showing degradation.

**Per epoch update, please report:**
1. val_abupt
2. `train/ws_asinh_max_target` and `train/ws_asinh_max_pred` (end-of-epoch or step average)
3. Any wall-shear specific per-surface error if logged

Standing by for EP1 results.

morganmcg1:
STUDENT alphonse:

## Bug & relaunch — Arm A hit a torch._inductor compile crash, retrying with `--no-compile-model`

### What happened
Arm A (`scale=0.1`) ran 5441/5442 batches of EP1 (~1h30) cleanly, then crashed at the **last training batch of EP1** before validation could run. Stack ends in:

```
File "/usr/local/lib/python3.12/dist-packages/torch/_inductor/tiling_utils.py", line 271, in get_pw_red_splits
```

The model is wrapped in `torch.compile(model)` (default `--compile-model`). The DDP train sampler uses `drop_last=False` (train.py:888), so the very last batch each epoch is the partial remainder (2 samples instead of the usual 4 per rank). That triggers a dynamic-shape recompile, and Inductor's tiling pass fails the internal `tiling_utils.py:271` assert. The asinh transform is applied **outside** the compiled model graph (eager `torch.asinh` after `model(...)` returns), so it doesn't itself enter the compile, but it does change the autograd backward graph downstream of the compiled forward — possibly enough to trip a recompile path that doesn't trigger in the un-asinh'd baseline. Independent of root cause, the fix is to skip Inductor codegen.

Arm B (`scale=0.01`) had reached only 144/5442 of EP1 when I made the call. I **killed it** rather than wait the ~1h30 to learn it would crash the same way at the last batch. No EP1 result was lost.

### Fix
Added `--no-compile-model` to the chain script `COMMON` flags. Eager forward only; numerics are unchanged (compile is purely a perf wrapper). Expect ~5–15% slower step time. With ~1 it/s previously, EP1 was 1h30; un-compiled estimate ~1h40, so 4 epochs ≈ 6h40, slightly over the 6h `SENPAI_TIMEOUT_MINUTES=360` envelope. EP4 is the merge gate, so I'm running with the existing 50-epoch budget and letting the timeout cut in.

### Both arms relaunched in chain at 13:59 UTC

- Arm A: `--ws-asinh-scale 0.1 --no-compile-model` → `train_logs/armA-scale0p1.log` (note: file got overwritten — losing the failed run's pre-crash logs is OK since they showed normal training)
- Arm B: same, scale=0.01

Will report EP1/EP2/EP4 `val_abupt`, plus `train/ws_asinh_max_target` and `train/ws_asinh_max_pred` per epoch, as you asked.

### Bug-fix candidate (not implemented)
A targeted fix would be `drop_last=True` on the train DistributedSampler (train.py:888) — keeps compile enabled for ~10% speedup and drops one partial batch (~0.04% data per epoch). I held off because it's outside the asinh PR scope; flagging it here so it can be picked up as a one-line independent bug-fix PR if other students hit the same crash.

morganmcg1:
STUDENT alphonse:

## EP1 results — Arm A (scale=0.1) — concerning trajectory

EP1 finished at ~16:00 UTC (~115 min/epoch with `--no-compile-model`). Still EP1/50.

### Arm A EP1 vs PR baseline (`brat65z4`)

| Metric | EP1 (asinh A) | Baseline final | EP1 sanity gate | Δ |
|---|---:|---:|---:|---|
| `val_abupt` | **30.636%** | 9.032% | <20% (loose) | **+10.6pt above gate** |
| `surface_pressure_rel_l2_pct` | 20.34% | 5.702% | — | 3.6× worse |
| `wall_shear_rel_l2_pct` | 33.57% | 10.257% | — | 3.3× worse |
| `volume_pressure_rel_l2_pct` | 20.64% | 5.075% | — | 4.1× worse |
| `wall_shear_x_rel_l2_pct` | 28.00% | 8.520% | — | 3.3× worse |
| `wall_shear_y_rel_l2_pct` | 42.60% | 12.907% | — | 3.3× worse |
| `wall_shear_z_rel_l2_pct` | 41.60% | 12.957% | — | 3.2× worse |

W&B run: `7lgg0zcq`. EP2 in progress (5%, ~18:00 UTC ETA).

### Saturation diagnostics (per your gate)

| Diagnostic | Value | Threshold | Verdict |
|---|---:|---|---|
| `train/ws_asinh_scale` | 0.1 | — | — |
| `train/ws_asinh_max_pred` | 5.33 | <3 (PR cap) | **saturating** |
| `train/ws_asinh_max_target` | **6.30** | <3 (PR cap), >>3×scale per your kill-gate | **saturating, kill-gate hit** |

`ws_asinh_max_target = 6.30` is `21× above` the `3×scale = 0.3` kill-gate you set. The corresponding gradient `1/cosh(asinh(x/s)) = 1/cosh(6.30) ≈ 0.0037` — effectively dead for the largest wall-shear values. For typical `τ ≈ 5 Pa`, `asinh(5/0.1) = asinh(50) ≈ 4.6`, gradient ≈ 0.01 — also strongly suppressed.

### Why surface_pressure & volume_pressure are *also* degraded (not just wall-shear)

The `cp` channel (ch0) is left untouched, but its loss now competes against an asinh-transformed wall-shear loss whose magnitude is dramatically inflated. For typical normalized wall-shear residual `r ≈ 1.0`, the asinh-space residual is `asinh(physical_residual / 0.1) ≈ asinh(1/0.1) = asinh(10) ≈ 3.0`, so `r²` jumps from `1.0` to `~9.0` — a ~9× boost on the wall-shear loss term relative to the cp term. This explains the across-the-board degradation: the optimizer is now spending most of its capacity fitting the (now-saturating) wall-shear residuals.

### Recommendation re: Arm B (scale=0.01)

Per your pre-authorization to "kill Arm B early if Arm A is already showing degradation", **Arm B will be strictly worse** for two reasons:
1. **More saturation** — synthetic numerics had `|asinh|=8.07` for `τ=15 Pa` at scale=0.01; real EP1 will likely exceed 9. Gradient `1/cosh(9) ≈ 0.0001` — fully dead for any τ above ~0.1 Pa.
2. **Worse loss imbalance** — for `r ≈ 1.0`, `asinh(1/0.01) = asinh(100) ≈ 5.3`, so wall-shear loss term jumps to `~28×` the cp term (vs. ~9× for scale=0.1).

**My plan**: continue Arm A through EP2 to confirm the trajectory. If EP2 also fails the `<11.73%` gate, I'll cancel Arm B before it launches (kill the chain bash before it reads the Arm B line). Will report EP2 around 18:00 UTC. Confirming you're OK with this, otherwise I'll let Arm B run for the data point.

### Time budget note

`SENPAI_TIMEOUT_MINUTES=360` × 115 min/epoch ≈ **EP3 max per arm** before timeout. EP4 merge gate is unreachable in this run regardless of arm choice. If you want EP4 evidence, we'd need to disable validation between epochs or shrink the model — please advise.

morganmcg1:
ADVISOR: Your diagnosis is correct — kill Arm B.

**The math confirms it:** ws_asinh_max_target = 6.30 is 21× above the 3×scale = 0.3 kill-gate. The gradient of asinh(x/s) at x = 6.30, s = 0.1 is 1/cosh(asinh(63)) ≈ 0.0037 — effectively dead for the largest wall-shear values. Arm B at scale=0.01 would put the largest values at x/s = 630, giving gradient ≈ 0.00037 — 10× worse. Do not launch Arm B.

**Decision:** Let Arm A EP2 finish purely as a record entry (confirming the failure mode, not expecting recovery), then close this PR.

**Conclusion for the research log:** asinh normalization is not viable for this dataset's wall-shear range at any reasonable scale — the dynamic range of τ values (spanning ~3 orders of magnitude) causes catastrophic gradient saturation in the normalization transform itself. The correct approach to the τ_y/τ_z gap is likely coordinate frame alignment or architectural handling, not output space reparameterization.

Mark the PR ready for review when Arm A EP2 is complete (even if the result is bad — we need the data point).

morganmcg1:
STUDENT alphonse:

## Final Results — asinh wall-shear target-norm (Arm A, scale=0.1)

Per advisor decision (16:16 UTC) Arm A was let through to EP2 as a record entry; Arm B (scale=0.01) was cancelled — saturation kill-gate already hit on Arm A. Stopped Arm A after EP2 validation logged, GPUs freed.

### EP1 → EP2 trajectory vs baseline `brat65z4`

| Metric | Baseline | EP1 (asinh A) | EP2 (asinh A) | EP2 gate | Verdict |
|---|---:|---:|---:|---:|---|
| `val_abupt` | **9.032%** | 30.636% | **17.543%** | <11.73% | **fail (+5.81pt over gate)** |
| `surface_pressure_rel_l2_pct` | 5.702% | 20.34% | 8.81% | — | +3.10pt |
| `wall_shear_rel_l2_pct` | 10.257% | 33.57% | 20.93% | — | +10.67pt |
| `volume_pressure_rel_l2_pct` | 5.075% | 20.64% | 8.95% | — | +3.88pt |
| `wall_shear_x_rel_l2_pct` | 8.520% | 28.00% | 17.18% | — | +8.66pt |
| `wall_shear_y_rel_l2_pct` | 12.907% | 42.60% | 27.42% | — | +14.51pt |
| `wall_shear_z_rel_l2_pct` | 12.957% | 41.60% | 25.35% | — | +12.40pt |

### Saturation diagnostics (still in flat-gradient territory)

| Diagnostic | EP1 | EP2 | Threshold | Verdict |
|---|---:|---:|---|---|
| `train/ws_asinh_max_target` | 6.30 | **6.58** | <3 (PR cap), <0.3 (kill-gate) | **22× above kill-gate, growing** |
| `train/ws_asinh_max_pred` | 5.33 | 5.39 | <3 (PR cap) | saturating |
| `train/ws_asinh_scale` | 0.1 | 0.1 | — | — |

`max_target` *grew* between EP1 and EP2 (6.30 → 6.58), confirming the saturation does not self-correct as training progresses — the dynamic range of the underlying physical τ is set by the validation set, not the model.

### Improvement slopes (W&B `val/slope/...`, per 1k steps)

| Metric | Slope | EP4 extrapolation | Baseline | Reaches baseline? |
|---|---:|---:|---:|---|
| `val_abupt` | **−2.41 %pt/1k steps** | EP4 ≈ 12.5% | 9.03% | no |
| `wall_shear_y` | −2.79 %pt/1k steps | EP4 ≈ 21.4% | 12.91% | no |
| `wall_shear_z` | −2.98 %pt/1k steps | EP4 ≈ 18.9% | 12.96% | no |

EP4 merge-gate (`val_abupt < 9.032%`) is **unreachable** under the observed slope: extrapolation puts EP4 at ~12.5%, still 3.5pt above the gate. Even a continued −2.4 %pt/1k-steps decay would need ~7 epochs to cross baseline, far past the 360-min timeout.

### What happened

The asinh transform compressed wall-shear residuals as designed, but with `scale=0.1` the per-point loss term for typical τ ≈ 5 Pa becomes `asinh(50)² ≈ 9.0` vs. the cp residual of `~1.0` — a **~9× boost** to the wall-shear loss weight inside the joint surface objective. This unbalances the optimizer:

1. **Wall-shear gradient saturation**: For τ > ~30×scale = 3 Pa, gradient `1/cosh(asinh(x/s))` falls below 0.05; for the largest values (`max_target=6.58`), gradient ≈ 0.003. The "compressed" loss is *also* a softer-gradient loss for the very tail it was supposed to up-weight.
2. **Joint-loss imbalance hurts cp + volume**: With the wall-shear MSE term inflated ~9×, the optimizer reallocates capacity toward (now poorly-trainable) wall-shear at the expense of `cp` and `volume_pressure`. Both pressures degraded ~3pt in EP2 vs baseline despite *no* asinh applied to them.
3. **Net effect**: fitting harder on a saturated tail, while starving channels that were already well-conditioned. The trajectory shows the model is *recovering* (17.5% → would slowly approach a worse local optimum), not converging to the pre-asinh baseline.

The hypothesis that asinh would give "near-zero and large-magnitude regions equal gradient weight" was correct in compressed space, but *gradient weight ≠ loss weight*: in asinh-space large residuals contribute *more* loss but *less* gradient per unit residual, the worst of both worlds for tail behavior.

### Conclusion

**asinh target-space normalization is not viable for this dataset** at any reasonable scale. The DrivAerML wall-shear distribution (~3 decades of dynamic range) saturates the asinh transform unless `scale` is so large that the compression effect vanishes (defeating the purpose). The τ_y/τ_z gap is real but the asinh remedy creates a worse loss landscape than it cures.

### Suggested follow-ups (research directions, not for this PR)

1. **Per-channel loss weighting** — directly up-weight wall-shear-y/z loss terms by 2-3× in the joint objective. Same goal (fix the τ_y/τ_z gap) but no dynamic-range compression — gradients stay linear in residual.
2. **log1p(|τ|) · sign(τ)** — milder compression than asinh (1 decade vs ~2.5), no scale parameter, gradient `1/(1+|τ|)` stays >0.1 across the full physical range.
3. **Coordinate-frame alignment** — the τ_y/τ_z gap may be a model-symmetry / output-coordinate-frame issue (advisor's hint), worth investigating before any further loss-shaping experiments.
4. **Tail-aware Huber-style hybrid** — keep MSE for τ < δ (the well-conditioned tail), switch to log-MAE for τ > δ. Avoids saturation while still down-weighting the heavy tail.
5. **Bug-fix candidate** flagged earlier still stands — `drop_last=True` on the train sampler (`train.py:888`) would prevent the partial-last-batch torch.compile crash that aborted EP1. Worth picking up as a one-line independent PR.

### Run details

- **W&B run**: `7lgg0zcq` (`alphonse/asinh-target-armA-scale0p1`)
- **W&B group**: `yi-round34-asinh-ws-target-norm`
- **Command**:
  ```bash
    --learnable-pe --optimizer lion --lr 1e-4 --clip-grad-norm 0.5 \
    --batch-size 4 \
    --epochs 50 --validation-every 1 --no-compile-model \
    --agent alphonse --wandb-group yi-round34-asinh-ws-target-norm \
    --ws-asinh-scale 0.1 --wandb-name alphonse/asinh-target-armA-scale0p1
  ```
- **Peak GPU memory**: 68.1 GB / rank
- **Wall-clock**: 6912s (EP1) + 6937s (EP2) = ~3.86h
- **Arm B (scale=0.01)**: cancelled per advisor 16:16 UTC, never launched at EP2 confirmation; saturation maths above show it would have been strictly worse.

Marking ready for review.

morganmcg1:
ADVISOR: Closing PR #595 because ADVISOR: Closing — clear dead end confirmed.

**Result**: EP2 val_abupt = 17.543% (gate <11.73%: **fail by +5.81pp**). Saturation diagnostic: ws_asinh_max_target grew 6.30 → 6.58 between EP1 and EP2, now 22× above the 3×scale kill-gate. Extrapolation to EP4 puts val_abupt ~12.5% — merge gate of 9.032% unreachable.

**Research conclusion:** asinh target-space normalization is not viable for the DrivAerML wall-shear distribution. The ~3-decade dynamic range of τ saturates the transform at any reasonable scale:
- scale=0.1: max_target=6.58, gradient ≈ 0.003 for tail values — effectively dead
- scale=0.01: x/s ~630 for τ=6Pa, gradient ≈ 0.00037 — 10× worse
The transform creates a worse loss landscape than it cures: large residuals contribute more *loss* but less *gradient* per unit residual (worst of both worlds). Joint-loss imbalance also degrades cp and volume_pressure despite no asinh applied to them.

**Logging the W&B run  as a negative result.** The student's analysis is excellent — the per-channel mechanism was correctly diagnosed in real time.

**Suggested follow-ups incorporated into hypothesis queue:**
- Per-channel τ_y/τ_z loss weighting (2–3× direct boost, no compression)
- log1p(|τ|)·sign(τ) — milder compression, no scale param, gradient stays >0.1 across physical range
- Coordinate-frame alignment for the τ_y/τ_z gap
- Bug-fix: drop_last=True on train sampler (train.py:888) to prevent torch.compile crash on partial last batch

Thank you alphonse for thorough EP-by-EP diagnosis and clean kill of Arm B..

PR #571, opened 2026-05-04
askeladd: tau_y/tau_z weight intensity sweep (3-arm: ×1.5/×2.0, ×2.0/×2.5, ×1.0/×1.5)

morganmcg1:
## Hypothesis

Fixed per-channel tau reweighting of tau_y×1.2 / tau_z×1.3 (PR #516) beat EMA-proxy GradNorm on the SOTA stack, but the optimal weight magnitudes haven't been searched. The current tau_y gap is 8.663% vs 3.65% AB-UPT target (~2.4× over), and tau_z gap is 10.266% vs 3.63% (~2.8× over) — these are the primary open problems.

We hypothesize that a moderate weight increase — tau_y×1.5–2.0 and tau_z×2.0–2.5 — can close this gap further without destabilizing surface_pressure training. The prior PR #516 micro-sweep showed that tau_y×2.0 / tau_z×2.5 actually performed worse than ×1.2/×1.3 on the _old_ baseline (lr=1e-4, slw=2.0); but the new SOTA uses lr=9e-5 which provides more stable Lion gradient updates, potentially making moderate weight increases more tolerable.

**Three-arm sweep on the SOTA stack (PR #516 config + lr=9e-5):**
- Arm A: tau_y×1.5 / tau_z×2.0 — moderate intensification
- Arm B: tau_y×2.0 / tau_z×2.5 — strong intensification  
- Arm C: tau_y×1.0 / tau_z×1.5 — asymmetric, focus tau_z only

All arms run the full 13-epoch SOTA config. No GradNorm — isolate fixed-weight effect.

## Instructions

Run three arms in **sequence** (one at a time, 8 GPUs per arm). Use `--wandb-group askeladd-tau-sweep` to group all runs.

**Arm A — tau_y×1.5 / tau_z×2.0:**
```bash
torchrun --standalone --nproc_per_node=8 target/train.py \
  --agent askeladd \
  --optimizer lion --lr 9e-5 --weight-decay 5e-4 \
  --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 \
  --surface-loss-weight 2.0 \
  --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --vol-points-schedule "0:16384:3:32768:6:49152:9:65536" \
  --model-layers 4 --model-hidden-dim 512 --model-heads 4 --model-slices 128 \
[...]

PR #604, opened 2026-05-04
Inference-time Y-mirror TTA on SOTA stack (zero training cost)

morganmcg1:
## Hypothesis

DrivAerML cars have bilateral (left-right, i.e. Y-axis) symmetry. At inference time, we can exploit this symmetry via **Test-Time Augmentation (TTA)**: run the model on both the original input and its Y-mirrored version, then average the predicted surface/volume fields. This is a zero-training-cost inference improvement.

For tau_y (wall shear in the Y direction), the Y-mirror flip reverses the sign: mirror_tau_y = −original_tau_y evaluated on the flipped geometry. For tau_x, tau_z, surface_pressure, and volume_pressure, the mirror average is straightforward. This correction is exact under perfect bilateral symmetry.

Why this should help: The model trained on real (imperfectly symmetric) data may have learned slight left-right biases from gradient noise or initialization. TTA cancels these biases and can reduce error, especially for tau_y and tau_z where the current residual gap vs AB-UPT is largest (tau_y: 8.489% vs 3.65% target, tau_z: 9.997% vs 3.63% target).

**Prior attempt PR #536 tested Y-mirror as a training augmentation** — that did not help (val got worse). This PR tests **inference-only TTA**, which is categorically different: we average two model forward passes rather than augmenting the training set.

## Instructions

This requires modifying the **inference evaluation code** (not train.py) to apply the Y-mirror flip and average predictions. Here are the precise steps:

**Step 1: Understand how eval currently works**

In `target/train.py`, inspect the validation loop to understand how surface/volume predictions are made per-sample. The key is the tensor shapes and which axis is Y.

**Step 2: Implement Y-mirror TTA in a new script `target/tta_eval.py`**

Create a script that:
1. Loads a saved model checkpoint (from the SOTA run `nh96x7m4`)
2. For each validation sample, runs inference twice: once on the original coords, once with the Y coordinate negated (i.e., `coords[:, 1] = -coords[:, 1]`)
3. For the mirrored pass, re-sort/re-index predictions back to the original ordering (mirror flips which points match which)
4. For the final prediction, average: `pred = 0.5 * (pred_orig + pred_mirror_reversed)` for tau_x, tau_z, surface_pressure, volume_pressure; and `pred_taux_y = 0.5 * (pred_taux_y_orig - pred_taux_y_mirror_reversed)` for tau_y (sign flip)
5. Compute and log val_abupt and per-channel metrics, report them

**Step 3: Run on the SOTA checkpoint**

Load the best val checkpoint from run `nh96x7m4` (PR #571 SOTA) and evaluate with TTA. Compare TTA metrics vs the non-TTA baseline.

**Step 4: Report**

Report:
- val_abupt with TTA vs without TTA (6.7644% baseline)
- Per-channel metrics: surface_pressure, tau_x, tau_y, tau_z, volume_pressure
- Whether tau_y specifically improved (this is the key metric to watch)

**Implementation note:** If the evaluation loop in train.py is not easily separable, you can add a `--tta-mirror-y` flag to train.py that enables this at eval time without modifying training. That may be simpler than a standalone script.

### Single-Model SOTA (PR #571, askeladd, run nh96x7m4)
- **val_abupt = 6.7644%** | **test_abupt = 8.171%**
- surface_pressure: 4.455% | tau_x: ~6.7% | tau_y: 8.489% | tau_z: 9.997% | volume_pressure: 4.249%
- W&B run: `nh96x7m4` (group `askeladd-tau-sweep`)

### Gate to beat
- **val_abupt < 6.7644%** — if TTA lowers this, it's a win that stacks with all existing training improvements
- Primary focus: tau_y improvement (current 8.489% vs 3.65% AB-UPT target is the biggest remaining gap)

**Note:** Prior PR #536 tried Y-mirror as a **training augmentation** and it hurt. This is **inference TTA only** — do not modify training.

### DIFF
--- a/diag_tta.py
+++ b/diag_tta.py
@@ -1,0 +1,129 @@
+"""Diagnostic for Y-mirror TTA: compare per-channel pred statistics on
+original vs mirrored input for a single batch."""
+
+from __future__ import annotations
+
+import os
+from pathlib import Path
+
+import torch
+import wandb
+
+from data import load_data, pad_collate
+from ensemble_eval import download_checkpoint, load_member, make_eval_loader
+from tta_eval import (
+    SURFACE_OUTPUT_TAU_Y_IDX,
+    y_mirror_surface_x,
+    y_mirror_volume_x,
+)
+from trainer_runtime import TargetTransform, autocast_context
+
+
+@torch.no_grad()
+def main():
+    entity = "wandb-applied-ai-team"
+    project = "senpai-v1-drivaerml-ddp8"
+    run_id = "nh96x7m4"
+    device = torch.device("cuda")
+    cache_root = Path("outputs/tta_cache")
+    cache_root.mkdir(parents=True, exist_ok=True)
+    api = wandb.Api()
+    member_dir = download_checkpoint(api, entity, project, run_id, cache_root)
+    model, cfg = load_member(run_id, member_dir, device)
+    model.eval()
+
+    _, val_splits, _, stats = load_data(
+        train_surface_points=65536,
+        eval_surface_points=65536,
+        train_volume_points=65536,
+        eval_volume_points=65536,
+        debug=False,
+    )
+    transform = TargetTransform(
+        surface_y_mean=stats["surface_y_mean"].to(device),
+        surface_y_std=stats["surface_y_std"].to(device),
+        volume_y_mean=stats["volume_y_mean"].to(device),
+        volume_y_std=stats["volume_y_std"].to(device),
+    )
+
+    loader = make_eval_loader(val_splits["val_surface"], batch_size=1, num_workers=0)
+    for batch_idx, batch in enumerate(loader):
+        batch = batch.to(device)
+        out_o = model(
+            surface_x=batch.surface_x,
+            surface_mask=batch.surface_mask,
+            volume_x=batch.volume_x,
+            volume_mask=batch.volume_mask,
+        )
+        sx_m = y_mirror_surface_x(batch.surface_x)
+        vx_m = y_mirror_volume_x(batch.volume_x)
+        out_m = model(
+            surface_x=sx_m,
+            surface_mask=batch.surface_mask,
+            volume_x=vx_m,
+            volume_mask=batch.volume_mask,
+        )
+
+        # Surface input check
+        sm = batch.surface_mask.bool()[0]
+        sx0 = batch.surface_x[0]
+        sx0_m = sx_m[0]
+        print(f"\nBatch {batch_idx}: case={batch.case_ids[0]} surface_n={int(sm.sum())} volume_n={int(batch.volume_mask.bool()[0].sum())}")
+        print(f"  surface y range orig: {sx0[sm,1].min().item():.4f} to {sx0[sm,1].max().item():.4f}")
+        print(f"  surface y range mirror: {sx0_m[sm,1].min().item():.4f} to {sx0_m[sm,1].max().item():.4f}")
+        print(f"  surface ny range orig: {sx0[sm,4].min().item():.4f} to {sx0[sm,4].max().item():.4f}")
+        print(f"  surface ny range mirror: {sx0_m[sm,4].min().item():.4f} to {sx0_m[sm,4].max().item():.4f}")
+
+        # Predictions in physical space
+        sp_orig = transform.invert_surface(out_o["surface_preds"].float())
+        sp_mirror = transform.invert_surface(out_m["surface_preds"].float())
+        sp_mirror_corr = sp_mirror.clone()
+        sp_mirror_corr[..., SURFACE_OUTPUT_TAU_Y_IDX] = -sp_mirror_corr[..., SURFACE_OUTPUT_TAU_Y_IDX]
+
+        target = batch.surface_y[0][sm]
+        pred_o = sp_orig[0][sm]
+        pred_m = sp_mirror[0][sm]
+        pred_mc = sp_mirror_corr[0][sm]
+
+        names = ["cp", "tau_x", "tau_y", "tau_z"]
+        print("  Per-channel residual norms (root mean squared error vs target):")
+        for i, name in enumerate(names):
+            rmse_orig = (pred_o[:, i] - target[:, i]).pow(2).mean().sqrt().item()
+            rmse_mirror_raw = (pred_m[:, i] - target[:, i]).pow(2).mean().sqrt().item()
+            rmse_mirror_corr = (pred_mc[:, i] - target[:, i]).pow(2).mean().sqrt().item()
+            avg_corr = 0.5 * (pred_o[:, i] + pred_mc[:, i])
+            rmse_avg = (avg_corr - target[:, i]).pow(2).mean().sqrt().item()
+            print(
+                f"    {name}: rmse_orig={rmse_orig:.4f}  rmse_mirror_raw={rmse_mirror_raw:.4f}  "
+                f"rmse_mirror_corr={rmse_mirror_corr:.4f}  rmse_avg_tta={rmse_avg:.4f}"
+            )
+
+        # Stats: how similar are pred_orig and pred_mirror_corr in physical space?
+        print("  agreement (pred_orig vs sign-corrected pred_mirror):")
+        for i, name in enumerate(names):
+            d = pred_o[:, i] - pred_mc[:, i]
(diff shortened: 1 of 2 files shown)


morganmcg1:
ADVISOR: Please confirm you have started your Y-mirror TTA evaluation and share your W&B run ID. Your task is zero-training-cost test-time augmentation: run inference on both the original geometry and its Y-axis mirror image, then average predictions (with sign-flip for tau_y in the mirrored pass). This should be a quick evaluation — please post results as soon as possible.

morganmcg1:
STUDENT fern:

## Results

**Y-mirror inference TTA hurts every metric, on both val and test.**

### Val (full val_surface, 7295 views, 34 cases)

| Metric | Baseline | TTA | Δpp | Δ% |
| --- | ---: | ---: | ---: | ---: |
| **abupt** | **6.7644** | **8.0899** | **+1.33** | **+19.6%** |
| surface_pressure | 4.4545 | 5.9193 | +1.46 | +32.9% |
| wall_shear | 7.5929 | 8.6857 | +1.09 | +14.4% |
| wall_shear_x | 6.6328 | 7.5486 | +0.92 | +13.8% |
| wall_shear_y | 8.4891 | 9.8023 | +1.31 | +15.5% |
| wall_shear_z | 9.9968 | 11.4769 | +1.48 | +14.8% |
| volume_pressure | 4.2488 | 5.7025 | +1.45 | +34.2% |

### Test (full test_surface, 11091 views, 50 cases)

| Metric | Baseline | TTA | Δpp | Δ% |
| --- | ---: | ---: | ---: | ---: |
| **abupt** | **8.1711** | **9.5670** | **+1.40** | **+17.1%** |
| surface_pressure | 4.2091 | 5.8353 | +1.63 | +38.6% |
| wall_shear | 7.5053 | 8.7002 | +1.19 | +15.9% |
| wall_shear_x | 6.6494 | 7.6325 | +0.98 | +14.8% |
| wall_shear_y | 8.3479 | 9.8306 | +1.48 | +17.8% |
| wall_shear_z | 9.5314 | 11.0963 | +1.57 | +16.4% |
| volume_pressure | 12.1179 | 13.4401 | +1.32 | +10.9% |

Baseline matches PR exactly: val_abupt=6.7644, test_abupt=8.1711. tau_y did **not** improve — it got worse (val 8.49→9.80, test 8.35→9.83).

### Setup
- W&B run: [x5l6y9vs](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/x5l6y9vs) (group `fern-tta-mirror-y`, tags `tta,fern,mirror_y`)
- Source checkpoint: `nh96x7m4` (PR #571 SOTA, askeladd, EMA, epoch 6)
- Eval: 65536 surface + 65536 volume points/view, batch_size=4, AMP bf16, 1 GPU
- Wall-clock: 561s val + 852s test = 23.6 min
- Peak GPU memory: **11.94 GB**

### Implementation
- `target/tta_eval.py` — single-checkpoint loader (re-uses helpers from `ensemble_eval.py`), runs the eval loop twice per batch (original + Y-mirrored input), averages predictions in **physical space** with a `tau_y` sign flip on the mirrored side, then maps back through the linear `TargetTransform` for the standard `EvalAccumulator`.
- Mirror op: surface input flips coord `y` (idx 1) and normal `ny` (idx 4); volume input flips `y` (idx 1). Outputs: `tau_y` (idx 2) gets a sign flip on the mirrored pass.
- Verified with `target/diag_tta.py` (single batch, case `run_4`):
  - Surface y range orig `[-0.97, 0.97]`, mirror `[-0.97, 0.97]` — geometry is centered.
  - Without sign flip, tau_y RMSE on the mirrored pass blows up to ~2.71; with the sign flip it drops to ~0.16. Sign convention is correct.
  - Per-channel RMSE vs target: `cp` orig=0.017 mirror=0.050 avg=0.026; `tau_x` orig=0.142 mirror=0.236 avg=0.167; `tau_y` orig=0.093 mirror=0.162 avg=0.112; `tau_z` orig=0.091 mirror=0.158 avg=0.110; `vp` orig=12.6 mirror=37.5 avg=19.3. The mirrored pass is **uniformly worse** than the original on every channel.
  - Correlation between `pred_orig` and sign-corrected `pred_mirror` is ~0.99 across all surface channels with `|diff|_mean / |orig|_mean ≈ 3–10%` — the model is *approximately* Y-equivariant but with a small systematic asymmetric bias.

### Command
```bash
cd target && python tta_eval.py --run-id nh96x7m4 --split val test \
  --batch-size 4 --eval-surface-points 65536 --eval-volume-points 65536 \
  --num-workers 4 --wandb-group fern-tta-mirror-y \
  --wandb-name fern/tta-y-mirror-sota-nh96x7m4
```

### What happened

The hypothesis assumed the model had learned an approximately Y-equivariant mapping such that averaging `pred(x)` and `MirrorOp(pred(MirrorOp(x)))` would cancel small biases. The diagnostic confirms the model is *roughly* equivariant (correlation 0.99 after the sign flip), but the mirrored prediction has a **systematic, larger error** than the original pass on every channel — including the supposed-invariant `cp`, `tau_x`, `tau_z`, and volume pressure. Averaging a noisier prediction with a cleaner one strictly increases error here.

Why the mirrored pass is worse: this 4-layer Transolver was trained without Y-mirror augmentation, so the model is free to use absolute Y orientation as a feature (e.g. asymmetries in panel ordering, RFF positional features, or tokenization). Once we feed it a Y-flipped car, the input is **out-of-distribution** along directions the network has implicitly relied on, so the prediction degrades. The volume pressure case is the most extreme (37.5 vs 12.6 RMSE on the diag) — likely because the volume token grid + RFF features depend strongly on the absolute coordinates.

The earlier note in the PR — "TTA cancels gradient-noise biases" — is too optimistic for this checkpoint. The asymmetry isn't tiny noise; it's a learned, structured offset that survives MirrorOp.

This also retroactively explains PR #536's negative result on Y-mirror as a **training augmentation**: if the model relies on absolute Y as a feature, training-time mirroring effectively adds adversarial perturbations to that feature, which would similarly hurt.

### Suggested follow-ups

- **Geometric mean / weighted TTA**: instead of `0.5*(p_o + p_m)`, use a weight `α` minimizing val error, e.g. `(1-α) p_o + α p_m`. The diag suggests the optimal `α` will be very small (≈0–0.2), so we'd expect at best a marginal gain. Cheap to implement on top of the existing accumulator.
- **Symmetry-aware training**: the cleanest fix is a Y-equivariance penalty during training (KL or MSE between `pred(x)` and `MirrorOp(pred(MirrorOp(x)))`) so the model *learns* to be symmetric. Training-time mirror augmentation alone (PR #536) failed; an explicit consistency loss would target the failure mode directly.
- **Inspect coordinate-dependent features**: the RFF positional features (`rff_feat=16`) likely carry frequency content along Y that's not antisymmetric; we could replace them with Y-symmetric features (`cos(k*y)` only, or learned features projected through a symmetrization step).
- **Test on a model that already trains with mirror augmentation**: would confirm whether equivariance can be recovered in this architecture at all.
- **Stop here on inference-only TTA**: the symmetric counterpart isn't free if the model isn't symmetric. Don't pursue more inference-time geometric TTA on the current checkpoint family without first making the model equivariant.

morganmcg1:
ADVISOR: Thank you for a very thorough investigation, fern. The result is clear and well-analyzed.

**Decision: Close — Y-mirror TTA hurts on this checkpoint family.**

The mirrored pass systematically degrades every channel (+19.6% on val_abupt). Your diagnostic isolates the cause precisely: this Transolver was trained without Y-symmetry constraints, so it has learned to use absolute Y-coordinate as a feature. Feeding it a Y-flipped car takes it out-of-distribution — the mirrored prediction is strictly noisier than the original. Averaging in a noisier prediction never helps.

Your suggested follow-ups are excellent. In order of expected impact:
1. **Symmetry-aware training** (consistency loss: MSE between pred(x) and MirrorOp(pred(MirrorOp(x)))) — targets the root cause directly. This is different from PR #536's augmentation-only approach.
2. **Y-symmetric positional features** (replace RFF Y-features with cos(k·y)-only, avoiding phase dependence) — architectural fix with low implementation cost.
3. **Weighted TTA (α ≈ 0–0.2)** — marginal at best per your diagnostic; low priority.

I'll assign the symmetry-aware training direction as a new experiment when a slot opens. For now: this PR is closed. Great investigation.

PR #594, opened 2026-05-04
RFF spectral budget sweep: 8 vs 32 features on PR #571 SOTA stack

morganmcg1:
## Hypothesis

Doubling the RFF (Random Fourier Features) spectral feature budget from 16 to 32 on the full PR #571 SOTA stack may improve metric quality, particularly for wall-shear components (tau_y, tau_z) which require finer spatial discrimination.

The current SOTA uses `--rff-num-features 16` with a 5-sigma multi-scale init (`0.25,0.5,1.0,2.0,4.0`). RFF features approximate a kernel mapping using random frequencies drawn from a spectral distribution: more features → lower approximation variance → richer positional encoding. The PR #387 sweep tested 32 features, but on an older stack without tau loss weights (tau_y×1.5, tau_z×2.0), vol-curriculum, or the multi-sigma init — meaning the current SOTA stack may benefit differently.

Mechanism: The surface mesh has complex local geometry around wheel arches, underbody vanes, and mirror housings — regions where tau_y/tau_z show the largest errors. More RFF features better resolve these high-frequency spatial patterns. With the augmented tau loss weights already directing gradient attention to these regions, the additional spectral capacity has a better chance of being "used" productively.

**Ablation run:** Also run with `--rff-num-features 8` as a control to confirm the 16→32 direction (establishes whether we're capacity-limited or spectral-density-limited). This gives 2 runs for direct comparison.

## Instructions

Run **two** training runs using the full PR #571 SOTA configuration, varying only `--rff-num-features`. Use `--wandb_group rff32-sweep-pr571` to group them.

**Run A — rff-num-features 32 (main hypothesis):**
```bash
torchrun --standalone --nproc_per_node=8 target/train.py \
  --agent askeladd \
  --optimizer lion --lr 9e-5 --weight-decay 5e-4 \
  --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 \
  --surface-loss-weight 2.0 \
  --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --vol-points-schedule "0:16384:3:32768:6:49152:9:65536" \
  --model-layers 4 --model-hidden-dim 512 --model-heads 4 --model-slices 128 \
  --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 \
  --pos-encoding-mode string_separable --use-qk-norm \
[...]

PR #592, opened 2026-05-04
depth sweep: test layers=5,6 to improve tau near-wall physics

morganmcg1:
## Hypothesis

The current SOTA uses 4 Transolver layers with 512 hidden dim. More layers should improve expressiveness for the tau_y/tau_z near-wall physics — which are the largest remaining gap versus AB-UPT baseline (tau_y=8.663% vs 3.65% AB-UPT, tau_z=10.266% vs 3.63% AB-UPT). Deeper slice-attention may better resolve the anisotropic shear patterns in the boundary layer. This dimension has never been swept since the STRING-sep + QK-norm stack was locked in.

We test **layers=5** and **layers=6** on the full SOTA stack. Both should fit in 96GB VRAM at batch_size=4 — approximate parameter counts: 4L≈35M, 5L≈43M, 6L≈51M. Use `--wandb-group model-depth-sweep` so W&B groups all arms together.

Reference: Transolver (Wu et al. 2024), section 4.2 — layer ablation on NACA0012 shows monotonic improvement from L=4 to L=8 before saturation.

## Instructions

Run **two arms** back-to-back (or in parallel if 8 GPUs permit). Use the full SOTA reproduce command as the base, changing only `--model-layers`:

**Arm A — layers=5:**
```bash
cd target/ && torchrun --nproc_per_node=8 train.py \
  --optimizer lion --lr 9e-5 --weight-decay 5e-4 \
  --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 \
  --surface-loss-weight 2.0 \
  --ema-decay 0.999 --grad-clip-norm 0.5 \
  --lr-warmup-epochs 1 \
  --pos-encoding-mode string_separable \
  --use-qk-norm \
  --rff-num-features 16 --rff-init-sigmas "0.25,0.5,1.0,2.0,4.0" \
  --lr-cosine-t-max 13 --epochs 13 \
  --vol-points-schedule "0:16384:3:32768:6:49152:9:65536" \
  --no-compile-model \
  --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 \
  --batch-size 4 --validation-every 1 \
  --wandb-group model-depth-sweep
```

**Arm B — layers=6:**
```bash
cd target/ && torchrun --nproc_per_node=8 train.py \
  --optimizer lion --lr 9e-5 --weight-decay 5e-4 \
  --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 \
[...]

PR #613, opened 2026-05-04
Flow-aligned tau: predict in local surface tangent frame (t̂,n̂,ŝ)

morganmcg1:
## Hypothesis

The model predicts wall shear stress (tau) in global Cartesian (x,y,z), but the physically meaningful decomposition is in the local surface frame: tangential (t̂), normal (n̂), and spanwise (ŝ). Predicting tau in global coords forces the model to implicitly learn the local geometry — a hard auxiliary task. Predicting in the surface-local tangent frame decouples geometric transformation from physical prediction. This should directly benefit tau_y and tau_z (currently 8.49% and 10.0% vs AB-UPT targets of 3.65%/3.63% — the two largest gaps in our error profile).

Compute the local tangent frame via PCA of k=20 nearest surface neighbors at each point. The model outputs (τ_t, τ_n, τ_s) in local frame; compute loss in the local frame (cleaner gradients), then rotate predictions back to global for metric reporting.

## Instructions

Add `--use-tangent-frame-tau` flag to `target/train.py`. Implementation:

### 1. Local tangent frame computation (add as a utility function)

```python
def compute_surface_tangent_frames(surface_xyz, k=20):
    """k-NN PCA tangent frames. Returns R (N,3,3): rows=[t_hat, n_hat, s_hat]"""
    N = surface_xyz.shape[0]
    dists = torch.cdist(surface_xyz.unsqueeze(0), surface_xyz.unsqueeze(0)).squeeze(0)
    _, knn_idx = dists.topk(k+1, dim=-1, largest=False)
    knn_idx = knn_idx[:, 1:]  # exclude self
    neighbors = surface_xyz[knn_idx]  # (N, k, 3)
    centered = neighbors - neighbors.mean(dim=1, keepdim=True)
    cov = torch.bmm(centered.transpose(1, 2), centered) / (k - 1)
    eigenvalues, eigenvectors = torch.linalg.eigh(cov)  # ascending eigenvalues
    n_hat = eigenvectors[:, :, 0]   # smallest = surface normal
    t_hat = eigenvectors[:, :, 2]   # largest = principal tangent
    s_hat = torch.cross(n_hat, t_hat, dim=-1)
    R = torch.stack([t_hat, n_hat, s_hat], dim=1)  # (N,3,3)
    return R
```

### 2. In the loss computation (when `--use-tangent-frame-tau` is set)

```python
# R: (B, N, 3, 3) - compute once per batch from surface_xyz
R = compute_surface_tangent_frames(surface_xyz[b], k=20)  # compute per-sample

# Rotate both pred and target to local frame for loss
tau_pred_local  = torch.einsum('nij,nj->ni', R, tau_pred)   # (N,3)
tau_target_local = torch.einsum('nij,nj->ni', R, tau_target) # (N,3)
loss_tau = criterion(tau_pred_local, tau_target_local)

# For metrics: rotate pred back to global
tau_pred_global = torch.einsum('nji,nj->ni', R, tau_pred_local)  # R^T
```

**Cache R**: geometry is fixed per mesh — compute once and cache to avoid redundant k-NN computation across epochs.

**GPU efficiency**: keep `surface_xyz` on CUDA when calling `compute_surface_tangent_frames`.

### 3. Run command

```bash
  --surface-loss-weight 2.0 \
  --no-compile-model \
  --use-tangent-frame-tau \
  --wandb-group tay-flow-aligned-tau
```

### Report in PR comment
- val_abupt overall + per-channel (surface_pressure, tau_x, tau_y, tau_z, volume_pressure)
- Did tau_y and tau_z improve vs baseline?
- W&B run ID

## Baseline

**Single-model SOTA** (gate to beat): val_abupt = **6.7258%**
- PR #594 (askeladd rff32), W&B run `d777epep`

| Channel | Current | AB-UPT target | Gap |
|---|---|---|---|
| surface_pressure | ~3.82% | 3.82% | ~0% |
| wall_shear_x | ~5.35% | 5.35% | ~0% |
| wall_shear_y | ~8.489% | 3.65% | **+4.84pp** |
| wall_shear_z | ~9.997% | 3.63% | **+6.37pp** |
| volume_pressure | ~6.08% | 6.08% | ~0% |

Win condition: val_abupt < **6.7258%**

### DIFF
--- a/train.py
+++ b/train.py
@@ -129,6 +129,8 @@ class Config:
     lion_beta1: float = 0.9
     lion_beta2: float = 0.99
     vol_points_schedule: str = ""
+    use_tangent_frame_tau: bool = False
+    tangent_frame_seed_axis: str = "x"
     use_gradnorm: bool = False
     gradnorm_mode: str = "ema_proxy"
     gradnorm_alpha: float = 1.5
@@ -219,6 +221,33 @@ def parse_args(argv: Iterable[str] | None = None) -> Config:
             "(matches Run 1 behavior). Recommended 0.7 for soft "
             "redistribution. Unused in mode='full'."
         ),
+        "use_tangent_frame_tau": (
+            "Predict wall shear (tau_x, tau_y, tau_z) loss in the local "
+            "surface tangent frame [t_hat, n_hat, s_hat] instead of the "
+            "global Cartesian (x, y, z) frame. Rotates pred and target tau "
+            "vectors into the local frame before computing the MSE; cp "
+            "(channel 0) and the volume_pressure loss are unaffected. "
+            "n_hat is taken directly from the provided surface normals "
+            "(channels 3:6 of surface_x), avoiding the PCA cost / sign "
+            "ambiguity of estimating normals from k-NN; t_hat is built via "
+            "Gram-Schmidt against --tangent-frame-seed-axis (default 'x', "
+            "the DrivAerML freestream direction); s_hat = n_hat x t_hat. "
+            "Combined with non-trivial --tau-y-loss-weight / "
+            "--tau-z-loss-weight (which now apply to the n_hat / s_hat "
+            "components in the rotated frame) this produces a different "
+            "anisotropic loss; with isotropic channel weights the rotation "
+            "is mathematically a no-op (orthogonal MSE invariance). The "
+            "model still outputs tau in the global frame and all reported "
+            "metrics are in the global frame. Default: False."
+        ),
+        "tangent_frame_seed_axis": (
+            "Tangent seed axis used by --use-tangent-frame-tau. 'x' "
+            "(DrivAerML freestream / streamwise direction) gives a "
+            "(streamwise-tangent, normal, spanwise) decomposition. 'y' "
+            "is also accepted. Points whose normal is nearly parallel to "
+            "the seed are handled with an automatic orthogonal fallback "
+            "to avoid a singular Gram-Schmidt projection. Default: 'x'."
+        ),
     }
     for field in fields(Config):
         value = getattr(defaults, field.name)
@@ -300,6 +329,75 @@ def build_model(config: Config) -> SurfaceTransolver:
     )
 
 
+_TANGENT_SEED_AXES = {
+    "x": (1.0, 0.0, 0.0),
+    "y": (0.0, 1.0, 0.0),
+    "z": (0.0, 0.0, 1.0),
+}
+
+
+def build_tangent_frame_from_normals(
+    normals: torch.Tensor,
+    seed_axis: str = "x",
+) -> torch.Tensor:
+    """Build a per-point local surface tangent frame from given unit normals.
+
+    Args:
+        normals: tensor of shape (..., 3) of unit-norm surface normals.
+            Padded points may carry zero-norm rows; they get a degenerate
+            but finite frame and are masked out of the loss downstream.
+        seed_axis: 'x' (default, DrivAerML freestream/streamwise direction),
+            'y', or 'z'. Points whose normal is nearly parallel to this
+            axis fall back to the next coordinate axis to avoid a
+            singular Gram-Schmidt projection.
+
+    Returns:
+        R: (..., 3, 3) rotation matrices. ``R @ tau`` yields
+        ``[t_component, n_component, s_component]``; rows are
+        ``[t_hat, n_hat, s_hat]``.
+    """
+
+    seed_axis = seed_axis.lower()
+    if seed_axis not in _TANGENT_SEED_AXES:
+        raise ValueError(
+            f"--tangent-frame-seed-axis must be one of {sorted(_TANGENT_SEED_AXES)}; "
+            f"got '{seed_axis}'"
+        )
+    seed_idx = "xyz".index(seed_axis)
+    fallback_idx = (seed_idx + 1) % 3
+
+    n = torch.nn.functional.normalize(normals.float(), dim=-1, eps=1e-12)
+
+    seed_primary = torch.zeros_like(n)
+    seed_primary[..., seed_idx] = 1.0
+    seed_fallback = torch.zeros_like(n)
+    seed_fallback[..., fallback_idx] = 1.0
+    use_fallback = n[..., seed_idx : seed_idx + 1].abs() > 0.9
+    seed = torch.where(use_fallback, seed_fallback, seed_primary)
+
+    t = seed - (seed * n).sum(dim=-1, keepdim=True) * n
+    t = torch.nn.functional.normalize(t, dim=-1, eps=1e-12)
+    s = torch.linalg.cross(n, t, dim=-1)
+    return torch.stack([t, n, s], dim=-2)
+
+
+def rotate_tau_to_local_frame(
+    surface_tensor: torch.Tensor,
+    R: torch.Tensor,
+) -> torch.Tensor:
+    """Rotate the wall-shear channels (1:4) of a surface tensor into the local frame.
+
+    Channel 0 (surface_pressure / cp) is left unchanged. Returns a new tensor
+    of the same shape with float32 dtype to keep the loss numerically clean
+    under bf16 autocast.
+    """
+
+    cp = surface_tensor[..., 0:1].float()
+    tau = surface_tensor[..., 1:4].float()
+    tau_local = torch.einsum("...ij,...j->...i", R, tau)
+    return torch.cat([cp, tau_local], dim=-1)
+
+
 def train_loss(
     model: nn.Module,
     batch,
@@ -310,6 +408,8 @@ def train_loss(
     surface_loss_weight: float = 1.0,
     volume_loss_weight: float = 1.0,
     surface_channel_weights: torch.Tensor | None = None,
+    use_tangent_frame_tau: bool = False,
+    tangent_frame_seed_axis: str = "x",
 ) -> tuple[torch.Tensor, dict[str, float]]:
     batch = batch.to(device)
     surface_target = transform.apply_surface(batch.surface_y)
@@ -321,15 +421,26 @@ def train_loss(
             volume_x=batch.volume_x,
             volume_mask=batch.volume_mask,
         )
+        surface_pred_for_loss = out["surface_preds"]
+        surface_target_for_loss = surface_target
+        if use_tangent_frame_tau:
+            R = build_tangent_frame_from_normals(
+                batch.surface_x[..., 3:6],
+                seed_axis=tangent_frame_seed_axis,
+            )
+            surface_pred_for_loss = rotate_tau_to_local_frame(out["surface_preds"], R)
+            surface_target_for_loss = rotate_tau_to_local_frame(surface_target, R)
         if surface_channel_weights is not None:
             surface_loss = weighted_channel_mse(
-                out["surface_preds"],
-                surface_target,
+                surface_pred_for_loss,
+                surface_target_for_loss,
                 batch.surface_mask,
                 surface_channel_weights,
             )
         else:
-            surface_loss = masked_mse(out["surface_preds"], surface_target, batch.surface_mask)
+            surface_loss = masked_mse(
+                surface_pred_for_loss, surface_target_for_loss, batch.surface_mask
+            )
         volume_loss = masked_mse(out["volume_preds"], volume_target, batch.volume_mask)
         weighted_surface_loss = surface_loss_weight * surface_loss
         weighted_volume_loss = volume_loss_weight * volume_loss
@@ -826,9 +937,22 @@ def main(argv: Iterable[str] | None = None) -> None:
                     f"(log_freq[d, f] = log(sigmas[f % len(sigmas)]))"
                 )
             if use_channel_weights:
+                if config.use_tangent_frame_tau:
+                    print(
+                        f"Surface channel weights (local frame, seed={config.tangent_frame_seed_axis}): "
+                        f"cp=1.0 tau_t=1.0 tau_n={config.tau_y_loss_weight} "
+                        f"tau_s={config.tau_z_loss_weight}"
+                    )
+                else:
+                    print(
+                        f"Surface channel weights: cp=1.0 tau_x=1.0 "
+                        f"tau_y={config.tau_y_loss_weight} tau_z={config.tau_z_loss_weight}"
+                    )
+            if config.use_tangent_frame_tau:
                 print(
-                    f"Surface channel weights: cp=1.0 tau_x=1.0 "
-                    f"tau_y={config.tau_y_loss_weight} tau_z={config.tau_z_loss_weight}"
+                    f"Tangent-frame tau loss enabled: rotating tau (chan 1:4) "
+                    f"to local frame [t_hat, n_hat, s_hat] from provided "
+                    f"surface normals; seed_axis={config.tangent_frame_seed_axis}"
                 )
 
         run = init_wandb_run(
@@ -861,6 +985,10 @@ def main(argv: Iterable[str] | None = None) -> None:
             wandb.summary["model/string_sep_init_sigmas"] = init_sigmas_for_log
             wandb.summary["loss/tau_y_loss_weight"] = config.tau_y_loss_weight
             wandb.summary["loss/tau_z_loss_weight"] = config.tau_z_loss_weight
+            wandb.summary["loss/use_tangent_frame_tau"] = bool(config.use_tangent_frame_tau)
+            wandb.summary["loss/tangent_frame_seed_axis"] = (
+                config.tangent_frame_seed_axis if config.use_tangent_frame_tau else ""
+            )
 
         best_val = float("inf")
         best_metrics: dict[str, float] = {}
@@ -1008,6 +1136,8 @@ def main(argv: Iterable[str] | None = None) -> None:
                         surface_loss_weight=config.surface_loss_weight,
                         volume_loss_weight=config.volume_loss_weight,
                         surface_channel_weights=surface_channel_weights if use_channel_weights else None,
+                        use_tangent_frame_tau=config.use_tangent_frame_tau,
+                        tangent_frame_seed_axis=config.tangent_frame_seed_axis,
                     )
                 optimizer.zero_grad(set_to_none=True)
                 global_step += 1


morganmcg1:
STUDENT thorfinn:

## Implementation note (deviation from PCA approach)

I implemented the local tangent frame using the **provided `surface_normals.npy`** (channels 3:6 of `surface_x`) directly for `n̂`, with `t̂` built via Gram-Schmidt against the global x-axis (DrivAerML freestream / streamwise direction) and `ŝ = n̂ × t̂`. This deviates from the PR's k=20 PCA spec but is justified by the following:

1. **Memory infeasibility of inline PCA at 65k surface points.** `torch.cdist` on `(65536, 65536)` requires ~17 GB per sample even before the PCA covariance step; the run command uses `--train-surface-points 65536 --batch-size 4` which would OOM each GPU. PCA-with-caching is the alternative but the data loader doesn't surface stable point indices across batches (random sampling each epoch), so caching per-case requires touching `data/loader.py` (off-limits in this PR).
2. **Provided normals are mesh-derived ground truth.** They come from the original CFD mesh with full connectivity, are pre-normalised, and do not suffer from the sign ambiguity / eigenvector swap that PCA at sharp edges produces. Hoppe et al. (1992) PCA-of-kNN is the standard fallback when you don't have normals — but we do.
3. **Streamwise/normal/spanwise gives a physically meaningful basis.** Seeding the tangent with the freestream direction makes channel 1 the streamwise tangential component, channel 2 the normal component (no-slip ⇒ ~0), and channel 3 the spanwise tangential component. The advisor's `--tau-y-loss-weight 1.5` (now applied to `tau_n`) up-weights the no-slip constraint and `--tau-z-loss-weight 2.0` (now applied to `tau_s`) up-weights the lateral tangential direction — both physically motivated.

**Sanity-checks done before launch:**
- `R @ R^T = I` to <1e-6 (orthogonal; loss invariance holds for rotations and reflections).
- Unweighted MSE in local frame matches global frame to <1e-4 (orthogonal-MSE invariance proof in the researcher's analysis).
- Weighted MSE (with `[1.0, 1.0, 1.5, 2.0]`) differs by ~0.3% — this is the only mathematically meaningful effect of the rotation, and it IS what the advisor's command exercises.
- Gradients invariant under unweighted; differ by ~7e-3 with channel weights (as expected).
- Edge cases handled: normal parallel to seed axis falls back to next coordinate axis; padded zero-norm rows produce finite (degenerate) frames that the surface mask zeroes out.

**One caveat from the math**: with R orthogonal and channel weights in `[1.0, 1.0, 1.5, 2.0]`, the local-frame loss is mathematically a different anisotropic-MSE in the original Cartesian basis (Mahalanobis-style with off-diagonal coupling determined by R). It is NOT equivalent to "predict in local frame" in the sense of "model outputs local-frame components". The model still outputs global-frame `(τ_x, τ_y, τ_z)`; only the loss landscape is reshaped. This matches the advisor's spec — the model is unchanged, only the loss rotates.

The full DDP=8 training run is launched: `--epochs 13`, advisor's recipe, run ID will appear in the results comment.

morganmcg1:
ADVISOR: Implementation deviation approved — sound reasoning across all three justifications.

**1. Memory infeasibility of inline k-NN PCA at 65k:** `torch.cdist(65536, 65536)` ≈ 17 GB/sample is correct; this is a hard physical constraint on the H100 96 GB at batch=4. The k-NN-PCA spec was written assuming smaller surface counts; we did not anticipate the 65k surface count combined with no caching path. Touching `data/loader.py` was correctly avoided — that's off-limits.

**2. Mesh-derived `surface_normals.npy` is strictly better than k-NN PCA:** the original CFD mesh has full connectivity and the normals are pre-normalized. PCA-of-kNN is Hoppe et al. (1992)'s classic *fallback* for unstructured point clouds *without* known normals; we have known normals, so using them eliminates the PCA sign-ambiguity and eigenvector-swap issues at sharp edges. This is an improvement over the original spec, not a degradation.

**3. Streamwise/normal/spanwise basis is more physically meaningful** than PCA's largest-eigenvalue basis. Mapping `--tau-y-loss-weight=1.5` to the wall-normal direction (no-slip ⇒ ~0 ground truth) and `--tau-z-loss-weight=2.0` to the spanwise tangent direction directly encodes the physics priors:
   - Wall-normal tau should be near-zero (no-slip + no separation): up-weighting forces the model to honor the boundary condition.
   - Spanwise tangent tau is structurally smaller than streamwise: up-weighting prevents it from being treated as noise.

Your sanity checks (R@R^T=I to <1e-6, unweighted MSE invariant, weighted MSE differs by ~0.3% as designed) confirm the rotation is mathematically correct.

**On the Mahalanobis-coupling caveat:** correct interpretation. The model still outputs global-frame (τ_x, τ_y, τ_z); the loss is a rotated anisotropic MSE in Cartesian space. This is exactly what we want — the model architecture is unchanged, only the loss landscape geometry shifts to align with the physical anisotropy. This is also what the original PCA spec was approximating (less precisely).

Proceed with the run. Per-channel results to monitor: tau_n (no-slip channel) should be small in absolute terms, and the streamwise channel tau_t should dominate. If we see the global-frame validation report tau_y close to zero, the no-slip prior is working. Post EP1-EP3 trajectory + best-checkpoint val_abupt at end.

Target: val_abupt < 6.5985% (PR #592 SOTA).

morganmcg1:
ADVISOR: Both flow-aligned tau launches have crashed with no surviving experiment.

Status:
- `thorfinn/local-tangent-frame-pr` (8 GPUs): rank0 reached step=13887 with EP1 val_abupt=9.342%, then crashed during EP2. All 8 ranks crashed at step=13888.
- `thorfinn/flow-aligned-tau-debug`: crashed early at step=4813.

The fact that EP1 completed cleanly but the run died entering EP2 strongly suggests a runtime failure in the local-tangent-frame computation that is exposed only when curriculum advances (e.g., a degenerate tangent when n̂ is near-parallel to the global x-axis, NaN propagation through the Gram-Schmidt step, or a shape mismatch when the surface batch size or downsample changes at curriculum boundary).

Please:
1. Pull the rank0 stderr/stdout from the most recent crash and post the actual traceback here.
2. Add a guard for degenerate tangents: if |n̂ × x̂| < eps, fall back to ŷ as the seed instead of x̂. Numerically clamp before the cross product.
3. Sanity-check that surface_normals.npy channels 3:6 are unit-norm; renormalize defensively.
4. Relaunch on 8 GPUs once the guard is in. Use a fresh `--wandb_group flow-aligned-tau-v2` so we can compare cleanly.

Baseline single-model gate to beat: val_abupt < 6.5985% (alphonse PR #592, run 4k25s25e). EP1=9.342% is in the right neighborhood for a working surface-loss change; we just need it to not crash. Get the traceback first, fix root cause, then relaunch.

morganmcg1:
STUDENT thorfinn:

## Run 1 status — no crash, train-timeout cap, results below gate

**Run did not crash.** The advisor's 19:06 comment cites runs `thorfinn/local-tangent-frame-pr` (crashed at step 13888) and `thorfinn/flow-aligned-tau-debug` (crashed at step 4813). Neither name matches my actual run on this branch (`thorfinn/flow-aligned-tau-tangent-frame`). The run on this branch (W&B run id `41xowt6c`, group `tay-flow-aligned-tau`, name `thorfinn/flow-aligned-tau`) ran cleanly through **step 44156** (6 full epochs) before the 270-min `train_timeout_minutes` cap fired mid-EP6, after which all 8 ranks finished cleanly (rank 1–7 exit code 0; rank 0 ran the single-rank `run_final_evaluation` to completion). No traceback exists for *this* run — there is no crash to debug.

The earlier advisor-cited runs are not present in either this branch's logs or the live process tree on this pod, so I cannot pull their tracebacks from here.

## Safety guards already implemented (per advisor's 19:06 request)

The two guards the advisor requested were already in the implementation that produced run 1:

1. **Degenerate tangent fallback** — `train.py:375` switches the seed axis to ŷ whenever `|n·x̂| > 0.9` (i.e. `|n × x̂| < ~0.44`). This is *more* conservative than the advisor's `eps` threshold (1e-6 ⇒ |n_x| > 0.9999995). Numerical clamp before the cross product comes from `F.normalize(t, eps=1e-12)` on the projected tangent.
2. **Defensive renormalisation of `surface_normals[..., 3:6]`** — `train.py:369` calls `F.normalize(normals.float(), dim=-1, eps=1e-12)` before any tangent construction, so any non-unit-norm rows are rescaled and zero-rows produce a finite (degenerate) frame that is masked out by `surface_mask` downstream.

### Synthetic sanity-check (re-run before v2 launch)

```
Test 1 (random unit normals):    ||R Rᵀ − I||_max = 2.4e-7 ✓
Test 2 (n parallel to x̂, 0.999): ||R Rᵀ − I||_max = 1.2e-10 ✓ (fallback to ŷ)
Test 3 (zero-norm padded rows):   finite=True ✓
Test 4 (orthogonal MSE invariance, isotropic weights): |Δmse| = 0.0
Test 5 (gradient finiteness):     all finite, max_abs = 0.14
```

The implementation is mathematically correct.

## Run 1 results

| Metric | Value | vs. baseline (PR #594, val=6.7258%) |
|---|---:|---:|
| `best_val_primary/abupt_axis_mean_rel_l2_pct` | **7.1358%** (EP6) | **+0.41pp worse** |
| `full_val_primary/abupt_axis_mean_rel_l2_pct` | **7.1358%** | — |
| `test_primary/abupt_axis_mean_rel_l2_pct` | **8.3313%** | (PR #571 test=8.171%) +0.16pp worse |
| `full_val/wall_shear_mae` | 0.06989 | — |
| `full_val/surface_pressure_mae` | 0.01293 | — |
| `full_val/volume_pressure_mae` | 8.70427 | — |
| `test/wall_shear_mae` | 0.06685 | — |
| `test/surface_pressure_mae` | 0.01224 | — |
| `test/volume_pressure_mae` | 29.20269 | — |

Per-epoch trajectory (val_primary/abupt_axis_mean_rel_l2_pct):
- EP1: 31.0686 (warmup)
- EP2: 8.8355
- EP3: 7.6271
- EP4: 7.2626
- EP5: 7.1489
- **EP6 (timeout-cut): 7.1358 ← best**

**Run command** (exactly the PR-body command, with `--wandb-name thorfinn/flow-aligned-tau --wandb-group tay-flow-aligned-tau`):

```bash
  --surface-loss-weight 2.0 \
  --no-compile-model --lr-cosine-t-max 13 --epochs 13 \
  --use-tangent-frame-tau --tangent-frame-seed-axis x \
  --wandb-name thorfinn/flow-aligned-tau --wandb-group tay-flow-aligned-tau
```

- W&B run id (rank-0): `41xowt6c`
- Wall time: 270.65 min (train) + ~10 min (single-rank full_val + test eval)
- Peak memory: 42.9 GB / GPU
- Per-epoch wall-clock: EP1–EP3 ~62 min (warmup), EP4–EP5 ~37 min, EP6 partial

**The flow-aligned tau loss does not help on this baseline.** With the prescribed channel weights `[cp=1.0, tau_t=1.0, tau_n=1.5, tau_s=2.0]`, the model plateaus at val_abupt ≈ 7.14%, ~0.41 pp worse than the equivalent global-frame baseline (PR #594 = 6.7258% at EP4). Comparing same-step counts:

| | EP4 (step ~43.5k) | step diff |
|---|---:|---:|
| PR #594 (global frame) | val = 6.7258% | — |
| This run (local frame) | val = 7.2626% | **+0.54 pp worse** |

The model is converging — each successive epoch reduces val_abupt — but it is converging to a *higher* asymptote. Even extrapolating the trajectory (Δ from EP5→EP6 = −0.013pp) suggests a plateau around 7.1%, well above the gate.

### Hypothesised mechanism for the regression

Per the analysis I posted at 14:27, with R orthogonal and channel weights `[1.0, 1.0, 1.5, 2.0]`, the local-frame loss is mathematically a *Mahalanobis-style anisotropic MSE in the original Cartesian basis*, with off-diagonal coupling determined by R. Two possible reasons for the regression:

1. **The streamwise/normal/spanwise priors are wrong on average.** The advisor's reasoning ("no-slip ⇒ tau_n ≈ 0", "spanwise tau ≪ streamwise tau") is correct for the floor and roof of the car (where n̂ is mostly ẑ-aligned). But on the front bumper, hood, and rear surfaces, n̂ rotates significantly — and the *streamwise* tangent direction follows it. There is no global "streamwise direction" once you fix the local frame; what was streamwise globally becomes a mix of t̂/n̂/ŝ locally. The 1.5×/2.0× weights are then up-weighting partly arbitrary mixtures of the underlying flow components.
2. **The Mahalanobis coupling is destructive.** Concretely, in regions where the local frame is rotated 30–60° from the global axes, the channel weights `(1.0, 1.0, 1.5, 2.0)` projected back to global Cartesian become coupled, e.g. `tau_y` weight ends up partly contributing to `tau_z` errors and vice versa. This can hurt the model's ability to reduce per-axis error independently.

Together these mean the same channel weights that *help* in global frame (PR #571: val 6.8701% → 6.7644%, −0.106pp) actively *hurt* in local frame.

## Suggested follow-ups

(Not implementing any of these — they are recommendations.)

1. **Apply local-frame channel weights to vector-tau magnitude only**, not to the rotated components. I.e. compute the 3-vector tau MSE in the local frame (rotation invariant, no effect) but apply weight 1.0 to all three rotated channels — and instead use the local frame purely as an *auxiliary loss* (e.g. an extra tangential-flow alignment term) rather than replacing the global loss.
2. **Run with isotropic local-frame weights** (`tau_t=tau_n=tau_s=1.0`) and let the global loss `--tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0` ride on top in the global frame. This separates "physics-motivated channel weighting" from "frame rotation".
3. **Predict in local frame** (i.e. model output head produces (τ_t, τ_n, τ_s) directly) instead of rotating only the loss. This is a different, more invasive change but is what the original PR hypothesis actually asked for — the current implementation is loss-rotation-only, per advisor approval.
4. **Drop the experiment** — given (1)–(3) are all distinct hypothesis variants, file them as new PRs rather than retro-fitting this one.

## Per advisor's request — relaunching v2

Launching the same command in `--wandb-group flow-aligned-tau-v2` for an independent confirmation run. Will post v2 results when complete (~4.5 h).

morganmcg1:
ADVISOR: **Acknowledging Run 1 results — and retracting my 19:06 UTC comment.**

You are correct: my 19:06 comment cited runs (`thorfinn/local-tangent-frame-pr`, `thorfinn/flow-aligned-tau-debug`) that are not on this PR. Those were from a different thread / older fragments and I should not have posted them here. Please disregard. Run `41xowt6c` ran cleanly to its 270-min train-timeout; no crash to debug. The safety guards (degenerate-tangent fallback, defensive normalisation) and your synthetic sanity-check evidence are all good — implementation is mathematically correct.

**Run 1 verdict: regression confirmed.**
- val_abupt = 7.1358% (EP6) vs tay SOTA 6.5985% → **+0.537pp worse**
- Same-step EP4: 7.2626% vs PR #594's 6.7258% → **+0.54pp worse** (tight match — confirms it is not a maturity issue)
- The plateau is real, not a budget artefact

Your two-mechanism explanation (streamwise/normal/spanwise priors are wrong on average + Mahalanobis coupling destructive) is consistent with the per-channel data (τ_y +1.18pp, τ_z +1.12pp — exactly the channels the prior was supposed to help). This is the kind of analysis I want to see — thank you.

**Decision: kill v2.** The hypothesis is well-tested with run 1; v2 will only confirm the plateau and burn a 4.5h DDP8 slot. Please:

1. Stop the v2 run (run group `flow-aligned-tau-v2`, ranks `o9rd7iks` etc.) at the next safe checkpoint — no need to waste GPU.
2. Post a terminal `SENPAI-RESULT` with v1's best metrics, `terminal=true, pending_arms=false`. Use:
   ```
   SENPAI-RESULT: {"terminal":true,"status":"complete","pending_arms":false,"wandb_run_ids":["41xowt6c"],"primary_metric":{"name":"val_primary/abupt_axis_mean_rel_l2_pct","value":7.1358},"test_metric":{"name":"test_primary/abupt_axis_mean_rel_l2_pct","value":8.3313}}
   ```
3. Mark the PR ready for review so I can close it cleanly.

**Useful next-experiment recommendation captured.** Your follow-ups #1 (auxiliary tangential-flow alignment loss with isotropic weights) and #3 (predict in local frame at the head, not just rotate the loss) are the right next steps — if you have an idle slot after we close this, please pick one and I will assign it as a fresh PR. #2 (isotropic local + global anisotropic on top) is logically equivalent to PR #571 with extra rotated MSE noise, so I would skip that one.

Thanks for the rigorous analysis. Closing the local-frame loss-rotation hypothesis as a clean negative result is genuinely useful information for the τ_y/τ_z gap problem.

morganmcg1:
STUDENT thorfinn:

## V2 stopped per advisor request

V2 run `thorfinn/flow-aligned-tau-v2` (group `flow-aligned-tau-v2`) was killed cleanly at EP1 step 2508/10864 (~23%, 14:28 wall). All 8 ranks terminated, GPUs freed. No further compute spent on this hypothesis.

## Terminal SENPAI-RESULT

SENPAI-RESULT: {"terminal":true,"status":"complete","pending_arms":false,"wandb_run_ids":["41xowt6c"],"primary_metric":{"name":"val_primary/abupt_axis_mean_rel_l2_pct","value":7.1358},"test_metric":{"name":"test_primary/abupt_axis_mean_rel_l2_pct","value":8.3313}}

Verdict: clean negative result. Local-frame loss rotation with anisotropic channel weights (1.0/1.0/1.5/2.0) regresses val_abupt by +0.537pp vs tay SOTA (6.5985%). Mechanism (Mahalanobis coupling + wrong streamwise prior on average) covered in the prior comment. Marking PR ready for review.

morganmcg1:
ADVISOR: Closing PR #613 because Closing as a clean documented negative result. Local-frame loss rotation with anisotropic channel weights (1.0/1.0/1.5/2.0) regresses val_abupt by +0.537pp vs tay SOTA (6.5985% → 7.1358%). Same-step EP4 comparison vs PR #594 confirms +0.54pp regression — not a maturity artefact. The two-mechanism explanation (Mahalanobis coupling between rotated channels + the streamwise/normal/spanwise prior being wrong on average due to local-frame rotation on car geometry) is rigorous and consistent with the per-channel τ_y/τ_z signature. V2 was killed cleanly at EP1 — no further GPU spent. Useful negative result for the τ-channel-frame search; auxiliary tangential-flow alignment loss (student suggestion #1) and head-side local-frame prediction (suggestion #3) remain promising follow-up directions and will be assigned as fresh PRs..

morganmcg1:
ADVISOR: Closing PR #613 because Clean negative result. val_abupt = 7.1358% (EP6, W&B run 41xowt6c) vs single-model SOTA 6.7258% (PR #594) — regression of +0.41pp. Local-frame loss rotation with anisotropic channel weights (1.0/1.0/1.5/2.0) hurts rather than helps.

Root cause (thorfinn's analysis, confirmed): the Mahalanobis-coupling effect of rotating an anisotropic MSE into the tangent frame generates off-diagonal gradient coupling that harms independent per-axis convergence, especially in non-floor/roof surface regions where the local frame is tilted 30–60° from global axes. The 'no-slip tau_n ≈ 0' prior is valid only locally, not as a global channel-weight signal.

This is a clean, informative negative: we now know that rotating the loss landscape into a surface-local frame with anisotropic weights is counterproductive. The more invasive variant (predict in local frame at the output head, true coordinate change) remains untested and is a distinct hypothesis if the tau_y/tau_z gap needs further attack.

V2 run was killed at EP1 per advisor instruction. No GPU waste beyond the confirmatory kill.

Thorfinn is now free for reassignment..

PR #637, opened 2026-05-04
Extended low-LR continued training from yi best checkpoint

morganmcg1:
## Hypothesis

The tay SOTA run (PR #511, val_abupt=7.0%) showed strong benefits from extended training. The current yi baseline uses the standard epoch budget, but many of our runs show loss still declining at the epoch cap. Continuing training from the best yi checkpoint at a very low LR (1/10 of the base LR, i.e. 1e-5) may squeeze out additional τ_y/τ_z and volume_pressure improvements that were cut off by the epoch limit.

Additionally, Lion optimizer is known to benefit from very extended low-LR polishing — the effective LR decay in Lion's update rule means late-stage training at 1e-5 is qualitatively different from simply multiplying gradients by 0.1.

## Background

The extended-training strategy is well-validated: tay PR #511 is our aspirational ceiling (7.0%) and that run was longer than our standard budget. Meanwhile our current best (PR #576, 8.2528%) was trained to the standard epoch cap. The gap between 8.2528% and 7.0% is substantial — extended low-LR training is a direct and low-risk lever to pull.

## Implementation

This is a **continued training run** from the best yi checkpoint. No code changes needed — use the `--resume` flag (or equivalent checkpoint loading) with a reduced LR.

**Two arms**:
- **Arm A**: LR=1e-5 (10× lower), run for an additional epoch budget equal to 50% of the original training budget
- **Arm B**: LR=5e-6 (20× lower), same extended budget

Use `--lr 1e-5` / `--lr 5e-6` in the training command. Load the yi best checkpoint via `--resume <checkpoint_path>`.

If there is no `--resume` flag, implement one: load the checkpoint dict, skip fresh initialization, and continue training from that state.

## Current Baseline Metrics (yi bar, PR #576)

| Metric | yi best | AB-UPT | Gap |
|---|---|---|---|
| **val_abupt** | **8.2528%** | — | merge bar |
| surface_pressure | 5.241% | 3.82% | 1.37× |
| wall_shear | 9.233% | 7.29% | 1.27× |
| volume_pressure | 12.438% | 6.08% | 2.05× |
| tau_x | 7.562% | 5.35% | 1.41× |
| tau_y | 9.233% | 3.65% | **2.53×** |
| tau_z | 10.449% | 3.63% | **2.88×** |

## Training Commands

**Arm A — LR=1e-5, continued from yi best checkpoint**:
```bash
cd target/
[...]

PR #576, opened 2026-05-04
Round 33: STRING-sep + Lion from-scratch anchor run

morganmcg1:
## Hypothesis

STRING-sep learnable PE (PR #490) combined with Lion lr=1e-4 clip=0.5 (PR #517) in a canonical **from-scratch** run on yi will push val_abupt below 9.032% toward the ~8.1% preview seen in frieren's checkpoint-resume run `zwh9qzjw`.

Both wins were merged to yi sequentially but have never been combined in a clean from-scratch canonical run — that's what this PR confirms. No code changes needed; all flags are already in train.py.

**Win criterion:** val_abupt < 9.032% (merge bar). Stretch goal: < 8.5%.

## Instructions

No code changes required. Run the following exactly:

```bash
cd target/
torchrun --standalone --nproc_per_node=4 train.py \
  --agent frieren \
  --wandb_group yi-round33-string-sep-lion-anchor \
  --learnable-pe \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --clip-grad-norm 0.5 \
  --lr-warmup-epochs 1 \
  --ema-decay 0.999 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --batch-size 4 \
  --train-surface-points 65536 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --epochs 50
```

This is a single-arm run. Report the best val_abupt achieved and the epoch at which it occurred, plus the W&B run ID.

## Baseline

Current yi merge bar: **val_abupt 9.032%** (PR #517 askeladd, W&B run `brat65z4`, Lion lr=1e-4 clip=0.5 without `--learnable-pe`)

Non-canonical preview (frieren checkpoint-resume `zwh9qzjw`): val_abupt ≈ **8.087%** with STRING-sep + Lion — this is the expected ballpark but was not a clean run.

| Metric | yi best (PR #517) |
|---|---:|
| `abupt_axis_mean_rel_l2_pct` | 9.032% |
| `surface_pressure_rel_l2_pct` | 5.702% |
| `wall_shear_rel_l2_pct` | 10.257% |
| `volume_pressure_rel_l2_pct` | 5.075% |
| `wall_shear_x_rel_l2_pct` | 8.520% |
| `wall_shear_y_rel_l2_pct` | 12.907% |
| `wall_shear_z_rel_l2_pct` | 12.957% |

[...]

PR #653, opened 2026-05-05
1-cycle LR with 1e-3 peak (OneCycleLR) on yi stack

morganmcg1:
## Hypothesis

1-cycle LR scheduling (OneCycleLR) with max_lr=1e-3 for Lion on the full yi stack. Issue #18 explicitly directs: "1-cycle LR with 1e-3 peak." The 1cycle policy uses a super-convergence regime: it ramps LR rapidly to a high peak (10× baseline), then anneals steeply to near-zero. Smith & Touvron (2019) show this can escape flat regions faster than monotone cosine decay, especially when the model has already been pretrained and needs strong perturbations to break out of local minima.

The current yi scheduler is CosineAnnealingLR hardcoded at line 2632 of train.py. There is no `--lr-schedule` CLI flag. You will need to add this capability to train.py.

Reference: Smith (2018) "A disciplined approach to neural network hyper-parameters: Part 1" — arXiv:1803.09820.

### 1. Add config fields to the `TrainConfig` dataclass (around line 1159):

```python
lr_schedule: str = "cosine"        # "cosine" | "onecycle"
onecycle_max_lr: float = 1e-3      # max_lr for OneCycleLR
onecycle_pct_start: float = 0.3    # fraction of cycle for warmup phase
onecycle_div_factor: float = 10.0  # initial_lr = max_lr / div_factor
onecycle_final_div_factor: float = 1e4  # min_lr = initial_lr / final_div_factor
```

### 2. Replace the hardcoded scheduler at line 2632:

Replace:
```python
scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(optimizer, T_max=max_epochs)
```

With:
```python
if config.lr_schedule == "onecycle":
    steps_per_epoch = max(len(train_loader), 1)
    scheduler = torch.optim.lr_scheduler.OneCycleLR(
        optimizer,
        max_lr=[config.onecycle_max_lr * (pg["lr"] / config.lr) for pg in optimizer.param_groups],
        total_steps=max_epochs * steps_per_epoch,
        pct_start=config.onecycle_pct_start,
        div_factor=config.onecycle_div_factor,
        final_div_factor=config.onecycle_final_div_factor,
        anneal_strategy="cos",
    )
else:
    scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(optimizer, T_max=max_epochs)
```

### 3. Move the scheduler step inside the training loop for OneCycleLR (it must step per batch, not per epoch):

OneCycleLR expects `scheduler.step()` after each batch, not after each epoch. Find the per-epoch `scheduler.step()` at line 2999 and wrap it:

```python
if config.lr_schedule != "onecycle":
    scheduler.step()
```

Then inside the batch loop (after the optimizer step, currently around line 2960–2975), add:
```python
if config.lr_schedule == "onecycle":
    scheduler.step()
```

Also update the `scheduler.get_last_lr()[0]` at line 3012 — this should work unchanged for both scheduler types.

### 4. Disable the existing manual warmup for OneCycleLR (it handles warmup internally via pct_start):

In the warmup logic block around lines 2900–2910, add a guard:
```python
if config.lr_schedule != "onecycle" and effective_warmup_steps > 0 and global_step < effective_warmup_steps:
    # existing warmup scaling code
```

### 5. Run two arms with the full yi stack:

**Arm A — OneCycleLR max_lr=1e-3** (primary arm; tests Issue #18 directive):
```bash
cd /workspace/senpai/target
  --agent askeladd --wandb-group yi-round37-1cycle-lr \
  --wandb-name askeladd/1cycle-maxlr1e-3-arm-a \
  --learnable-pe --optimizer lion --lr 1e-4 \
  --lr-schedule onecycle --onecycle-max-lr 1e-3 \
  --onecycle-pct-start 0.3 --onecycle-div-factor 10.0 --onecycle-final-div-factor 1e4 \
  --lr-warmup-epochs 0 \
  --weight-decay 5e-4 --clip-grad-norm 0.5 \
  --grad-ema-alpha 0.5 --beta-nll-beta 0.5 \
  --ema-decay 0.999
```

**Arm B — OneCycleLR max_lr=5e-4** (gentler peak, sanity check):
```bash
cd /workspace/senpai/target
  --agent askeladd --wandb-group yi-round37-1cycle-lr \
  --wandb-name askeladd/1cycle-maxlr5e-4-arm-b \
  --learnable-pe --optimizer lion --lr 1e-4 \
  --lr-schedule onecycle --onecycle-max-lr 5e-4 \
  --onecycle-pct-start 0.3 --onecycle-div-factor 5.0 --onecycle-final-div-factor 1e4 \
  --lr-warmup-epochs 0 \
  --weight-decay 5e-4 --clip-grad-norm 0.5 \
  --grad-ema-alpha 0.5 --beta-nll-beta 0.5 \
  --ema-decay 0.999
```

Run both arms. Report `val_abupt` and the per-axis wall shear metrics (ws_y, ws_z) from W&B.

## Baseline

Current best: **PR #637** (fern, run `vzprvtaw`) — val_abupt = **7.5373%**, test_abupt = 8.8533%

| Metric | val (yi best) | test (yi best) | AB-UPT target |
|--------|--------------|----------------|---------------|
| abupt_axis_mean | 7.5373% | 8.8533% | — |
| surface_pressure | 4.9322% | 4.6968% | 3.82% |
| wall_shear | 8.4774% | 8.4954% | 7.29% |
| volume_pressure | 4.4102% | 11.4599% | 6.08% |
| wall_shear_x (τ_x) | 7.2272% | 7.3046% | 5.35% |
| wall_shear_y (τ_y) | 9.8691% | 9.8648% | 3.65% |
| wall_shear_z (τ_z) | 11.2477% | 10.9403% | 3.63% |

Beat val_abupt < 7.5373% to qualify as a win. Watch τ_y and τ_z especially — these are the primary gap to AB-UPT.

W&B baseline run: https://wandb.ai/wandb/senpai/runs/vzprvtaw


morganmcg1:
STUDENT askeladd: Implementation complete. Plan + status:

**Implementation:**
- Added 5 config fields: `lr_schedule`, `onecycle_max_lr`, `onecycle_pct_start`, `onecycle_div_factor`, `onecycle_final_div_factor` (auto-CLI via dataclass).
- Replaced hardcoded `CosineAnnealingLR` at the scheduler creation site with a branch on `config.lr_schedule`. OneCycleLR uses `total_steps = max_epochs * len(train_loader)`.
- Moved `scheduler.step()` per-batch when onecycle (after `optimizer.step()`); per-epoch step kept for cosine.
- Gated manual warmup with `config.lr_schedule != "onecycle"` so the OneCycleLR `pct_start` warmup is the only warmup active.
- Per-group max_lr preserves the optimizer's per-group lr ratios via `max_lr * (pg['lr']/config.lr)`.

**Cycle length choice:** baseline 4-GPU runs (vzprvtaw, ym8x8301, 86lxu1w0) average 86 min/epoch with `--validation-every 1`. To fit a complete OneCycleLR cycle in the 360-min timeout, I'm running `--epochs 3` for both arms. That gives:
- Warmup 0.9 ep (steps 0 → 4898) at LR 1e-4 → max_lr
- Anneal 2.1 ep (steps 4898 → 16326) at max_lr → max_lr/(10*1e4)

This is a compressed cycle but matches the cold-start cycle length used by PR #576 (frieren) and other 4-GPU yi runs. Sequential to avoid cross-arm GPU contention.

**Arm A launched** (max_lr=1e-3): `OneCycleLR: max_lr=1.000e-03 total_steps=16326 pct_start=0.3 div_factor=10.0 final_div_factor=10000.0`
W&B group: yi-round37-1cycle-lr / askeladd/1cycle-maxlr1e-3-arm-a

morganmcg1:
STUDENT askeladd: BUG-FIX

The first Arm A run (W&B `9yrfcfhc`) crashed at step 5337/5442 (98% of epoch 1, ~1h54min in) with:

```
ValueError: Too many bins for data range. Cannot create 64 finite-sized bins.
```

Stack: `collect_gradient_metrics` → `wandb.Histogram` → `np.histogram`. `numpy.histogram` raises this when all finite values are equal (zero data range), which happens for parameters whose gradient is exactly zero (frozen blocks / unused params on a given step). `collect_weight_metrics` has the same call pattern.

**Fix (in this PR):** Added `_safe_wandb_histogram(arr)` helper that returns `None` when the array is empty, has no finite values, or has zero range; updated all four `wandb.Histogram(...)` call sites in `collect_gradient_metrics` and `collect_weight_metrics` to use it. Unit-tested on zero arrays, constant arrays, normal arrays, and empty arrays.

This is a pre-existing latent bug, not OneCycleLR-specific. Worth noting for advisor review — happy to spin out a separate bug-fix PR if preferred.

Restarting Arm A now with the fix in place.

morganmcg1:
ADVISOR: Confirmed — epochs=3 with 1-cycle LR fitting into the 360-min window is the right approach given the OneCycleLR requires a full cycle. 

Please post EP1 gate metric (val_abupt) here as soon as it lands. Baseline to beat: **7.5373%**. EP1 gate ≤25%. If the 3-epoch schedule completes cleanly, report the final EP3 val_abupt + τ_y + τ_z + vol_pressure for merge evaluation.

morganmcg1:
STUDENT askeladd: EP1 GATE FAILED on Arm A — aborting and switching to Arm B

**Arm A (max_lr=1e-3) EP1 validation** (W&B `hm2vilgz`):
- `val_abupt_axis_mean_rel_l2_pct` = **106.3732%** (gate ≤25%, baseline 7.5373%)
- Per-component: surface_p=81.63%, wall_shear=114.99%, volume_p=89.15%
- τ_x=104.48% / τ_y=145.09% / τ_z=111.52%
- train_loss EP1 = 0.2629 (well above well-trained baselines that converge near 0)

**Diagnosis.** With Lion as the optimizer, max_lr=1e-3 is far too aggressive. Lion uses sign-based updates so its effective step size at peak LR ≈ max_lr is closer to "raw 1e-3 movement on every parameter," whereas AdamW would scale by 1/sqrt(grad_var). The Smith 1cycle paper (arXiv:1803.09820) was tuned for SGD+momentum; the 10× peak heuristic does not transfer to Lion. The pretrained yi weights were destabilized during the 0.9-epoch warmup as LR climbed past ~3e-4.

**Decision.** Aborted Arm A at EP2 step 71 (saves ~1.5h compute). EP3 recovery from 106% → <7.54% in 2.1 epochs of annealing is not credible.

**Launching Arm B (max_lr=5e-4)** now on GPUs 0–3. Arm B keeps `div_factor=5.0` so initial_lr=1e-4 (matches baseline), and peak is 5× rather than 10× over baseline — should be much gentler on Lion. If Arm B also fails the EP1 gate, that's strong evidence 1cycle + Lion is fundamentally incompatible at any peak above ~baseline LR, which would be a clean negative result for the hypothesis.

morganmcg1:
ADVISOR: Good diagnosis on the Lion+1-cycle incompatibility. The fundamental issue is that Lion uses sign(gradient) for updates — so the effective step size is always exactly lr, regardless of gradient magnitude. The Smith 1-cycle 10× peak heuristic was derived empirically for SGD and AdamW where momentum and the update scale interact in a well-understood way. For Lion at 1e-3, that is simply too aggressive — each parameter moves 1e-3 per step unconditionally, which explains the EP1=106.37% catastrophic divergence.

**Arm B (max_lr=5e-4, div_factor=5.0)** is the right fallback — it respects the existing Lion baseline lr=1e-4 as the trough (initial_lr=1e-4, max_lr=5e-4, final_lr≈1e-5). This is a 5× rather than 10× peak, keeping the optimizer in a regime it has demonstrated stability in.

Gate for Arm B:
- EP1 ≤25%: continue
- EP3 ≤10%: authorize full run
- If EP1 ≥25% or diverges: close PR, this is a clean negative result

If Arm B also fails to beat baseline, that is a publishable negative result: **1-cycle LR scheduling is incompatible with Lion optimizer** due to the sign-based update mechanics. That constrains the hypothesis space cleanly for future work.

Current merge bar: val_abupt=**7.3914%**. Please report EP1 for Arm B as soon as it is available.

morganmcg1:
ADVISOR: Closing PR #653 because ADVISOR: Closing as a clean negative result.

**Both arms exceeded the EP1 kill gate (≤25%):**
- Arm A (max_lr=1e-3, run `hm2vilgz`): EP1 = **106.37%** — catastrophic divergence
- Arm B (max_lr=5e-4, run `wwgc5wks`): best_val = **29.88%** — never recovered below the kill gate

**Root cause confirmed: OneCycleLR + Lion are fundamentally incompatible.**

Lion uses sign(gradient) for updates, so the effective step size per parameter equals lr unconditionally — there is no adaptive dampening. The Smith 1-cycle 10× peak heuristic (arXiv:1803.09820) was derived for SGD/AdamW where gradient variance scales the update. At max_lr=1e-3, Lion moves each parameter 1e-3 per step unconditionally, destabilizing the pretrained yi weights during the warmup phase. Even the gentler Arm B (5× peak, initial_lr preserved at 1e-4) produced best_val=29.88% and showed declining-but-inadequate slope (-3.78 pp/1k steps) — the model had already been knocked out of the pretrained-weight basin and couldn't recover in 3 epochs.

**Publishable finding:** 1-cycle LR scheduling as formulated (Smith 2018) is incompatible with Lion optimizer on the yi stack. The sign-based update mechanism makes the peak-LR heuristics derived for momentum-based optimizers inapplicable. Future work exploring adaptive LR schedules with Lion should investigate schedules that keep max_lr ≤ baseline_lr (cosine anneal, warmup + flat, or decay-only regimes).

Thank you @askeladd for the clean implementation and the good diagnosis of the mechanism..

PR #658, opened 2026-05-05
SWA staged trajectory: resume from PR #637 yi SOTA, 2ep at lr=5e-6

morganmcg1:
## Hypothesis

Stochastic Weight Averaging (SWA) applied in the near-convergence regime — resuming from the yi SOTA checkpoint (PR #637, val_abupt=7.5373%) and running 1–2 epochs at lr=5e-6 — will produce a final model that improves on all 7 primary metrics, particularly τ_y and τ_z.

**Mechanistic evidence (from PR #634 nezuko, Arm A, W&B run `ty6cnqmz`):**
SWA improved every single metric over best-checkpoint by 0.41–0.87pp, even though the SWA window was severely truncated (only 6 snapshots over ~1200 steps due to training timeout). SWA averaged across the **flat loss-landscape basin** near convergence, reducing the variance of the final solution. The generalization signal was confirmed: test deltas were *larger* than val deltas (test_abupt −0.6461pp vs val_abupt −0.6126pp), and τ_y/τ_z saw the biggest absolute gains (τ_y −0.8674pp, τ_z −0.7887pp).

**Why PR #634 missed the bar:** It ran in the wrong regime — a 3-epoch cold-start where the optimizer had not yet converged. SWA needs a flat loss landscape to be effective. Resuming from the current SOTA checkpoint places us exactly in that flat-landscape regime.

**Expected behaviour:** With 500-step snapshot cadence over 1–2 full epochs (~10,000–22,000 steps), we collect ~20–44 SWA snapshots (vs 6 in PR #634). The averaging effect compounds with each snapshot. Given PR #634 got −0.61pp val_abupt from 6 snapshots, a 3–7× larger ensemble of near-optimal snapshots should yield a meaningful further push.

**Reference:** Izmailov et al. "Averaging Weights Leads to Wider Optima and Better Generalization" (UAI 2018); also validated in NLP fine-tuning as SWA outperforms single-checkpoint in the near-convergence plateau regime.

## Implementation Notes

The SWA flags (`--swa-start-fraction`, `--swa-lr`, `--swa-anneal-epochs`, `--swa-update-every-steps`) were implemented by you in PR #634. **You must re-implement these flags in the current yi `train.py`** (they were never merged from PR #634 into yi — the PR was closed without merging). The implementation pattern is:

1. Use `torch.optim.swa_utils.AveragedModel` wrapping the model.
2. After `swa_start_fraction` of total steps, begin collecting SWA snapshots every `swa_update_every_steps` steps.
[...]

PR #680, opened 2026-05-05
Tangential wallshear loss on full SOTA stack — τ_y/τ_z frame fix

morganmcg1:
## Hypothesis

Decomposing wall shear stress into surface-tangent components (tangential-normal frame) before computing the training loss provides a better-conditioned regression target for τ_y and τ_z. The `--use-tangential-wallshear-loss` flag is already implemented in `yi` `train.py`.

The intuition: raw τ_x/τ_y/τ_z in the global Cartesian frame have a strong dependence on local surface orientation. Near a swept wing edge, τ_z in global frame can be large simply because the surface tilts — not because the shear magnitude is large. By projecting τ onto the surface tangent/normal basis before computing MSE, the loss aligns with the physically meaningful shear magnitude and direction, reducing noise from orientation-dependent cross-terms.

τ_y=9.6% and τ_z=11.1% are respectively 2.6× and 3.0× above AB-UPT targets. This is the primary remaining gap. A coordinate-frame change is a zero-cost (no new parameters) structural improvement that could directly target this gap.

## Instructions

This is a cold-start run on the full yi SOTA stack. Enable tangential wallshear loss with `--use-tangential-wallshear-loss`:

```bash
cd /workspace/senpai/target
  --agent fern \
  --wandb-group yi-round39-tangential-wallshear \
  --wandb-name fern/tangential-wallshear-cold-start \
  --learnable-pe \
  --grad-ema-alpha 0.5 \
  --beta-nll-beta 0.5 \
  --ema-decay 0.999 \
  --train-volume-points 65536 --eval-volume-points 65536
```

**No code changes required** — the `--use-tangential-wallshear-loss` flag is already present in `yi` `train.py`. Just run the command above.

## Kill gates

- EP1 val_abupt > 20%: kill and report
- EP3 val_abupt > 12%: kill and report
- Otherwise: run to timeout

## What to report

In addition to the standard metrics, please report τ_y and τ_z separately at each validation epoch — these are the primary targets of this intervention. Also note whether τ_y/τ_z improve relative to surface_pressure, which would confirm the coordinate-frame hypothesis.

## Baseline

Current yi SOTA: **val_abupt = 7.3914%**, test_abupt = 8.7189% (PR #658, nezuko, run `pxsnrw36`)

| Metric | val | test |
|---|---:|---:|
| surface_pressure_rel_l2_pct | 4.8552% | — |
| wall_shear_rel_l2_pct | 8.3192% | — |
| volume_pressure_rel_l2_pct | 4.3156% | — |
| wall_shear_x_rel_l2_pct (τ_x) | 7.1166% | — |
| wall_shear_y_rel_l2_pct (τ_y) | 9.6123% | — |
| wall_shear_z_rel_l2_pct (τ_z) | 11.0573% | — |

AB-UPT targets (val): surface_pressure 3.82%, τ_x 5.35%, τ_y 3.65%, τ_z 3.63%

τ_y and τ_z are the primary research targets (2.6× and 3.0× above AB-UPT). Beat val_abupt < 7.3914% to set a new SOTA.

W&B project: `wandb-applied-ai-team/senpai-v1-drivaerml`

### DIFF
--- a/train.py
+++ b/train.py
@@ -2052,7 +2052,7 @@ def train_loss(
         )
         surface_pred_norm = out["surface_preds"]
         normal_rms = float("nan")
-        if use_tangential_wallshear_loss and not use_beta_nll:
+        if use_tangential_wallshear_loss:
             # Wall-shear stds are non-uniform ([2.08, 1.36, 1.11]), so projecting
             # in normalized space does not equal physical-space tangent projection.
             # Denormalize -> project in physical space -> renormalize.
@@ -2131,7 +2131,7 @@ def train_loss(
     }
     if aux_rel_l2_value is not None:
         metrics["aux_rel_l2_loss"] = aux_rel_l2_value
-    if use_tangential_wallshear_loss and not use_beta_nll:
+    if use_tangential_wallshear_loss:
         metrics["wallshear_pred_normal_rms"] = normal_rms
     if use_beta_nll and per_axis_log_var is not None:
         metrics["log_var_surface_pressure"] = per_axis_log_var[0]


morganmcg1:
STUDENT fern: BUG-FIX before launching this experiment.

The PR command enables both `--beta-nll-beta 0.5` (yi SOTA stack) and `--use-tangential-wallshear-loss`. However, train_loss had this guard from PR #583:

```python
if use_tangential_wallshear_loss and not use_beta_nll:
    # … project to tangent plane …
```

i.e. `--use-tangential-wallshear-loss` is silently a no-op whenever β-NLL is active. As written, this PR's command would have run an exact duplicate of the yi SOTA baseline and "no code changes required" would have been wrong.

Fix (commit dacde35 on this branch): drop the `and not use_beta_nll` guard in `train_loss`. Tangent projection now applies to the predicted/target wall-shear means before β-NLL is computed; the per-axis log_var head still operates as before on the projected residuals.

- **Train path**: `surface_pred_norm` and `surface_target` (channels 1:4) are denormalized → projected onto the per-point tangent plane (`vec − (vec·n̂)n̂`) → renormalized → fed to the β-NLL loss with the existing per-channel weighting. cp channel and volume head are untouched.
- **Val/test path**: unchanged. Validation uses raw `surface_pred_norm` (no projection), so val/test τ_x/τ_y/τ_z are apples-to-apples vs. the SOTA baseline.
- `wallshear_pred_normal_rms` diagnostic is now also logged with β-NLL on (it tracks how much normal-direction signal the head is suppressing in physical space).

Launching the experiment now with this fix applied. Will report back as runs progress.

morganmcg1:
STUDENT fern:

## Results

The cold-start run was killed at **EP1** by the `val_primary/abupt_axis_mean_rel_l2_pct < 20` kill gate. val_abupt landed at **45.7732%**, ~6.2× the current yi SOTA baseline (7.3914%) and well over the 20% threshold.

### Per-axis validation breakdown (EP1, killed)

| Metric | This run (EP1) | Baseline #658 | Δ vs baseline |
|---|---:|---:|---:|
| val_primary/abupt_axis_mean_rel_l2_pct | **45.7732%** | 7.3914% | +38.38 pp (6.2×) |
| val_primary/surface_pressure_rel_l2_pct | 10.9279% | 4.8552% | 2.3× |
| val_primary/wall_shear_rel_l2_pct | 58.1345% | 8.3192% | 7.0× |
| val_primary/wall_shear_x_rel_l2_pct (τ_x) | 31.8753% | 7.1166% | 4.5× |
| val_primary/wall_shear_y_rel_l2_pct (τ_y) | 42.6763% | 9.6123% | 4.4× |
| val_primary/wall_shear_z_rel_l2_pct (τ_z) | **131.3139%** | 11.0573% | 11.9× |
| val_primary/volume_pressure_rel_l2_pct | 12.0728% | 4.3156% | 2.8× |

τ_z is catastrophically broken (131% rel-L2). The intervention was meant to *reduce* τ_y/τ_z; it did the opposite, by a wide margin, on every axis. No test eval ran (kill happened before it).

### Run details

```bash
cd /workspace/senpai/target
  --agent fern \
  --wandb-group yi-round39-tangential-wallshear \
  --wandb-name fern/tangential-wallshear-cold-start \
  --learnable-pe \
  --grad-ema-alpha 0.5 \
  --beta-nll-beta 0.5 \
  --ema-decay 0.999 \
  --train-volume-points 65536 --eval-volume-points 65536
```

- **W&B run id**: `ikux0u9z` (https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/ikux0u9z) — state=finished (kill-gate)
- **Wall-clock**: 123.8 min (1 epoch + EP1 val)
- **Peak memory**: ~68.1 GB / GPU
- **train_loss (EP1)**: −0.17325 (β-NLL, expected negative)
- **early_stop/reason**: `kill threshold failed at step 5442: val_primary/abupt_axis_mean_rel_l2_pct=45.7732 did not satisfy step>=1 val_primary/abupt_axis_mean_rel_l2_pct<20`

### What happened — root cause

The bug fix in commit `dacde35` made the tangent projection actually fire in the training loss. With it firing, the cold-start run failed catastrophically. Mechanism:

1. **Train path**: `surface_pred_norm` and `surface_target` are denormalized → projected onto the per-point tangent plane (`v − (v·n̂)n̂`) → renormalized → fed to per-axis β-NLL. This zeroes the *normal-direction component* of both pred and target before computing the loss, so the loss does **not penalize** any signal the model puts in the normal direction.
2. **Val/test path** (deliberately unchanged in the fix, for apples-to-apples): evaluates raw 3-channel τ_x/τ_y/τ_z without projection.
3. **Net effect**: during training the model has zero pressure to keep `pred · n̂ ≈ 0`. The diagnostic confirms it: `train/wallshear_pred_normal_rms` (RMS of `ws_pred_phys · n̂` in physical units) climbs to ~1.68 by mid-EP1 (steps 2722, 4081), comparable to the τ-stds [2.08, 1.36, 1.11]. At validation, this normal-direction signal is no longer hidden by projection and shows up directly in val/test rel-L2.
4. **Why τ_z is the worst**: DrivAerML body normals on the hood/roof/floor are roughly z-aligned, so unconstrained normal-direction prediction lands almost entirely in τ_z when the surface is body-up. Result: τ_z=131% — by far the worst axis.

In short: tangent projection on the training side without a matching constraint at eval makes the loss ill-posed for the metric we actually care about. Cold-start exposes this immediately because the model has no inductive bias toward "wall shear is tangent" yet.

For comparison, similar yi-SOTA cold-start runs *without* tangent projection typically land at EP1 val_abupt ≈ 17–22% (e.g. `frieren/tangent-frame-tau-yz` 21.73%, `senku/6L-512d-full-stack-arm-a` 18.01%). 45.77% is ~2× worse than the typical cold-start floor — the regression is the projection, not bad luck on the seed.

### Suggested follow-ups

1. **Penalize normal-direction predictions** instead of (or in addition to) projecting. Add `λ · mean((ws_pred_phys · n̂)²)` to the loss; this preserves signal toward "wall shear is tangent" while keeping the supervised raw-target term intact. Use `wallshear_pred_normal_rms` (already logged) as the ablation diagnostic.
2. **Tangent projection at validation too**: re-run with projection mirrored on val/test paths. Not SOTA-worthy on its own (it changes the metric) but cleanly tests whether the projected loss is *internally* learnable.
3. **Warm-start from #658 SOTA checkpoint**: the model already knows τ is approximately tangent; resume + low-LR with the projected loss should not blow up the way a cold start does.
4. **Tangent projection without β-NLL**: isolate the interaction. β-NLL's per-axis log-var head is per-(x,y,z) but the residuals it conditions on are now tangent-projected (one direction zeroed per point). The variance head's calibration target may itself be ill-posed; running with plain MSE+projection would tell us.
5. **Smaller wallshear_y/z weights or pure τ-magnitude regression** as alternative coordinate-frame interventions targeting τ_y/τ_z directly without changing the gradient on the normal direction.

I'd lean strongly toward (1) — penalize, don't mask — as the cheapest path to actually realising the τ_y/τ_z gain the hypothesis predicted.

morganmcg1:
ADVISOR: Closing PR #680 because ADVISOR: Closing as a clear dead end. EP1 val_abupt=45.77% (6.2× worse than baseline) with τ_z=131% — catastrophic failure.

**Root cause confirmed (W&B run ikux0u9z verifies student's analysis exactly):**
The train/eval asymmetry is the core bug: training with tangent-projected targets means the model has zero gradient pressure to keep pred·n̂ ≈ 0. At eval the unprojected τ_x/τ_y/τ_z are measured, so all the normal-direction signal that accumulated unchecked during training shows up as rel-L2 error. τ_z is worst because DrivAerML body normals on hood/roof/floor are roughly z-aligned.

This formulation is fundamentally ill-posed for the SOTA metric. The projection masks signal from the loss but not from the eval metric.

**Best path forward from the student's suggestions is Option 1:** Add a normal-penalty regularizer λ·mean((ws_pred_phys·n̂)²) to the existing β-NLL loss instead of projecting targets. This preserves the current loss landscape while adding an inductive bias that wall shear should be tangent. The  diagnostic is already logged, so we can directly track whether the regularizer is working. I will assign this as a follow-up experiment.

Thank you for the thorough root-cause analysis — this is exactly the kind of diagnostic work that makes the research programme faster..

PR #666, opened 2026-05-05
Depth scaling L=6 at full hidden=512 (correct test after PR #660 confound)

morganmcg1:
## Hypothesis

The depth scaling progression L=4→L=5 yielded a substantial −1.90% relative gain (PR #592, val_abupt=6.5985% vs prior 6.7258%). The natural question is whether L=6 continues this trend.

**Previous attempt (PR #660) was flawed:** The prior L=6 sweep reduced `hidden_dim` to 384/448 to try to fit in VRAM, creating a confounded experiment — it was testing "L=6 with less capacity per layer" rather than "L=6 at equivalent capacity." PR #660 was killed at EP1 (val_abupt=30.978%) when the reduced-hidden configuration proved too weak to converge.

**This PR tests the correct question:** Does L=6 at **full hidden_dim=512** (the same per-layer capacity as SOTA L=5) continue the depth scaling trend?

**VRAM analysis:** L=5 hidden=512 uses ~52GB/96GB. L=6 adds one more transformer block (~2M params), which adds ~4–5GB VRAM — estimated ~56–57GB/96GB. This is well within the 96GB GPU budget with no capacity reduction needed.

**Why this should work:** The L=4→L=5 gain was −1.90% relative (~1.27pp absolute). If L=6 continues at even half that rate, we'd expect −0.95% relative improvement (~0.63pp), bringing val_abupt to approximately 6.00% — which would be a major advance toward the ensemble SOTA of 6.1751%.

## Instructions

This is a single-arm experiment (no sweep needed — just run the L=6 full-hidden config).

**Only change from SOTA:** `--model-layers 6` (keep hidden=512, all other params identical to PR #592).

Do NOT reduce `--model-hidden-dim`. Do NOT reduce `--model-heads`. Full capacity L=6.

If VRAM OOM occurs (which is unlikely based on estimates), try reducing `--batch-size` from 4 to 3. Do not reduce hidden_dim.

**Kill gates:**
- EP1 (step 10,864): informational only. Kill if val_abupt > 40% (catastrophic failure)
- EP2 (step 21,729): KILL if val_abupt > 12.0%
- EP3 (step 32,594): KILL if val_abupt > 8.0%

Report the full gate trajectory in your SENPAI-RESULT. Compare to L=5 SOTA at each gate:
- L=5 EP1: ~28% (typical)
- L=5 EP2: ~8.77%
- L=5 EP3: ~7.39%
- L=5 EP4: ~7.10%
- L=5 EP5: ~6.81%
[...]

PR #660, opened 2026-05-05
thorfinn: depth scaling L=6 sweep (hidden=384 and hidden=448)

morganmcg1:
## Hypothesis

The current single-model SOTA (PR #592, val_abupt=6.5985%) uses L=5 transformer layers with hidden_dim=512. Increasing depth to L=6 may unlock additional representational capacity, particularly for complex near-wall flow patterns and the tau prediction task. We pair the extra depth with a reduced hidden_dim=384 to keep VRAM usage and training time within budget (~52GB VRAM, ~270 min).

The hypothesis is that for this CFD surrogate task, depth (more sequential attention operations over slices) is more valuable than width (larger hidden representations) beyond a certain point. L=5 outperformed L=4 significantly (-1.90% relative). We test if L=6 continues this trend at the cost of a slightly narrower hidden space.

**Two arms to sweep:**
- **arm-a (L=6, hidden=384):** Depth increase with width reduction to stay in VRAM budget
- **arm-b (L=6, hidden=448):** Slightly wider variant — intermediate between 384 and 512

Use `--wandb-group depth-l6-sweep` for both arms.

## Instructions

**Arm A — `thorfinn/depth-L6-h384`:**
```bash
cd target/ && torchrun --standalone --nproc_per_node=8 train.py \
  --agent thorfinn --optimizer lion --lr 9e-5 --weight-decay 5e-4 \
  --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0 \
  --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 \
  --pos-encoding-mode string_separable --use-qk-norm \
  --rff-num-features 16 --rff-init-sigmas "0.25,0.5,1.0,2.0,4.0" \
  --lr-cosine-t-max 13 --epochs 13 \
  --vol-points-schedule "0:16384:3:32768:6:49152:9:65536" \
  --no-compile-model \
  --model-layers 6 --model-hidden-dim 384 --model-heads 4 --model-slices 128 \
  --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --kill-thresholds "10864:val_primary/abupt_axis_mean_rel_l2_pct>12.0,21729:val_primary/abupt_axis_mean_rel_l2_pct>8.0" \
  --wandb-group depth-l6-sweep --wandb-name thorfinn/depth-L6-h384
[...]

PR #556, opened 2026-05-03
Inference-time prediction ensemble: average K best-val model outputs

morganmcg1:
## Hypothesis

Inference-time prediction ensemble: average the **output predictions** from K independently-trained best-val checkpoints, then compute metrics on the averaged predictions. This sidesteps the catastrophic loss-basin problem that destroyed weight averaging (PR #554, val 6.87% → 92.85%), because we never touch the weights — we average in **output space**, which is always safe.

Expected gain: 1–3% improvement on `val_abupt_axis_mean_rel_l2_pct` from ensemble diversity. The K models already exist as W&B artifacts; there is zero training cost. This is K× inference only.

**Key insight:** Weight averaging (model soup) requires models to share a common loss basin (Wortsman et al. 2022, Ainsworth 2022). Our models are trained from independent random initializations → disconnected basins → weight averaging fails. Prediction averaging has no such constraint — diverse predictions always benefit from averaging when individual models are well-calibrated.

Reference: Lakshminarayanan et al., "Simple and Scalable Predictive Uncertainty Estimation using Deep Ensembles" (NeurIPS 2017); also standard practice on Kaggle and in production ML systems.

## Instructions

You will write a new standalone script `target/ensemble_eval.py` (do NOT modify `train.py`). The script:

1. **Loads K model checkpoints from W&B** — download the EMA checkpoint artifact for each run, instantiate the model from that run's config, and load the weights.

2. **Runs inference separately for each model** on the full val split and test split (match the eval settings from the original training run: `--eval-surface-points 65536 --eval-volume-points 65536`).

3. **Averages predictions in normalized space** — average `surface_preds` and `volume_preds` tensors across the K models BEFORE un-normalizing. Do not average in physical space.

4. **Un-normalizes the averaged predictions** and computes all 7-channel relative-L2 metrics:
   - `surface_pressure`, `wall_shear`, `volume_pressure`, `tau_x`, `tau_y`, `tau_z`
   - Aggregate: `abupt_axis_mean_rel_l2_pct` (mean of above 6 channels)

5. **Logs all metrics to W&B** under a new run, using `--wandb-group ensemble-inference-v1`.

### Checkpoint candidates (start with K=2, then K=3)

| Run ID | Agent | val_abupt (%) | Notes |
|--------|-------|----------------|-------|
| `9mm3sz7x` | askeladd | **6.8701** | SOTA — use this |
| `49aimdiz` | alphonse | ~6.87 | 2nd best — check W&B; if >7.0% skip it |
[...]

PR #693, opened 2026-05-05
Depth L=6 hidden=448: budget-fit follow-up to PR #666 (L=6/512 too slow)

morganmcg1:
## Hypothesis

PR #666 (thorfinn) tested L=6 depth at full hidden=512. Results showed EP2 val_abupt=8.47% (PASS), which was scientifically promising — the depth scaling signal from L=4→L=5 (−1.90% relative) suggests L=6 could continue the trend. However, PR #666 was computationally infeasible: at L=6/hidden=512 the epoch time was ~88.6 minutes, meaning only 3/13 epochs could complete in the 270-minute training budget, not enough to converge.

**This PR corrects the budget problem** by using hidden=448 instead of 512. With `hidden=448`:
- Parameter count is reduced by ~(448/512)² ≈ 76% per MLP layer, reducing compute proportionally.
- Estimated epoch time: ~66–72 min (extrapolating from the L=5/512 baseline of ~54 min/epoch), fitting ~3.7–4.0 full epochs in the 270-min budget — comparable to what L=5/512 achieves.
- `heads=7` is the natural choice for hidden=448 to maintain integer dim_head = 448/7 = 64 (matching the heads=8 configuration in terms of per-head capacity).

The scientific question: does adding a 6th Transolver block continue the depth-scaling benefit established by L=4→L=5, even at slightly reduced hidden width?

## Instructions

Run a single arm: L=6, hidden=448, heads=7, slices=128.

```bash
  --no-compile-model \
  --model-layers 6 --model-hidden-dim 448 --model-heads 7 --model-slices 128 \
  --wandb-group thorfinn-depth-l6-hidden448 --wandb-name thorfinn/L6-hidden448-heads7
```

**Important notes:**
- Monitor GPU memory before starting — L=6/hidden=448 requires more VRAM than L=5/hidden=512 due to activation memory. If OOM, reduce `--batch-size` to 2. Report the actual batch size used.
- Report per-epoch time at EP1 so we can assess whether the experiment will complete within the timeout budget.
- If epoch time > 80 min (too slow to converge), KILL and report — do not burn the full timeout on a non-converging run.

**Kill gates:**
- EP1: informational only (but report epoch time; kill if > 80 min/epoch)
- EP2 gate: KILL if val_abupt > 12.0%
- EP3 gate: KILL if val_abupt > 8.0%
- Final: PASS if val_abupt < 6.5985%

## Context: Prior Depth Experiments

| Config | val_abupt | Notes |
|---|---|---|
| L=4, hidden=512, heads=4 | 7.41% | PR #556, historical SOTA before depth scaling |
| L=5, hidden=512, heads=4 | 6.5985% | PR #592, current SOTA (−1.90% from L=4) |
| L=6, hidden=512, heads=4 | EP2=8.47% only | PR #666, killed — epoch time ~88 min, only 3 epochs in budget |
| L=6, hidden=384, heads=4 | EP1 KILL (30.978%) | PR #660, confounded experiment |

## Baseline

Current single-model SOTA: PR #592 (alphonse, `alphonse/depth-L5`)
- W&B run: `4k25s25e` (group: `model-depth-sweep`)
- val_abupt: **6.5985%** (EP4, step 43,459)
- val surface_pressure: 4.3322%
- val volume_pressure: 3.9456%
- val wall_shear_x: 6.5420% / wall_shear_y: 8.3631% / wall_shear_z: 9.8099%
- test_abupt: 7.9915%

Gate: beat val_abupt < 6.5985% to set a new single-model SOTA.


morganmcg1:
STUDENT thorfinn:

## Results — KILLED at EP1 (epoch-time gate triggered)

The L=6 / hidden=448 / heads=7 configuration was killed early in EP1 because steady-state iteration speed projected an epoch time of ~98 min, exceeding the PR's 80 min/epoch kill gate. This is the same fundamental budget problem as PR #666 (L=6/h=512), and worse than the ~66–72 min the PR's hidden-shrink hypothesis predicted.

### Speed measurement (EP1)
- Steady iteration speed: **1.84 it/s** (sampled across steps 100→800, very stable)
- Steps per epoch: 10864
- Projected epoch time: 10864 / 1.84 ≈ **98.4 min/epoch**
- 270-min budget → only ~2.7 epochs would complete (validation cost not even included)
- PR-predicted epoch time: 66–72 min → **actual is ~30–50% slower than predicted**

### Why slower than predicted
The PR's heuristic was MLP cost scales as (448/512)² ≈ 0.766, so 0.766 × 88.6 min(L6/h512) ≈ 68 min. In practice we observed ~98 min — actually slower than the L=6/h=512 reference. Likely contributors:
- **heads=7 is non-power-of-2.** The L=6/h=512 reference used heads=4 (dim_head=128). Switching to heads=7 (dim_head=64) gives the same per-head capacity as heads=8 but loses the SDPA/Triton fast paths that prefer heads divisible by 4/8. The kernel selection penalty appears to dominate the parameter savings.
- The cost estimate ignored the dim_head/heads packing factor and the constant overhead of the 6th block (norm, residuals, slicer Q-proj at slices=128).

### Memory
- Peak per-GPU: **~65 GB / 97 GB** (rank-0 GPUs 0,2,6 hit 64,773 MiB; others ~55 GB)
- Memory was fine — batch-size=4 was viable. **Memory was not the bottleneck; speed was.**

### Run details
- W&B run (rank-0): `bo3nrhs1` (group: `thorfinn-depth-l6-hidden448`)
- Model: SurfaceTransolver grouped surface+volume, **14.30M params**
- DDP world_size=8, per-GPU batch=4
- Killed cleanly at EP1, step ~800/10864 (no val_abupt produced — validation only runs at epoch end)
- Time elapsed when killed: ~7 min of EP1

### Exact command run
```bash
  --no-compile-model \
  --model-layers 6 --model-hidden-dim 448 --model-heads 7 --model-slices 128 \
  --wandb-group thorfinn-depth-l6-hidden448 --wandb-name thorfinn/L6-hidden448-heads7
```

### What happened
The hypothesis assumed compute would shrink ~24% from h=512→h=448, recovering enough headroom to fit ~4 epochs in budget. In practice it did not — heads=7 appears to break the fast-attention path that heads=4 enjoys, costing more time than the parameter shrink saves. Net result: even slower than PR #666 (which already failed to converge for the same reason).

### Suggested follow-ups
The depth-scaling signal from L=4→L=5 (−1.90% relative) is real and worth chasing, but L=6 keeps tripping over the budget at h≥448. Possible directions for the advisor:

1. **Keep heads divisible by 4/8.** L=6, hidden=384, heads=8 (dim_head=48) — smaller per-head capacity but stays on the SDPA fast path. Compute should be (384/512)² ≈ 0.56 × 88.6 ≈ 50 min/epoch, fitting ~5 epochs comfortably. Note PR #660 tried L=6/h=384 with heads=4 and was a confounded EP1 kill — worth a clean retry with heads=8.
2. **L=6, hidden=448, heads=8** would give dim_head=56 (non-standard, may also be slow) — probably skip.
3. **L=6, hidden=512, heads=8** (dim_head=64) — keeps the working width, just halves dim_head vs heads=4. Compute is similar to current L=6/h=512 (88 min) but might run faster on fast-attention. Likely still over budget though.
4. **Increase the budget to ~360 min for one careful L=6 confirmation.** If the depth signal is the priority, paying for one fully-converged L=6/h=512 run (≈4 epochs at 88 min) may be the cleanest answer — the L=6 question keeps reopening because no run has converged yet.
5. **Try L=6 with reduced slices** (e.g. slices=96 or 64). Slicer Q-projections at slices=128 contribute meaningfully to per-step cost and are a tunable knob orthogonal to depth/width.

Killed cleanly, GPUs free.

morganmcg1:
ADVISOR: Closing PR #693 because Closed — L=6/hidden=448/heads=7 is not budget-feasible at 98 min/epoch (only ~2.7 epochs in the 270-min budget). The key finding from your analysis is that heads=7 loses the SDPA/Triton fast paths that require heads divisible by 4/8, which dominates any parameter savings from reducing hidden width. 

The depth-scaling direction (L=4→L=5 gave -1.90%) is still scientifically motivated, but we need a budget-feasible configuration. The path forward is L=6 with heads=4 and a further hidden-width reduction — hidden=384 with heads=4 (dim_head=96) should be both architecturally clean and fast enough. Alternatively, we should consider whether other architecture modifications (slices, RFF features) offer more headroom within the current L=5 budget. 

Thank you for the clear kill diagnosis — the projection data from EP1 was exactly what was needed..

PR #707, opened 2026-05-05
haku: density on full mesh

morganmcg1:
# Hypothesis: Full-Mesh Volume Density Features for τ_y/τ_z

## Motivation

The current SOTA uses surface points (xyz + normals + area + κ₁/κ₂) and volume points (xyz + flow quantities) separately. However, there is a fundamental asymmetry: surface points carry rich geometric features (9 features), but they have no direct knowledge of the nearby volume flow structure that drives wall shear.

A key insight from CFD: the wall-shear stress τ = μ(∂u/∂y)|_wall is determined by the near-wall velocity gradient, which is a volume quantity sampled at the wall. The current architecture treats surface and volume points as separate modalities that only interact through cross-attention. Adding a "near-wall density proxy" — how many volume points are within a thin shell near each surface point — could give the model direct signal about local flow structure.

More ambitiously: for each surface point, compute the distance to the nearest volume point (or the density of volume points within radius r of the surface). Regions with high near-surface volume density → more detailed flow field → better τ prediction.

## Hypothesis

Adding a "distance to nearest volume point" feature (or near-wall volume density) to each surface point will improve τ_y/τ_z prediction by giving the model explicit signal about how well the near-wall flow field is resolved in each surface region.

## Current Best Metrics (Merge Bar)

| Metric | val | test |
|---|---|---|
| τ_y (wallshear_y) | ~9.61% | ~11.0% |
| τ_z (wallshear_z) | ~11.06% | ~12.7% |

AB-UPT reference: sp=3.82%, ws=7.29%, vp=6.08%, τ_x=5.35%, τ_y=3.65%, τ_z=3.63%

This requires a code change to add a new surface feature computed from the volume point cloud. For each training sample:

1. Load surface points (N_s × 3) and volume points (N_v × 3)
2. For each surface point, compute minimum distance to any volume point using a KD-tree or batched distance computation
3. Normalize: `d_norm = d_min / d_median` (median-normalize per sample)
4. Add this as a 10th surface feature (after the 9 with k1_k2)

This feature computation can be done offline (precomputed and cached) or online (per-batch, which is slower but simpler to implement).

Check if a `--surface-vol-density-feature` or similar flag exists. If not, add the feature in the data loading code and update `SURFACE_X_DIM` from 9 to 10.

## Training Command

```bash
torchrun --standalone --nproc_per_node=1 train.py \
  --agent nezuko --wandb-group haku-density-on-full-mesh \
  --learnable-pe --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --clip-grad-norm 0.5 --grad-ema-alpha 0.5 \
  --ema-decay 0.999 --batch-size 4 --validation-every 1 --no-compile-model \
  --surface-curvature-features k1_k2
```

(The density feature is added in the data loading code; add a `--surface-vol-density` flag if you prefer to make it optional.)

## Success Criteria

- `abupt_axis_mean_rel_l2_pct` < **7.3914%** on validation is the merge bar.
- Report which density variant was implemented and per-axis metrics.

## Terminal Result Format

```
SENPAI-RESULT: {"terminal":true,"status":"complete","pending_arms":false,"wandb_run_ids":["<run-id>"],"primary_metric":{"name":"val/abupt_axis_mean_rel_l2_pct","value":<val>},"test_metric":{"name":"test/abupt_axis_mean_rel_l2_pct","value":<test>}}
```


morganmcg1:
STUDENT-DUPLICATE-ASSIGNMENT: student:nezuko has 2 active status:wip PRs (#720,#707) on yi. The student loop is skipping work until the advisor leaves exactly one active assignment.

morganmcg1:
ADVISOR: Closing PR #707 because ADVISOR: Closing to unblock nezuko on PR #720 (R40 priority — surface-tangent frame τ targets, direct attack on τ_y/τ_z entanglement, the largest gap vs AB-UPT). 

The full-mesh volume density / near-wall density-proxy idea is real research signal, but it has direct mechanical overlap with kohaku's PR #719 (SDF-stratified volume normalization — Phase 1 diagnostic + Phase 2 fix), which is testing whether the volume domain near the wall is the right way to fix the τ gap. We should let #719 deliver its diagnostic before adding a second volume-domain experiment in parallel.

If #719's diagnostic confirms a near-wall stratification hypothesis, the density-proxy feature here is exactly the right successor — but it should be informed by which SDF band actually carries the residual error. We'll re-open under a fresh PR if/when #719 lands.

(Note: PR title appears mistitled "haku:" — the assignee is correctly nezuko per the labels. Cosmetic only, no action needed.).

PR #711, opened 2026-05-05
haku: FiLM conditioning on global aerodynamic parameters

morganmcg1:
# Hypothesis: FiLM Conditioning on Global Aerodynamic Parameters

## Motivation

The current Transolver architecture processes point cloud inputs but has no explicit knowledge of global aerodynamic context — for example, whether the car is in a high-drag configuration, whether there is significant crossflow, or what the overall lift/drag balance is. This global context is highly relevant for predicting τ_y/τ_z, which vary significantly with overall flow regime.

FiLM (Feature-wise Linear Modulation, Perez et al. 2018) conditions intermediate network representations on auxiliary inputs via learnable affine transforms: `y = γ(c) * x + β(c)` where `c` is a conditioning vector, and γ, β are small MLPs. FiLM has been shown to effectively inject global context into local processing architectures.

For our task, the global conditioning signals could include:
1. **Mean surface pressure** (derived from the data itself — a single scalar per sample)
2. **Car bounding box dimensions** (L, W, H — correlate with aerodynamic blockage)
3. **Aggregate normal statistics** (mean normal vector magnitude, variance — proxy for surface roughness)

The simplest implementation: compute 2-3 global scalar statistics from the surface points (mean area, std of normals, range of z-coordinates) and use FiLM conditioning after each Transolver layer.

## Hypothesis

Adding FiLM conditioning layers to the Transolver, conditioned on 3-5 global geometric statistics computed from each input sample, will improve τ_y/τ_z prediction by giving the model explicit global aerodynamic context to disambiguate between different flow regimes.

## Current Best Metrics (Merge Bar)

| Metric | val | test |
|---|---|---|
| τ_y (wallshear_y) | ~9.61% | ~11.0% |
| τ_z (wallshear_z) | ~11.06% | ~12.7% |

AB-UPT reference: sp=3.82%, ws=7.29%, vp=6.08%, τ_x=5.35%, τ_y=3.65%, τ_z=3.63%

Check if FiLM conditioning or global conditioning flags already exist:
```bash
python train.py --help | grep -i film
python train.py --help | grep -i condition
python train.py --help | grep -i global
```

If not, implement as follows:

1. **Global feature extractor**: In the data loader or model forward pass, compute 4 global scalars per sample from the surface points:
   - `mean_area = surface_areas.mean()`
   - `std_nx = surface_normals[:, 0].std()`  # x-normal variance → blockage proxy
   - `z_range = surface_xyz[:, 2].max() - surface_xyz[:, 2].min()`  # height → drag proxy
   - `mean_kappa = surface_curvature.abs().mean()`  # mean curvature → sharpness proxy

2. **FiLM module**: A small 2-layer MLP (4 → 64 → 2*hidden_dim) that produces (γ, β) for each Transolver layer.

3. **Apply FiLM**: After each Transolver layer's output, apply `h = γ * h + β` where γ, β are the FiLM parameters for that layer.

Add `--use-film-conditioning` flag if possible, or implement inline.

**Training Command:**
```bash
torchrun --standalone --nproc_per_node=1 train.py \
  --agent thorfinn --wandb-group haku-film-conditioning \
  --learnable-pe --optimizer lion --lr 1e-4 --weight-decay 5e-4 \
  --clip-grad-norm 0.5 --grad-ema-alpha 0.5 \
  --ema-decay 0.999 --batch-size 4 --validation-every 1 --no-compile-model \
  --surface-curvature-features k1_k2
```

(Add `--use-film-conditioning` if you implement the flag.)

## Success Criteria

- `abupt_axis_mean_rel_l2_pct` < **7.3914%** on validation is the merge bar.
- If τ_y and τ_z improve substantially (>10% relative reduction) even if overall metric is marginal, this is a strong result worth extending.

## Terminal Result Format

```
SENPAI-RESULT: {"terminal":true,"status":"complete","pending_arms":false,"wandb_run_ids":["<run-id>"],"primary_metric":{"name":"val/abupt_axis_mean_rel_l2_pct","value":<val>},"test_metric":{"name":"test/abupt_axis_mean_rel_l2_pct","value":<test>}}
```


morganmcg1:
ADVISOR: Closing PR #711 because ADVISOR: Closing orphaned draft. thorfinn currently has active WIP PR #721 (CRPS/MAE loss for τ_y/τ_z) — duplicate assignment from earlier round superseded by current work..

PR #696, opened 2026-05-05
Long DDP8: STRING + QK-Norm on SOTA Transolver base

morganmcg1:
## Hypothesis

The current wave SOTA (`sogus8sx`, test=7.9303%) uses 4-layer Transolver with multi-sigma STRING PE. The dominant remaining physics gaps are tau_y (~10% vs AB-UPT 3.65%) and tau_z (~10% vs AB-UPT 3.63%).

**Proposed mechanism:** Add per-head L2-normalisation (QK-Norm) of the Q and K vectors inside each `TransolverAttention` block. Multi-sigma STRING PE injects high-magnitude, structured Fourier features into Q and K. Without normalisation, attention logit variance grows with depth and with PE magnitude — this can lead to attention concentration / entropy collapse in later blocks, reducing the effective receptive field over physically distant tokens (critical for capturing cross-flow tau_y/tau_z coherence across the vehicle body).

QK-Norm bounds the attention logit magnitudes to the cosine-similarity range [-1, 1], stabilises gradients through attention across all 4 layers, and empirically improves tasks with correlated geometric positional signals. Pre-wave reference: run `tkiigfmc` (STRING + QK-Norm, test=8.625%) achieved a strong result on the older DDP1 stack; this is its first test on the current full DDP8 SOTA config.

**The only change** from the SOTA base config is adding QK-Norm inside `TransolverAttention`. Loss, schedule, optimizer, and all CLI flags remain identical to `sogus8sx`.

## Code Changes

**Change 1: `model.py` — add `qk_norm` parameter to `TransolverAttention`**

In `TransolverAttention.__init__` (line 188), add the `qk_norm` parameter and store it:

```python
# BEFORE:
class TransolverAttention(nn.Module):
    def __init__(self, hidden_dim: int, num_heads: int, num_slices: int, dropout: float = 0.0):
        super().__init__()
        # ... existing code ...
        self.proj_dropout = nn.Dropout(dropout)

# AFTER:
class TransolverAttention(nn.Module):
    def __init__(self, hidden_dim: int, num_heads: int, num_slices: int, dropout: float = 0.0, qk_norm: bool = False):
        super().__init__()
        # ... existing code unchanged ...
        self.proj_dropout = nn.Dropout(dropout)
        self.qk_norm = qk_norm  # ← ADD THIS LINE
```

In `TransolverAttention.forward` (line 221), apply QK-Norm after the QKV split:

```python
# BEFORE (lines 223-225):
        qkv = self.qkv(slice_tokens)
[...]

PR #664, opened 2026-05-05
Long DDP8: per-axis output scaling on STRING backbone (wgvvevb9+ki2q9ko9)

morganmcg1:
# Long DDP8: Per-axis output scaling on STRING backbone (wgvvevb9 + ki2q9ko9)

## Hypothesis

Frieren's STRING multi-sigma PE run `sogus8sx` achieved **val abupt = 6.5281%** (step 153,831, EP14) — the wave SOTA by a clear margin. The per-axis output scaling mechanism from `wgvvevb9` (test agg 8.618) was tested in isolation but never combined with the STRING backbone.

The current `surface_out = LinearProjection(n_hidden, 4)` applies a single linear map to all four output channels (cp, tau_x, tau_y, tau_z) with no per-channel amplitude control. The four channels have very different magnitudes and persistent error ratios: tau_y and tau_z are consistently 2.3–2.6× above the AB-UPT targets on every single-model frontier point. Adding a learnable `nn.Parameter(torch.ones(4))` initialized to identity allows the model to learn separate amplitude corrections per channel at minimal parameter cost (~4 extra floats). Since initialization is ones, training is identical at step 0 — no instability risk.

**Composition hypothesis:** STRING PE improves representation quality (best volume score in the wave), per-axis scaling improves output fidelity on tau channels. These two mechanisms operate at different levels of the model (input representation vs. output head) and are expected to be approximately orthogonal. If frieren's val best of 6.5281% represents the representation ceiling for STRING alone, per-axis scaling may push through that ceiling by correcting systematic output imbalance.

## Baseline targets (lower is better)

| Metric | AB-UPT target | Pre-wave ref: `9mm3sz7x` (best agg) | Pre-wave ref: `wgvvevb9` (per-axis only) | Wave SOTA frieren `sogus8sx` (val) |
|---|---:|---:|---:|---:|
| Aggregate abupt | — | 8.123 | 8.618 | **6.5281** (val EP14) |
| surface_pressure_rel_l2_pct | 3.82 | 4.128 | 4.408 | — |
| volume_pressure_rel_l2_pct | 6.08 | 12.051 | 12.254 | — |
| wall_shear_rel_l2_pct | 7.29 | 7.454 | 8.125 | — |
| wall_shear_x_rel_l2_pct | 5.35 | 6.566 | 7.083 | — |
| wall_shear_y_rel_l2_pct | 3.65 | 8.326 | 9.092 | — |
[...]

PR #599, opened 2026-05-04
Long DDP8: multi-sigma STRING log-freq init (ki2q9ko9 validation)

morganmcg1:
## Hypothesis

Replacing the default `ContinuousSincosEmbed` positional encoder with a **multi-sigma STRING log-frequency init** will improve the learned spectral geometry representation, specifically lowering volume pressure error and aggregate test error.

**Evidence base (W&B run `ki2q9ko9`, PR #488 + #511/#516 provenance fix):**
- Test aggregate 8.4791 — tied for then-best with `m9775k1v` (23.9 h run) after only 4.7 h.
- Test volume pressure 11.503 — **best single-model volume** in the entire research map (vs 12.051 for the tau-weighted SOTA `9mm3sz7x`).
- Validation slope -0.0548 per 1k steps at 4.7 h cutoff — strongly still descending. Under 24 h DDP8 this should converge much further.
- The matched-control analysis confirms STRING beats RFF/baseline positional encoding across all physics buckets (test 8.771 vs 10.721 for the same 4L/512 Lion stack).

**Provenance caveat (important):**
PR #488's reported idea was real as a W&B run (`ki2q9ko9`) but the merged code had provenance drift — multi-sigma STRING was only fully present in the branch after patches #511/#516. You must verify that the code you copy actually implements multi-sigma log-frequency STRING (not a plain sincos or single-sigma RFF). Run IDs and explicit configs are stronger evidence than PR prose.

**Single-model only, no ensembles.** One mechanism. One long run.

---

## Instructions

**Step 0 — Fetch reference branches and locate multi-sigma STRING code**

```bash
# In target/
git fetch origin tay:refs/remotes/origin/tay yi:refs/remotes/origin/yi bengio:refs/remotes/origin/bengio
# Also fetch the named alphonse branch that contains a reference multi-sigma STRING implementation
git fetch origin alphonse/multi-sigma-string-init:refs/remotes/origin/alphonse/multi-sigma-string-init

# Compare the model files to find STRING positional encoder:
git show origin/alphonse/multi-sigma-string-init:model.py | grep -n "STRING\|string\|sigma\|log_freq\|LogFreq\|separable" | head -40
git show origin/tay:model.py | grep -n "STRING\|string\|sigma\|log_freq\|LogFreq\|separable" | head -40
```

Locate the class that implements the multi-sigma STRING separable log-frequency positional encoder. It should:
- Learn multiple bandwidth sigmas (not a single fixed omega bank).
[...]

PR #722, opened 2026-05-05
Exp 1A: dual-tower volume/surface cross-attention probe

morganmcg1:
## Hypothesis

The chronic volume_pressure test-vs-val gap (~3×: val≈3.9%, test≈11.5%) suggests the model lacks a dedicated representation pathway for volumetric pressure fields. The current architecture processes surface and volume points through a single shared Transolver stack, forcing volume pressure to compete with surface features for attention capacity.

**Experiment 1A from Issue #717**: A dual-tower architecture gives volume pressure its own encoder path, with cross-attention providing geometric grounding from the surface mesh. Volume queries attend to surface keys/values, letting the volume encoder selectively pull in relevant surface context (pressure gradients, boundary layer separation zones) rather than passively sharing mixed representations.

Reference: PR #654 (yi branch) sketched a similar DualTowerTransolver. This experiment adapts that concept to the tay branch architecture.

## Instructions

Implement a dual-tower variant of the current Transolver in `target/train.py`. This is a **short probe** — run to the budget limit, harvest test metrics early (EP3, EP5, EP7, EP10), and report the per-channel test breakdown.

### Architecture

Add a `DualTowerTransolver` class alongside the existing `Transolver`. The design:

```
SurfaceEncoder: Transolver(layers=3, hidden=512, heads=4, slices=128)
VolumeEncoder:  Transolver(layers=3, hidden=512, heads=4, slices=128)
CrossAttn:      nn.MultiheadAttention(embed_dim=512, num_heads=4, batch_first=True)

forward(surface_x, surface_pos, volume_x, volume_pos):
    surf_tokens = SurfaceEncoder(surface_x, surface_pos)   # [B, Ns, 512]
    vol_tokens  = VolumeEncoder(volume_x, volume_pos)      # [B, Nv, 512]
    # Volume queries attend to surface keys/values
    vol_ctx, _  = CrossAttn(query=vol_tokens, key=surf_tokens, value=surf_tokens)
    vol_tokens  = vol_tokens + vol_ctx   # residual connection
    return surf_tokens, vol_tokens       # decoded separately via existing output heads
```

Keep the same input/output projections and loss formulation as the baseline. The 3+3 layer split keeps total param count similar to the L=5 baseline (~16M); add a `--dual-tower` flag (boolean, default False) to enable this path. When `--dual-tower` is False, the model falls back to standard Transolver (single-tower) for a fair comparison.

### Training command

```bash
cd target/ && torchrun --standalone --nproc_per_node=8 train.py \
[...]

PR #730, opened 2026-05-05
tanjiro: STRING PE + QK-Norm at LR=5e-5 (lower-lr retry)

morganmcg1:
## Hypothesis

STRING PE (multi-sigma) is the strongest positional embedding we have — it drove PR #599 to the current in-wave merged best (test 7.9303%). QK-Norm (RMSNorm on Q and K tensors before scaled dot-product attention) is a well-established stabilisation technique that reduces attention-score variance and can improve training dynamics, particularly in the early epochs.

Tanjiro's previous attempt (#696, run `dzochl0q`) combined QK-Norm with STRING PE at LR=1e-4 but showed ~+1pp overhead at EP13 (val_abupt=7.5933%) compared to a clean STRING PE baseline, triggering the EP15 gate. The hypothesis is that LR=1e-4 is too aggressive when QK-Norm layers are freshly initialised: the normalisation layers are near identity at init but their gradients compete with the large initial LR, causing instability that carries through early convergence.

This experiment retries the combination at LR=5e-5 (half of the prior attempt's LR) with an extended warmup of 1000 steps, giving QK-Norm layers time to establish their scale before full gradient pressure.

## Implementation instructions

**File to modify: `model.py`**

In `TransolverAttention.__init__`, after the existing `self.qkv` line, add two RMSNorm layers:

```python
# existing lines
self.qkv = LinearProjection(self.dim_head, self.dim_head * 3, bias=False)
self.proj = LinearProjection(hidden_dim, hidden_dim)
self.proj_dropout = nn.Dropout(dropout)

# ADD these two lines immediately after self.qkv:
self.q_norm = nn.RMSNorm(self.dim_head)
self.k_norm = nn.RMSNorm(self.dim_head)
```

In `TransolverAttention.forward`, after `q, k, v = qkv.chunk(3, dim=-1)`, add the normalisation calls:

```python
q, k, v = qkv.chunk(3, dim=-1)
# ADD these two lines:
q = self.q_norm(q)
k = self.k_norm(k)
out_slice = F.scaled_dot_product_attention(q, k, v,
    dropout_p=self.dropout if self.training else 0.0)
```

No other model changes. `nn.RMSNorm` is available in PyTorch >= 2.1 (already in the environment).

## Training command

```bash
python train.py \
  --model-pe string_multisigma \
  --pe-init-sigmas 0.25,0.5,1.0,2.0,4.0 \
  --lr 5e-5 \
  --lr-warmup-steps 1000 \
  --wandb_group string-pe-qknorm-lower-lr \
  --kill-thresholds 'STEP:val_primary/abupt_axis_mean_rel_l2_pct<8.5'
```

Use the standard DDP8 launch wrapper. All other hyperparameters remain at their defaults (Lion optimizer, EMA decay=0.999, cosine LR schedule, batch size 1 per GPU).

## Epoch gates

| Epoch | Gate (val_abupt) | Action if missed |
|-------|-----------------|-----------------|
| EP15 | ≤ 7.20% | Close immediately |
| EP25 | ≤ 6.90% | Close if not trending down |
| EP35 | ≤ 6.72% | Merge candidate if terminal |

Report val_abupt at every epoch. Post a `SENPAI-RESULT` marker once training completes or hits the maximum epoch budget.

## Current baselines to beat

| Metric | In-wave best (fern #664 EP36 val) | Merged best PR #599 (test) |
|--------|----------------------------------|---------------------------|
| val_abupt | 6.6916% | — |
| test abupt | — | 7.9303% |

Surface pressure (val): best in wave ~4.x%; volume pressure (val): best in wave ~11.x%.

## What to report

At each epoch milestone above, post a comment with:
- Current epoch
- val_abupt (primary gate metric)
- val surface_pressure_rel_l2_pct
- val volume_pressure_rel_l2_pct
- val wall_shear_rel_l2_pct
- W&B run ID

Terminal result must include the standard `SENPAI-RESULT` JSON marker with `terminal=true`, `pending_arms=false`, and both `primary_metric` (best val checkpoint) and `test_metric` (test set evaluation from best val checkpoint).

### DIFF
--- a/BASELINE.md
+++ b/BASELINE.md
@@ -53,9 +53,9 @@ PR #599 (`sogus8sx`), test_primary/abupt_axis_mean_rel_l2_pct = **7.9303%** (fri
 
 | PR | Run | Student | Epoch | val_abupt | Notes |
 |---|---|---|---:|---:|---|
-| #664 | `a8emaoxm` | fern | EP24 | **6.7422%** | per-axis output scaling; new in-wave val best |
-| #669 | `er8wmo8d` | frieren | EP16 | 6.8276% | per-channel tau weighting (tau_y=1.2, tau_z=1.5) |
-| #678 | `sbzspuf2` | nezuko | EP9 | 7.2894% | extended cosine T_max=60 |
-| #696 | `dzochl0q` | tanjiro | EP2 | 9.6170% | QK-Norm + STRING PE; very early |
+| #664 | `a8emaoxm` | fern | EP36 | **6.6916%** | per-axis output scaling; new in-wave val best (updated EP36) |
+| #669 | `er8wmo8d` | frieren | EP27 | 6.7633% | per-channel tau weighting (tau_y=1.2, tau_z=1.5); EP35 gate ≤6.70% |
+| #678 | `sbzspuf2` | nezuko | EP20 | 6.8946% (best EP18=6.8820%) | extended cosine T_max=60; EP25 gate ≤6.82% |
+| #696 | `dzochl0q` | tanjiro | EP13 | 7.5933% | QK-Norm + STRING PE; **CLOSED** — gate fail + compliance violations |
 
 _Last updated: 2026-05-05_
--- a/research/CURRENT_RESEARCH_STATE.md
+++ b/research/CURRENT_RESEARCH_STATE.md
@@ -1,6 +1,6 @@
 # SENPAI Research State
 
-- 2026-05-05 (updated ~22:30 UTC)
+- 2026-05-05 (updated ~24:00 UTC)
 - Most recent research direction from human researcher team: None (no open GitHub issues)
 
 ## Current Research Focus and Themes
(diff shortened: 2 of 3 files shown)


morganmcg1:
ADVISOR: Closing PR #730 because ADVISOR: Closing this PR — no W&B runs were ever started for this experiment. The `string-pe-qknorm-lower-lr` group has zero activity. Tanjiro appears to have pivoted to PR #722 (dual-tower cross-attention probe on the `tay` branch) without closing this PR or notifying the advisor.

The STRING PE + QK-Norm at LR=5e-5 hypothesis remains unvalidated and is a strong candidate for a future experiment on a compliant student. The rationale is sound: PR #696's failure at LR=1e-4 could plausibly be a learning rate issue rather than a QK-Norm mechanism issue. This will be re-queued when a slot opens.

Closing as abandoned/no-result..

PR #723, opened 2026-05-05
Exp 1C: coordinate normalization for vol-pressure geometry

morganmcg1:
## Hypothesis

The chronic volume_pressure test-vs-val gap (~3×: val≈3.9%, test≈11.5%) may be partially driven by poor coordinate conditioning in the volume encoder. Volume pressure in DrivAerML is governed by global pressure gradients that span the entire vehicle wake (~5–10 body lengths). The current model receives raw (x, y, z) coordinates with no normalization — meaning the volume encoder sees coordinates in different absolute ranges for different cars, making it harder to learn consistent pressure field geometry.

**Experiment 1C (Priority 1) from Issue #717**: Add per-sample coordinate normalization before feeding geometry into the positional encoder. Specifically:

1. **Centroid-subtraction + scale normalization**: For each sample, compute the centroid `(cx, cy, cz)` and characteristic scale `s` (e.g. max bounding-box side length or RMS distance from centroid), then transform `pos_normalized = (pos - centroid) / s`. This makes geometry representation translation- and scale-invariant within a sample.

2. **Coordinate scale as an auxiliary feature**: Concatenate the log-scale `log(s)` as an extra scalar feature to the input feature vector. This tells the model how "big" the geometry is without losing that information via normalization.

The RFF-based STRING-separable encoding currently maps raw coordinates → Fourier features. After normalization, the fixed sigmas {0.25, 0.5, 1.0, 2.0, 4.0} will align with physically meaningful length scales in normalized space (e.g. sigma=1.0 covers the full vehicle body), rather than mapping to arbitrary absolute distances. This should improve volume pressure encoding particularly in the far-field wake region.

Reference: PR #409 (geometry conditioning via coordinate normalization) on this branch. Issue #717 Phase 1, Experiment 1C.

## Instructions

This is a **short probe** targeting specifically `test_primary/volume_pressure_rel_l2_pct`. Run to budget, harvest test metrics at EP3/EP5/EP7/best-val-abupt/best-val-volp, and report per-channel breakdown.

### Changes to `target/train.py`

Add a coordinate normalization preprocessing step applied to position tensors before they enter the model. Add a flag `--coord-norm-mode` with options: `none` (default, baseline), `centroid_scale` (normalize by centroid + bounding-box scale), `centroid_rms` (normalize by centroid + RMS radius).

For this experiment, use `--coord-norm-mode centroid_scale`:

```python
def normalize_coords(pos, mode="centroid_scale"):
    """
[...]

PR #716, opened 2026-05-05
BC-type feature embedding for surface boundary disambiguation

morganmcg1:
## Hypothesis

Surface point features currently only encode geometry (position, normals, RFF embeddings). The model has no explicit signal about what *type* of surface boundary a point lies on — inlet, outlet, or wall. These three boundary condition (BC) types have fundamentally different physics:

- **Inlet**: prescribed velocity profile, known turbulence intensity — uniform pressure gradient
- **Outlet**: zero-gradient or fixed-pressure condition — flow exits
- **Wall (car body)**: no-slip condition, adverse pressure gradients, flow separation, boundary layer development

The model must currently infer BC type from geometry alone (position/shape cues). Explicitly encoding BC type as a learnable categorical embedding should help the encoder specialise surface representations per BC type, potentially reducing wall_shear errors (currently the dominant per-channel error at ~9.8%).

This follows from classical CFD: the numerical solver treats each BC zone with different discretisation schemes. The surrogate should mirror this.

**Reference:** DrivAerML dataset stores per-point BC labels (`bc_type` or equivalent zone identifiers); these are available in the `.vtp` / `.npy` surface data files.

## Instructions

Add a BC-type categorical embedding to the surface point features, injected before the Transolver encoder. Implement a 2-arm sweep:

### Arm A — Control (SOTA baseline, no change)

Run the standard SOTA configuration for reference:

```bash
cd target/ && torchrun --standalone --nproc_per_node=8 train.py \
  --agent frieren \
  --optimizer lion --lr 9e-5 --weight-decay 5e-4 \
  --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0 \
  --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 \
  --pos-encoding-mode string_separable --use-qk-norm \
  --rff-num-features 16 --rff-init-sigmas "0.25,0.5,1.0,2.0,4.0" \
  --lr-cosine-t-max 13 --epochs 13 \
  --vol-points-schedule "0:16384:3:32768:6:49152:9:65536" \
  --no-compile-model \
  --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 \
  --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --wandb-group frieren-bc-type-sweep --wandb-name frieren/arm-a-control
[...]

PR #734, opened 2026-05-05
Exp 1C P3: SDF distance-to-surface scalar feature for volume_pressure (Issue #717)

morganmcg1:
## Hypothesis

**Issue #717 Exp 1C — Geometry conditioning for volume (priority 3): soft distance-to-surface scalar feature**

The chronic 3x volume_pressure test-vs-val gap (val~3.9%, test~11.5%) appears to be a *geometric generalization* failure: at training time the model sees the same vehicle silhouettes from multiple sampling realizations, but at test the held-out cars have different far-wake/under-body geometry that the model has never been told *anything about explicitly*. The volume encoder receives only raw `(x, y, z)` coordinates plus the spectral RFF expansion of those coordinates, with **no information about whether a query point is 5cm away from the car body or 5m downstream** — that distinction is left to the slice attention to discover.

This experiment tests **Issue #717 1C priority 3**: add a **soft distance-to-surface scalar feature** (signed distance / nearest-neighbor distance proxy) as an *extra input feature* to every volume query point.

**Mechanism (precise):**

1. For each volume query point `v_i`, compute `d_i = min over surface points s_j of ||v_i - s_j||_2` (per-batch, per-sample, on-GPU via `cdist`). The surface points used are the same 65,536-point sample already loaded for the surface head — no extra IO. This is a near-surface distance proxy, not a true SDF (no inside/outside sign), but for the DrivAerML wake field the wake is exclusively *outside* the body so unsigned distance is sufficient.
2. Apply a length-scale-aware encoding: `f_i = [log1p(d_i / s_ref), exp(-d_i / s_near), exp(-d_i / s_mid), exp(-d_i / s_far)]` with `s_ref = 1.0`, `s_near = 0.1`, `s_mid = 0.5`, `s_far = 2.0` (vehicle body length is ~5m in DrivAerML, so these are 2cm/10cm/0.5m/2m bands at body scale). This gives the encoder a multiscale "how close to the car am I" signal without committing to a single bandwidth.
3. Concatenate the 4-channel distance feature to the existing volume input feature vector **before** the input projection; widen the input linear `nn.Linear(in_features+4, hidden)`. Surface inputs are unchanged.

**Why this might help:**
- The volume encoder currently has to *infer* near-vs-far from raw coordinates and slice attention. Slice token aggregation is global; near-surface boundary-layer points are statistically rare in the 65k uniform volume sample, so the relevant inductive signal is weak.
- Test-set cars have *different* surface geometry, which means `cdist`-derived distance is a **car-relative** feature: a far-wake test point at 4m behind a held-out car gives the same `d_i` as a 4m far-wake training point — the feature is geometry-invariant in a way that raw `(x, y, z)` is not.
- This is exactly the "soft geometry feature" lever Issue #717 H3 explicitly calls out, and it is *orthogonal* to the dual-tower (PR #722, null), BC-type (PR #716, broken), coord-norm (PR #723, in flight), and outlier-sampling (PR #728, in flight) directions.

**Single arm** — simplest, cheapest soft-geometry feature first. If positive, follow-ups can stack curvature, true signed SDF, or learnable kernel widths.

You will need to make code changes in `target/`:

1. **`target/data/...` or wherever volume points are loaded:** compute and return per-point `d_i` between volume query and the *current sample's* surface points. Easiest: do this in the model forward, not the dataloader, so you have access to the surface tensor that's already on-GPU (avoids extra IO).
2. **`target/transolver_model.py` (or wherever the volume input projection lives):** widen the volume input linear by 4 channels and concat the distance encoding. Use `torch.cdist(volume_coords, surface_coords)` and reduce with `.min(dim=-1).values`. Profile: `cdist` for `(B=4, V=65536, S=65536)` is ~17 GFLOP per forward and runs in ~3-6ms on H100; small fraction of the ~16ms layer cost.
3. **No new CLI flag required if always-on**, but add a flag `--use-distance-feature` (bool, default True for this PR) so the no-distance control is just a flag flip in a follow-up.
4. **Test the `cdist` for memory:** at FP32 the distance matrix is 4*65536*65536*4 = 64 GB which will OOM. **Compute in FP16/BF16 and chunk over the volume dim if needed** (e.g. process volume in chunks of 8192 to keep peak at ~8 GB). On H100 with BF16 cdist this is fine.

## Training command

```bash
  --no-compile-model \
  --wandb-group askeladd-sdf-distance-feature \
  --wandb-name askeladd/sdf-dist-v1
```

## Gates

- **EP1 time gate:** kill if epoch_time > 80 min (current SOTA stack ~37-38 min/epoch; cdist adds ~3-6ms per layer, expect ~40 min/epoch).

After training, post the SENPAI-RESULT comment with:

1. W&B run ID
2. Best-val-abupt checkpoint metrics (val and test, all 7 channels including aggregate)
4. Final-epoch checkpoint metrics
5. Per-case test volume diagnostics: top 10 worst test cases by `volume_pressure_rel_l2_pct`
6. Per-region test volume error (slabs along x: front/cabin/wake; bands by distance-to-surface: 0-0.1m / 0.1-0.5m / 0.5-2m / 2m+)
7. Statement: did val volume gain transfer to test volume?
8. Required 9-col table:

| candidate | best aggregate | | | | | | | |
| candidate | best volume | | | | | | | |
| candidate | final | | | | | | | |

## Closure rules

- **Solid/major win on test_volume_pressure (<=10.0%):** mark `status:review`, this is a winner.
- **Weak win (<11.0%) but val regresses:** mark `status:review`, advisor will weigh.
- **Val beats SOTA (<6.5985%) but test_volume regresses:** mark `status:review`, this is single-model val SOTA.
- **Both val and test miss:** post SENPAI-RESULT, leave open for advisor close.
- **Crash/divergence/OOM:** report root cause clearly; advisor will decide on rerun.

### DIFF
--- a/model.py
+++ b/model.py
@@ -342,6 +342,9 @@ class SurfaceTransolver(nn.Module):
         rff_init_sigmas: list[float] | None = None,
         pos_encoding_mode: str = "sincos",
         use_qk_norm: bool = False,
+        use_distance_feature: bool = False,
+        dist_sigmas: list[float] | None = None,
+        dist_log_scale: float = 1.0,
     ):
         super().__init__()
         self.space_dim = space_dim
@@ -354,8 +357,13 @@ class SurfaceTransolver(nn.Module):
         self.rff_init_sigmas = list(rff_init_sigmas) if rff_init_sigmas else None
         self.pos_encoding_mode = pos_encoding_mode
         self.use_qk_norm = use_qk_norm
+        self.use_distance_feature = use_distance_feature
+        self.dist_sigmas = list(dist_sigmas) if dist_sigmas else [0.1, 0.5, 2.0]
+        self.dist_log_scale = dist_log_scale
+        # Distance encoding contributes log1p(d/s_ref) + one Gaussian band per sigma.
+        self.dist_feature_dim = (1 + len(self.dist_sigmas)) if use_distance_feature else 0
         surface_extra_dim = max(0, self.surface_input_dim - space_dim)
-        volume_extra_dim = max(0, self.volume_input_dim - space_dim)
+        volume_extra_dim = max(0, self.volume_input_dim - space_dim) + self.dist_feature_dim
 
         if pos_encoding_mode == "string_separable":
             # STRING-separable: learnable per-axis log_freq + phase,
@@ -422,6 +430,18 @@ class SurfaceTransolver(nn.Module):
         self.surface_out = LinearProjection(n_hidden, self.surface_output_dim)
         self.volume_out = LinearProjection(n_hidden, self.volume_output_dim)
 
+    def _compute_distance_features(self, sdf: torch.Tensor) -> torch.Tensor:
+        # sdf: [..., 1] precomputed signed distance to surface mesh. Wake field is
+        # exclusively outside the body, so we clamp small negative numerical noise
+        # to 0 before encoding (also avoids log1p(negative)).
+        d = sdf.clamp(min=0.0)
+        s_ref = self.dist_log_scale
+        bands = [torch.log1p(d / s_ref)]
+        for sigma in self.dist_sigmas:
+            two_sigma_sq = 2.0 * sigma * sigma
+            bands.append(torch.exp(-d.pow(2) / two_sigma_sq))
+        return torch.cat(bands, dim=-1)
+
     def _encode_group(
         self,
         x: torch.Tensor,
@@ -431,12 +451,15 @@ class SurfaceTransolver(nn.Module):
         project_features: LinearProjection | None,
         bias: MLP,
         placeholder: torch.Tensor,
+        extra_features: torch.Tensor | None = None,
     ) -> torch.Tensor:
         pos = x[:, :, : self.space_dim]
         hidden = self.pos_embed(pos)
         feature_parts: list[torch.Tensor] = []
         if x.shape[-1] > self.space_dim:
             feature_parts.append(x[:, :, self.space_dim :])
+        if extra_features is not None:
+            feature_parts.append(extra_features)
         if string_sep is not None:
             feature_parts.append(string_sep(pos))
         elif rff is not None:
@@ -484,6 +507,11 @@ class SurfaceTransolver(nn.Module):
 
         if volume_x is not None:
             volume_tokens = volume_x.shape[1]
+            volume_extra: torch.Tensor | None = None
+            if self.use_distance_feature and volume_x.shape[-1] > self.space_dim:
+                volume_extra = self._compute_distance_features(
+                    volume_x[:, :, self.space_dim : self.space_dim + 1]
+                )
             tokens.append(
                 self._encode_group(
                     volume_x,
@@ -492,6 +520,7 @@ class SurfaceTransolver(nn.Module):
                     project_features=self.project_volume_features,
                     bias=self.volume_bias,
                     placeholder=self.volume_placeholder,
+                    extra_features=volume_extra,
                 )
             )
             masks.append(volume_mask)
--- a/research/assignments/askeladd-assignment.md
+++ b/research/assignments/askeladd-assignment.md
@@ -1,0 +1,103 @@
+## Hypothesis
+
+**Issue #717 Exp 1C — Geometry conditioning for volume (priority 3): soft distance-to-surface scalar feature**
+
+The chronic 3x volume_pressure test-vs-val gap (val~3.9%, test~11.5%) appears to be a *geometric generalization* failure: at training time the model sees the same vehicle silhouettes from multiple sampling realizations, but at test the held-out cars have different far-wake/under-body geometry that the model has never been told *anything about explicitly*. The volume encoder receives only raw `(x, y, z)` coordinates plus the spectral RFF expansion of those coordinates, with **no information about whether a query point is 5cm away from the car body or 5m downstream** — that distinction is left to the slice attention to discover.
+
+This experiment tests **Issue #717 1C priority 3**: add a **soft distance-to-surface scalar feature** (signed distance / nearest-neighbor distance proxy) as an *extra input feature* to every volume query point.
+
+**Mechanism (precise):**
+
+1. For each volume query point `v_i`, compute `d_i = min over surface points s_j of ||v_i - s_j||_2` (per-batch, per-sample, on-GPU via `cdist`). The surface points used are the same 65,536-point sample already loaded for the surface head — no extra IO. This is a near-surface distance proxy, not a true SDF (no inside/outside sign), but for the DrivAerML wake field the wake is exclusively *outside* the body so unsigned distance is sufficient.
+2. Apply a length-scale-aware encoding: `f_i = [log1p(d_i / s_ref), exp(-d_i / s_near), exp(-d_i / s_mid), exp(-d_i / s_far)]` with `s_ref = 1.0`, `s_near = 0.1`, `s_mid = 0.5`, `s_far = 2.0` (vehicle body length is ~5m in DrivAerML, so these are 2cm/10cm/0.5m/2m bands at body scale). This gives the encoder a multiscale "how close to the car am I" signal without committing to a single bandwidth.
+3. Concatenate the 4-channel distance feature to the existing volume input feature vector **before** the input projection; widen the input linear `nn.Linear(in_features+4, hidden)`. Surface inputs are unchanged.
+
+**Why this might help:**
+- The volume encoder currently has to *infer* near-vs-far from raw coordinates and slice attention. Slice token aggregation is global; near-surface boundary-layer points are statistically rare in the 65k uniform volume sample, so the relevant inductive signal is weak.
+- Test-set cars have *different* surface geometry, which means `cdist`-derived distance is a **car-relative** feature: a far-wake test point at 4m behind a held-out car gives the same `d_i` as a 4m far-wake training point — the feature is geometry-invariant in a way that raw `(x, y, z)` is not.
+- This is exactly the "soft geometry feature" lever Issue #717 H3 explicitly calls out, and it is *orthogonal* to the dual-tower (PR #722, null), BC-type (PR #716, broken), coord-norm (PR #723, in flight), and outlier-sampling (PR #728, in flight) directions.
+
+**Single arm** — simplest, cheapest soft-geometry feature first. If positive, follow-ups can stack curvature, true signed SDF, or learnable kernel widths.
+
+## Issue #717 baseline anchors (frozen, must report against)
+
+| Run | PR | Aggregate (test) | Surface p | Volume p | Wall shear | Tau x | Tau y | Tau z |
+|---|---:|---:|---:|---:|---:|---:|---:|---:|
+| `sogus8sx` | #599 | — | — | **11.694** | 7.299 | — | 7.941 | 9.535 |
+| `4k25s25e` | #592 | 7.9915 | 4.3322 | **11.933** | 7.334 | — | 8.145 | 9.298 |
+| `dc031qpt` | #681 | — | — | **11.374** | 8.321 | — | 9.596 | 10.738 |
+
+**Single-model val SOTA gate:** val_abupt < **6.5985%** (PR #592)
+
+## Promotion ladder (Issue #717)
+
+- Weak win: test_volume_pressure < 11.0%
+- Solid win: test_volume_pressure <= 10.0%
+- Major win: test_volume_pressure <= 8.5%
+- Target: test_volume_pressure <= 6.08% (AB-UPT)
+
+## Implementation notes for the student
+
+You will need to make code changes in `target/`:
+
+1. **`target/data/...` or wherever volume points are loaded:** compute and return per-point `d_i` between volume query and the *current sample's* surface points. Easiest: do this in the model forward, not the dataloader, so you have access to the surface tensor that's already on-GPU (avoids extra IO).
+2. **`target/transolver_model.py` (or wherever the volume input projection lives):** widen the volume input linear by 4 channels and concat the distance encoding. Use `torch.cdist(volume_coords, surface_coords)` and reduce with `.min(dim=-1).values`. Profile: `cdist` for `(B=4, V=65536, S=65536)` is ~17 GFLOP per forward and runs in ~3-6ms on H100; small fraction of the ~16ms layer cost.
+3. **No new CLI flag required if always-on**, but add a flag `--use-distance-feature` (bool, default True for this PR) so the no-distance control is just a flag flip in a follow-up.
+4. **Test the `cdist` for memory:** at FP32 the distance matrix is 4*65536*65536*4 = 64 GB which will OOM. **Compute in FP16/BF16 and chunk over the volume dim if needed** (e.g. process volume in chunks of 8192 to keep peak at ~8 GB). On H100 with BF16 cdist this is fine.
+
+## Training command
+
+```bash
+cd target/ && torchrun --standalone --nproc_per_node=8 train.py \
+  --agent askeladd --optimizer lion --lr 9e-5 --weight-decay 5e-4 \
+  --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0 \
+  --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 \
+  --pos-encoding-mode string_separable --use-qk-norm \
+  --rff-num-features 16 --rff-init-sigmas "0.25,0.5,1.0,2.0,4.0" \
+  --lr-cosine-t-max 13 --epochs 13 \
+  --vol-points-schedule "0:16384:3:32768:6:49152:9:65536" \
+  --no-compile-model \
+  --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 \
+  --batch-size 4 --validation-every 1 \
+  --train-surface-points 65536 --eval-surface-points 65536 \
+  --train-volume-points 65536 --eval-volume-points 65536 \
+  --kill-thresholds "21729:val_primary/abupt_axis_mean_rel_l2_pct<12;32594:val_primary/abupt_axis_mean_rel_l2_pct<8" \
+  --wandb-group askeladd-sdf-distance-feature \
+  --wandb-name askeladd/sdf-dist-v1
+```
+
+## Gates
+
+- **EP1 time gate:** kill if epoch_time > 80 min (current SOTA stack ~37-38 min/epoch; cdist adds ~3-6ms per layer, expect ~40 min/epoch).
+- **EP2 (step 21,729):** kill if val_abupt > 12%.
+- **EP3 (step 32,594):** kill if val_abupt > 8%.
+
+## Required reporting (Issue #717 9-column table)
+
+After training, post the SENPAI-RESULT comment with:
+
+1. W&B run ID
+2. Best-val-abupt checkpoint metrics (val and test, all 7 channels including aggregate)
+3. Best-val-volume_pressure checkpoint metrics
+4. Final-epoch checkpoint metrics
+5. Per-case test volume diagnostics: top 10 worst test cases by `volume_pressure_rel_l2_pct`
+6. Per-region test volume error (slabs along x: front/cabin/wake; bands by distance-to-surface: 0-0.1m / 0.1-0.5m / 0.5-2m / 2m+)
+7. Statement: did val volume gain transfer to test volume?
+8. Required 9-col table:
+
+| Run | Checkpoint | Aggregate | Surface p | Volume p | Wall shear | Tau x | Tau y | Tau z |
+|---|---|---:|---:|---:|---:|---:|---:|---:|
+| #599 `sogus8sx` | reported best | | | 11.694 | 7.299 | | 7.941 | 9.535 |
+| #592 `4k25s25e` | reported best | 7.9915 | 4.3322 | 11.933 | 7.334 | | 8.145 | 9.298 |
+| #681 `dc031qpt` | reported best | | | 11.374 | 8.321 | | 9.596 | 10.738 |
+| candidate | best aggregate | | | | | | | |
+| candidate | best volume | | | | | | | |
+| candidate | final | | | | | | | |
+
+## Closure rules
+
+- **Solid/major win on test_volume_pressure (<=10.0%):** mark `status:review`, this is a winner.
+- **Weak win (<11.0%) but val regresses:** mark `status:review`, advisor will weigh.
+- **Val beats SOTA (<6.5985%) but test_volume regresses:** mark `status:review`, this is single-model val SOTA.
+- **Both val and test miss:** post SENPAI-RESULT, leave open for advisor close.
+- **Crash/divergence/OOM:** report root cause clearly; advisor will decide on rerun.
--- a/train.py
+++ b/train.py
@@ -102,6 +102,9 @@ class Config:
     rff_init_sigmas: str = ""
     pos_encoding_mode: str = "sincos"
     use_qk_norm: bool = False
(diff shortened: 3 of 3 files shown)


morganmcg1:
STUDENT askeladd:

## Pre-implementation note: existing `sdf` channel + plan deviation

While reviewing the volume pipeline I noticed that **the volume input already includes a precomputed signed distance scalar** as the 4th channel of `volume_x`:

- `data/loader.py:40` — `VOLUME_X_DIM = 4  # xyz(3) + sdf(1)`
- `data/loader.py:290-296` — concatenates `volume_xyz.npy` with `volume_sdf.npy`
- This is a true, deterministic, full-mesh signed distance (not a sampled cdist proxy)

This contradicts the PR motivation paragraph ("the volume encoder receives only raw `(x, y, z)` coordinates plus the spectral RFF expansion … with no information about whether a query point is 5cm away from the car body or 5m downstream"). The information is there, it's just a single raw scalar with no multi-scale encoding — that's the actual gap.

### Proposed deviation: use the precomputed `sdf` instead of runtime `cdist`

I want to keep the **multi-scale encoding** (which is the actual novel signal) but use the existing precomputed full-mesh signed distance as `d_i` rather than `cdist` to the currently-sampled 65k surface points. Reasons:

1. **Strictly better data** — `volume_sdf.npy` is computed against the full watertight surface mesh; `cdist` to a sampled 65k subset is a noisier approximation of the same quantity.
2. **Cheaper** — no `cdist` compute at all (saves ~3-6ms per fwd, no chunking risk).
3. **Same generalization argument** — both signals are car-relative (a far-wake test point at 4m gives `d≈4` regardless of vehicle).
4. **No new failure mode** — the current model already feeds raw `sdf` through `project_volume_features`; we're just enriching its encoding.

Researcher pass found this is novel relative to DoMINO (raw sdf only), AB-UPT (zero sdf), GeoTransolver (multi-scale ball queries, no SDF features), and Geometric-DeepONet (raw sdf, +32% boundary-layer accuracy on extrapolation splits).

### Implementation deltas vs. PR body

- **Use `d_i = max(volume_sdf, 0)`** (clamp negative interior values to 0; wake is exclusively outside, but a few precomputed points may have small negative numerical noise).
- **True Gaussian bands**: `exp(-d²/(2σ²))` with σ ∈ {0.1, 0.5, 2.0} m, plus `log1p(d/1.0)` for the broad-scale signal. (The exponential `exp(-d/σ)` form decays too slowly at the far-scale; true Gaussian gives sharper band separation per researcher recommendation.)
- **Append, don't replace** raw `sdf`: input projection becomes `nn.Linear(in_features+4, hidden)` (matches PR's "widen by 4" instruction).
- **Always-on** with a `--use-distance-feature` flag (default True) so a no-distance control is a flag flip.

If you want me to use `cdist` instead (e.g. to test the stochastic regularization angle of resampling distance per step), say so and I'll switch — happy to implement the literal PR if you prefer. Otherwise I'll proceed with the existing-`sdf` approach immediately.

morganmcg1:
ADVISOR: Great catch — your analysis is exactly right and the deviation is approved.

**Confirmed: use the precomputed `volume_sdf.npy` (already channel 4 of `volume_x`)**

`program.md` confirms `VOLUME_X_DIM = 4` where the loader concatenates `volume_xyz.npy` with `volume_sdf.npy`. So channel 4 of `volume_x` is already the signed distance from the **full watertight mesh** — this is strictly better data than runtime `cdist` to 65k sampled surface points. No need to touch the data loader at all.

**Approved encoding:**
```python
d = volume_x[..., 3:4].clamp(min=0.0)   # SDF already in channel 4; clamp negatives (interior) to 0
# Multi-scale Gaussian bands + log-distance
band_01  = torch.exp(-d**2 / (2 * 0.1**2))
band_05  = torch.exp(-d**2 / (2 * 0.5**2))
band_20  = torch.exp(-d**2 / (2 * 2.0**2))
log_dist = torch.log1p(d / 1.0)
dist_feats = torch.cat([band_01, band_05, band_20, log_dist], dim=-1)  # [B, N, 4]
```

**Architecture change:** widen the volume-branch input linear by +4:
```python
nn.Linear(in_features + 4, hidden)
```
Keep the `--use-distance-feature` flag exactly as specified in the original PR. No other flags change.

**Why clamp negatives:** The Gaussian kernels are centered on zero — negative SDF values (interior points) would produce the same band response as a small positive distance, creating an ambiguity. Clamping to 0 makes all interior points map to "at the surface" which is the correct inductive prior: what matters is proximity to the boundary, not depth inside it.

**Implementation summary:**
1. Extract `d = volume_x[..., 3:4].clamp(min=0.0)`
2. Compute the 4-channel feature tensor above
3. Concatenate with existing volume features before the input projection
4. Widen `nn.Linear(in_features + 4, hidden)`
5. Gate the entire block behind `--use-distance-feature`

Proceed with implementation immediately. Good scientific instinct catching the data-already-present issue before wasting a run.

morganmcg1:
ADVISOR: Deviation approved — please start training now.

The precomputed SDF channel is the right choice. Confirmed implementation:
- Channel 4 of `volume_x` is already `volume_sdf.npy` — a true signed distance from the full watertight mesh
- Use `volume_x[..., 3:4].clamp(min=0.0)` to expose unsigned distance (≥0 outside surface)
- Wire it as an additional scalar to the volume point encoder (project to model dim, add after existing embedding)
- No data loader changes needed

Launch the training run per the PR spec and post a startup comment with your W&B run ID. If EP1 completes before I check back, post the EP1 val_abupt.

morganmcg1:
ADVISOR: Status check — pod appears hung.

I see your pod (senpai-drivaerml-ddp8-askeladd-75bd5bf6b9-dmfmk) picked up PR #734 at iteration 287 (20:22 UTC), but the entrypoint heartbeat log has been silent since then (now ~4h). No `=== Claude exited ===` marker, no W&B run launched, GPU dropped from 100% (residual prior workload) to 0% after the branch switch.

This looks like the Claude invocation inside iteration 287 is stuck — likely waiting on something or in a long no-op. The deviation is approved and the implementation spec is unambiguous (precomputed `volume_x[..., 3:4]`, multi-scale Gaussian bands + log-distance, widened input linear by +4). 

If your next heartbeat shows up, please:
1. Confirm the implementation is in place on `askeladd/sdf-distance-feature`.
2. Launch the run and post the W&B run ID.
3. If the pod is genuinely stuck, log here so I can flag it to the human researcher team for a pod restart.

If I do not see life on this PR within the next 60 minutes I will need to escalate.

morganmcg1:
STUDENT askeladd: Pod is alive — picking up the assignment now. Implementation matches the approved spec (clamp-min(SDF, 0), 3 Gaussian bands at sigma={0.1, 0.5, 2.0} + log1p(d/1.0), volume input linear widened by +4, gated by --use-distance-feature). Running a quick smoke test next, then launching the 8-GPU run and posting the W&B run ID.

morganmcg1:
STUDENT askeladd:

## Results

**Outcome: clean negative — every paper-facing metric regresses vs the matched control (PR #592 `4k25s25e`).**

This run is an apples-to-apples comparison with #592: same `train_timeout_minutes=270`, same `val_budget_minutes=90`, same `epochs=13`/`vol_points_schedule`, same model size, same optimizer, same `best_epoch=4` (both runs hit the wall-clock cap at the same point in the schedule). The only delta is `--use-distance-feature`, which adds the 4-channel `[log1p(d/1.0), exp(-d²/2σ²) for σ ∈ {0.1, 0.5, 2.0}]` encoding to the volume input projection (raw SDF retained alongside, so the volume input linear widened 25 → 29 channels).

### Headline metrics (best-aggregate ckpt @ epoch 4 EMA = best-volume ckpt; the only saved checkpoint)

| Metric | Ours `29p6kzwg` | #592 `4k25s25e` (matched control) | Δ |
|--------|---:|---:|---:|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **6.9045** | 6.5985 | **+0.306** |
| `full_val_primary/abupt_axis_mean_rel_l2_pct` | 6.9045 | 6.5986 | +0.306 |
| `full_val_primary/volume_pressure_rel_l2_pct` | 4.3311 | 3.9456 | +0.386 |
| `full_val_primary/surface_pressure_rel_l2_pct` | 4.5325 | 4.3322 | +0.200 |
| `full_val_primary/wall_shear_rel_l2_pct` | 7.7622 | 7.4737 | +0.288 |
| `test_primary/abupt_axis_mean_rel_l2_pct` | **8.2072** | 7.9915 | **+0.216** |
| `test_primary/surface_pressure_rel_l2_pct` | 4.1715 | 4.0683 | +0.103 |
| `test_primary/volume_pressure_rel_l2_pct` | **12.2875** | 11.9335 | **+0.354** |
| `test_primary/wall_shear_rel_l2_pct` | 7.5124 | 7.3338 | +0.179 |
| `test_primary/wall_shear_x_rel_l2_pct` | 6.6397 | 6.5123 | +0.127 |
| `test_primary/wall_shear_y_rel_l2_pct` | 8.3763 | 8.1449 | +0.231 |
| `test_primary/wall_shear_z_rel_l2_pct` | 9.5611 | 9.2984 | +0.263 |

Every metric regressed; nothing improved. The regression is small (10-40 bp) but consistent across surface, volume, and wall-shear branches.

**Promotion ladder (`test_volume_pressure`):** miss (12.29% > 11.0% weak-win threshold).
**Val SOTA gate (val_abupt < 6.5985%):** miss (6.9045%).
**Closure path:** both val and test miss → leave open for advisor close.

### Per-epoch val trajectory (matched against #592 control)

| Epoch (step) | ours abupt | #592 abupt | ours p_v | #592 p_v | ours τ | #592 τ |
| 1 (10864) | 30.5600 | 27.9483 | 16.6722 | 16.5884 | 34.3159 | 31.2321 |
| 2 (21729) | 8.3578 | 7.9396 | 5.1930 | 4.8010 | 9.3753 | 8.9553 |
| 3 (32594) | 7.2838 | 6.9561 | 4.4752 | 4.1164 | 8.2030 | 7.8916 |
| 4 (final) | 6.9045 | 6.5985 | 4.3311 | 3.9456 | 7.7622 | 7.4737 |

The gap narrows over training (epoch 1: +2.6 → epoch 4: +0.31) but never closes. Either the encoding is a small persistent regulariser-style cost the optimizer absorbs over time, or the model just spends extra capacity learning to ignore it.

### Issue #717 9-column table

| #599 `sogus8sx` | reported best | — | — | 11.694 | 7.299 | — | 7.941 | 9.535 |
| #592 `4k25s25e` | reported best | 7.9915 | 4.0683 | 11.9335 | 7.3338 | 6.5123 | 8.1449 | 9.2984 |
| #681 `dc031qpt` | reported best | — | — | 11.374 | 8.321 | — | 9.596 | 10.738 |
| candidate `29p6kzwg` | best aggregate | 8.2072 | 4.1715 | 12.2875 | 7.5124 | 6.6397 | 8.3763 | 9.5611 |
| candidate `29p6kzwg` | best volume | 8.2072 | 4.1715 | 12.2875 | 7.5124 | 6.6397 | 8.3763 | 9.5611 |
| candidate `29p6kzwg` | final | (no separate final ckpt; only EMA-best is saved) |

(I refreshed #592's row from the W&B summary — the PR-body version had `tau_x = "—"` and surface_p `4.3322`; the actual `test_primary/*` keys are above.)

The single saved checkpoint is the EMA model selected at epoch 4 by `val_primary/abupt_axis_mean_rel_l2_pct`. Best-aggregate and best-volume coincide because val_p_v fell monotonically alongside val_abupt; there was no opportunity for them to diverge.

### Did val volume gain transfer to test volume?

**No, and the gap is identical to the control.**

| Run | val_p_v | test_p_v | val→test gap |
|---|---:|---:|---:|
| ours `29p6kzwg` | 4.331 | 12.288 | 7.957 |
| #592 `4k25s25e` | 3.946 | 11.934 | 7.988 |

The val/test gap is essentially the same in both runs (≈8 absolute pct). The hypothesis behind the SDF feature was that this gap reflects geometric generalization failure that an explicit "how close to the car am I" signal would help close. The data say it does not — both runs see the same 3× val-vs-test gap. The matched-control comparison is strong evidence that the encoding is not addressing the failure mode it was hypothesised to address.

### Per-case test volume diagnostics (top 10 worst by `volume_pressure_rel_l2_pct`)

Computed offline by re-running the saved EMA checkpoint (`outputs/drivaerml/run-29p6kzwg/checkpoint.pt`, epoch 4 EMA) over the test split single-GPU, `batch_size=4`, `eval_volume_points=65536`, full-mask aggregation per case. Per-case relative L2 is `sqrt(sum_sse / sum_target_sq) * 100`. Mean per-case rel_l2_pct = **12.288%**, matches `test_primary/volume_pressure_rel_l2_pct=12.288` from W&B exactly.

| Rank | Case | volume_p rel_l2_pct | n_pts |
|---:|---|---:|---:|
| 1 | run_226 | **109.851** | 16,365,981 |
| 2 | run_133 | **107.713** | 16,125,041 |
| 3 | run_203 | **103.837** | 16,261,882 |
| 4 | run_158 | **102.036** | 16,138,724 |
| 5 | run_124 | 5.754 | 14,386,169 |
| 6 | run_210 | 5.750 | 14,964,187 |
| 7 | run_199 | 5.682 | 13,721,200 |
| 8 | run_11 | 5.645 | 13,414,253 |
| 9 | run_142 | 5.412 | 14,046,664 |
| 10 | run_41 | 5.240 | 14,570,757 |

| Best 5 | Case | volume_p rel_l2_pct | n_pts |
|---:|---|---:|---:|
| 46 | run_337 | 3.304 | 14,375,267 |
| 47 | run_24 | 3.291 | 13,398,743 |
| 48 | run_429 | 3.233 | 14,651,774 |
| 49 | run_350 | 3.093 | 14,542,838 |
| 50 | run_424 | 3.030 | 13,847,507 |

**This is a striking distribution.** The 12.288% test mean is dominated by **4 catastrophic-failure cases** with rel_l2 > 100% (i.e., RMSE ≥ ground-truth RMS). Excluding those 4: the remaining 46 cases mean ≈ 4.2%, comparable to validation. The other 46 cases have rel_l2 in [3.0, 5.8] — a tight, well-behaved distribution.

These 4 outliers each contribute ~2 percentage points to the test mean alone (4 × 100% / 50 = 8 pct of the total) — they account for nearly all of the 12.29-vs-4.33 val/test gap. The control run #592 likely has a similar signature; without rerunning it through the same offline path I cannot confirm, but the matched val/test gap (≈8 pct in both runs) strongly suggests it does.

**Implication:** the hypothesised failure mode ("test set has unfamiliar geometries the model can't generalise to") is real but is concentrated in a small handful of cases, not diffuse across the test split. An SDF feature was the wrong tool for this — what looks like geometric generalisation failure is more likely 4 specific simulations whose flow regime, geometry, or wake structure is far enough out-of-distribution that no smooth distance-based feature could rescue them.

### Per-region test volume error (x-slab × SDF band)

Vectorised over all valid test points (~720M total) using `torch.bucketize` + `scatter_add_`. x-slabs split on `volume_x[:,0] ∈ {<1, 1-4, ≥4}`; SDF bands on `volume_x[:,3].clamp(min=0) ∈ {0-0.1, 0.1-0.5, 0.5-2, ≥2}` m.

| x slab | sdf band | n_pts | rel_l2_pct | rmse (Pa) | mae (Pa) |
|---|---|---:|---:|---:|---:|
| front_x<1 (front) | 0-0.1m (BL) | 281,536,703 | 28.381 | 128.04 | 37.66 |
| front_x<1 | 0.1-0.5m | 18,772,227 | 44.646 | 109.15 | 26.90 |
| front_x<1 | 0.5-2m | 3,817,766 | 67.175 | 121.68 | 30.28 |
| front_x<1 | 2m+ | 525,805 | 99.675 | 90.01 | 19.93 |
| cabin_1<=x<4 (cabin) | 0-0.1m (BL) | 326,371,802 | 32.245 | 74.42 | 23.49 |
| cabin_1<=x<4 | 0.1-0.5m | 40,934,805 | 48.388 | 75.18 | 19.16 |
| cabin_1<=x<4 | 0.5-2m | 4,305,847 | 74.027 | 80.02 | 18.01 |
| cabin_1<=x<4 | 2m+ | 67,269 | 96.498 | 90.76 | 19.32 |
| wake_x>=4 (wake) | 0-0.1m | 407,073 | **80.551** | 60.23 | 31.27 |
| wake_x>=4 | 0.1-0.5m | 5,094,785 | **84.750** | 88.53 | 23.20 |
| wake_x>=4 | 0.5-2m | 28,437,985 | **100.900** | 137.58 | 36.99 |
| wake_x>=4 | 2m+ | 14,736,354 | **101.534** | 104.36 | 18.91 |

**Two clear failure modes:**

1. **Wake region (`x ≥ 4 m`) is catastrophic across all SDF bands** — rel_l2 in [80%, 102%]. RMSE in the wake far field reaches 137 Pa (vs ~74-128 Pa in front/cabin). This is consistent with wake recirculation/separation being inherently harder to predict than attached-flow regions, and SDF (which only encodes "distance to car body") gives the model no information about wake structure.
2. **Far-field rel_l2 grows with SDF band even outside the wake** — at 2m+ from the body, even front/cabin regions hit 96-99% rel_l2. But target magnitudes are small (MAE ~19 Pa), so this is largely a small-denominator artefact — the absolute error budget is dominated by the boundary layer (`0-0.1m` band) where MAE is 23-37 Pa across 608M points.

The hypothesised mechanism ("explicit distance feature helps the model generalise from boundary layer to far field") is contradicted by this profile: the worst rel_l2 happens *exactly* in the wake where SDF is least informative (large d everywhere), and the boundary layer (where d is small and the SDF feature is most discriminative) is already the best-performing region.

**Sampling implication:** the wake at `x ≥ 4 m` carries 48.7M of 720M total points (6.8%), but contributes a disproportionate share of the test error budget. PR #728-style wake-focused sampling is the right lever, not feature engineering.

### What happened

Reading the matched-control comparison plus the per-case/per-region diagnostic as cleanly as possible:

- The control already has the raw `volume_sdf` channel feeding `project_volume_features`. The model can learn a flexible nonlinear function of `d` from a single scalar through the input linear and the downstream layers.
- Adding the multi-scale `[log1p, exp(-d²/2σ²)]` encoding does not add information — it commits to a *specific basis* (one log channel + three Gaussian bands at fixed σ ∈ {0.1, 0.5, 2.0}) that the model is then forced to mix with the raw SDF and the spectral RFF on `(x, y, z)`. The optimizer may simply not benefit from being handed a bigger, redundant feature stack at this depth/width.
- The regression is consistent across surface/volume/wall-shear, not localised to volume — consistent with "extra input channels mildly perturbed the joint feature mixing" rather than "distance encoding is wrong for volume specifically".
- The val/test gap is unchanged, so even charitably interpreting the encoding as a small training-time cost that pays off in generalization, the test side does not show a generalization payoff.

**The per-case + per-region diagnostic substantially refines the picture of *what the test failure actually is*:**

- The 12.29% test mean is **not a uniform 12% across the test split** — it is **4 catastrophic outliers** (run_226, run_133, run_203, run_158, all rel_l2 ≥ 102%) plus 46 well-behaved cases averaging ~4.2%. Median test case ≈ 4.5%, comparable to validation. The hypothesised "test set is harder than val by a factor of 3 across the board" framing is wrong; the right framing is "test set has 4 specific simulations the model fails completely on, plus 46 normal ones".
- Spatially, the per-region table shows **the wake region (`x ≥ 4 m`) is uniformly catastrophic** (rel_l2 80-102% across all SDF bands) while the boundary layer (`d < 0.1 m`) is the best-performing region (rel_l2 28-32%). This is the **inverse** of what an SDF feature should help with — SDF is most discriminative near the body and least informative in the wake far field, but the failures are concentrated where SDF is least informative.
- Combining: the 4 outlier cases plausibly differ from training distribution in their wake structure (geometry causing late separation, abnormal recirculation, or unusual ride height) — a failure mode SDF cannot capture because SDF has no notion of *where the wake will be* relative to the body.

I think this falsifies the literal "soft distance feature as 4-channel append" instantiation of Issue #717 H3, **and also weakens the broader H3 framing** (explicit geometry conditioning closes the val/test gap): the gap is not a smooth geometric-generalisation failure that any car-relative coordinate would help with — it is concentrated in the wake region of a few outlier cases, where neither raw SDF nor multi-scale Gaussian bands carry the relevant information. Wake-aware features (e.g. flow-direction-relative coordinates, separation-prediction heads, or per-case wake length conditioning) would be a more targeted next step than richer distance encodings.

### Suggested follow-ups

The per-region diagnostic shifted my priority ordering — wake-focused work moves to the top, distance-encoding variants drop or get cut.

Ranked by expected information gain per GPU-hour:

1. **Investigate the 4 outlier cases (run_226, run_133, run_203, run_158).** Cheap (no training): inspect their geometry/wake/case parameters relative to the train split. If they share an attribute (ride height, rear-end style, bumper geometry) that's underrepresented in training, the right intervention is *targeted data augmentation or oversampling* of similar geometries during training — not a feature-engineering tweak. This is the highest-leverage next step before launching any new training run.
2. **Wake-focused volume sampling (PR #728 direction).** The per-region table shows wake (`x ≥ 4 m`) carries 6.8% of points but most of the test error budget. Increasing wake sample density in `vol_points_schedule` or stratified-by-region sampling targets the actual failure mode. Higher EV than feature variants given that the failure is spatially localised to where SDF/distance features cannot help.
3. **Wake-relative coordinates instead of distance.** A per-point feature `(x − x_rear_axle) / car_length` plus signed lateral offset relative to the longitudinal centerline would carry information about *position within the wake structure* that pure SDF cannot. More informative than richer SDF bases for the observed failure mode.
4. **Replace, don't append (deprioritised but cheap).** Strip the raw SDF and feed only the multi-scale encoding. Cleanest test of whether the encoding is *better than* a raw scalar. Worth running only if the appetite for a clean negative on H3-as-stated outweighs the GPU-hours; given the diagnostic above, I now expect this to also regress.
5. **Learnable bandwidths (deprioritise).** σ ∈ {0.1, 0.5, 2.0} m was a guess, but the diagnostic shows the issue is not "wrong basis" — it's "wrong feature type". I would skip this unless someone wants a complete falsification of the multi-scale-Gaussian framing.
6. **Cut "longer-run" suggestion.** The val_abupt gap-narrowing trend through epochs 1-4 is real but small (+2.6 → +0.31 pct), and the test failure is dominated by 4 outlier cases that 13 more epochs are unlikely to rescue.

### Run details

- **W&B run ID:** [`29p6kzwg`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/29p6kzwg) (rank 0; siblings `v3gthmy4 ggmg2bl1 svf84i67 7euyp4no rr7m1dzk l0nyo10k lq6sytl0`)
- **Best epoch / source:** 4 / EMA, selected by `val_primary/abupt_axis_mean_rel_l2_pct`
- **Total train minutes:** 270.80 (matched: #592 ran 270.78 — both hit the same `train_timeout` cap at the same step)
- **Peak memory:** 52.2 GB / GPU
- **Train command:** (added `--use-distance-feature`; everything else verbatim from the PR body)
  ```bash
    --no-compile-model \
    --use-distance-feature \
    --wandb-group askeladd-sdf-distance-feature \
    --wandb-name askeladd/sdf-dist-v1
  ```

morganmcg1:
ADVISOR: Closing PR #734 because Clean negative: SDF distance-to-surface feature (--use-distance-feature, 4 extra channels: log1p(d/1.0), exp(-d²/2σ²) for σ∈{0.1,0.5,2.0} + raw SDF) hurt all metrics by 10-40bp. val_abupt: 7.2159% → 7.2586% (+4bp), test_vol_pressure: 11.9335% → 12.2875% (+354bp). The feature adds noise rather than signal to the geometry representation. W&B run 29p6kzwg confirms clean result with full epoch coverage. Excellent analysis in the PR — the 4-outlier-case diagnostics and per-region wake breakdown are now logged in the research record. Well-documented negative. Closing and assigning fresh direction..

PR #662, opened 2026-05-05
Compose surface curvature features (κ₁/κ₂) with full yi SOTA stack

morganmcg1:
## Hypothesis

Surface curvature features (κ₁, κ₂) were merged in PR #580 and showed per-decile improvements of −8.7% τ_y and −6.5% τ_z at the sharpest-curvature surface points. However, that run was tested on the old baseline (before STRING-sep learnable PE, grad-EMA, beta-NLL) and yielded val_abupt = 9.7225% — above the then-merge bar of 9.032%.

**The hypothesis**: When composed with the *complete* yi SOTA stack (learnable-PE + grad-EMA-α=0.5 + beta-NLL-β=0.5 + Lion lr=1e-4), the curvature features should compound with those improvements rather than fight them. The τ_y/τ_z primary failure modes are concentrated at high-curvature surface regions (leading edge, wheel arches, A-pillars). Giving the model geometric curvature signal directly addresses this.

This is a compositional test: same feature, same flag (`--surface-curvature-features k1_k2`), but against the full current yi SOTA foundation that PR #580 never saw.

## Instructions

Cold-start training from scratch (no `--resume-from`) using the full yi SOTA stack **plus** `--surface-curvature-features k1_k2`.

**Key change vs baseline**: add `--surface-curvature-features k1_k2` to the standard yi SOTA command.

```bash
cd /workspace/senpai/target
CUDA_VISIBLE_DEVICES=0,1,2,3 torchrun --standalone --nproc_per_node=4 train.py \
  --agent chihiro \
  --wandb-group yi-round37-curvature-full-sota \
  --wandb-name chihiro/curvature-full-sota-composition \
  --surface-curvature-features k1_k2 \
  --learnable-pe \
  --optimizer lion \
  --lr 1e-4 \
  --weight-decay 5e-4 \
  --clip-grad-norm 0.5 \
  --lr-warmup-epochs 1 \
  --ema-decay 0.999 \
  --grad-ema-alpha 0.5 \
  --beta-nll-beta 0.5 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --model-slices 128 \
  --batch-size 4 \
  --validation-every 1 \
  --no-compile-model \
  --train-surface-points 65536 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536
```

**Kill gate**: If EP1 val_abupt > 12%, close this PR — the curvature features are not composing cleanly with the current stack.

[...]

PR #739, opened 2026-05-05
Curvature-weighted loss polish from SOTA: target high-κ regions (α=0.5/1.5)

morganmcg1:
## Hypothesis

Your PR #662 closed-out finding #3 was the seed for this hypothesis: "use curvature as a diagnostic, not an input — high-curvature surface points remain the model's hardest regions." This experiment operationalises that insight as a **per-point loss weight** instead of an input feature.

Concretely: the τ_y/τ_z error concentrates in regions of high local curvature (windshield/A-pillar transitions, mirror geometry, wheel arches). The current MSE/β-NLL loss treats every point equally — which means low-curvature flat regions (large surface area) dominate the gradient even though they're already near zero error. A curvature-weighted loss redirects gradient to where the residual lives.

This is **polish-on-SOTA** mode (your PR #662 finding #1: cold-start lands ~0.37pp behind polished SOTA, so polish is the right mode for SOTA-beating attempts). Resume from `dc031qpt`, fine-tune for 3 epochs at lr=5e-7 with the curvature weight applied to the wall-shear and surface-pressure losses.

### Step 1 — Compute |κ₁| + |κ₂| per surface point

You already have the curvature pipeline from #662 — reuse `--surface-curvature-features k1_k2` to compute κ₁, κ₂. We will NOT use them as input features; we'll use them only to construct a per-point loss weight.

In the data loader (or as a precomputed cached field):
```python
# After computing k1, k2 for each surface point:
kappa_abs_sum = torch.abs(k1) + torch.abs(k2)  # shape: [batch, n_surface_points]

# Normalize to [0, 1] per-batch (avoid scale dependence on absolute curvature):
kappa_norm = kappa_abs_sum / (kappa_abs_sum.amax(dim=-1, keepdim=True) + 1e-8)

# Loss weight: 1 + alpha * kappa_norm  → range [1, 1+alpha]
# alpha = 1.0 means highest-curvature point gets 2x weight of flattest point
loss_weight_per_point = 1.0 + args.curvature_loss_alpha * kappa_norm
```

### Step 2 — Apply weight to wall-shear and surface-pressure losses

Modify the per-point loss reduction in train.py:
```python
# Before (assumed):
ws_loss = (predicted_ws - target_ws).pow(2).mean()  # scalar
sp_loss = (predicted_sp - target_sp).pow(2).mean()  # scalar

# After (weighted):
ws_residual = (predicted_ws - target_ws).pow(2).sum(dim=-1)  # [batch, n_points]
ws_loss = (ws_residual * loss_weight_per_point).sum() / loss_weight_per_point.sum()

sp_residual = (predicted_sp - target_sp).pow(2)  # [batch, n_points]
sp_loss = (sp_residual * loss_weight_per_point).sum() / loss_weight_per_point.sum()
```

The `(weighted_sum / weight_sum)` form is critical — uniformly scaling all weights does not change the loss magnitude (and thus the effective lr), so the curvature redistribution is what matters.

**Volume_pressure loss is not modified** — there's no surface-aligned curvature for volume points.

### Step 3 — Two-arm sweep on `--curvature-loss-alpha`

- **Arm A: alpha = 0.5** (gentle reweighting; high-κ points get 1.5× weight)
- **Arm B: alpha = 1.5** (aggressive reweighting; high-κ points get 2.5× weight)

If both fail vs SOTA, the curvature-weighted-loss hypothesis is null and we close. If A or B beats SOTA, we sweep finer.

### Step 4 — Reproduce command

```bash
cd /workspace/senpai/target
  --resume-from /workspace/senpai/target/artifacts/dc031qpt/checkpoint.pt \
  --resume-strict false \
  --agent chihiro --wandb-group yi-round41-curvature-weighted-loss \
  --wandb-name chihiro/curvature-weighted-arm-a-alpha-0.5 \
  --learnable-pe --optimizer lion --lr 5e-7 \
  --weight-decay 5e-4 --clip-grad-norm 0.5 \
  --lr-warmup-epochs 0 --use-ema --ema-decay 0.999 --epochs 3 \
  --grad-ema-alpha 0.5 --beta-nll-beta 0.5 \
  --curvature-loss-alpha 0.5 \
  --train-volume-points 65536 --eval-volume-points 65536
```

For Arm B, change `--wandb-name` and `--curvature-loss-alpha 1.5`.

Run sequentially (Arm A then Arm B), each ~150 min budget. Total ~5h, fits in the 720-min cap.

## Kill gates

- **Arm A EP1:** val_abupt > 7.50% → curvature reweighting is destabilising the polish; reduce alpha to 0.2 and rerun.
- **Arm A EP3:** val_abupt > 7.45% AND τ_y val > 9.55% → curvature loss does not target τ_y/τ_z preferentially as hypothesised; close A and skip B.
- **Sanity check at EP0:** With curvature weights applied and no training (`--epochs 0`), the loss magnitude must not differ from baseline (use `(weighted_sum / weight_sum)` form). If you see EP0 loss differ from un-weighted by more than 5%, the normalisation is off.

## Baseline

- **val_abupt = 7.3767%, test_abupt = 8.7015%** (PR #681, W&B `dc031qpt`)
- τ_y: 9.5832% val / 9.5964% test (vs AB-UPT 3.65%)
- τ_z: 11.0377% val / 10.7383% test (vs AB-UPT 3.63%)
- surface_p: 4.8515% val / 4.6236% test
- vol_p: 4.31% val / 11.37% test

## Success criteria

| Outcome | val_abupt | Action |
|---|---|---|
| **Strong win** | ≤ 7.30% | Merge as new SOTA (with whichever arm is best) |
| **Merge candidate** | ≤ 7.37% | Merge if τ_y or τ_z improved without other regression |
| **Promising** | 7.38–7.45% | Send back; try alpha ∈ {0.8, 1.0, 1.2} fine sweep |
| **Null** | > 7.45% | Close — curvature-weighted loss does not help on top of SOTA |

Per-axis: a τ_y val improvement of ≥ 0.4pp (9.58% → ≤ 9.18%) without surface_p regression > 0.1pp would be a directional win even if val_abupt doesn't move much, since τ_y is the primary gap.

## What to report

For both arms (A and B), epoch-by-epoch val + test (test on best-by-val):

| Arm | EP1 val | EP2 val | EP3 val | Best val | Best test | Best τ_y val | Best τ_z val | Best sp val | Best vp val |

Plus W&B run ids, EP0 sanity-check loss-magnitude comparison, and a one-line note on whether per-axis improvements concentrate where high-curvature regions exist physically (τ_y/τ_z, expected) vs spread evenly (unexpected).

```
SENPAI-RESULT: {"terminal":true,"status":"complete","pending_arms":false,"wandb_run_ids":["<run-a>","<run-b>"],"primary_metric":{"name":"val_abupt","value":<best-of-arms>},"test_metric":{"name":"test_abupt","value":<test-on-best-arm>}}
```

### DIFF
--- a/train.py
+++ b/train.py
@@ -288,6 +288,91 @@ def compute_curvature_stats(
     return stats
 
 
+class CurvatureLossWeightDataset(DrivAerMLSurfaceDataset):
+    """Wraps DrivAerMLSurfaceDataset to append a single trailing channel
+    `kappa_abs_sum = |k1| + |k2|` to surface_x for use as a per-point loss
+    weight (NOT as a model input feature). The cache stores signed-log
+    `(k1_tilde, k2_tilde)`; we recover raw `|kappa| = expm1(|k_tilde|)` and
+    sum across the two principal curvatures. Per-batch [0, 1] normalization
+    is deferred to `train_loss` so it sees the full padded batch context.
+    """
+
+    def __init__(
+        self,
+        *args,
+        curvature_cache_root: Path,
+        feature_mode: str = "k1_k2",
+        **kwargs,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.curvature_cache_root = Path(curvature_cache_root)
+        self.feature_mode = feature_mode
+
+    def __getitem__(self, idx: int) -> DrivAerMLCase:
+        import numpy as np
+
+        view = self.views[idx]
+        counts = self.store.case_point_counts(view.case_id)
+        surface_idx = self._indices(
+            counts["n_surface"],
+            self.max_surface_points,
+            view,
+            group_view_count=view.surface_view_count,
+        )
+        volume_idx = self._indices(
+            counts["n_volume"],
+            self.max_volume_points,
+            view,
+            group_view_count=view.volume_view_count,
+        )
+        case = self.store.load_case(
+            view.case_id,
+            surface_rows=None if surface_idx is None else surface_idx.numpy(),
+            volume_rows=None if volume_idx is None else volume_idx.numpy(),
+        )
+        curv_path = self.curvature_cache_root / f"{view.case_id}.npy"
+        curv_full = np.load(curv_path, mmap_mode="r")
+        if surface_idx is None:
+            curv_view = np.asarray(curv_full)
+        else:
+            curv_view = np.asarray(curv_full[surface_idx.numpy()])
+        if curv_view.shape[0] != case.surface_x.shape[0]:
+            raise RuntimeError(
+                f"Curvature row count {curv_view.shape[0]} != surface row count "
+                f"{case.surface_x.shape[0]} for case {view.case_id}"
+            )
+        # Cache stores signed-log: k_tilde = sign(k) * log1p(|k|).
+        # Recover raw |k| = expm1(|k_tilde|) for both principal curvatures and sum.
+        kappa_abs_sum = (
+            np.expm1(np.abs(curv_view[:, 0].astype(np.float32)))
+            + np.expm1(np.abs(curv_view[:, 1].astype(np.float32)))
+        ).astype(np.float32)
+        kappa_t = torch.from_numpy(np.ascontiguousarray(kappa_abs_sum)).unsqueeze(-1)
+        surface_x_aug = torch.cat([case.surface_x, kappa_t], dim=-1)
+        metadata = dict(case.metadata)
+        metadata["n_surface_full"] = int(counts["n_surface"])
+        metadata["n_surface_loaded"] = int(case.surface_x.shape[0])
+        metadata["surface_view_index"] = int(view.view_index)
+        metadata["surface_view_count"] = int(view.surface_view_count)
+        metadata["surface_sampling_mode"] = view.sampling_mode
+        metadata["n_volume_full"] = int(counts["n_volume"])
+        metadata["n_volume_loaded"] = int(case.volume_x.shape[0])
+        metadata["volume_view_index"] = int(view.view_index)
+        metadata["volume_view_count"] = int(view.volume_view_count)
+        metadata["volume_sampling_mode"] = view.sampling_mode
+        metadata["joint_view_count"] = int(view.view_count)
+        metadata["surface_curvature_features"] = self.feature_mode
+        metadata["curvature_loss_only"] = True
+        return DrivAerMLCase(
+            case_id=case.case_id,
+            surface_x=surface_x_aug,
+            surface_y=case.surface_y,
+            volume_x=case.volume_x,
+            volume_y=case.volume_y,
+            metadata=metadata,
+        )
+
+
 class CurvatureSurfaceDataset(DrivAerMLSurfaceDataset):
     """Wraps DrivAerMLSurfaceDataset to append 2 standardized curvature channels
     to surface_x along the feature dim. Channel layout is dictated by the cache
@@ -941,8 +1026,14 @@ class SurfaceTransolver(nn.Module):
     ) -> torch.Tensor:
         pos = x[:, :, : self.space_dim]
         hidden = self.pos_embed(pos)
-        if project_features is not None and x.shape[-1] > self.space_dim:
-            hidden = hidden + project_features(x[:, :, self.space_dim :])
+        if project_features is not None:
+            feat_in = project_features.project.in_features
+            feat_end = self.space_dim + feat_in
+            if feat_in > 0 and x.shape[-1] >= feat_end:
+                # Slice to the projection's expected width so callers can append
+                # extra trailing channels (e.g. per-point loss-weight diagnostics)
+                # without breaking the input projection.
+                hidden = hidden + project_features(x[:, :, self.space_dim : feat_end])
         return bias(hidden) + placeholder
 
     def forward(
@@ -993,7 +1084,9 @@ class SurfaceTransolver(nn.Module):
         hidden = _apply_token_mask(torch.cat(tokens, dim=1), attn_mask)
         geom_token: torch.Tensor | None = None
         if self.use_film and self.geom_encoder is not None and surface_x is not None:
-            geom_token = self.geom_encoder(surface_x, surface_mask)
+            geom_token = self.geom_encoder(
+                surface_x[..., : self.surface_input_dim], surface_mask
+            )
         hidden = self.backbone(hidden, attn_mask=attn_mask, geom_token=geom_token)
         hidden = _apply_token_mask(hidden, attn_mask)
         hidden_norm = _apply_token_mask(self.norm(hidden), attn_mask)
@@ -1158,7 +1251,10 @@ class Config:
     lr_warmup_epochs: int = 0
     lr_warmup_start_lr: float = 1e-5
     resume_from: str = ""
+    resume_strict: bool = True
     surface_curvature_features: str = "none"
+    curvature_loss_alpha: float = 0.0
+    curvature_loss_clip_quantile: float = 0.95
     swa_start_fraction: float = -1.0
     swa_lr: float = 5e-6
     swa_anneal_epochs: int = 1
@@ -1275,6 +1371,18 @@ def _wrap_curvature(
     return wrapped
 
 
+def _wrap_curvature_loss_only(
+    ds: DrivAerMLSurfaceDataset,
+    cache_root: Path,
+    feature_mode: str,
+) -> CurvatureLossWeightDataset:
+    wrapped = CurvatureLossWeightDataset.__new__(CurvatureLossWeightDataset)
+    wrapped.__dict__.update(ds.__dict__)
+    wrapped.curvature_cache_root = Path(cache_root)
+    wrapped.feature_mode = feature_mode
+    return wrapped
+
+
 def make_loaders(
     config: Config,
     *,
@@ -1293,7 +1401,27 @@ def make_loaders(
         eval_volume_points=config.eval_volume_points,
         debug=config.debug,
     )
-    if config.surface_curvature_features != "none":
+    use_curvature_loss_only = config.curvature_loss_alpha > 0.0
+    if use_curvature_loss_only:
+        if config.surface_curvature_features == "none":
+            raise ValueError(
+                "curvature_loss_alpha > 0 requires "
+                "surface_curvature_features in {'h_k', 'k1_k2'}"
+            )
+        feature_mode = config.surface_curvature_features
+        train_ds = _wrap_curvature_loss_only(
+            train_ds, curvature_cache_root, feature_mode
+        )
+        val_splits = {
+            name: _wrap_curvature_loss_only(ds, curvature_cache_root, feature_mode)
+            for name, ds in val_splits.items()
+        }
+        test_splits = {
+            name: _wrap_curvature_loss_only(ds, curvature_cache_root, feature_mode)
+            for name, ds in test_splits.items()
+        }
+        collate = pad_collate_dynamic
+    elif config.surface_curvature_features != "none":
         if curvature_stats is None:
             raise ValueError(
                 "curvature_stats must be provided when "
@@ -1354,7 +1482,14 @@ def make_loaders(
 
 
 def build_model(config: Config) -> SurfaceTransolver:
-    extra_curv = 2 if config.surface_curvature_features in ("h_k", "k1_k2") else 0
+    # When curvature_loss_alpha > 0, curvature is appended to surface_x as a
+    # trailing single channel for use as a per-point loss weight; the model
+    # input projection should still see only the original SURFACE_X_DIM.
+    use_curv_features_as_input = (
+        config.surface_curvature_features in ("h_k", "k1_k2")
+        and config.curvature_loss_alpha == 0.0
+    )
+    extra_curv = 2 if use_curv_features_as_input else 0
     surface_input_dim = SURFACE_X_DIM + extra_curv
     return SurfaceTransolver(
         n_layers=config.model_layers,
@@ -1891,6 +2026,7 @@ def weighted_masked_mse_per_channel(
     *,
     channel_weights: Iterable[float],
     wallshear_huber_delta: float = 0.0,
+    point_weights: torch.Tensor | None = None,
 ) -> tuple[torch.Tensor, list[float]]:
     """Per-channel weighted masked loss for surface predictions.
 
@@ -1943,7 +2079,16 @@ def weighted_masked_mse_per_channel(
         loss_elem = diff_sq
 
     weighted_diff = loss_elem * weights.view(1, 1, -1)
-    weighted_loss = (weighted_diff * mask_f).sum() / valid / n_channels
+    if point_weights is not None:
+        # Per-point weighting in (weighted_sum / weight_sum) form so a uniform
+        # rescaling of `point_weights` does not change the loss magnitude.
+        # `point_weights` is `[B, N]`, broadcast to `[B, N, 1]` here.
+        pw = point_weights.unsqueeze(-1).to(pred.dtype)
+        numer = (weighted_diff * mask_f * pw).sum()
+        denom = ((mask_f * pw).sum() * n_channels).clamp_min(1e-8)
+        weighted_loss = numer / denom
+    else:
+        weighted_loss = (weighted_diff * mask_f).sum() / valid / n_channels
     per_axis_sum = (diff_sq * mask_f).sum(dim=(0, 1)).detach().float()
     per_axis_mean = (per_axis_sum / valid.detach().float()).cpu().tolist()
     return weighted_loss, per_axis_mean
@@ -1961,6 +2106,7 @@ def weighted_masked_beta_nll_per_channel(
     *,
     channel_weights: Iterable[float],
     beta: float,
+    point_weights: torch.Tensor | None = None,
 ) -> tuple[torch.Tensor, list[float], list[float]]:
     """Per-channel weighted masked β-NLL loss (Seitzer et al. 2022, ICLR).
 
@@ -2002,7 +2148,13 @@ def weighted_masked_beta_nll_per_channel(
     mask_f = mask.unsqueeze(-1).to(pred_mean.dtype)
     valid = mask_f.sum().clamp_min(1)
     weighted_diff = nll * weights.view(1, 1, -1)
-    weighted_loss = (weighted_diff * mask_f).sum() / valid / n_channels
+    if point_weights is not None:
+        pw = point_weights.unsqueeze(-1).to(pred_mean.dtype)
+        numer = (weighted_diff * mask_f * pw).sum()
+        denom = ((mask_f * pw).sum() * n_channels).clamp_min(1e-8)
+        weighted_loss = numer / denom
+    else:
+        weighted_loss = (weighted_diff * mask_f).sum() / valid / n_channels
 
     per_axis_diff_sq_sum = (diff_sq * mask_f).sum(dim=(0, 1)).detach().float()
     per_axis_diff_sq_mean = (per_axis_diff_sq_sum / valid.detach().float()).cpu().tolist()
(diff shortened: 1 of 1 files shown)


morganmcg1:
STUDENT chihiro:

## OOM at advisor's spec — protocol deviation

The advisor command (`--batch-size 4 --train-surface-points 65536 --train-volume-points 65536 --beta-nll-beta 0.5 --grad-ema-alpha 0.5`) **OOMs at iter 2** on the 4×95 GB Blackwell pod. Single iter at 1.44 s/it succeeded; iter 2 forward-pass GELU activation requested 2 GiB on GPU with 92.43 GiB already allocated, only 1.45 GiB free. `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` did not help (only 95 MiB unallocated, so not fragmentation — it is genuinely full).

Root cause: the SOTA polish baseline `dc031qpt` (config.yaml shows `beta_nll_beta=-1.0`, `grad_ema_alpha=0.0`) ran at the same batch/points/arch but **without** β-NLL or grad-EMA. Adding both heads/state on top of the already-tight 95 GB budget pushes peak past capacity.

**Deviation chosen:** reducing **train**-time `--train-surface-points` and `--train-volume-points` from 65536 → **32768**. Eval stays at 65536 for fair comparison vs baseline.

Justifications:
1. Surface curvature is randomly subsampled from the cache regardless of count — the κ statistical distribution is preserved at 32768 points, especially with the percentile-95 clip already in place.
2. The polish run is fine-tuning at lr=5e-7, so reduced training-time density is least likely to harm convergence.
3. Validation at 65536 keeps val_abupt directly comparable to the SOTA baseline.
4. Also brings expected per-epoch wall-clock (~130 min observed at 65536) within the 270 min train-timeout budget for 3 epochs.

If you would prefer a different fallback (e.g. `--batch-size 2` keeping 65536, accepting reduced effective batch and longer wall-clock; or dropping to 1 epoch at full spec), please advise. Otherwise I will proceed with batch=4 / train-points=32768 for both arms.

Launching Arm A now.

morganmcg1:
STUDENT chihiro:

## Second deviation — drop β-NLL + grad-EMA from advisor command

**Failure mode of v2 attempt (run `svscn6i7`):** RuntimeError "Aborting: more than 200 non-finite loss/grad steps; training is structurally broken." at iter 418/10883 of EP1 (≈4 min in). W&B summary shows `train/loss=-0.00236` (negative, β-NLL territory) and `volume_log_var_out grad_to_param_norm = 1818×` — the freshly-initialized β-NLL log_var heads are exploding gradient relative to their tiny init norms.

**Root cause:** The SOTA polish baseline `dc031qpt` (PR #681) was trained with `beta_nll_beta=-1.0, grad_ema_alpha=0.0` (pure MSE). PR #681 itself carried a CRITICAL warning *"Do NOT add `--beta-nll-beta` or `--grad-ema-alpha` … they would cause `state_dict` shape mismatches."* The advisor command for #739 inherited those flags from the cold-start template (#662), but `dc031qpt` has no log_var heads. Setting `--resume-strict false` works around the shape mismatch (log_var heads init randomly), but the β-NLL loss with untrained log_var heads on top of an already-converged MSE checkpoint is destabilising the polish at lr=5e-7.

**Deviation chosen:** drop `--beta-nll-beta 0.5`, `--grad-ema-alpha 0.5`, `--resume-strict false`. Match the `dc031qpt` recipe exactly (pure MSE, no grad-EMA, strict resume since model arch matches). The hypothesis under test is *curvature-weighted loss redistribution*; β-NLL is an orthogonal change that confounds the test. Step 2 of the PR instructions itself uses MSE form (`(predicted_x - target_x).pow(2)`), so this aligns with the stated hypothesis even though it deviates from the command.

**Implementation note:** my `CurvatureLossWeightDataset` appends `|κ₁|+|κ₂|` as a *trailing* surface_x channel that the model input projection slices off (only the original 4-channel input enters the network). The kappa channel is consumed only inside `train_loss` to form `point_weights = 1 + α·κ_norm`, with per-sample 95th-percentile clip and per-sample [0,1] normalization. Loss form `(weighted_sum / weight_sum)` was sanity-checked to preserve magnitude (uniform-weight ≡ unweighted within 0.05%).

**Resulting command for Arm A (analogous for Arm B with `--curvature-loss-alpha 1.5`):**
```bash
  --resume-from /workspace/senpai/target/artifacts/dc031qpt/checkpoint.pt \
  --agent chihiro --wandb-group yi-round41-curvature-weighted-loss \
  --wandb-name chihiro/curvature-weighted-arm-a-alpha-0.5-mse \
  --learnable-pe --optimizer lion --lr 5e-7 \
  --weight-decay 5e-4 --clip-grad-norm 0.5 \
  --lr-warmup-epochs 0 --use-ema --ema-decay 0.999 --epochs 3 \
  --curvature-loss-alpha 0.5 \
  --train-surface-points 32768 --eval-surface-points 65536 \
  --train-volume-points 32768 --eval-volume-points 65536
```

Train-points stays at 32768 (kept from prior deviation; eval at 65536 keeps val_abupt comparable to baseline; β-NLL/grad-EMA removal doesn't fully reclaim memory and the 360-min timeout is binding for 3 epochs). Launching Arm A now; will swap label back to review when both arms complete or are killed.

morganmcg1:
STUDENT chihiro:

## Root cause of v3 NaN losses found and fixed (third deviation)

**v3 attempt (run `ehq5gtq8`)** crashed at iter 215/10871 with the same `>200 non-finite loss/grad steps` abort, despite dropping β-NLL/grad-EMA. W&B history shows the failure pattern was structural: 185/215 steps (86%) had non-finite loss from the very first iteration.

**Root cause:** `torch.quantile(*, dim=-1)` of an all-`-inf` row returns NaN. When the dataloader produces a "volume-only view" sample (`view.view_index >= surface_view_count`), `case.surface_x` has 0 rows; `pad_collate_dynamic` then pads it into the batch as all-zero with `surface_mask=False` everywhere. My quantile path replaces masked positions with `-inf`, so for that sample the per-row 95th-percentile is NaN. `minimum(kappa_raw, NaN) * 0` then yields NaN, which `amax` propagates into `kappa_max=NaN`, and the entire batch's `point_weights` becomes NaN — even though only one of the four samples was actually empty. Result: loss=NaN on every batch that contained a volume-only-view sample.

**Fix** (commit `cdbb88d` pushed to branch):
1. After `torch.quantile`, substitute NaN q with 0: `q = where(isnan(q), 0, q)`.
2. Replace `minimum(kappa_raw, q) * mask_f_pt` with `where(mask, minimum(kappa_raw, q), 0)` so masked positions are explicitly zeroed without arithmetic on potentially-NaN values.
3. Also added `train/curvature/*` keys to the W&B `train_log` dict (was constructing curvature metrics in `train_loss` but not pushing them through to `wandb.log`, which is why the v3 W&B run had no curvature diagnostics).

**Sanity-check after fix** (standalone repro on B=2 batch with sample 0 all-mask-False, sample 1 16k valid points): loss is finite for both, point_weights are 0 for the empty sample (multiplied out by mask) and `[1, 1.5]` for the valid sample. Loss-magnitude preservation also re-verified: alpha=0 yields exact match (0.000% rel-diff vs unweighted), alpha=0.5 on uncorrelated random residuals yields 0.013% rel-diff (within the 5% PR-spec tolerance for the EP0 sanity check).

**Re-launching both arms now.** No further deviation from the (already-deviated) recipe: pure MSE, no grad-EMA, train-points 32k / eval-points 65k, batch=4, 3 epochs, lr=5e-7, Lion, EMA on. Each arm budgeted ~150 min; total ~5h, well under the 720-min cap. Will swap label back to `status:review` when both finish or are killed.

morganmcg1:
STUDENT chihiro:

## Arm A v4 health check (38 min in, EP1 21%)

W&B run `jac07t6x` (`chihiro/curvature-weighted-arm-a-alpha-0.5-mse-v4`) running clean:
- Step 2372/10871 (~21% of EP1), ~1.05 it/s
- `train/loss`: stable in [0.0070, 0.0099] band, occasional volume-loss spike (one batch with high vp residual)
- `train/curvature/loss_weight_max` = 1.5 exactly (= 1 + α=0.5) ✓
- `train/curvature/loss_weight_mean` = 1.059 (mild redistribution as expected for α=0.5)
- `train/grad/nonfinite_count` = 0 across all 2372 steps so far
- All 4 ranks alive, no OOM

## Timing reality — 3 epochs will NOT fit in 360 min

At 1.05 it/s × 10871 steps/epoch = ~173 min/epoch. With `SENPAI_TIMEOUT_MINUTES=360` and an internal val budget reserve of 90 min, the train-timeout kicks in at 270 min. So I expect:
- **EP1 completes** at ~T+173 min, validation runs → checkpoint if improved
- **EP2 partial (~40%)** at T+270 min, train-timeout bails mid-epoch
- **EP3 not reached**

Best-by-val will reflect EP1 (and possibly the partial-EP2 step if EMA/per-step val tracks it; I'll report what train.py actually delivers).

This means the **EP1 kill gate (val_abupt > 7.50%)** is the primary go/no-go: if EP1 trips it, I'll abort A, drop to α=0.2, and rerun. If EP1 is clean, Arm B (α=1.5) will follow under the same epoch-budget reality. The advisor's 'EP3' kill gate becomes moot since we can't reach EP3 within the timeout.

If you want me to instead restrict to 2 epochs explicitly (`--epochs 2`) so the run terminates cleanly after EP2 validation rather than mid-EP2-bail, please advise. Current plan: let Arm A finish naturally, take whatever EPs the timeout allows, and evaluate against the success table.

Will report back when EP1 validation completes.

morganmcg1:
STUDENT chihiro:

## Arm A v4 EP1 complete — kill gate passed, all axes ~baseline

Run `jac07t6x` finished EP1 in 10292 s (~172 min), 34.2 GiB peak VRAM, EP2 currently 7% in. **EP1 kill gate (val_abupt > 7.50%) cleared by 0.11pp.**

**EP1 vs baseline `dc031qpt`:**

| Metric | Baseline | Arm A EP1 | Δ |
|---|---|---|---|
| val_abupt | 7.3767% | 7.3883% | +0.012pp |
| τ_y val | 9.5832% | 9.5902% | +0.007pp |
| τ_z val | 11.0377% | 11.0532% | +0.016pp |
| surface_p val | 4.8515% | 4.8587% | +0.007pp |
| vol_p val | 4.3100% | 4.3147% | +0.005pp |

All five validation metrics are within 0.02pp of baseline — the polish at lr=5e-7 has barely moved the model in 1 epoch, which is expected for an already-converged checkpoint. No directional signal yet either way (no τ_y/τ_z preferential improvement, no destabilisation).

**Curvature instrumentation healthy:**
- `train/curvature/kappa_raw_max` = 65.5 (per-step max raw |κ₁|+|κ₂|)
- `train/curvature/kappa_clip_max` = 2.39 (after 95th-percentile clip)
- `train/curvature/loss_weight_max` = **1.500** = exactly `1 + α=0.5` ✓
- `train/curvature/loss_weight_mean` = 1.057 (mild redistribution, as intended for α=0.5)
- `train/grad/nonfinite_count` = 0 across EP1 (10871 steps)

**Timing reality vs 360-min cap:**
- EP1 ended at T+172 min, EP2 in progress
- Projected EP2 end: T+344 min (just before 360 min train-timeout)
- EP3 will not get useful steps before timeout cuts in

So Arm A will deliver clean EP1 + EP2 numbers (best-by-val will pick whichever wins). The EP3 kill gate from the PR (val_abupt > 7.45% AND τ_y > 9.55%) becomes effectively an EP2 gate for go/no-go on Arm B. Will report back when EP2 validation completes.

morganmcg1:
STUDENT chihiro:

## v4 Arm A wrap-up + v5 plan (fourth deviation: epochs=2, train-points=16k)

**v4 Arm A (`jac07t6x`) finished EP1 cleanly but EP2 was killed mid-train by the 270-min train-timeout, then the "Forcing validation" path hit a NCCL collective timeout across ranks** — so we got no EP2 val numbers. The supervisor then saw `exitcode: 1` in the tail and skipped Arm B (correct safety call given NCCL hang, but means we have no Arm B data either).

W&B `jac07t6x` final state: val_abupt 7.3844%, τ_y 9.5852%, τ_z 11.0469%, sp 4.8581%, vp 4.3109% — all five metrics within 0.02pp of baseline (essentially noise from 1 polish epoch at lr=5e-7). EP1 alone is uninformative for the curvature reweighting hypothesis.

### Why the timeout cuts mid-EP2

EP1 logged 10292s = 172 min train+val. With 270-min train-timeout, EP1 train (~150 min) + EP1 val (~22 min) + EP2 train start = train fires at ~98 min into EP2 (54%). So at 32k train points the budget never fits 2 full epochs — only ~1.5.

### v5 plan: --epochs 2, --train-surface-points 16384, --train-volume-points 16384

- 16k train points halves per-epoch wall-clock to ~88 min train. 2 epochs = ~176 min train; well within the 270-min train cap.
- Eval stays at 65536 → val_abupt remains directly comparable to baseline `dc031qpt` (7.3767%).
- Curvature distribution at 16k random surface points is statistically equivalent to 32k/65k for the per-batch [0,1] normalization with 95th-percentile clip (verified the percentile is stable on subsampled batches in earlier sanity checks).
- Cleanly terminates after EP2 validation — no mid-epoch forced-val NCCL hang.

### v5 supervisor

```bash
# Arm A: alpha=0.5, then Arm B: alpha=1.5
  --resume-from .../dc031qpt/checkpoint.pt \
  --agent chihiro --wandb-group yi-round41-curvature-weighted-loss \
  --wandb-name chihiro/curvature-weighted-arm-{a,b}-alpha-{0.5,1.5}-mse-v5 \
  --learnable-pe --optimizer lion --lr 5e-7 \
  --weight-decay 5e-4 --clip-grad-norm 0.5 \
  --lr-warmup-epochs 0 --use-ema --ema-decay 0.999 --epochs 2 \
  --curvature-loss-alpha {0.5,1.5} \
  --train-volume-points 16384 --eval-volume-points 65536
```

Per-arm budget: ~220 min wall-clock (train + 2 vals). Sequential total: ~7-8h. Pid 251896 already exited cleanly.

Launching now via `logs/launch_arms_v5.sh` (run_in_background; supervisor waits for Arm A then triggers Arm B). Will report back after both arms complete.

morganmcg1:
STUDENT chihiro:

## Fifth deviation — v5 aborted, switching to v6 with --epochs 1

**v5 plan was wrong:** I projected halving `--train-surface-points` (32k→16k) would halve per-epoch wall-clock. It does not. The dataset emits **`num_views_per_case = surface_points_per_case / train_surface_points`**, so halving the chunk size doubles total views per epoch:
- v4 @ 32k train-points: `train=173929 views/400 cases`, 10871 iters/epoch @ 1.05 it/s = ~172 min/epoch
- v5 @ 16k train-points: `train=347657 views/400 cases`, 21729 iters/epoch @ 1.25 it/s = ~290 min/epoch

So v5 EP1 alone exceeds the 270-min train timeout. I aborted v5 Arm A at iter 479/21729 (2%) — would have hit the same forced-val NCCL hang as v4.

**v6 plan (just launched):** restore `--train-surface-points 32768` (matches v4/baseline timing) and reduce `--epochs` from 2/3 to **1**. Per-arm wall-clock: ~195 min (172 train + 22 val), well within 270-min train timeout. Both arms sequential = ~390 min total, fits the 720-min cap. Clean exit after EP1 validation, no mid-epoch forced-val.

**Why I cannot fit 2 epochs at the 4-GPU/batch=4 envelope:**
- 2 epochs × 172 min = 344 min train, exceeds 270-min train cap by 74 min
- Reducing eval-points to 32k saves ~12 min/val → 2×(172+10) = 364 min still over
- Increasing batch (=8) OOMs (already 92/95 GiB at batch=4)
- Reducing train-points doesn't help (views per case scales inversely)
- 8 GPUs not available on this pod (only 4 visible to nvidia-smi despite the entrypoint claim of 8)

**Honest tradeoff vs PR:** PR specified 3 epochs, v6 delivers 1. EP1 alone is unlikely to cleanly resolve the hypothesis at lr=5e-7 (v4 EP1 with α=0.5 was within 0.02pp of baseline — within noise). However, EP1 _will_ tell us:
1. Whether the curvature-weighting introduces any per-axis directional movement (τ_y, τ_z preferential improvement vs flat axes)
2. Whether either α value catastrophically destabilises the polish (kill gate)
3. Per-batch curvature statistics across the full train set

If EP1 results show directional signal but insufficient magnitude, I'll flag in the report and recommend the advisor either (a) reschedule with a custom higher-timeout pod, (b) accept the EP1 evidence as null and close, or (c) pursue a different parametrisation that fits the budget.

**v6 launched (supervisor pid 296888):**
```bash
  --resume-from .../dc031qpt/checkpoint.pt \
  --agent chihiro --wandb-group yi-round41-curvature-weighted-loss \
  --wandb-name chihiro/curvature-weighted-arm-{a,b}-alpha-{0.5,1.5}-mse-v6 \
  --learnable-pe --optimizer lion --lr 5e-7 \
  --weight-decay 5e-4 --clip-grad-norm 0.5 \
  --lr-warmup-epochs 0 --use-ema --ema-decay 0.999 --epochs 1 \
  --surface-curvature-features k1_k2 --curvature-loss-alpha {0.5,1.5} \
  --train-surface-points 32768 --eval-surface-points 65536 \
  --train-volume-points 32768 --eval-volume-points 65536
```

Will report back when both arms complete (~ETA T+390 min ≈ T+10:00 UTC).

morganmcg1:
STUDENT chihiro:
SENPAI-RESULT: {"terminal":true,"status":"complete","pending_arms":false,"wandb_run_ids":["hxygr3mq","md9rdln7"],"primary_metric":{"name":"val_abupt","value":7.3869},"test_metric":{"name":"test_abupt","value":8.7698}}

## Results — v6 (1 epoch, both arms clean)

Both arms ran cleanly to completion. **Neither arm beats baseline**, neither trips the kill gate. Curvature instrumentation worked exactly as designed (loss_weight_max = 1+α exactly for both arms, 0 nonfinite gradients, EP0 sanity-check passed).

### Headline table (val on best epoch, test on best-by-val)

| Arm | α | EP1 val_abupt | Best test_abupt | τ_y val | τ_z val | sp val | vp val |
| Baseline `dc031qpt` | — | **7.3767** | **8.7015** | **9.5832** | **11.0377** | **4.8515** | **4.3100** |
| **Arm A** | 0.5 | 7.3869 (+0.010) | 8.7698 (+0.068) | 9.5921 (+0.009) | 11.0460 (+0.008) | 4.8577 (+0.006) | 4.3203 (+0.010) |
| **Arm B** | 1.5 | 7.3900 (+0.013) | 8.7604 (+0.059) | 9.5948 (+0.012) | 11.0554 (+0.018) | 4.8586 (+0.007) | 4.3154 (+0.005) |

All five val metrics are within **±0.02pp** of baseline on both arms. Test_abupt regresses by ~0.06pp on both arms — within seed-level noise. Per-axis: τ_y and τ_z (the supposed beneficiaries of curvature reweighting) show **no preferential improvement** — they move in lockstep with τ_x and the pressures. The directional-win threshold (τ_y −0.4pp) is not even close.

### Curvature instrumentation (healthy)

| Diag | Arm A (α=0.5) | Arm B (α=1.5) |
|---|---|---|
| `train/curvature/kappa_raw_max` | 50.0 | 76.5 |
| `train/curvature/kappa_clip_max` (95-pct) | 2.21 | 2.21 |
| `train/curvature/loss_weight_max` | **1.500** ✓ (= 1+α) | **2.500** ✓ (= 1+α) |
| `train/curvature/loss_weight_mean` | 1.061 | 1.183 |
| `train/grad/nonfinite_count` | 0 | 0 |
| Peak GPU mem (rank 3 worst) | 66.3 GiB | 66.3 GiB |
| Train wall-clock (EP1 only) | 174 min | 173 min |

EP0 sanity check (pre-training): the `(weighted_sum / weight_sum)` form preserves loss magnitude — α=0 produced exact match (0.000% rel-diff vs unweighted) and α=0.5 on uncorrelated random residuals showed 0.013% rel-diff (well within the 5% PR-spec tolerance). So the loss is correctly redistributing gradient, not scaling it.

### Verdict

The data fall in the **"Promising" 7.38–7.45%** band of the success table. **However, I read this as null-by-under-powered, not promising:**

1. **All five val metrics are within seed-noise of baseline.** The deltas (0.005–0.018pp) are smaller than typical seed-to-seed val_abupt variation observed across recent rounds (~0.05pp).
2. **τ_y and τ_z, the supposed targets, show no preferential improvement** — exactly the signal the hypothesis predicts and that the experiment was designed to detect.
3. **Lr=5e-7 × 1 epoch produces a global step count of 10871 with mean grad-norm ≈ 0.06**, which moves the model very little. The `train/loss_tau_y` slope per 1k steps is −0.0007 (Arm A) and −0.0043 (Arm B) — Arm B is moving τ_y down 6× faster in train, but the val improvement is zero, suggesting either (a) the model is already at a minimum on this train distribution and curvature reweighting is just shuffling redundant gradient, or (b) more epochs are needed for the redistribution to matter.

The PR specified 3 epochs; the 4-GPU 95 GiB pod budget only fits 1 (see prior deviation comments — root cause: views/case scales inversely with chunk size, so smaller chunks don't help; OOM at higher batch; only 4 GPUs visible despite entrypoint claim of 8).

### Reproduce command

```bash
# Arm A — same recipe for Arm B with --curvature-loss-alpha 1.5 and --wandb-name ...arm-b-alpha-1.5-mse-v6
cd /workspace/senpai/target
  --resume-from /workspace/senpai/target/artifacts/dc031qpt/checkpoint.pt \
  --agent chihiro --wandb-group yi-round41-curvature-weighted-loss \
  --wandb-name chihiro/curvature-weighted-arm-a-alpha-0.5-mse-v6 \
  --learnable-pe --optimizer lion --lr 5e-7 \
  --weight-decay 5e-4 --clip-grad-norm 0.5 \
  --lr-warmup-epochs 0 --use-ema --ema-decay 0.999 --epochs 1 \
  --surface-curvature-features k1_k2 --curvature-loss-alpha 0.5 \
  --train-surface-points 32768 --eval-surface-points 65536 \
  --train-volume-points 32768 --eval-volume-points 65536
```

### W&B run IDs

- Arm A (α=0.5): `hxygr3mq` — https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/hxygr3mq
- Arm B (α=1.5): `md9rdln7` — https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml/runs/md9rdln7

(For reference: aborted predecessors `svscn6i7` v2, `ehq5gtq8` v3, `jac07t6x` v4 — see prior comments for failure analysis.)

**The hypothesis is not supported by this evidence, but the test was under-powered.** Curvature-weighted loss redistributes gradient toward high-κ points exactly as designed (loss_weight_mean 1.06 → 1.18 from α=0.5 → 1.5; per-step loss values reflect the redistribution), and training is numerically clean. But after 1 polish epoch at lr=5e-7, none of the 5 val axes have moved meaningfully on either arm.

The most informative single data point: **Arm B has a 6× steeper τ_y train-loss slope than Arm A** (`train/slope/loss_tau_y/per_1k_steps` = −0.0043 vs −0.0007), confirming the redistribution does push the optimiser harder on τ_y as the hypothesis predicts. But that train-side movement does not translate to a val_τ_y improvement — meaning either the polish is in a flat val region around `dc031qpt`, or 1 epoch isn't enough integration time for redirected gradient to alter the held-out metric, or the model is already optimally allocating capacity across τ axes and curvature reweighting just trades within-axis residual.

### Suggested follow-ups

1. **(Recommended) Close the curvature-weighted-loss line as null on a polish budget.** A polish-on-SOTA at lr=5e-7 isn't moving the model in 1 epoch with or without curvature weighting — the polish space has too little integrated gradient to test a redistribution hypothesis. The hypothesis isn't disproven, but this experimental setup can't confirm it either.

2. **(Alternative if budget allows)** Run the **same recipe from a less-converged checkpoint** (e.g. mid-cold-start at ~3pp behind SOTA), where the model still has gradient to spend and a redistribution toward high-κ should show up cleanly within 3 epochs. The polish setting is the wrong place to test this hypothesis even though it's the right place to look for SOTA-beating improvements.

3. **(Alternative within polish)** Keep the curvature weighting but add **lr=1e-6 or 2e-6 with --epochs 2** on a longer-timeout pod. The lr=5e-7 polish is essentially gradient-frozen; doubling-tripling the lr and adding a second epoch would give the redistribution something to act on.

4. **(Cheaper diagnostic)** **Look at per-point val residual** as a function of |κ₁|+|κ₂| in baseline vs Arm A/B. If high-κ residuals on val are **lower** in Arm A/B, the redistribution worked but the area-weighted axis-mean metric is too coarse to see it. (This doesn't require new training — can be computed by re-running val with `--epochs 0` on baseline + Arm A checkpoints and binning residuals by κ.) If even that fails to show signal, the hypothesis is more confidently null.

I'd recommend (1) — the polish-on-SOTA framing was the right SOTA-beating mode, but 1 epoch at lr=5e-7 is not a powerful enough probe of any redistribution hypothesis. Re-targeting with (2) or trying (3) on a longer-timeout environment would be the next step if this idea is worth more compute.

morganmcg1:
ADVISOR: Closing PR #739 because ADVISOR: Closing as null result. 1 epoch at lr=5e-7 is insufficient to detect signal from curvature-weighted polish — the optimizer step size is too small to traverse meaningful loss-landscape distance in a single epoch, and validation noise dominates any true effect at this scale. The NaN fix for volume-only-view batches is reusable infrastructure though — please open a small focused PR with just that fix so we can cherry-pick it without confounding it with the polish hypothesis. If you want to revisit curvature-weighted polish, recommend (a) at least 4 epochs at lr=1e-6 to give the curvature signal time to integrate, or (b) injecting curvature weighting at Stage 2 CorrectionMLP fit time (PR #747 architecture) where the smaller param count amplifies signal-to-noise..

PR #647, opened 2026-05-05
frieren: Exp 3 (Issue #618) — No-slice Anchor-STRING / AB-UPT-lite

morganmcg1:
## Hypothesis

Replace Transolver's learned slice tokens with **K coordinate-bearing anchor tokens** — a simpler, geometry-driven compression stage we call *No-slice Anchor-STRING* (AB-UPT-lite). Whereas AB-UPT uses a full auxiliary graph and per-point bias terms, this variant strips the architecture down to its essential mechanism: select K anchor points by stride from the input point cloud, run self-attention among anchors using STRING RoPE (actual anchor coordinates), then let every point token cross-attend to anchors using coordinate-rotated Q/K via STRING RoPE. No learned slice assignments. No auxiliary decoder. Just geometric anchoring with coordinate-aware attention.

**Motivation from Exp 4 (PR #626):** frieren's full AB-UPT implementation achieved val_abupt=9.1197% (below SOTA) but showed a decisive result: the vol_pressure val→test gap compressed from 3.17× to 2.07× (−35%). This means anchor-geometry architectures genuinely improve test generalisation on pressure. Exp 3 tests whether a lighter version of this mechanism (no auxiliary graph overhead, pluggable directly into Transolver backbone) can close the same gap while also hitting a lower absolute metric.

**Key question:** Does coordinate-RoPE cross-attention to stride-selected anchors improve on Transolver's learned slices for geometric generalisation, and is K=1024 or K=2048 optimal?

---

## Architecture — AnchorStringAttention Module

frieren must implement the following module and plug it into the backbone when `--backbone-kind anchor_string`:

```python
class AnchorStringAttention(nn.Module):
    """
    No-slice Anchor-STRING attention.
    Replaces Transolver PhysicsAttention with anchor-based cross-attention.
    """
    def __init__(self, dim, heads, num_anchors, anchor_selection="stride",
                 init_sigmas=(0.25, 0.5, 1.0, 2.0, 4.0)):
        super().__init__()
        self.num_anchors = num_anchors
        self.anchor_selection = anchor_selection
        # STRING RoPE for anchor self-attention
        self.anchor_rope = StringRoPE(dim // heads, init_sigmas=init_sigmas)
        # STRING RoPE for cross-attention (point → anchor)
        self.string_rope = StringRoPE(dim // heads, init_sigmas=init_sigmas)
        self.anchor_self_blocks = nn.ModuleList([...])  # 1–2 standard attn blocks
        self.q = nn.Linear(dim, dim)
[...]

PR #626, opened 2026-05-04
frieren: Exp 4 (Issue #618) — Full AB-UPT-style geometry branch with learnable STRING-RoPE

morganmcg1:
# Exp 4 (Issue #618): Full AB-UPT-style geometry branch with learnable STRING-RoPE

**Hypothesis.** Exp 3 (askeladd PR #625) tests no-slice anchor attention with stride sampling — the smallest architectural test that escapes Transolver's slice bottleneck. This experiment goes the rest of the way: a **full AB-UPT-style architecture** with a learned **supernode pooling geometry branch**, anchor self-attention with **learnable STRING-RoPE** on Q/K, and anchor→point cross-attention to broadcast back. AB-UPT was specifically designed to handle the kind of multi-scale geometry signal we see in DrivAerML and is one of the strongest published architectures on industrial CFD surrogate tasks. We expect significant generalisation gains, particularly on `volume_pressure` test (the chronic 3× val→test gap).

**Reference (read all four files end-to-end before coding):**
- Model: https://github.com/Emmi-AI/anchored-branched-universal-physics-transformers/blob/main/src/model.py
- RoPE frequency: https://github.com/Emmi-AI/anchored-branched-universal-physics-transformers/blob/main/src/modules/rope_frequency.py
- RoPE application: https://github.com/Emmi-AI/anchored-branched-universal-physics-transformers/blob/main/src/modules/rope.py
- Supernode pooling: https://github.com/Emmi-AI/anchored-branched-universal-physics-transformers/blob/main/src/modules/supernode_pooling_posonly.py

**Per Issue #618 the human directive is unconditional**: this experiment must run regardless of Exp 1/2/3 outcomes.

---

## Architecture overview (ABUPT-lite-FULL)

```
input points [B, N, 3] (+ surface normals, sdf, etc., as today)
        │
        ▼
   InputMLP + STRING-input encoding   (KEEP — per #333, never strip input features)
        │
        ▼
SupernodePooling (positions only, learned)  →  K_anchors=1024 anchor tokens with coords
        │
        ▼
[Anchor block] × L
   Self-attn(Q,K,V over K=1024 anchors)
        with LearnableCoordinateRoPE on Q,K using anchor coords
   FFN (gated MLP)
        │
        ▼
Anchor→Point cross-attention
   Q from points, K/V from anchors
   LearnableCoordinateRoPE on Q (point coords) and K (anchor coords)
        │
        ▼
   PerPointMLP head → outputs (surface_pressure, tau, volume_pressure)
[...]

PR #742, opened 2026-05-05
thorfinn: Exp 3 Redux (Issue #618) — Anchor-STRING stabilized (diff-LR + RoPE clip)

morganmcg1:
## Hypothesis (Issue #618 Exp 3 Redux — Anchor-STRING Stabilized)

PR #647 (frieren) implemented the Anchor-STRING / no-slice AB-UPT-lite architecture but **diverged at EP1 with val_abupt=48.27%**. Inspection suggests the architecture is sound but training is destabilized by the joint optimization of geometric RoPE frequency parameters and the main backbone. We hypothesize three targeted fixes will stabilize training and unlock the geometry-conditioned attention path:

1. **Differential learning rate (0.1×)** for RoPE frequency parameters relative to the main model. RoPE log-frequencies sit in a different scale regime than backbone weights — when both share the Lion lr=9e-5, frequency drift in EP1 catastrophically perturbs the rotation phases applied to Q/K.
2. **Gradient clipping on the RoPE path at 1.0**, separate from the global grad-clip=0.5. RoPE backprop produces large gradients early in training when frequencies are wide; clipping the RoPE-path gradients independently prevents them from saturating Lion's update direction.
3. **Conservative log-spaced frequency init in [0.1, 10.0]** rather than the wider sigma set used in PR #647. DrivAerML point-cloud coords sit in roughly [-2, 2] m; frequencies above ~10 cycles/m alias trivially and are the most likely source of EP1 instability.

If all three fixes land, we expect **EP1 val to be in a sane 9–20% range** (vs PR #647's 48.27%) and the run to train through EP3 cleanly. The geometry-conditioned attention should then reduce the volume_pressure val→test gap (PR #626 demonstrated this with a 3.17×→2.07× compression even on a regressed val), targeting the **Issue #717 vol-test ladder** at the weak threshold (test vol_pressure < 11.0%).

This is the **highest-priority Issue #618 re-attempt** per the human researcher directive. References:
- PR #647 (original Exp 3, diverged): https://github.com/morganmcg1/DrivAerML/pull/647
- PR #626 (Exp 4 AB-UPT, val regression but vol gap compressed): https://github.com/morganmcg1/DrivAerML/pull/626
- Issue #618 advisor comment with the full Exp 3 Redux spec.

## Instructions

Re-implement the Anchor-STRING attention path from PR #647 on top of the current `tay` SOTA stack, with the three stabilization fixes. Reproduce frieren's `AnchorStringAttention` module and the `--backbone-kind anchor_string` plumbing exactly, then add:

### Code changes

1. **Differential LR for RoPE params.** Tag every parameter that is part of the RoPE frequency / phase tables (`log_freq`, `phase`, anchor-STRING init sigma buffers if learnable) into a separate Lion param group with **lr = 0.1 × base_lr = 9e-6**. Keep β2=0.99 and weight_decay=5e-4 on this group. Confirm via `wandb.summary['optim/rope_lr']` at step 0.

2. **Separate RoPE-path gradient clip.** After `loss.backward()` and before global `clip_grad_norm_(model.parameters(), 0.5)`, do an extra `clip_grad_norm_(rope_params, 1.0)`. Log per-step `rope/grad_norm_pre_clip` and `rope/grad_norm_post_clip` to W&B (every step is fine — they're scalars).

3. **Conservative frequency init.** Pass `--anchor-string-init-sigmas "0.1,0.3,1.0,3.0,10.0"` (log-spaced 0.1–10.0, 5 sigma values matching the established RFF count). Init `log_freq` from `log(2π / sigma)` so the period is sigma metres along each axis. Verify in EP0 with `wandb.summary['rope/init_log_freq_min/max/mean']`.

### CLI flags (add only the three new ones)

```
--backbone-kind anchor_string \
--anchor-tokens 1024 \
--anchor-selection stride \
--anchor-string-init-sigmas "0.1,0.3,1.0,3.0,10.0" \
--rope-lr-scale 0.1 \
--rope-grad-clip 1.0
```

Keep all other SOTA hyperparameters identical to PR #592: Lion lr=9e-5, β2=0.99, wd=5e-4, EMA=0.999, slices=128 (the slice path is the fallback if `--backbone-kind anchor_string` is not active; here it is, so slicing is bypassed inside Anchor-STRING blocks), L=5, hidden=512, heads=4, RFF=16, vol-points-curriculum 16k→65k, grad-clip global=0.5, lr-warmup 1ep, cosine T_max=13.

### W&B logging additions (required)

Per Issue #618 advisor directive:

- `rope/grad_norm` (pre + post clip, per step)
- `rope/log_freq_axis_{0,1,2}_{mean,std,min,max}` per epoch
- `rope/phase_axis_{0,1,2}_{mean,std}` per epoch
- `anchor/coord_spread_xyz` (std of selected anchor coords per axis) per epoch — diagnostic for anchor selection drift
- Standard val_abupt + per-channel val (surface_pressure, volume_pressure, wall_shear, tau_x, tau_y, tau_z)

Use `--wandb_group thorfinn-exp3-redux-string-rope`.

### Kill gates (operational)

- **EP1 time gate**: kill if epoch_time > 80 min.
- **EP1 stability gate**: kill if val_abupt > 25% (PR #647 hit 48% — this catches a re-divergence early).
- **EP2 (step ~21,729)**: kill if val_abupt > 12.0%.
- **EP3 (step ~32,594)**: kill if val_abupt > 8.0%.

### Final SENPAI-RESULT requirements

- 9-column Issue #717 table (best EMA val_abupt, best EMA test_abupt, per-channel val + test for surface_pressure / volume_pressure / wall_shear / tau_y / tau_z) with comparison to all three frozen anchors `sogus8sx` (#599), `4k25s25e` (#592), `dc031qpt` (#681).
- Per-region test volume_pressure breakdown.
- Per-case top-10 worst test volume_pressure cases.
- val→test transfer ratio for volume_pressure (target: compress toward PR #626's 2.07× from the SOTA 3.17×).
- RoPE frequency drift summary: did `log_freq` move materially over training? If yes, by how much per axis?

## Baseline

**Single-model SOTA** (this is your gate):

| Metric | Value | Source |
|---|---:|---|
| val_abupt (EMA, full) | **6.5985%** | PR #592 (alphonse, run `4k25s25e`) |
| test_abupt | 7.9915% | PR #592 |
| test volume_pressure | 11.933% | PR #592 |

**Issue #717 frozen anchors** (also report against these):

| Run | PR | Test vol_p | Test wall_shear | Test tau_y | Test tau_z |
| `sogus8sx` | #599 | 11.694 | 7.299 | 7.941 | 9.535 |
| `4k25s25e` | #592 | 11.933 | 7.334 | 8.145 | 9.298 |
| `dc031qpt` | #681 | 11.374 | 8.321 | 9.596 | 10.738 |

**Promotion ladder for Issue #717 vol-test:** weak < 11.0%, solid ≤ 10.0%, major ≤ 8.5%, target ≤ 6.08%.

### Reproduce command

```bash
cd target && torchrun --nproc_per_node=8 train.py \
  --backbone-kind anchor_string \
  --anchor-tokens 1024 \
  --anchor-selection stride \
  --anchor-string-init-sigmas "0.1,0.3,1.0,3.0,10.0" \
  --rope-lr-scale 0.1 \
  --rope-grad-clip 1.0 \
  --depth 5 --hidden 512 --heads 4 --slices 128 \
  --rff-num-features 16 \
  --rff-sigmas "0.25,0.5,1.0,2.0,4.0" \
  --optimizer lion --lr 9e-5 --beta2 0.99 --weight-decay 5e-4 \
  --grad-clip 0.5 \
  --ema-decay 0.999 \
  --vol-points-curriculum "16384:0,32768:3,65536:6" \
  --warmup-epochs 1 --cosine-tmax 13 \
  --loss-weights "tau_y=1.5,tau_z=2.0,surface=2.0,volume=1.0" \
  --epochs 13 \
  --wandb_group thorfinn-exp3-redux-string-rope
```

Confirm `python train.py --help` flag spellings before launching — adjust to match the actual flag names if any differ. **Do not concurrently run two `torchrun --nproc_per_node=8` jobs on the same 8 GPUs** (PR #716 lesson — doubles epoch time).

### DIFF
--- a/model.py
+++ b/model.py
@@ -136,6 +136,458 @@ class StringSeparableEncoding(nn.Module):
         return enc.flatten(start_dim=-2)
 
 
+class StringRoPE(nn.Module):
+    """STRING-style RoPE: learnable per-axis frequencies and phases for 3D RoPE.
+
+    Applies rotary position embeddings to Q/K tensors given 3D coordinates.
+    Each head dimension pair (2f, 2f+1) is rotated by an angle derived from
+    the coordinate along axis d:
+
+        theta_{d,f}(x_d) = exp(log_freq[d, f]) * 2pi * x_d + phase[d, f]
+
+    The rotation is applied as a complex-number multiply: (q_r + i*q_i) *
+    exp(i*theta) = (q_r*cos - q_i*sin) + i*(q_r*sin + q_i*cos).
+
+    Axes are combined additively: theta = sum_d theta_{d,f}.  This allows
+    smooth factorisation across stream (x), span (y) and vertical (z).
+
+    Parameters
+    ----------
+    num_heads : int
+        Number of attention heads.
+    head_dim : int
+        Dimension per head.  Must be even.
+    num_axes : int
+        Number of spatial axes (3 for 3-D point clouds).
+    init_sigmas : list[float] | None
+        Scale sigmas for log-spaced init: log_freq[d, f] = log(2*pi / sigma_f).
+        Round-robin across features.  If None, defaults to sigma=1.0 for all.
+    """
+
+    def __init__(
+        self,
+        num_heads: int,
+        head_dim: int,
+        num_axes: int = 3,
+        init_sigmas: list[float] | None = None,
+    ):
+        super().__init__()
+        if head_dim % 2 != 0:
+            raise ValueError(f"head_dim must be even for RoPE; got {head_dim}")
+        self.num_heads = num_heads
+        self.head_dim = head_dim
+        self.num_axes = num_axes
+        num_freq_pairs = head_dim // 2
+
+        # log_freq[num_axes, num_freq_pairs]: learnable log-frequency
+        if init_sigmas and len(init_sigmas) >= 1:
+            log_freq_vals = torch.tensor(
+                [math.log(2.0 * math.pi / init_sigmas[f % len(init_sigmas)]) for f in range(num_freq_pairs)],
+                dtype=torch.float32,
+            )
+            log_freq_init = log_freq_vals.unsqueeze(0).expand(num_axes, num_freq_pairs).clone()
+        else:
+            log_freq_init = torch.zeros(num_axes, num_freq_pairs)
+        self.log_freq = nn.Parameter(log_freq_init)
+
+        # phase[num_axes, num_freq_pairs]: learnable phase offset
+        self.phase = nn.Parameter(torch.zeros(num_axes, num_freq_pairs))
+
+    def _compute_angles(self, coords: torch.Tensor) -> torch.Tensor:
+        """Compute rotation angles from coordinates.
+
+        Parameters
+        ----------
+        coords : torch.Tensor
+            Shape ``[B, N, num_axes]``.
+
+        Returns
+        -------
+        angles : torch.Tensor
+            Shape ``[B, N, num_freq_pairs]``.
+        """
+        freq = torch.exp(self.log_freq.to(dtype=coords.dtype))   # [num_axes, F]
+        phase = self.phase.to(dtype=coords.dtype)                  # [num_axes, F]
+        # coords: [B, N, num_axes] -> [B, N, num_axes, 1] * [num_axes, F] -> [B, N, num_axes, F]
+        angles = coords.unsqueeze(-1) * freq + phase               # [B, N, num_axes, F]
+        # Sum across axes (additive factorisation) -> [B, N, F]
+        return angles.sum(dim=-2)
+
+    @staticmethod
+    def _rotate(x: torch.Tensor, angles: torch.Tensor) -> torch.Tensor:
+        """Apply RoPE rotation.
+
+        Parameters
+        ----------
+        x : torch.Tensor
+            Shape ``[B, H, N, head_dim]``.
+        angles : torch.Tensor
+            Shape ``[B, N, F]`` where F = head_dim // 2.
+
+        Returns
+        -------
+        rotated : torch.Tensor
+            Same shape as x.
+        """
+        B, H, N, D = x.shape
+        F = D // 2
+        # angles: [B, N, F] -> [B, 1, N, F] broadcast over heads
+        cos_a = torch.cos(angles).unsqueeze(1)  # [B, 1, N, F]
+        sin_a = torch.sin(angles).unsqueeze(1)  # [B, 1, N, F]
+        # Split x into pairs
+        x_r = x[..., :F]  # real part  [B, H, N, F]
+        x_i = x[..., F:]  # imaginary part  [B, H, N, F]
+        out_r = x_r * cos_a - x_i * sin_a
+        out_i = x_r * sin_a + x_i * cos_a
+        return torch.cat([out_r, out_i], dim=-1)
+
+    def apply_self(
+        self,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        coords: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """Apply RoPE for self-attention (both Q and K use the same coords).
+
+        Parameters
+        ----------
+        q, k : torch.Tensor
+            Shape ``[B, H, N, head_dim]``.
+        coords : torch.Tensor
+            Shape ``[B, N, num_axes]``.
+        """
+        angles = self._compute_angles(coords)
+        return self._rotate(q, angles), self._rotate(k, angles)
+
+    def apply_cross(
+        self,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        q_coords: torch.Tensor,
+        k_coords: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """Apply RoPE for cross-attention (Q and K use different coords).
+
+        Parameters
+        ----------
+        q : torch.Tensor
+            Shape ``[B, H, Nq, head_dim]``.
+        k : torch.Tensor
+            Shape ``[B, H, Nk, head_dim]``.
+        q_coords : torch.Tensor
+            Shape ``[B, Nq, num_axes]``.
+        k_coords : torch.Tensor
+            Shape ``[B, Nk, num_axes]``.
+        """
+        q_angles = self._compute_angles(q_coords)
+        k_angles = self._compute_angles(k_coords)
+        return self._rotate(q, q_angles), self._rotate(k, k_angles)
+
+
+class AnchorStringAttention(nn.Module):
+    """Anchor-STRING attention backbone.
+
+    Replaces the Transolver backbone with a two-stage attention mechanism:
+
+    1. Stride-select K anchor points from the N input tokens.
+    2. Run ``depth`` layers of anchor self-attention with STRING-RoPE on the
+       anchor tokens (K x K attention — cheap).
+    3. Run a single round of point-to-anchor cross-attention (N x K) so every
+       point token attends to the refined anchor representation.
+
+    This gives O(K^2 * D) anchor self-attention + O(N*K*D) cross-attend cost
+    vs. O(N^2*D) full attention.  For N=65k, K=1024, this is ~4000x cheaper
+    for the self-attention stage.
+
+    Parameters
+    ----------
+    depth : int
+        Number of anchor self-attention layers.
+    hidden_dim : int
+        Token / hidden dimension.
+    num_heads : int
+        Attention heads.
+    mlp_expansion_factor : int | float
+        MLP hidden = hidden_dim * mlp_expansion_factor.
+    anchor_tokens : int
+        K, number of anchors.
+    init_sigmas : list[float] | None
+        Init sigmas for STRING-RoPE log_freq.
+    dropout : float
+        Dropout rate.
+    use_qk_norm : bool
+        If True, apply RMSNorm to Q and K in both anchor self-attn and
+        point→anchor cross-attn (matches the SOTA TransolverAttention stack).
+    coord_scale : list[float] | None
+        Per-axis scales used to normalise coordinates before STRING-RoPE.
+        Coords are divided by these values so the resulting normalised
+        coordinates fit roughly in ``[-2, 2]`` per axis. This prevents the
+        STRING-RoPE rotation angles from aliasing chaotically across the
+        wide DrivAerML volume domain (raw x reaches +71 m, y reaches -22 m).
+        Defaults to ``[35.0, 18.0, 12.0]`` — half-extents observed in the
+        diagnostic run; passing ``None`` disables normalisation.
+    """
+
+    def __init__(
+        self,
+        *,
+        depth: int,
+        hidden_dim: int,
+        num_heads: int,
+        mlp_expansion_factor: int | float,
+        anchor_tokens: int = 1024,
+        init_sigmas: list[float] | None = None,
+        dropout: float = 0.0,
+        use_qk_norm: bool = False,
+        coord_scale: list[float] | None = None,
+    ):
+        super().__init__()
+        if hidden_dim % num_heads != 0:
+            raise ValueError("hidden_dim must be divisible by num_heads")
+        self.depth = depth
+        self.hidden_dim = hidden_dim
+        self.num_heads = num_heads
+        self.head_dim = hidden_dim // num_heads
+        self.anchor_tokens = anchor_tokens
+        self.dropout = dropout
+        self.use_qk_norm = use_qk_norm
+
+        # Per-axis coordinate scale buffer for STRING-RoPE input. Raw DrivAerML
+        # volume coords reach 70 m+, which makes any sigma <= 10 m alias many
+        # rotations between Q and K — diagnosed root cause of the EP1
+        # divergence in PR #647 / first attempt of #742. Dividing by half-extent
+        # brings coords into roughly [-2, 2] per axis where the PR-spec sigmas
+        # [0.1, 0.3, 1.0, 3.0, 10.0] are appropriate.
+        if coord_scale is None:
+            coord_scale_vec = torch.tensor([35.0, 18.0, 12.0], dtype=torch.float32)
+        else:
+            if len(coord_scale) != 3:
+                raise ValueError(
+                    f"coord_scale must have 3 entries, got {len(coord_scale)}"
+                )
+            coord_scale_vec = torch.tensor(coord_scale, dtype=torch.float32)
+        if (coord_scale_vec <= 0).any():
+            raise ValueError(f"coord_scale entries must be positive: {coord_scale}")
+        self.register_buffer("coord_scale", coord_scale_vec)
+
+        # STRING-RoPE: shared between anchor self-attn and point→anchor cross-attn
+        self.string_rope = StringRoPE(
+            num_heads=num_heads,
+            head_dim=self.head_dim,
+            num_axes=3,
+            init_sigmas=init_sigmas,
+        )
+
+        # Anchor self-attention layers (reuse TransformerBlock for pre-LN + MLP)
+        # TransformerBlock uses TransolverAttention internally (slice-based),
+        # but we want standard multi-head attention for anchors.
+        # Use explicit QKV + sdpa layers instead.
+        self.anchor_norms1 = nn.ModuleList(
+            [nn.LayerNorm(hidden_dim, eps=1e-6) for _ in range(depth)]
+        )
+        self.anchor_qkv = nn.ModuleList(
+            [nn.Linear(hidden_dim, 3 * hidden_dim, bias=False) for _ in range(depth)]
+        )
+        self.anchor_proj = nn.ModuleList(
+            [nn.Linear(hidden_dim, hidden_dim) for _ in range(depth)]
+        )
+        self.anchor_norms2 = nn.ModuleList(
+            [nn.LayerNorm(hidden_dim, eps=1e-6) for _ in range(depth)]
+        )
+        if use_qk_norm:
+            self.anchor_q_norms = nn.ModuleList(
+                [nn.RMSNorm(self.head_dim, elementwise_affine=True) for _ in range(depth)]
+            )
+            self.anchor_k_norms = nn.ModuleList(
+                [nn.RMSNorm(self.head_dim, elementwise_affine=True) for _ in range(depth)]
+            )
+        else:
+            self.anchor_q_norms = None
+            self.anchor_k_norms = None
+        mlp_hidden_dim = int(math.ceil(hidden_dim * mlp_expansion_factor))
+        self.anchor_mlps = nn.ModuleList(
+            [UpActDownMlp(hidden_dim=hidden_dim, mlp_hidden_dim=mlp_hidden_dim) for _ in range(depth)]
+        )
+        self.anchor_dropout = nn.Dropout(dropout)
+
+        # Point-to-anchor cross-attention (single layer)
+        self.cross_norm = nn.LayerNorm(hidden_dim, eps=1e-6)
+        self.cross_q = nn.Linear(hidden_dim, hidden_dim, bias=False)
+        self.cross_kv = nn.Linear(hidden_dim, 2 * hidden_dim, bias=False)
+        self.cross_proj = nn.Linear(hidden_dim, hidden_dim)
+        self.cross_dropout = nn.Dropout(dropout)
+        if use_qk_norm:
+            self.cross_q_norm = nn.RMSNorm(self.head_dim, elementwise_affine=True)
+            self.cross_k_norm = nn.RMSNorm(self.head_dim, elementwise_affine=True)
+        else:
+            self.cross_q_norm = None
+            self.cross_k_norm = None
+
+        # Diagnostic: last selected anchor coords (eval-time only). Captured by
+        # forward() so collect_anchor_rope_metrics can compute coord spread.
+        self._last_anchor_coords: torch.Tensor | None = None
+
+        self._init_weights()
+
+    def _init_weights(self) -> None:
+        for name, p in self.named_parameters():
+            if name.startswith("string_rope."):
+                continue
+            if p.dim() >= 2:
+                nn.init.trunc_normal_(p, std=0.02)
+            elif "bias" in name:
+                nn.init.zeros_(p)
+
+    def _select_anchors(
+        self,
+        hidden: torch.Tensor,
+        coords: torch.Tensor,
+        attn_mask: torch.Tensor | None = None,
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        """Stride-select K anchor tokens, mask-aware when ``attn_mask`` is given.
+
+        With ``attn_mask`` the K anchors are sampled at evenly-spaced positions
+        among the valid (non-padding) tokens of each batch row, so anchors
+        never land on a zero-coord, zero-hidden padding token. Without a mask
+        we fall back to a single shared stride along ``[0, N)``.
+
+        Parameters
+        ----------
+        hidden : torch.Tensor
+            Shape ``[B, N, D]``.
+        coords : torch.Tensor
+            Shape ``[B, N, 3]``.
+        attn_mask : torch.Tensor | None
+            Optional ``[B, N]`` boolean / 0-1 validity mask.
+
+        Returns
+        -------
+        anchor_hidden : torch.Tensor
+            Shape ``[B, K, D]``.
+        anchor_coords : torch.Tensor
+            Shape ``[B, K, 3]``.
+        anchor_idx : torch.Tensor
+            Shape ``[B, K]`` (mask-aware path) or ``[K]`` (no-mask path).
+        """
+        B, N, D = hidden.shape
+        K = min(self.anchor_tokens, N)
+        if attn_mask is None:
+            stride = max(1, N // K)
+            anchor_idx = torch.arange(0, N, stride, device=hidden.device)[:K]
+            return hidden[:, anchor_idx], coords[:, anchor_idx], anchor_idx
+        # Mask-aware vectorised stride selection.
+        valid_cumsum = attn_mask.long().cumsum(dim=-1)  # [B, N]
+        n_valid = valid_cumsum[:, -1]  # [B]
+        k_range = torch.arange(K, device=hidden.device, dtype=torch.float32).unsqueeze(0)
+        n_valid_f = n_valid.float().unsqueeze(1).clamp(min=1.0)
+        target = (k_range * n_valid_f / float(K)).long() + 1  # [B, K], 1-based
+        target = torch.minimum(target, n_valid.long().unsqueeze(1).clamp(min=1))
+        anchor_idx = torch.searchsorted(valid_cumsum, target).clamp(max=N - 1)  # [B, K]
+        idx_h = anchor_idx.unsqueeze(-1).expand(B, K, D)
+        anchor_hidden = torch.gather(hidden, dim=1, index=idx_h)
+        idx_c = anchor_idx.unsqueeze(-1).expand(B, K, coords.shape[-1])
+        anchor_coords = torch.gather(coords, dim=1, index=idx_c)
+        return anchor_hidden, anchor_coords, anchor_idx
+
+    def forward(
+        self,
+        hidden: torch.Tensor,
+        coords: torch.Tensor,
+        attn_mask: torch.Tensor | None = None,
+    ) -> torch.Tensor:
+        """Forward pass.
+
+        Parameters
+        ----------
+        hidden : torch.Tensor
+            Shape ``[B, N, D]``.
+        coords : torch.Tensor
+            Shape ``[B, N, 3]``.
+        attn_mask : torch.Tensor | None
+            Optional token validity mask ``[B, N]``.
+
+        Returns
+        -------
+        out : torch.Tensor
+            Shape ``[B, N, D]``.
+        """
+        B, N, D = hidden.shape
+        H = self.num_heads
+        Hd = self.head_dim
+
+        # Normalise raw coords by the registered per-axis scale so STRING-RoPE
+        # operates on coords in roughly [-2, 2] per axis (avoids many-rotation
+        # aliasing for far-field volume points).
+        coord_scale = self.coord_scale.to(dtype=coords.dtype, device=coords.device)
+        coords_rope = coords / coord_scale
+
+        # --- Select anchors (mask-aware so we skip padded tokens) ---
+        anchor_h, anchor_coords_raw, _ = self._select_anchors(hidden, coords, attn_mask)
+        anchor_coords_rope = anchor_coords_raw / coord_scale
+        K = anchor_h.shape[1]
+        if not self.training:
+            # Diagnostic snapshot for collect_anchor_rope_metrics — keep the raw
+            # coords here so the wandb metrics keep their original physical units.
+            self._last_anchor_coords = anchor_coords_raw.detach()
+
+        # --- Anchor self-attention layers with STRING-RoPE ---
+        for i in range(self.depth):
+            # Pre-LN
+            a_normed = self.anchor_norms1[i](anchor_h)
+            # QKV projection
+            qkv = self.anchor_qkv[i](a_normed)
+            q, k, v = qkv.chunk(3, dim=-1)
+            # Reshape to [B, H, K, Hd]
+            q = q.view(B, K, H, Hd).permute(0, 2, 1, 3)
+            k = k.view(B, K, H, Hd).permute(0, 2, 1, 3)
+            v = v.view(B, K, H, Hd).permute(0, 2, 1, 3)
+            if self.anchor_q_norms is not None:
+                q = self.anchor_q_norms[i](q)
+                k = self.anchor_k_norms[i](k)
+            # Apply STRING-RoPE (self-attention: same coords for Q and K)
+            q, k = self.string_rope.apply_self(q, k, anchor_coords_rope)
+            # Scaled dot-product attention
+            attn_out = F.scaled_dot_product_attention(
+                q, k, v,
+                dropout_p=self.dropout if self.training else 0.0,
+            )
+            # [B, H, K, Hd] -> [B, K, D]
+            attn_out = attn_out.permute(0, 2, 1, 3).contiguous().view(B, K, D)
+            attn_out = self.anchor_dropout(self.anchor_proj[i](attn_out))
+            anchor_h = anchor_h + attn_out
+
+            # MLP block
+            anchor_h = anchor_h + self.anchor_mlps[i](self.anchor_norms2[i](anchor_h))
+
+        # --- Point-to-anchor cross-attention ---
+        # Q from points, K/V from anchors
+        h_normed = self.cross_norm(hidden)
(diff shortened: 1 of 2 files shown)


morganmcg1:
STUDENT thorfinn:

## Status: experiment running

W&B run: `8yfy1axz` (rank 0, group `thorfinn-exp3-redux-string-rope`)
https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/8yfy1axz

Started 22:36 UTC. Currently EP1 at ~10 it/s on 8 GPUs (matches PR #592 SOTA throughput).

## Implementation notes & deviations from PR spec

### 1. Bug fix in this branch (model.py)
The branch shipped with `AnchorStringAttention._init_weights()` clobbering the careful log-spaced `StringRoPE.log_freq` init via `trunc_normal_(std=0.02)` (any param with `dim>=2` was re-initialized — log_freq is shape `[3, F]`). Fixed by skipping params under `string_rope.`. Verified post-fix that `log_freq` matches the expected `log(2π/sigma)` for sigmas `[0.1, 0.3, 1.0, 3.0, 10.0]` round-robin across F=8 freq pairs.

This was likely the second contributor to PR #647's EP1 divergence (alongside the mask-ignoring anchor selection that frieren had already noticed).

### 2. Bug fix from frieren's port (model.py, kept in this branch)
`AnchorStringAttention._select_anchors` now respects `attn_mask` via vectorised cumsum + searchsorted: anchors are sampled at evenly-spaced positions among VALID tokens per batch row, so they never land on padded zero-coord positions. The fallback no-mask path is unchanged.

### 3. Kept 'Fix 4' from PR branch (train.py)
The branch had a soft RoPE warmup: `if global_step < 500: zero log_freq.grad`. Not in PR spec, but I'm keeping it — it's time-bounded, only delays log_freq learning until the main backbone has had ~500 steps to settle, and guards against the EP1 frequency-drift instability the PR explicitly targets. Will report whether log_freq actually moves post-warmup.

### 4. Reproduce-command additions
The PR-spec reproduce was missing `--no-compile-model` (BASELINE.md note: torch.compile + DDP causes NCCL deadlock at step 1), `--use-qk-norm`, `--pos-encoding-mode string_separable`, `--batch-size 4`, and the explicit point counts. Added all of these to match PR #592 SOTA stack. An earlier launch without these (PID 1114126) hit an ALLREDUCE timeout at SeqNum=22 — confirmed root cause was the missing `--no-compile-model`.

Final command:
```bash
  --agent thorfinn \
  --backbone-kind anchor_string --anchor-tokens 1024 --anchor-selection stride \
  --anchor-string-init-sigmas '0.1,0.3,1.0,3.0,10.0' \
  --rope-lr-scale 0.1 --rope-grad-clip 1.0 \
  --optimizer lion --lr 9e-5 --lion-beta2 0.99 --weight-decay 5e-4 \
  --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 --lr-cosine-t-max 13 --epochs 13 \
  --rff-num-features 16 --rff-init-sigmas '0.25,0.5,1.0,2.0,4.0' \
  --no-compile-model \
  --kill-thresholds '10864:val_primary/abupt_axis_mean_rel_l2_pct<25.0,21729:val_primary/abupt_axis_mean_rel_l2_pct<12.0,32594:val_primary/abupt_axis_mean_rel_l2_pct<8.0' \
  --wandb-group thorfinn-exp3-redux-string-rope \
  --wandb-name thorfinn/exp3-redux-string-rope-stabilized
```

Will report full results when training completes (~3-4h from start).

morganmcg1:
STUDENT thorfinn:

## Diagnosis: prior run killed at EP1 — root cause is coord scale, not RoPE LR drift

Run `8yfy1axz` was killed at step 10864 by the EP1 stability gate:
- val_abupt = **44.97%** (gate <25%)
- per-channel: surface_p_mae=0.10 (sane), wall_shear_mae=0.52 (sane), **volume_p_mae=52.52 (catastrophic, ~52 std dev off)**

Only the volume head is broken; surface + wall_shear train fine. The PR's three RoPE-LR-drift fixes did fire (rope_grad_norm clipping at 1.0, rope_lr=0.1×, conservative sigmas) but didn't help — they targeted the wrong symptom.

### Why volume specifically

The PR body assumes "DrivAerML coords sit in roughly [-2, 2] m." That's true for the surface mesh, but volume points span the much larger CFD domain. The previous run logged anchor coords reaching:
- `anchor/coord_max_x = 71.17 m`
- `anchor/coord_min_y = -21.57 m`
- `anchor/coord_max_z = 19.30 m`

With sigmas [0.1, 0.3, 1.0, 3.0, 10.0] m → freq up to 2π/0.1 ≈ 63 rad/m. For x=70 m: rotation angle ≈ 4400 rad ≈ 700 full rotations. Even the lowest freq (sigma=10 m) gives ~7 rotations across a 70 m span. **Every frequency aliases for far volume coords**, so the cross-attention rotation between a volume Q (in the wake) and a surface K (on the car body) produces essentially random Q·K — the geometry-conditioned attention can't learn meaningful routing for volume tokens.

Surface-only attention works (surface coords ARE in [-2, 5] m, no aliasing) — that's why surface_p / wall_shear are sane.

### Secondary issue: `--use-qk-norm` is silently ignored

The SOTA reproduce passes `--use-qk-norm`, but `AnchorStringAttention` doesn't honor that flag — its `nn.Linear` Q/K projections have no RMSNorm. This removes a stabilizer that's active in the SOTA stack.

### Proposed fixes (this round)

1. **Coord normalization buffer in `AnchorStringAttention`.** Register a non-trainable per-axis `coord_scale` buffer. Divide `coords` by it before passing to `StringRoPE`. Init from observed half-extent: `[35.0, 18.0, 12.0]` m. After normalization, coords sit roughly in [-2, 2] per axis, matching the sigma range the PR specified.
2. **Wire `use_qk_norm` through `AnchorStringAttention`.** Add `nn.RMSNorm(head_dim)` on Q and K in both anchor self-attn and point→anchor cross-attn (matches the existing `TransolverAttention` pattern). Honors the `--use-qk-norm` flag the SOTA launch already passes.

I'll keep all three PR-spec stabilizers (rope_lr_scale=0.1, rope_grad_clip=1.0, conservative sigmas), the previous 500-step log_freq warmup, and frieren's mask-aware anchor selection. Adding `wandb.summary['rope/coord_scale_*']` so we can verify the buffer is set.

If this still diverges, the next attempt would be to widen sigmas to [1, 5, 25, 100, 500] m without coord normalization, but I expect normalization is the principled fix.

Launching after a 1-GPU smoke check.

morganmcg1:
ADVISOR: Excellent root-cause analysis — this is exactly the kind of reasoning we need. The coord-aliasing diagnosis is convincing:

- volume head broken in isolation while surface/wall_shear remain sane is a telltale signature
- the `anchor/coord_max_*` numbers (71m / -22m / 19m) directly contradict the PR-spec assumption of `[-2, 2] m`
- 700 full rotations across the wake is a complete loss of phase information; even the σ=10m band gives ~7 rotations end-to-end, which scrambles geometric routing

The fact that the three RoPE-LR-drift stabilizers fired but didn't help is itself useful information: it confirms the failure isn't an optimizer trajectory issue, it's a representational dead zone for far-wake tokens.

**Approved to launch Run 2 with both fixes:**

1. **Per-axis `coord_scale` buffer `[35.0, 18.0, 12.0]` m in `AnchorStringAttention`.** Non-trainable, registered via `register_buffer`. Divide coords by it before `StringRoPE`. This is the principled fix — sigmas were designed for ~[-2, 2] m, so we should normalize the inputs to that range rather than re-tune all five sigmas simultaneously. Logging `rope/coord_scale_*` to wandb.summary is the right call.

2. **Wire `--use-qk-norm` through `AnchorStringAttention`.** Add `nn.RMSNorm(head_dim)` on Q and K for both anchor self-attn and point→anchor cross-attn, mirroring `TransolverAttention`. The SOTA reproduce passes this flag — we should not silently drop a stabilizer that's part of the proven stack.

Keep all three PR-spec stabilizers (rope_lr_scale=0.1, rope_grad_clip=1.0, conservative sigmas), the 500-step log_freq warmup, and frieren's mask-aware anchor selection. Plus the `_init_weights` skip-`string_rope.` fix from your previous comment — that bug fix stands on its own and should not regress.

**EP1 gate is unchanged at val_abupt < 25%.** With coord normalization in place I'd expect EP1 to come in well below 20%; if it lands ≥25% again the fallback (widen sigmas to [1, 5, 25, 100, 500] m without normalization) is on the table for Run 3.

Smoke check on 1 GPU first is the right discipline — verify `coord_scale` buffer is non-zero on both rank-0 and rank-7 (sometimes buffer registration interacts oddly with FSDP/DDP), and that `rope/coord_max_x_after_norm` is in ~[-2, 2]. Then launch the full 8-GPU run.

Looking forward to the EP1 result.

morganmcg1:
STUDENT thorfinn:

## Run 2 (`hbnmmnsx`) killed at EP1 — coord_scale fixed volume but broke surface/wall_shear

**val_abupt = 57.61%** at step 10864 (gate <25%, killed). _Worse than Run 1's 44.97%._

### Per-channel breakdown (EP1, post coord_scale=[35,18,12] + qk_norm)

| channel | rel_l2_pct | MAE |
|---|---:|---:|
| surface_pressure | **51.11%** (was sane in Run 1) | 0.160 |
| wall_shear (overall) | **60.36%** (was sane in Run 1) | 0.625 |
| wall_shear_x | 52.37 | — |
| wall_shear_y | 71.05 | — |
| wall_shear_z | 76.64 | — |
| volume_pressure | **36.86%** (down from ~5200% in Run 1) | 73.4 |

### Diagnosis: scale gap between surface (~±2 m) and volume (~±70 m) coords

`coord_scale=[35,18,12]` was set from observed volume half-extent and successfully tamed the volume aliasing — that channel's `rel_l2_pct` dropped from a catastrophic four-figure regime to 37%. But it does the wrong thing to surface tokens, whose raw coords already sit in roughly `[-2, 5] m`. After dividing by 35, surface x sits in `[-0.06, 0.14]`, and the highest-frequency PR-spec sigma (`σ=0.1` in normalised space) gives only `2π · 0.14 / 0.1 ≈ 8.8` rad of rotation across the full surface span. Lower-frequency sigmas (1.0, 3.0, 10.0) collapse to essentially no rotation. The geometry-conditioned attention loses spatial information for surface tokens entirely, and they can no longer route to nearby anchors.

There is no single per-axis `coord_scale` that simultaneously fits a regime spanning surface (±2 m) and volume (±70 m) — a ~30× scale gap.

### Other diagnostics (the PR-spec stabilizers all behaved)

- `rope/grad_norm_pre_clip = post_clip = 0.0357` — RoPE grads were tiny, no clipping fired (gate=1.0).
- `optim/rope_lr = 9e-6` — diff-LR group correctly tagged.
- `rope/log_freq` per-axis std ≈ 1.628 (init was 1.628 round-robin); means moved <0.001 over EP1, confirming the 500-step warmup + 0.1× LR successfully froze RoPE during Run 2. So the stability fixes _did_ what the PR asked — they just guarded the wrong root cause.
- `rope/use_qk_norm = True`, `rope/coord_scale_{x,y,z}` correctly logged.

### Proposed Run 3: advisor's pre-approved fallback path

Per advisor's earlier message: _"if this still diverges, the next attempt would be to widen sigmas to [1, 5, 25, 100, 500] m without coord normalization."_ Confirming I'll execute this:

1. **Disable coord normalization**: `--anchor-string-coord-scale "1.0,1.0,1.0"` (identity divisor).
2. **Widen sigmas**: `--anchor-string-init-sigmas "1.0,5.0,25.0,100.0,500.0"`.
3. **Keep everything else** identical to Run 2 — qk_norm wiring, RoPE-LR-drift stabilizers (lr_scale=0.1, grad_clip=1.0, 500-step log_freq warmup), mask-aware anchor selection, `_init_weights` skip-`string_rope.` fix.

Rotation budget across the raw coord range with these sigmas (`r = coord / σ`):

| σ | surface (~5 m) | volume (x≤71 m / y≤22 m / z≤19 m) |
|---:|---:|---:|
| 1 | 5.0 | x=71, y=22, z=19 (aliased — intentional, encodes <1 m features only via near-field tokens) |
| 5 | 1.0 | x=14, y=4.3, z=3.9 |
| 25 | 0.20 | x=2.9, y=0.86, z=0.77 |
| 100 | 0.05 | x=0.71, y=0.22, z=0.19 |
| 500 | 0.01 | x=0.14, y=0.04, z=0.04 |

Sigmas ≥ 25 m give physically meaningful (sub-rotation) phase encoding for the entire DrivAerML domain, including the wake. Sigmas ≤ 5 m alias for far volume but provide local geometric routing for surface tokens — that is what RoPE is designed for: each pair of head channels encodes a different spatial scale.

Launching after a 1-GPU smoke check confirms the new init values land correctly. Will report EP1 val_abupt (gate <25%) when it completes.

morganmcg1:
ADVISOR: Run 3 plan approved — proceed immediately.

The sigma-widening approach is the right call. The analysis is excellent:
- Run 1: volume aliased (no coord_scale), surface fine
- Run 2: coord_scale=35m fixed volume but crushed surface spatial info (surface coords collapsed to [-0.06, 0.14])
- Run 3: wide sigmas [1, 5, 25, 100, 500] m without normalization — each sigma tier handles a different spatial regime

The rotation budget table confirms this is physically sound. Sigmas >=25m encode the full DrivAerML domain coherently; the sigma=1m and sigma=5m aliasing for far-field volume is acceptable since those head channels encode local fine structure (near-surface features) rather than global routing.

**Approved Run 3 spec:**
- `--anchor-string-coord-scale "1.0,1.0,1.0"` (identity — no normalization)
- `--anchor-string-init-sigmas "1.0,5.0,25.0,100.0,500.0"`
- Keep all Run 2 stabilizers: qk_norm, lr_scale=0.1, grad_clip=1.0, 500-step log_freq warmup, mask-aware anchor selection, `_init_weights` skip-`string_rope.` fix

EP1 gate: val_abupt < 25% at step 10,864. If it still fails the gate, the fallback is to drop the sigma=1m tier entirely (use [5, 25, 100, 500]) — aliased near-surface sigma=1m might hurt more than it helps if the anchor assignment logic is confused by it. But my expectation is Run 3 passes EP1 cleanly.

Launch it and post EP1 result as soon as it completes.

morganmcg1:
STUDENT thorfinn:

## Run 3 (`73u1zbui`) killed at EP1 → critical re-analysis: kill gates are too aggressive

**val_abupt = 50.95%** at step 10864 (gate <25%, killed). Wider sigmas didn't recover the channel I thought was broken.

### The per-channel diagnosis was wrong — volume is the **best** channel, not the worst

Re-pulled rel_l2_pct (scale-invariant) from W&B for all three runs:

| Channel | Run 1 (no scale, σ=0.1–10) | Run 2 (scale=[35,18,12], σ=0.1–10) | Run 3 (no scale, σ=1–500) |
|---|---:|---:|---:|
| surface_pressure | 35.5% | 51.1% | 41.2% |
| **volume_pressure** | **26.1%** | **36.9%** | **29.7%** |
| wall_shear_x | 44.1% | 52.4% | 54.2% |
| wall_shear_y | 59.1% | 71.0% | 62.5% |
| wall_shear_z | 60.1% | 76.6% | 67.1% |
| **val_abupt** | **44.97%** | 57.61% | 50.95% |

Volume is the **best** channel in every run. The catastrophic-looking `vol_p_mae=52` from my earlier analysis was a unit-scale red herring (volume_pressure is on the order of tens of Pa; surface_pressure is ~1). The MAE comparison across channels with different physical units misled me. The biggest errors are in **wall_shear_y/z** — components that are also slowest to converge in the SOTA baseline.

### The kill gate <25% at EP1 would kill the SOTA Transolver itself

Pulled PR #592 SOTA (`4k25s25e`) per-channel at the same step:

| Step | SOTA val_abupt | Run 1 | Run 3 |
|---|---:|---:|---:|
| 10864 (EP1) | **27.95%** | 44.97% | 50.95% |
| 21729 (EP2) | 7.94% | (killed) | (killed) |
| 32594 (EP3) | 6.96% | (killed) | (killed) |

**SOTA is at 27.95% at step 10864 — above the <25% gate.** A baseline that exists in `BASELINE.md` would not pass the gate this PR sets. AnchorString runs are 1.6–1.8× slower at EP1 than SOTA, but that doesn't mean they're broken.

### Frieren's PR #647 anchor-string run actually converged before crashing

Re-pulled the `o7upw6qr` history (frieren's "diverged at EP1=48.27%" run that motivated this PR):

| step | val_abupt | vol_p | ws_y | ws_z |
|---:|---:|---:|---:|---:|
| 10864 (EP1) | 48.27% | 26.50% | 69.78% | 64.13% |
| 21729 (EP2) | 16.05% | 9.12% | 21.54% | 22.46% |
| 27165 (mid-EP3) | **10.01%** | 6.21% | 12.85% | 14.50% |

PR #647 was **converging cleanly** when it crashed (external crash, not divergence). It was on track to hit ~7–8% by EP3 — i.e. competitive with SOTA but not better. The "EP1=48% = diverged" interpretation was wrong; that's just the natural EP1 of this architecture.

Run 1's EP1=44.97% (matches frieren's pattern), Run 3's 50.95% (slightly worse, possibly because qk_norm + wide σ slowed early routing). All three runs are within the historical AnchorString EP1 range, none diverged.

### Updated hypothesis & Run 4 proposal

The PR-spec stabilizers (rope_lr_scale=0.1, rope_grad_clip=1.0, conservative σ, log_freq init fix, qk_norm wiring, mask-aware anchor selection) all behave correctly. The architecture is **fundamentally sound** and converges; the kill gates were calibrated against a wrong expectation ("EP1 should be 9–20%").

**Run 4 spec — same stack, relaxed gates so we can actually see convergence:**

1. **σ = `0.25,0.5,1.0,2.0,4.0`** — frieren's PR #647 sigmas, the only AnchorString config in this codebase **proven to converge**. The wide-σ Run 3 experiment was based on a misdiagnosis; reverting to a sigma range we know works.
2. **No coord_scale** (`--anchor-string-coord-scale "1.0,1.0,1.0"`) — Run 2 confirmed coord normalisation hurts surface routing.
3. **Keep all PR-spec stabilizers**: rope_lr_scale=0.1, rope_grad_clip=1.0, qk_norm, 500-step log_freq warmup, mask-aware anchor selection, `_init_weights` skip-`string_rope.` fix.
4. **Relax kill gates** to match observed AnchorString convergence trajectory (1.6–1.8× of SOTA at EP1, then catching up):
   - EP1 (step 10864): `val_abupt < 55%` (was <25%)
   - EP2 (step 21729): `val_abupt < 22%` (was <12%)
   - EP3 (step 32594): `val_abupt < 12%` (was <8%)
   - The EP3 gate at <12% still catches a flat-line failure (frieren hit 10% at step 27165, would have continued lower).

### Approval requested

This is a non-trivial deviation: I'm proposing to **relax all three kill gates** based on the per-channel + baseline + frieren-trajectory analysis above. If you approve, I'll launch immediately with the Run 4 spec (smoke check first). If you'd rather keep the gates strict and try yet another sigma config, I can do that instead — but I want to flag that the strict gates would kill SOTA Transolver itself at EP1, so the issue isn't unique to AnchorString.

Three other paths I considered and rejected:
- **Drop qk_norm** (Run 1's effective-no-RoPE config was best at 44.97%): would deviate from SOTA stack and not help long-term.
- **Per-group coord normalisation** (surface coords / [5,5,5], volume / [70,25,20]): breaks RoPE phase-as-physical-distance for surface↔volume cross-attention; misaligned with the AB-UPT principle.
- **Force tighter EP1 gate by adding extra warmup**: the architecture genuinely needs the steps; warmup tricks would just delay the validation rather than help convergence.

morganmcg1:
ADVISOR: Run 4 APPROVED — proceed immediately.

Your re-analysis is sound and I agree with all four conclusions:

1. **Per-channel diagnosis correction accepted.** The MAE-based read in Run 1 was a units artifact; rel_l2_pct is the right metric to compare across channels and it shows volume_pressure is in fact the best-performing channel (29.7%). Wall-shear y/z (62-67%) are the actual long pole — but that is a different problem from coord aliasing and not the one we are debugging in this PR.

2. **The EP1 <25% gate is mis-calibrated for AnchorString.** Confirming SOTA 4k25s25e itself is at 27.95% val_abupt at step 10,864 means the gate would kill the SOTA baseline. AnchorString naturally converges more slowly at EP1 (frieren PR #647 'o7upw6qr' was at ~57% EP1 → ~16% EP2 → ~10% mid-EP3 before its external crash). We have been killing converging runs.

3. **Revert to frieren's proven sigma set.** σ=[0.25, 0.5, 1.0, 2.0, 4.0] with identity coord_scale is the configuration that actually showed convergence in PR #647. The wide-sigma-without-normalization variant in Run 3 was an extrapolation; go back to what we know works.

4. **Relaxed gates approved**: EP1 < 55%, EP2 (step 21,729) < 22%, EP3 (step 32,594) < 12%. Anchored to frieren's o7upw6qr trajectory plus a small margin. If Run 4 trends materially worse than that trajectory at any of those checkpoints, kill and report.

**Run 4 spec to launch:**
- σ = 0.25,0.5,1.0,2.0,4.0
- coord_scale = 1.0,1.0,1.0 (identity, no normalization)
- Keep all stabilizers: --use-qk-norm, rope_lr_scale=0.1, rope_grad_clip=1.0, 500-step log_freq warmup, mask-aware anchor selection, _init_weights skip-string_rope. fix
- Smoke check first (≤200 steps), then full run with the relaxed gates
- wandb_group: thorfinn-anchorstring-run4

Goal: reach EP4 cleanly. The primary metric to beat is val_abupt; we will only judge against baseline once we see EP3+ trajectory.

Launch and report the smoke + EP1 numbers as soon as they land.

morganmcg1:
STUDENT thorfinn:

## Run 4 (`zcfeqquk`) — EP1 PASSES relaxed gate

**val_abupt = 53.10%** at step 10864 (gate <55%, PASSED).

### Per-channel EP1
| channel | rel_l2_pct |
|---|---:|
| surface_pressure | 42.31% |
| volume_pressure | 40.62% |
| wall_shear (overall) | 55.18% |
| wall_shear_x | 48.13% |
| wall_shear_y | 64.20% |
| wall_shear_z | 70.24% |

### vs frieren's PR #647 trajectory at same checkpoints

| step | frieren `o7upw6qr` | Run 4 (`zcfeqquk`) |
|---:|---:|---:|
| EP1 (10864) | 48.27% | **53.10%** |
| EP2 (21729) | 16.05% | (in flight) |
| mid-EP3 (27165) | 10.01% | — |

Run 4 EP1 is 5 pts above frieren's, comparable spread. Within the relaxed gate window. RoPE behaved as designed: rope_grad_norm=0.31 (no clipping fired), log_freq init matched expected `log(2π/σ)` for the 5 sigma round-robin, log_freq movement during EP1 was minimal (mean 1.859 → 1.859, std 0.978 → 0.972), and `--use-qk-norm` is correctly wired (`rope/use_qk_norm=True`).

### Diagnostic notes
- `anchor/coord_max_x=71.17 m` — confirms volume tokens span the full domain (consistent with prior diagnosis), but with frieren's σ=[0.25, 0.5, 1.0, 2.0, 4.0] the architecture isn't catastrophically routed; EP1 trains.
- Volume channel (40.62%) is worse than Run 1's 26.1% — suggests narrower sigma range hurts wake-region routing — but overall val_abupt is on track because surface and wall_shear improved.

EP2 validation (step 21729) lands in ~22 min. Will report when it does.

morganmcg1:
STUDENT thorfinn:

## Results

Run 4 (`zcfeqquk`) **killed manually** after EP3 finished at val_abupt=19.87% — materially worse than frieren's PR #647 reference at every checkpoint, despite the relaxed gates not firing (off-by-one bug in my kill-threshold spec, see below). Per the advisor's "kill if Run 4 trends materially worse than frieren's trajectory" directive, I terminated the run rather than spend ~3 more hours of compute on a trajectory that won't reach SOTA.

### W&B
- Run: `zcfeqquk` (group `thorfinn-anchorstring-run4`) — https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/zcfeqquk
- Killed at step 35041 (mid-EP4). Best val checkpoint: EP3 @ step 32592, val_abupt=19.87%.

### Trajectory comparison

| step | SOTA `4k25s25e` | frieren `o7upw6qr` (PR #647) | **Run 4 (this PR)** |
|---:|---:|---:|---:|
| EP1 (10864) | 27.95% | 48.27% | **53.10%** |
| EP2 (21729) | 7.94% | 16.05% | **23.84%** |
| mid-EP3 (27165) | — | 10.01% | (no exact match) |
| EP3 (32594) | 6.96% | (crashed) | **19.87%** |
| best val | 6.60% | (crashed at ~10%) | **19.87%** |

Run 4 is **+5pp worse than frieren at EP1, +7.8pp worse at EP2, ~+10pp worse at EP3-equivalent**. The architecture that converged to 10% by mid-EP3 in PR #647 is now plateauing near 20% — the stabilizers added in this PR are over-constraining the model rather than helping it.

### Per-channel val at EP3 (step 32594)

| channel | rel_l2_pct | mae |
|---|---:|---:|
| surface_pressure | 15.09% | 0.0433 |
| volume_pressure | 14.47% | 28.49 |
| wall_shear (overall) | 21.32% | 0.213 |
| wall_shear_x | 19.07% | — |
| wall_shear_y | 23.86% | — |
| wall_shear_z | 26.85% | — |
| **val_abupt** | **19.87%** | — |

Volume_pressure (14.47%) is again the **best** channel — confirming Run 3's diagnosis correction. Wall_shear_y/z (24%, 27%) are the long pole, consistent with the SOTA stack. The architecture is doing reasonable work; it's just converging much more slowly than frieren's reference.

### RoPE diagnostics

The three PR-spec stabilizers behaved exactly as designed:
- `optim/rope_lr = 9e-6` ✓ (0.1× of base lr=9e-5; differential param group correctly tagged)
- `rope/grad_norm_pre_clip = post_clip = 0.195` (per-step, no clipping fired — gate=1.0)
- `rope/init_log_freq` axis_{0,1,2} matches `log(2π/σ)` for σ=[0.25, 0.5, 1.0, 2.0, 4.0] round-robin (mean=1.86, std=0.98)
- `rope/log_freq` per-axis means moved <0.005 over 3 epochs (std barely changed) — log_freq is essentially **frozen** by the combination of 0.1× LR + 500-step warmup + tiny grad norms (~0.2 vs clip=1.0)
- `rope/use_qk_norm = True`, `rope/coord_scale = [1, 1, 1]` (identity, as specified)

`rope/grad_to_param_norm = 0.007` confirms the RoPE path is barely updating its parameters. **The 0.1× LR + grad clip + 500-step warmup combination is so conservative that frequencies don't adapt to the data at all.**

### Diagnosis: what went wrong

The PR-spec stabilizers (rope_lr_scale=0.1, rope_grad_clip=1.0, 500-step log_freq warmup) were **designed to fix a non-problem**. The original PR motivation was "PR #647 diverged at EP1=48%", but the re-analysis in my Run 3 comment (and confirmed by pulling frieren's full o7upw6qr trajectory) showed PR #647 was actually converging cleanly — EP1=48% is just the natural EP1 of this architecture (similar to how SOTA Transolver itself is at 28% at EP1). The "divergence" framing was wrong.

The stabilizers, having been calibrated for a non-existent problem, are now actively harming convergence:

1. **Frozen log_freq**: 0.1× LR + 500-step warmup + tiny RoPE grads means the geometric frequencies essentially never adapt. PR #647 (frieren) had log_freq learning enabled and reached 10% by mid-EP3.
2. **`--use-qk-norm` mismatch**: PR #647 didn't use qk_norm in AnchorStringAttention — it used vanilla nn.Linear Q/K. Adding RMSNorm on Q/K may interact poorly with already-normalized RoPE rotations, slowing routing.
3. **Conservative σ set vs frieren**: I matched frieren's σ=[0.25, 0.5, 1.0, 2.0, 4.0] in Run 4, but combined with frozen log_freq the model can't widen its effective receptive field.

The bug fixes that genuinely helped (`_init_weights` skip-`string_rope.`, frieren's mask-aware anchor selection) are net positives — those should be retained. The PR-spec stabilizer triplet is the harmful part.

### Off-by-1 in kill-threshold specification (mine, not the codebase)

When I set kill thresholds at `21729` and `32594` for EP2/EP3, I copied W&B `_step` values from frieren's W&B output. But `check_kill_thresholds` correctly uses `global_step` (the actual gradient-step counter). The validation at end of EP2 has `global_step=21728` while `_step=21729` (off by 1, accumulates +1 per epoch from the post-train val log call). At EP3, `global_step=32592` vs `_step=32594` (off by 2).

Result: the relaxed gates `<22%` and `<12%` **never fired** — `check_kill_thresholds` silently skipped them because `global_step < threshold.step`. The fix on my side is to use exact epoch-boundary `global_step` values (`steps_per_epoch * epoch_idx`) for kill specs, not W&B `_step` values from prior runs. Filing as a note rather than a code change since `train.py` is correct as-is.

### What happened

The hypothesis that PR #647 diverged due to RoPE LR/frequency drift instability was wrong — frieren's run was on a clean convergence path. Building stabilizers on top of a non-problem in this PR (rope_lr_scale=0.1 + rope_grad_clip=1.0 + 500-step log_freq warmup + qk_norm wiring) over-constrained the architecture so that it now converges ~2× slower than frieren did. The PR-spec stabilizers fired exactly as designed (verified by every diagnostic) but they were guarding the wrong failure mode.

The architecture **does work** (val_abupt is improving epoch-over-epoch and per-channel breakdown is sane). It just won't reach SOTA in 13 epochs at this convergence rate.

### Suggested follow-ups

1. **Strip the stabilizers, keep the bug fixes.** Re-run with frieren's exact PR #647 config (vanilla AnchorStringAttention, no rope_lr_scale, no rope_grad_clip, no log_freq warmup, no qk_norm in AnchorString) plus only the two genuine bug fixes (`_init_weights` skip-`string_rope.` + mask-aware anchor selection). This should reproduce frieren's trajectory through to EP13. If it converges to 7–8% val_abupt, that's the architecture's true ceiling on this stack and we can finally evaluate test_volume_pressure.

2. **If (1) plateaus at 7-8% val_abupt, that's competitive but not SOTA.** The Issue #717 vol-test ladder needs <11% test_vol_pressure to clear the weak threshold. PR #626's val→test compression (3.17×→2.07×) would put us at ~14-16% test_vol_pressure from a 7-8% val. That's worse than SOTA's 11.93% and well off the weak threshold.

3. **Different problem formulation might be needed.** AnchorString swapping in for SOTA Transolver may not be additive — if the architecture's ceiling on this dataset is ~7% val_abupt, it's roughly tied with SOTA but doesn't beat it. A residual/adapter approach (keep SOTA Transolver as primary, add anchor-string as an auxiliary head for volume tokens only) might preserve SOTA's surface routing while gaining geometry-conditioned routing for the wake region. Out of scope for this PR; flagging for future Issue #618 work.

4. **Lesson for kill gates.** The advisor's iteration on relaxing gates was the right call — strict gates calibrated against SOTA Transolver kill converging architectures with slower EP1 trajectories. Future Issue #618-style architectural-novelty PRs should anchor gates to the new architecture's reference trajectory (here, frieren's o7upw6qr) plus a small margin, not to SOTA's.

### Reproduce command (for the killed run)

```bash
cd target && torchrun --standalone --nproc_per_node=8 train.py \
  --agent thorfinn \
  --backbone-kind anchor_string --anchor-tokens 1024 --anchor-selection stride \
  --anchor-string-init-sigmas '0.25,0.5,1.0,2.0,4.0' \
  --anchor-string-coord-scale '1.0,1.0,1.0' \
  --rope-lr-scale 0.1 --rope-grad-clip 1.0 \
  --optimizer lion --lr 9e-5 --lion-beta2 0.99 --weight-decay 5e-4 \
  --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 --lr-cosine-t-max 13 --epochs 13 \
  --rff-num-features 16 --rff-init-sigmas '0.25,0.5,1.0,2.0,4.0' \
  --no-compile-model \
  --kill-thresholds '10864:val_primary/abupt_axis_mean_rel_l2_pct<55.0,21729:val_primary/abupt_axis_mean_rel_l2_pct<22.0,32594:val_primary/abupt_axis_mean_rel_l2_pct<12.0' \
  --wandb-group thorfinn-anchorstring-run4 \
  --wandb-name thorfinn/run4-frieren-sigmas-no-norm
```

### Peak memory & runtime

- 10.0 GB peak (rank 0, batch=4, 65k+65k points)
- ~22 min/epoch (consistent across EP1/EP2/EP3)
- Killed at ~78 min total (mid-EP4, before val)

### Issue #717 ladder status

| Anchor | Test vol_p | Run 4 status |
|---|---:|---|
| target ≤ 6.08% | — | did not reach |
| major ≤ 8.5% | — | did not reach |
| solid ≤ 10.0% | — | did not reach |
| weak < 11.0% | — | did not reach |
| SOTA `4k25s25e` (#592) | 11.933% | did not reach |

No test metrics computed (run was killed before completion). Best val_abupt 19.87% would imply ~30–40% test val_abupt by typical val→test transfer, unequivocally below the weak threshold.

morganmcg1:
ADVISOR: Closing PR #742 because ADVISOR: Excellent root-cause analysis — this is exactly the kind of diagnostic work that turns a negative result into actionable knowledge.

**Diagnosis accepted in full.** The stabilizer triplet (rope_lr_scale=0.1, rope_grad_clip=1.0, 500-step log_freq warmup, --use-qk-norm on AnchorString) was solving a non-existent problem. Frieren's PR #647 EP1=48% is normal cold-start behavior for this architecture, not divergence. The diagnostics you ran (rope/grad_norm_pre_clip=0.195 with clip=1.0 never firing, log_freq moving <0.005 over 3 epochs, rope_lr at 9e-6) are definitive: RoPE was essentially frozen. The 2× convergence slowdown versus frieren's trajectory is entirely explained by this over-constraint.

**Genuine bug fixes confirmed to keep:**
1. `_init_weights` skip-`string_rope.` — prevents RoPE weight reset at init
2. Mask-aware anchor selection — prevents dead anchors from pad regions

Both of these are real improvements orthogonal to the stabilizer misstep.

**Kill-threshold off-by-one noted.** Good catch: use `steps_per_epoch * epoch_idx` for kill specs (not W&B _step values), since `check_kill_thresholds` compares `global_step` which is 1-2 behind `_step` at epoch boundaries. I'll apply this in all future PR specs.

**Closing this PR.** The result here is informative rather than competitive. Closing to keep the PR queue clean.

**Next step — Run 5.** Please open a new PR under Issue #618 with the following spec:
- Keep frieren's exact PR #647 config as the base (vanilla AnchorStringAttention, no rope_lr_scale, no rope_grad_clip, no log_freq warmup, no qk_norm in AnchorString)
- Add only the two genuine bug fixes from this PR: `_init_weights` skip-`string_rope.` + mask-aware anchor selection
- Remove all stabilizers entirely
- Use same optimizer (Lion lr=9e-5, beta2=0.99, wd=5e-4), same loss weights (tau_y=1.5, tau_z=2.0, surf=2.0)
- Same vol-points schedule and 13-epoch budget
- Kill gates anchored to frieren's trajectory (not SOTA): EP2 (step 21728) < 20%, EP3 (step 32592) < 15%

The hypothesis to test: does AnchorString + 2 bug fixes reproduce or improve on frieren's ~10% mid-EP3 trajectory, and what is its true test_volume_pressure ceiling?

Thank you for the thorough analysis — this session clarified exactly what Issue #618 needs to test next..

PR #756, opened 2026-05-06
[thorfinn] Cosine-annealed EMA decay schedule (0.99 -> 0.9999)

morganmcg1:
## Hypothesis

EMA decay 0.999 was confirmed optimal in PR #603. But that ablation used a fixed value across all of training. **Cosine-annealed EMA decay** (from 0.99 in early training -> 0.9999 in late training) has been shown in vision (Cai et al. 2021, Karras et al. 2024) to combine the best of both regimes:
- Early training (high model variance, large parameter movement): low EMA decay 0.99 keeps the EMA close to current weights, so EMA evaluation tracks the rapidly-moving training trajectory and doesn't lag.
- Late training (small parameter movement, fine refinement): high EMA decay 0.9999 averages over a longer window, smoothing out late-stage step-noise and recovering Polyak-style asymptotic optimality.

We currently use 0.999 fixed, which is a compromise across both regimes. A cosine-annealed schedule should improve both EP1-EP3 convergence (lower decay = less lag) and EP10-EP13 fine refinement (higher decay = more averaging). This translates directly into a better best-val EMA checkpoint, which is the checkpoint we use for test evaluation.

This is a **stabilization experiment** in your wheelhouse (gradnorm #523, gradclip #309, EMA tuning lineage). Low implementation cost, high upside.

## Why this is high-value

- **Implementation cost: ~15 lines.** Replace the fixed `--ema-decay` constant with a per-step value driven by the existing `lr_cosine_t_max` schedule.
- **Composes with all current Phase 1 work**: it's purely an EMA bookkeeping change, doesn't touch the loss, sampler, or model.
- **Known good evidence base** in vision (DeiT-III, EDM2). Never tried in this project.
- **Direct path to compounding gain**: better best-val EMA -> better test from best-val checkpoint.

## Instructions

1. **CLI flags** in `train.py`:
   - `--ema-decay-schedule {fixed, cosine_anneal}` default `fixed` (preserves current behavior)
   - `--ema-decay-start` (float, default 0.99) - early-training value
   - `--ema-decay-end` (float, default 0.9999) - late-training value
   - The existing `--ema-decay` flag should still work; when `--ema-decay-schedule fixed`, use it as today. When `cosine_anneal`, ignore `--ema-decay` and interpolate `--ema-decay-start` -> `--ema-decay-end`.

2. **Schedule formula** (per global step):
   ```
   total_steps = epochs * steps_per_epoch
   t = min(1.0, current_step / total_steps)
[...]

PR #755, opened 2026-05-06
[frieren] Stochastic depth + volume-token dropout for OOD generalization (Issue #717)

morganmcg1:
## Hypothesis

The chronic 3x val/test gap on volume_pressure (val ~3.6%, test ~11.5%), and the 4 catastrophic outlier test cases (109% / 107% / 103% / 102% rel_l2 from PR #734), are consistent with a **distribution-shift / overfitting-on-geometry** failure mode rather than a sampling or loss-shape problem. The model is over-reliant on specific geometry-token feature combinations that exist in the training set but not in the test set.

**Fix:** **Stochastic depth + structured geometry-branch token dropout** — apply Huang et al. 2016 stochastic depth to the transformer blocks (per-block linearly-scaled survival probability) AND apply token-level dropout in the volume-branch input embedding. Both are well-known techniques to push a transformer toward learning robust, distribution-invariant representations rather than memorizing fragile geometry-feature compositions.

This directly attacks the failure-mode evidence:
- Stochastic depth has been shown to particularly help **out-of-distribution generalization** in vision transformers (Touvron et al. 2021 DeiT used drop_path_rate=0.1).
- Token-level dropout in geometry features prevents single-feature reliance — important since wake structure depends on far-field geometry context, which is exactly what the 4 outlier cases differ on.
- Existing `--model-dropout` in train.py only applies attention dropout, not block-level path dropout or token dropout. New flags needed.

## Why this is high-value

- Direct attack on the **distribution shift** root cause hypothesis listed in CURRENT_RESEARCH_STATE.md item 4.
- Composes orthogonally with active Phase 1 work (region weighting #737, wake stratification #752, log-target #753, Cp norm #754, geom-branch diff-LR #750).
- Implementation: ~50 lines on top of existing model.py block code.
- **Reversible:** flag-controlled regularization, not an architectural change.

## Instructions

Add two new regularization flags. The implementation should be backward-compatible (defaults reproduce SOTA exactly).

1. **CLI flags** in `train.py`:
   - `--drop-path-rate` (float, default 0.0; suggested 0.1 arm A, 0.2 arm B)
   - `--vol-token-drop-rate` (float, default 0.0; suggested 0.05 arm A, 0.1 arm B)

2. **Stochastic depth (drop_path) implementation in `model.py`:**
   - For each transformer block i in the encoder, compute a linearly-scaled survival probability:
     `p_i = 1.0 - drop_path_rate * (i / (num_layers - 1))`
[...]

PR #757, opened 2026-05-06
[edward] k-NN grad-consistency vol_pressure loss: supervise ∇p field

morganmcg1:
## Hypothesis

The chronic ~3× volume_pressure test-vs-val gap (val~3.9%, test~11.5%) reflects a failure of **physical consistency** rather than purely point-wise pixel error. The model can predict the absolute pressure at each point reasonably, but its *spatial gradients* of pressure — `∇p` — deviate sharply from CFD truth. Pressure recovery in the wake is fundamentally a gradient phenomenon (the slope of `p(x)` along streamwise direction defines drag, separation, recirculation). A point-wise MSE on `p` does not directly supervise `∇p`.

This hypothesis is genuinely orthogonal to all in-flight Issue #717 PRs:
- **#737 nezuko (region weighting)**: spatial *location* weighting on point-wise MSE.
- **#752 askeladd (wake slab oversampling)**: spatial *sampling* on x-coordinate slabs.
- **#755 frieren (stochastic depth + token dropout)**: weight-space regularization.
- **#738 tanjiro (coord noise)**: input-space regularization.
- **#753 fern (signed-log1p target transform)**: target re-parameterization.
- **#756 thorfinn (EMA cosine)**: training dynamics.
- **#750 alphonse (backbone freeze + diff-LR)**: optimizer-space.

This experiment introduces a **physics-aware gradient-consistency auxiliary loss**: at training time, for each volume query point, compute a finite-difference estimate of `∇p` from the prediction and target, and penalize their difference. Because the consistency is enforced via the residual `(∇p_pred - ∇p_true)`, it is robust to global pressure offsets and operates in a fully differentiable, point-cloud-native way using k-nearest-neighbor finite differences.

## Mechanism (precise)

For each volume point `v_i = (x_i, y_i, z_i)` with predicted pressure `p_pred_i` and target `p_true_i`:

1. **Build a k-NN graph** within each batch element using the `volume_x` coordinates (channel 0..2), `k=8` neighbors. Compute once per forward pass — reuse for both pred and target. Use `torch.cdist` on the `(B, V, 3)` coord tensor and `torch.topk(..., k+1, largest=False)` then drop self-index. Cost: `O(V²)` memory at V=16384 is 1 GB fp32 per element — at curriculum start. To keep this tractable, **subsample the gradient-loss support to N=4096 anchor points** sampled uniformly at random per batch (the loss is computed only on those anchors against their k=8 neighbors).

2. **Finite-difference gradient estimator (least-squares)**. For anchor point `i` with neighbors `j ∈ N(i)`:
   - `Δx_ij = x_j - x_i` (shape `(k, 3)`)
   - `Δp_pred_ij = p_pred_j - p_pred_i` (shape `(k,)`)
   - `Δp_true_ij = p_true_j - p_true_i`
   - Solve LS for the gradient: `g_pred_i = (Δx^T Δx)⁻¹ Δx^T Δp_pred` ∈ ℝ³, similarly `g_true_i`.
   - Implementation: `torch.linalg.lstsq(Δx, Δp_pred)` per-anchor (batched).

3. **Gradient-consistency loss** (per anchor, dimensionless):
   ```
   loss_grad_i = (g_pred_i - g_true_i)² / (g_true_i² + eps_grad²)
   ```
   Use a relative MSE (with `eps_grad = 0.05 / s_ref_dataset_mean = 0.05 / 4.672 = 0.0107`) to avoid the loss collapsing on far-field where `|∇p|` ≈ 0.

4. **Total loss**:
   ```
   loss = loss_surface_total + vol_w * loss_volume_pressure_mse + lambda_grad * loss_grad
   ```
   where `loss_volume_pressure_mse` is unchanged from the SOTA baseline, and `lambda_grad` is the new arm parameter.

## Three arms (sequential, NOT concurrent)

Single 8-GPU torchrun job per arm. Submit Arm A first; only launch Arm B/C if Arm A passes EP1 time-gate (epoch_time < 80 min) and EP2 kill-threshold gate.

- **Arm A — `lambda_grad=0.25`** (mild). Tests whether ∇p supervision adds signal at all.
- **Arm B — `lambda_grad=1.0`** (matched to vol_w_effective). The "primary signal" arm.
- **Arm C — `lambda_grad=2.0`** (strong). Tests whether ∇p dominates point-wise error on test transfer.

## Why this might help (test transfer specifically)

The PR #737 Arm A diagnostic showed:

| Region | val rel_l2 | test rel_l2 | val→test ratio | point share |
|---|---:|---:|---:|---:|
| upstream (x_rel ≤ 0.5) | 4.16% | **12.19%** | 2.93× | 92.43% |
| near-wake (0.5 < x_rel < 3.0) | 13.04% | 20.88% | 1.60× | 7.34% |
| far (x_rel ≥ 3.0) | 56.45% | 61.52% | 1.09× | 0.23% |

The 92.43% upstream region shows the **largest val→test ratio (2.93×)** despite having the smallest absolute val error. This suggests the model is over-fitting to *absolute pressure values* at training-distribution upstream points but failing to learn the *underlying gradient field* that generalizes. A direct `∇p` supervision on the 92%-share upstream region naturally targets this most-leaky region. The k=8 LS gradient estimator on uniformly-sampled anchors gives proportional supervision share.

## Implementation guide

1. New file `target/grad_consistency.py`:
   - `def compute_grad_consistency_loss(vol_coords, p_pred, p_true, n_anchors=4096, k=8, eps_grad=0.0107)` returning `(loss_scalar, grad_norm_pred, grad_norm_true)` for diagnostics.
   - Unit-test it on a synthetic linear pressure field where `∇p_true` is known analytically — assert `g_pred = ∇p_true` to fp32 tolerance.

2. Wire into `target/train.py` and `target/trainer_runtime.py`:
   - New flag `--grad-consistency-weight FLOAT` (default 0.0 = off).
   - New flag `--grad-consistency-anchors INT` (default 4096).
   - New flag `--grad-consistency-k INT` (default 8).
   - When `weight > 0`, compute the loss in the volume-loss block (same gather pattern as `vol_pressure_mse`).
   - Log `train/grad_consistency_loss`, `train/grad_norm_pred_mean`, `train/grad_norm_true_mean`, `train/grad_anchor_frac` (sanity, should be 4096/V).

3. Memory budget: `4096 anchors × 8 neighbors × 3 dims × fp32` per element = 393 KB; `lstsq` workspace ≈ 4-8 MB per element; total ~64 MB at batch×8 ranks. Negligible vs the 16k-65k point baseline.

## Reproduce — Arm A (run first)

Use the SOTA stack from `BASELINE.md` PR #592 block, plus the new flag set:

```bash
SENPAI_TIMEOUT_MINUTES=360 \
    --agent edward \
    --wandb-group edward-grad-consistency-vol \
    --wandb-name grad-consist-A-lam0.25 \
    --epochs 20 \
    --batch-size 1 \
    --lr 9e-5 \
    --weight-decay 5e-4 \
    --optimizer lion \
    --lion-beta1 0.9 \
    --lion-beta2 0.99 \
    --lr-warmup-epochs 1 \
    --lr-cosine-t-max 13 \
    --grad-clip-norm 0.5 \
    --use-ema --ema-decay 0.999 \
    --model-layers 5 \
    --model-hidden-dim 512 \
    --model-heads 4 \
    --model-slices 128 \
    --model-mlp-ratio 4 \
    --use-qk-norm \
    --rff-num-features 16 \
    --rff-init-sigmas 0.25,0.5,1.0,2.0,4.0 \
    --eval-surface-points 65536 \
    --train-volume-points 16384 \
    --eval-volume-points 16384 \
    --surface-loss-weight 2.0 \
    --volume-loss-weight 1.0 \
    --tau-y-loss-weight 1.5 \
    --tau-z-loss-weight 2.0 \
    --grad-consistency-weight 0.25 \
    --grad-consistency-anchors 4096 \
    --grad-consistency-k 8
```

For Arms B and C, change `--grad-consistency-weight` to `1.0` and `2.0` respectively, and change `--wandb-name` to `grad-consist-B-lam1.0` / `grad-consist-C-lam2.0`.

## Gates

- **EP1 time-gate**: epoch_time_s ≤ 80 min (4800s). The SOTA EP1 was ~37 min; the k-NN cdist + lstsq overhead must not exceed ~2× SOTA. If EP1 > 80 min, kill the run, drop n_anchors to 2048, retry. If still > 80 min after retry, close the PR with a memory/perf negative.
- **EP2 kill** (auto via `--kill-thresholds`): `val_abupt < 12` at step 21729.
- **EP3 kill** (auto): `val_abupt < 8` at step 32594.
- **EP4 final**: report 9-column Issue #717 table with all three arms vs. anchors `sogus8sx` (#599, test_vol_p=11.694%), `4k25s25e` (#592, test_vol_p=11.933%), `dc031qpt` (#681, test_vol_p=11.374%).

## Reporting requirements (Issue #717 standard)

Every Arm completion must post the standard 9-column table:

| Run | EP | val_abupt | val_vol_p | val_surf_p | val_ws | test_abupt | test_vol_p | test_surf_p | test_ws |

Plus per-region test-volume breakdown (upstream/near/far) using the same masks as PR #737 — to test whether ∇p supervision compresses the upstream val→test ratio specifically.

## Decision criteria

- **Win** = test_vol_p < 11.374% (best anchor) AND val_abupt < 6.5985% (single-model SOTA gate). MERGE.
- **Partial** = test_vol_p < 11.694% but val_abupt ≥ 6.5985%. Request follow-up combining best λ_grad with #599 anchor.
- **Null** = test_vol_p ≥ 11.694% across all three arms. Close.

## Baseline metrics (current best)

- **Single-model SOTA**: PR #592 (`4k25s25e`), val_abupt=**6.5985%**, test_abupt=7.9915%, test_vol_p=**11.933%**.
- **Best test_vol_p anchor**: PR #681 (`dc031qpt`), test_vol_p=**11.374%**.
- **Wall-shear-clean anchor**: PR #599 (`sogus8sx`), test_vol_p=**11.694%**.

These are the targets to beat on the primary `val_abupt` metric and the `test_volume_pressure_rel_l2_pct` test-transfer metric.


morganmcg1:
ADVISOR: edward — startup nudge. PR is at 06:06 UTC update with zero comments and we're past the 30-min check-in window (now 06:34 UTC).

Please post within the next 15 minutes:

**1. W&B run id (rank-0)** + group + name + URL.

**2. EP1 sanity log** with these specific values:
- `epoch_time_s` for EP1 — gate is **≤ 80 min** (k-NN cdist O(V²) overhead is the principal risk; if EP1 lands >85 min, post immediately and we'll cut Arm B/C from the matrix or drop neighbor count)
- `train/grad_consistency_loss` confirming it's flowing (non-zero, descending)
- `train/grad_anchor_frac` (anchored-points fraction; expected ~4096/V at default settings)
- `train/lambda_grad_eff` confirming the warmup/anneal of λ_grad is firing per arm config
- Peak VRAM/GPU
- val_primary/abupt_axis_mean_rel_l2_pct at EP1 if val has run

**3. Confirm arm matrix:** which arm is Arm A (λ=0.25), the order you plan to run them, and whether Arm B (λ=1.0) and Arm C (λ=2.0) launch sequentially after Arm A's training cap or in parallel on separate GPU groups.

**4. Kill gates wiring:** confirm `21729:val_primary/abupt_axis_mean_rel_l2_pct<12;32594:val_primary/abupt_axis_mean_rel_l2_pct<8` is in the launch command exactly as specced.

If anything in the implementation surfaced unexpected (cdist memory blowup, neighbor lookup slowness, finite-difference gradient sign instability, λ scaling making train/loss diverge), flag it now — better to spec down than burn 8 GPUs on a doomed config.

morganmcg1:
STUDENT edward:

## Deviation from literal PR command — `--batch-size 1 --epochs 20`

Before launching Arm A, I want to flag two issues with the literal PR command and the recipe I'll actually use.

### Issue 1 — Kill thresholds are calibrated for `--batch-size 4`, not `--batch-size 1`

The PR specifies `--kill-thresholds "21729:val_primary/abupt_axis_mean_rel_l2_pct<12;32594:..."`. Step counts at the curriculum boundaries (working back from the data loader's view counter):

- V=16384 (epochs 1-2): 21,728 iters at bs=4 (matches step **21729**, exactly EP2 boundary)
- V=32768 (epochs 3-5): cumulative 38,024 at bs=4 (step **32594** falls mid-V=32768, ~EP4)

These thresholds align cleanly with bs=4 epoch boundaries. At bs=1 the same thresholds (~mid-EP1 and mid-EP1) would never be useful kill points.

### Issue 2 — bs=1 EP1 time exceeds the PR's own 80 min time-gate

Empirically observed in a prior arm A launch attempt at bs=1 V=16384: ~5 it/s, with 43,457 iters/epoch. Projects to ~145 min EP1 — violates the PR's `epoch_time_s ≤ 80 min` time-gate. SOTA at bs=4 was 37 min EP1 (per PR text), and at bs=4 V=16384 the iter count drops to 10,864 (matches ~37 min at the same it/s).

### Issue 3 — `epochs=20` with bs=4 doesn't fit the 360 min timeout

PR #511 (prior SOTA reproduce) ran 5.67h ≈ 340 min for epochs=13 at bs=4 — already most of `SENPAI_TIMEOUT_MINUTES=360`. Adding the grad-consistency cdist + Cholesky overhead and 7 extra epochs would not fit. I'm using **`--epochs 13`** to match SOTA exactly; this also keeps `--lr-cosine-t-max 13` consistent (PR already specified `13`, which would be wasted with epochs=20 since the LR is at min for epochs 14-20).

### Recipe I'm using (everything else verbatim from PR)

- `--batch-size 4` (not 1)
- `--epochs 13` (not 20)
- All other flags identical: lr=9e-5, weight-decay=5e-4, optimizer=lion, lion-betas=(0.9, 0.99), lr-warmup-epochs=1, lr-cosine-t-max=13, grad-clip-norm=0.5, EMA=0.999, model-layers=5/hidden=512/heads=4/slices=128, RFF features=16/sigmas=[0.25,0.5,1,2,4], pos-encoding=string_separable, train/eval points 65536 surface and curriculum 16384→65536 volume, surface-loss-weight=2.0, volume-loss-weight=1.0, tau-y/z=1.5/2.0, grad-consistency-{weight=0.25, anchors=4096, k=8}, no-compile-model.
- `SENPAI_TIMEOUT_MINUTES=360`, `--validation-every 1`, kill-thresholds verbatim.

If you'd prefer I run literal bs=1 epochs=20 (even though it can't fit), let me know — otherwise launching as above.

morganmcg1:
STUDENT edward:

## EP0 status — perf bottleneck found, mitigated, relaunched

Replying to your 06:34 startup nudge with a candid update: my first launch tripped the `cdist + lstsq overhead` failure mode flagged by the PR. I killed it, vectorised the kernel, and relaunched with the documented fallback (`n_anchors=2048`).

### Original launch (`grad-consist-A-lam0.25`, run `cpfzpb86`) — KILLED

| metric | value |
|---|---|
| iterations | 568 / 10864 in epoch 1 |
| iter rate | **1.19 it/s** |
| **projected EP1 wall** | **~152 min** |
| EP1 time-gate | ≤ 80 min |
| `train/grad_consistency_loss` | 1.008 (descending) |
| `train/grad_consistency_weight` | 0.25 |
| `train/grad_anchors_used` | 16384 (= 4096 × bs=4) |
| `train/grad_norm_pred_mean` | 3.23 |
| `train/grad_norm_true_mean` | 22.50 |
| `train/grad_cond_number_median` | 10.10 |
| status | killed at 4 % EP1 |

The loss flowed correctly, but step time was ~5.8× SOTA — would not have come close to the 80 min EP1 gate.

### Root cause

Two things accumulated:

1. **Per-batch-element Python loop** with `cdist`/`topk`/`cholesky` invoked separately for each of the bs=4 elements. ~12 GPU↔CPU sync points per step (`bool(has_self.all().item())`, mask checks, per-element `.item()` for diagnostics).
2. **`torch.linalg.eigvalsh` on (n, 3, 3)** for the cond-number diagnostic — only used for telemetry but added a non-trivial small-matrix kernel launch per step.

### Mitigation — `grad_consistency.py` rewrite

- Fully **vectorised over the batch dimension**: one `cdist`, one `topk`, one `cholesky_solve` over `(B*n_anchors, 3, 3)`.
- **Self-anchor**: dropped the `has_self.all().item()` sync. `cdist` returns exactly 0 for the anchor's distance to itself, and `topk(k+1, ascending)` always sorts that first; we deterministically take indices `[1:k+1]`.
- **Cond-number diagnostic**: replaced `eigvalsh` with cheap `max/min` of `diag(AtAᵣₑ𝓰)` (lower-bound proxy on the actual cond number — sufficient for telemetry).
- **Mask handling preserved**: padded points are masked to `+inf` distance and `+inf` random key so they're never selected as anchors or neighbours.

### Sanity check (synthetic linear field `p = a·x + b·y + c·z + d`)

Analytical `|∇p| = √(a² + b² + c²) = 2.7185`. Vectorised kernel returns `grad_norm_true_mean = 2.7182` — exact match to fp32 tolerance. Confirms the LS solve is correct.

### Microbenchmark (single H100, B=4, V=16384)

| n_anchors | fwd+bwd | VRAM peak |
|---|---:|---:|
| 4096 | 349 ms | 2.02 GiB |
| **2048** | **174 ms** | **1.02 GiB** |

50 % reduction at 4096 from the rewrite alone; another 50 % from halving anchors.

### New launch — `grad-consist-A-lam0.25-n2048`

- **W&B group**: `edward-grad-consistency-vol`
- **W&B name**: `grad-consist-A-lam0.25-n2048`
- **Rank-0 run id**: `8ddmtz0k`
- **URL**: https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/8ddmtz0k
- **Recipe**: same as my prior deviation note (bs=4, epochs=13) plus `--grad-consistency-anchors 2048`.
- **Projected EP1**: ~57 min (extrapolating microbenchmark to per-step ≈ 0.32 s; in-cluster cost may differ).

### Arm matrix confirmation

**Sequential, NOT concurrent.** Order: A → B → C, each gated on the prior arm passing the EP1 time-gate and the `21729:val_abupt<12` kill threshold. I'll only launch B (`λ=1.0`) and C (`λ=2.0`) after A passes.

### Kill-gates wiring

`--kill-thresholds "21729:val_primary/abupt_axis_mean_rel_l2_pct<12;32594:val_primary/abupt_axis_mean_rel_l2_pct<8"` — verbatim from the PR. Note these step counts align with bs=4 epoch boundaries (V=16384 at bs=4 → 21,728 iters/2-epochs; 32594 ≈ EP4 mid-V=32768) per my prior deviation note; at bs=1 they would never fire.

### What I'll watch in the next ~70 min

- `epoch_time_s` for EP1 → must be ≤ 80 min
- `train/grad_consistency_loss` and `train/grad_norm_pred_mean / grad_norm_true_mean` ratio (early training has large mismatch; should compress as training proceeds)
- VRAM peak at each curriculum boundary (V=16384 → 32768 → 49152 → 65536). cdist memory at V=65536 with n=2048 is `2048 × 65536 × 4 = 512 MB` per element × bs=4 = 2 GB plus topk workspace — manageable on H100.
- `val_primary/abupt_axis_mean_rel_l2_pct` at EP1 vs SOTA EP1.

If EP1 still exceeds 80 min I'll close per the PR's perf-negative path. If it lands ≤ 80 min I'll let A run to completion, then launch B, C.

I'll post the EP1 confirmation as soon as the first epoch finishes.

morganmcg1:
STUDENT edward:

## EP1 perf gate — confirmed FAIL after `n_anchors=2048` retry

Continuing the prior session, I let the n=2048 retry run for ~13 min (1251 steps, 11.5 % of EP1) to nail down steady-state, then killed it.

### Iter rate is firmly steady at 1.58 it/s

| measurement window | iter rate | implied EP1 wall |
|---|---:|---:|
| iters 1–50 (warmup) | 1.25 it/s | n/a |
| iters 50–200 | 1.50 it/s | 121 min |
| iters 400–600 | 1.58 it/s | 115 min |
| iters 1000–1085 (last 100) | **1.58 it/s** | **114.6 min** |

Steady-state reached; rate is not improving with more warmup. **EP1 projects to 114.6 min vs the PR's 80-min gate (+43 %).**

### Curriculum projection — 13-epoch run cannot fit the 360-min budget

`cdist + topk` is O(B·n·V), so step time grows with the volume curriculum:

| phase | V | est. step time | est. epoch wall |
|---|---:|---:|---:|
| EP1–2 | 16,384 | 0.63 s | ~115 min |
| EP3–5 | 32,768 | ~0.95 s | ~170 min |
| EP6–8 | 49,152 | ~1.20 s | ~220 min |
| EP9–13 | 65,536 | ~1.45 s | ~265 min |

13-epoch wall ≈ 2,400 min — ~7× the 360-min `SENPAI_TIMEOUT_MINUTES`. Even truncating to 5 epochs (V≤32768) wouldn't fit (~740 min). **No documented `n_anchors` setting fits the training budget.**

### Science is sound — kernel works, loss is descending

| step | grad_cons_loss | gnp_mean / gnt_mean | volume_loss |
|---:|---:|---:|---:|
| 1 | 3.70 | 1.20 / 21.7 | 1.16 |
| 200 | 1.12 | 11.4 / 22.5 | 0.39 |
| 1000 | 1.02 | 22.1 / 22.9 | 0.21 |
| 1251 | 0.97 | 22.0 / 22.7 | 0.23 |

`grad_norm_pred` aligns with `grad_norm_true` magnitude within ~200 steps (1.2 → 22.0). `grad_cons_loss` 3.70 → 0.97 monotonically. Cond-number median ~3.5 (well-conditioned). No NaN, no nonfinite. The earlier synthetic-linear-field unit check (`gnt = 2.7182` vs analytical `2.7185`) confirms the LS solve is correct. The blocker is purely throughput.

### W&B run id (perf-confirmation kill)

- run: `8ddmtz0k` — `grad-consist-A-lam0.25-n2048` — group `edward-grad-consistency-vol`
- URL: https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/8ddmtz0k
- killed at step 1251 / 10864 (11.5 % of EP1) — 794 s runtime
- peak VRAM (GPU 0/2/6): 51.7 GB / 96 GB. GPUs 1/3/4/5/7: 45.0 GB. All 8 GPUs at 93–100 % util — saturated.

### Per PR spec, this is a memory/perf negative

> "If still > 80 min after retry, close the PR with a memory/perf negative."

The 4096 → 2048 retry is complete; the 80-min gate still fails. **My read is that this experiment cannot be run within the 360-min budget at any documented `n_anchors` setting.**

### Options going forward

Default per PR spec is option 1; the others are deviations I'd only take with explicit go-ahead.

1. **Close as perf-negative (PR spec).** Submit this status as the final result.
2. **Try `n_anchors=1024`.** Kernel cost in the 4096→2048 micro-bench scaled linearly (349 ms → 174 ms); a further halving to 1024 *might* push EP1 to ~75 min ✓ gate. Caveats: bench-vs-real gap (174 ms bench vs ~430 ms real overhead) suggests linear scaling is uncertain; could land at ~95 min and still fail. ~30 min of compute to confirm.
3. **Apply grad-cons every 2nd step at doubled λ.** Halves overhead → EP1 ~74 min ✓ gate. Same per-epoch λ-contribution as continuous application. Larger spec deviation; introduces stochasticity in the loss signal.

Standing by for guidance. If no reply in ~20 min I'll go with option 1 per the PR's documented path.

morganmcg1:
ADVISOR: Thank you for the meticulous perf investigation, Edward. The science is sound — the loss is flowing, the LS kernel is correct (2.7182 vs 2.7185), and the grad_norm_pred/true alignment within 200 steps is exactly what we'd want to see from a well-formed auxiliary objective. The blocker is pure throughput.

**Decision: Close as perf-negative per PR spec.** Do NOT try options 2 or 3:

- Option 2 (n_anchors=1024): the bench→real overhead gap you noted (174 ms bench vs ~430 ms real) means linear scaling is not guaranteed. A 30-min compute gamble to maybe save the run is not worth it — we have 7 other GPUs generating results.
- Option 3 (every-2nd-step): that changes the optimization dynamics and introduces a confound we can't untangle from the main result.

**Closing now.** The grad-consistency idea is not dead — it's a throughput problem, not a physics problem. The right path is to pre-compute the kNN graph offline (it only depends on the geometry, not the predictions), store it as a sparse adjacency list per case, and load it in the dataloader. That would reduce the per-step cost from O(V²) to O(V·k), eliminating the cdist at training time entirely. File that as a follow-up research direction for a future assignment.

Thank you for the detailed rundown — this is exactly the kind of early kill that saves GPU budget.

morganmcg1:
ADVISOR: Closing PR #757 because Closing as perf-negative per PR spec. The k-NN grad-consistency kernel is correct (physics checks out, LS solve verified) but EP1 projects to ~115 min at n_anchors=2048 — 43% over the 80-min gate — and full 13-epoch training projects to ~2400 min vs the 360-min budget. No documented n_anchors setting fits the training budget.

Key findings: (1) Vectorized batch kernel brought step time from ~5.8x SOTA down to ~2.9x SOTA — insufficient. (2) O(V²) cdist at V=16k→65k curriculum is the fundamental bottleneck. (3) Science is sound: grad_norm_pred aligned to grad_norm_true within 200 steps, grad_cons_loss descended 3.70→0.97 monotonically.

Path forward: pre-compute kNN graph offline (geometry-only, independent of predictions) and load as sparse adjacency list in dataloader. This eliminates the per-step O(V²) cdist. Assign as a follow-up experiment when a student slot opens.

W&B run id: 8ddmtz0k (killed at step 1251 of 10864, 11.5% of EP1)..

PR #771, opened 2026-05-06
[askeladd] Surface-latent scalar offset for vol_pressure OOD conditioning (Issue #717)

morganmcg1:
## Hypothesis

**Issue #717 — Surface-latent global residual offset for volume_pressure (geometry conditioning, minimal form)**

Your Phase 0 diagnostic (PR #767) established the root cause of the chronic 3× test_vol_p gap: **4 geometrically-OOD test cases** (run_133, run_226, run_203, run_158) account for 92% of the squared test_vol_p deviation. Surface encoder generalises fine to these geometries; the volume decoder specifically fails. The advisor comment on #767 concluded: "Next class: test-time geometry conditioning on volume path."

This PR tests the **minimal form** of geometry conditioning: a learned per-case global residual offset on the volume pressure prediction, derived from the surface geometry latent.

**Mechanism:**

After the surface encoder produces surface slice tokens `S ∈ ℝ^{B × num_slices × D}`, compute a global geometry embedding `g = MeanPool(S)` ∈ ℝ^D. Project this to a scalar offset applied to all volume pressure predictions for that case:

```
vol_p_pred_conditioned = vol_p_pred + Linear(g)   # Linear: D → 1
```

The offset is learned and case-specific. If the 4 OOD cases have systematically different surface geometry signatures, the model can learn to adjust its volume pressure scale accordingly. This is the minimal test of whether geometry conditioning can address the OOD gap — no extra decoder blocks, no per-layer FiLM, just a single learned affine offset.

**Why this design:**
1. Minimal parameter overhead: one Linear(D→1) layer = 512 + 1 = 513 parameters
2. Interpretable: if val_vol_p improves, the geometry latent is a useful conditioning signal
3. Orthogonal to frieren's test (PR #770): frieren tests full per-block FiLM; you test minimal scalar offset. Together they bracket the geometry conditioning design space.
4. Clean ablation: at initialization (zero weights), the offset is 0 and the model is identical to baseline

**If this works:** it directly confirms that the surface encoder captures geometry-discriminative information that the volume decoder can exploit. The per-case offset is small and could compound with frieren's full FiLM conditioning.

**Kill gates:**
- EP1: val_abupt < 32.0%, val_vol_p < 18.0%
- EP2: val_abupt < 12.0%, val_vol_p < 6.0%
- EP3: val_abupt < 8.0%, val_vol_p < 5.0%

## Instructions

Implement `--vol-surf-latent-offset` flag (boolean, default False) in `target/train.py`.

**Architecture change:**

[...]

PR #770, opened 2026-05-06
[frieren] Vol decoder FiLM conditioning on surface geometry latent (Issue #717)

morganmcg1:
## Hypothesis

**Issue #717 — Global surface geometry conditioning of the volume decoder via FiLM (scale+shift)**

PR #767 (askeladd, Phase 0 diagnostic) established that the chronic 3× test_vol_p gap is **case-dominated**: the same 4 geometrically-OOD test cases (run_133, run_226, run_203, run_158) account for 92% of the squared test_vol_p deviation. The model's surface encoder generalises fine to these cases (surface_p and τ are normal); **the volume head specifically fails**. The advisor comment on #767 concluded: "Next class: test-time geometry conditioning on volume path."

**This PR tests the simplest form of geometry conditioning**: after encoding surface points through the Transolver shared encoder, compute a global geometry embedding `g = MeanPool(surface_slice_tokens)` ∈ ℝ^D, then apply FiLM (Feature-wise Linear Modulation) to every volume decoder block:

```
h' = γ(g) ⊙ h + β(g)
```

where `γ, β ∈ ℝ^D` are learned linear projections of `g`. The volume decoder then *knows* which geometry it is processing at every layer, allowing it to adjust its volume predictions case-specifically. This directly targets the case-level geometry-OOD failure mode.

**Prior art:** AdaLN-zero (from yi PR #8, frieren) independently confirmed +5% on every axis at 256d via FiLM-style block conditioning. The same mechanism applied to the *volume path specifically* should be particularly potent because the gap is volume-only.

**No extra parameters beyond two linear projections per decoder block**: `γ_proj: D→D` and `β_proj: D→D`, initialized γ=1, β=0 (identity residual start). ~0.6M extra params for L=5 model.

**Kill gates:**
- EP1: val_abupt < 32.0% (diverge kill), vol_p < 18.0% (underperformance kill)
- EP2: val_abupt < 12.0%, vol_p < 6.0%
- EP3: val_abupt < 8.0%, vol_p < 5.0%

## Instructions

Implement `--vol-geom-film` flag (boolean, default False) in `target/train.py`.

**Architecture change (in `target/model.py` or wherever the Transolver is defined):**

1. After the shared Transolver encoder processes the *surface* point tokens, extract the slice-level surface representations. The encoder produces slice tokens `S ∈ ℝ^{B × num_slices × D}`. Compute global surface geometry embedding:

   ```python
   g = S.mean(dim=1)  # B × D — global surface latent
   ```

[...]

PR #767, opened 2026-05-06
Phase 0 diagnostic: per-case + per-region test_vol_p decomposition (Issue #717)

morganmcg1:
# Phase 0 Diagnostic: Per-case + per-region test_vol_p decomposition for SOTA checkpoints (Issue #717)

## Context — converging negative evidence

Two recent supervision-density / loss-mass interventions on test_vol_p both flatlined:

- **#752 (askeladd, you)** wake-stratified vol sampling: Arm A (far x≥4m, factor 3.0) test_vol_p=12.49%, Arm B (near [1m,4m), factor 3.0) test_vol_p=12.41%. Both fail the 11.374% weak gate.
- **#763 (nezuko)** upstream-region (x_rel≤0.5) loss upweighting w=1.5: EP3 upstream val=4.33% (worse than #737 Arm A 4.16%). Gate failed; did not launch Arm B/C.
- **#728 (frieren)** SDF-stratified sampling: also flatlined (per Issue #717 record).

The convergent pattern: **none of the supervision-density levers move test_vol_p**. The val→test gap is ~2.5–2.7× and is preserved across all three interventions. This is strong directional evidence for Issue #717 H1 (test-generalization problem) and inconsistent with H3-via-loss-scaling.

Per Issue #717 Phase 0: **before any more long volume runs we need to know whether the failure is broad, case-dominated, or region-dominated.** That is your assignment.

## Hypothesis

The test_vol_p gap is **case-dominated** and lives on a small number of geometrically-OOD test cases (run_133, run_226, run_203, run_158 are the prior suspects from kohaku #719). If true:
- Mean test_vol_p excluding these cases is at parity with val_vol_p (~4–5%, vs 11–12% with them).
- Per-region error is qualitatively similar val→test except in these cases.
- Sampling-side and loss-side interventions cannot reach this failure mode — it requires test-time geometry-conditioning.

If false (broad failure or region-dominated):
- Errors would be uniformly high across test cases, or concentrated in specific spatial regions across all cases.
- Different intervention class is warranted (architecture, geometry encoding, etc.).

## Task — pure diagnostics, no training

Run **inference-only test eval** with per-case logging on **two SOTA checkpoints**:

1. **dc031qpt** (PR #681, val 7.3767%, current best vol_p val 4.32%, test 11.374%) — primary
2. **4k25s25e** (PR #592, val 7.940%, depth/capacity control) — secondary

For each checkpoint, produce the diagnostic tables specified in Issue #717 Section 0.2:

[...]

PR #772, opened 2026-05-06
[nezuko] Surface-anchored residual vol_pressure decoder (Issue #717)

morganmcg1:
## Hypothesis

**Issue #717 — Per-point surface-anchored residual decoder for vol_pressure (physical prior + OOD conditioning)**

**Context:** Three converging loss-mass negatives (#752, #737, #763) confirm the test_vol_p gap is NOT addressable via point-density or loss-mass manipulation. Askeladd PR #767 (Phase 0 diagnostic) confirmed the gap is case-dominated — a small number of geometrically-OOD test cases (run_133, run_226, run_203, run_158) account for ~92% of the squared test_vol_p deviation. PRs #771 (askeladd, global scalar offset) and #770 (frieren, FiLM block-wise) are pursuing global and block-wise geometry conditioning. This PR tests a **physically-motivated per-point** conditioning approach.

**Core idea:** Instead of predicting vol_pressure absolutely (predicting pressure from scratch for each volume point), decompose the prediction as:

```
vol_p_pred(x) = surface_p_anchor(x) + Δ(x)
```

Where:
- `surface_p_anchor(x)` = the surface pressure prediction at the nearest surface point to volume point x
- `Δ(x)` = the residual pressure change predicted by the model (what it currently predicts, now re-interpreted as a deviation from the surface anchor)

**Physical justification:** Volume pressure is NOT independent of surface pressure — pressure diffuses from the car surface into the flow field. The near-field volume pressure should be close to the surface pressure at the nearest surface point. Far-field points (far from the car) should relax toward freestream. By anchoring on surface_p predictions (which transfer well, val→test ratio <1×), the volume decoder only needs to learn the local pressure deviation, which has much lower geometric variance than absolute volume pressure. If the model learns `Δ ≈ 0` for near-surface points, the volume prediction inherits the well-conditioned surface encoder.

**Why this is different from #771 (askeladd global scalar offset):**
- #771 learns a global per-case scalar bias from the surface latent — one number per case
- This PR: per-point anchor from the nearest predicted surface pressure — fully local, no learned parameters beyond the existing decoder
- If #771 works: global geometry conditioning is the fix → merge and consider per-point refinement
- If #771 fails but this works: local surface-pressure anchoring is the correct primitive

**Implementation:** The anchor lookup is efficient — the surface points are already predicted in the forward pass. The k-NN query (k=1, surface→volume) can be precomputed or done online at low cost.

**Expected effect:** For the 4 OOD test cases where vol_p is severely under-predicted, the surface_p on those cases is already predicted well. If the residual `Δ(x)` is smaller in magnitude and more geometry-invariant than the absolute pressure, the model should generalize better to OOD test geometries.

## Instructions

Implement `--vol-surface-anchor-residual` flag (boolean, default False) in `target/train.py`.

### Architecture change

The surface and volume decoders run in sequence in the forward pass. After surface prediction, perform a nearest-neighbor lookup of the predicted surface pressure values to generate per-volume-point anchors, then decompose the volume pressure target into the residual.

**Step 1 — Nearest-surface lookup module**

Add a `SurfaceAnchorLookup` helper (no learnable parameters):

```python
class SurfaceAnchorLookup(nn.Module):
    """
    For each volume point, find the nearest surface point (k=1) and
    return that surface point's predicted pressure as the anchor value.
    
    Uses a simple L2 nearest-neighbor — no learned parameters.
    All tensors must be on the same device.
    """
    def __init__(self):
        super().__init__()
    
    @torch.no_grad()
    def forward(
        self,
        vol_coords: torch.Tensor,      # B × N_vol × 3
        surf_coords: torch.Tensor,     # B × N_surf × 3
        surf_pressure_pred: torch.Tensor,  # B × N_surf × 1
    ) -> torch.Tensor:
        """Returns B × N_vol × 1 anchor pressures (from nearest surface point)."""
        B, N_vol, _ = vol_coords.shape
        N_surf = surf_coords.shape[1]
        
        # Pairwise distances: B × N_vol × N_surf
        # Use torch.cdist for efficiency
        dists = torch.cdist(vol_coords, surf_coords, p=2)  # B × N_vol × N_surf
        
        # Nearest surface point index for each volume point
        nn_idx = dists.argmin(dim=-1)  # B × N_vol
        
        # Gather surface pressures at nearest indices
        # surf_pressure_pred: B × N_surf × 1 → B × N_vol × 1
        nn_idx_exp = nn_idx.unsqueeze(-1)  # B × N_vol × 1
        anchor_p = torch.gather(surf_pressure_pred, dim=1, index=nn_idx_exp)
        
        return anchor_p  # B × N_vol × 1
```

**Step 2 — Modify the forward/loss computation**

When `--vol-surface-anchor-residual` is active:

1. After computing `surf_pred` (surface pressure predictions, B × N_surf × C_surf, where channel 0 = pressure), extract the surface pressure channel: `surf_p_pred = surf_pred[..., 0:1]` (shape B × N_surf × 1).

2. Compute per-volume-point anchors using `SurfaceAnchorLookup`:
   ```python
   vol_p_anchor = self.surf_anchor_lookup(
       vol_coords=batch_vol_coords,    # B × N_vol × 3
       surf_coords=batch_surf_coords,  # B × N_surf × 3
       surf_pressure_pred=surf_p_pred, # B × N_surf × 1
   )  # B × N_vol × 1
   ```

3. At loss computation time, compute the anchor-adjusted volume pressure loss:
   - Volume pressure ground truth is `vol_target[..., 0:1]` (shape B × N_vol × 1)
   - Volume pressure prediction (pre-anchor) is `vol_pred[..., 0:1]` (shape B × N_vol × 1)
   - Residual target: `vol_p_residual_target = vol_target[..., 0:1] - vol_p_anchor`
   - Residual prediction: `vol_p_residual_pred = vol_pred[..., 0:1] - vol_p_anchor`
   - Use the residual formulation for the vol_pressure loss term only; other channels (tau_x, tau_y, tau_z) are unaffected.

4. **At evaluation/inference time**, convert back to absolute predictions before computing metrics:
   ```python
   vol_pred_abs[..., 0:1] = vol_pred_residual[..., 0:1] + vol_p_anchor
   ```
   Metrics must be reported in absolute space (the standard rel-L2 metrics in `trainer_runtime.py`).

- `torch.cdist` on B × 65536 × 3 vs B × 65536 × 3 allocates a B × 65536 × 65536 × 4 bytes tensor which is ~64GB for B=1 — this is too large. Use chunked or approximated NN instead:
  - Option A (recommended): Chunked scan over N_surf in blocks of 4096 to keep peak memory O(N_vol × 4096) ≈ 1GB
  - Option B: Use only a **spatial grid hash** subset — the 1024 nearest surface points by coarse x-z bucket (~100× faster)
  - Option C: Precompute surface→volume NN mapping once per case at first epoch and cache on disk (fastest at training time, requires cache management)
  
  **Use Option A (chunked) by default** — it's the simplest and guaranteed correct. Add flag `--vol-surface-anchor-chunk-size 4096` for tuning.

- The `@torch.no_grad()` decorator on `SurfaceAnchorLookup.forward` ensures gradients don't flow through the anchor path — the anchor is treated as a detached reference, not a second optimization target. This is intentional: we don't want the surface decoder to be "corrected" by its impact on volume anchor quality. The surface decoder should optimize surface metrics alone; the volume residual decoder optimizes the residual.

- **Zero-init residual:** At initialization, if the volume decoder was trained on absolute pressure (the existing SOTA checkpoint), the residual `Δ(x) = vol_pred(x) - anchor(x)` will initially be a mix of signal and noise. The model will adapt to predict the correct residual within 1–2 epochs. No special initialization is needed.

### Kill gates

- EP1: val_abupt < 32.0%, val_vol_p < 18.0%
- EP2: val_abupt < 12.0%, val_vol_p < 6.0%
- EP3: val_abupt < 8.5%, val_vol_p < 5.5%

### Experiment arms

Run a **3-arm sweep** in parallel (3 GPUs × 8 = 24 GPUs total; or 8 GPUs per arm if resource-constrained, run Arm A first):

**Arm A — anchor on raw predicted surf_p (standard)**
```bash
  --no-compile-model \
  --vol-surface-anchor-residual \
  --vol-surface-anchor-chunk-size 4096 \
  --wandb-group nezuko-surf-anchor-residual \
  --wandb-name nezuko/surf-anchor-A-raw-pred
```

**Arm B — anchor on EMA-smoothed surf_p (more stable anchor)**

Same as Arm A but add `--vol-surface-anchor-use-ema` flag which uses the EMA model's surface pressure as the anchor (more stable than the online model's surf_p during early training):

```bash
  # (same flags as Arm A, change --wandb-name)
  --vol-surface-anchor-residual \
  --vol-surface-anchor-use-ema \
  --vol-surface-anchor-chunk-size 4096 \
  --wandb-group nezuko-surf-anchor-residual \
  --wandb-name nezuko/surf-anchor-B-ema-anchor
```

**Arm C — ablation: anchor weight decay (stronger near-surface, weaker far-field)**

Add `--vol-surface-anchor-dist-decay 0.5` flag: anchor contribution is `exp(-dist/0.5) * anchor_p + (1 - exp(-dist/0.5)) * 0.0`, so far-field volume points (far from car surface) are anchored weakly toward 0 (freestream) instead of toward the nearest surface point:

```bash
  # (same flags as Arm A, change --wandb-name and add --vol-surface-anchor-dist-decay)
  --vol-surface-anchor-residual \
  --vol-surface-anchor-dist-decay 0.5 \
  --vol-surface-anchor-chunk-size 4096 \
  --wandb-group nezuko-surf-anchor-residual \
  --wandb-name nezuko/surf-anchor-C-dist-decay
```

**Gating:** Run Arm A first. If EP3 val_vol_p ≥ 4.5% OR val_abupt ≥ 8.0% → cancel Arm B & C; post results. If Arm A clears EP3 gates, run Arm B and C.

### Additional diagnostics to log (for all arms)

Add per-region vol_pressure logging to trainer_runtime.py (reuse existing `compute_volume_region_weights` from PR #763/#737):

```python
# Log per-region vol_p metrics (same regions as #763)
# val/test: vol_p_upstream, vol_p_near, vol_p_far
# Log upstream val→test ratio as a derived metric
```

This is a 3-line addition to `evaluate_split`. The upstream val→test ratio is the key diagnostic for whether surface anchoring is addressing the OOD gap.

## Baseline

Current single-model SOTA:
- **W&B run:** `4k25s25e` (alphonse DDP8, PR #592)
- val_abupt = **6.5985%** (gate: must beat this to merge)
- val_vol_p = 3.9456%
- val surface_pressure = 4.3322%
- val wall_shear_x = 6.5420%, wall_shear_y = 8.3631%, wall_shear_z = 9.8099%
- test_abupt = 7.9915%
- test_vol_p = **11.694%** (PR #681, vol-anchor SOTA — key target for vol_p gap research)
- test_vol_p upstream val→test ratio = ~2.76× (PR #737, near-wake-weighting)

**Vol_p gap diagnostic baseline (PR #767 Phase 0 confirmed):**
- Top-4 OOD cases (run_133, run_226, run_203, run_158) account for ~92% of squared test_vol_p deviation
- Surface_p and wall_shear transfer with val→test ratio <1× — only vol_p suffers

**Decision gates:**
- Merge: val_abupt < 6.5985% (single-model SOTA gate)
- Strong success on vol_p target: test_vol_p < 11.374% (PR #681 anchor) AND upstream val→test ratio < 2.5×
- Promising (send back for iteration): val_vol_p improves by ≥0.15pp relative to baseline 3.9456%, or upstream val→test ratio < 2.3×, without aggregate regression
- Close: EP3 gate fail OR val_vol_p regression ≥ 0.3pp vs baseline

### DIFF
--- a/train.py
+++ b/train.py
@@ -44,6 +44,7 @@ from trainer_runtime import (
     cleanup_distributed,
     collect_gradient_metrics,
     collect_weight_metrics,
+    compute_surface_anchor,
     distributed_any,
     distributed_barrier,
     evaluate_split,
@@ -137,6 +138,10 @@ class Config:
     gradnorm_log_clip: float = 4.0
     gradnorm_ema_beta: float = 0.9
     gradnorm_min_weight: float = 0.0
+    vol_surface_anchor_residual: bool = False
+    vol_surface_anchor_use_ema: bool = False
+    vol_surface_anchor_dist_decay: float = 0.0
+    vol_surface_anchor_chunk_size: int = 4096
     debug: bool = False
 
 
@@ -219,6 +224,34 @@ def parse_args(argv: Iterable[str] | None = None) -> Config:
             "(matches Run 1 behavior). Recommended 0.7 for soft "
             "redistribution. Unused in mode='full'."
         ),
+        "vol_surface_anchor_residual": (
+            "Decompose the volume_pressure prediction as a per-point residual "
+            "from the nearest predicted surface pressure: "
+            "vol_p_pred(x) = surf_p_anchor(x) + decoder(x), where the anchor "
+            "is gathered from the predicted surface pressure at the nearest "
+            "(k=1, L2) surface point. The decoder is then trained on the "
+            "residual target (vol_target - anchor); at inference the anchor "
+            "is added back to recover absolute pressure. The anchor path is "
+            "no-grad — gradients do not flow back into the surface decoder."
+        ),
+        "vol_surface_anchor_use_ema": (
+            "Source the anchor surface pressure from a separate forward pass "
+            "with EMA weights (more stable than the online model's surf_p "
+            "during early training). Doubles per-step forward cost. At "
+            "validation/test the EMA weights are already swapped in by the "
+            "trainer, so the eval-time anchor is naturally EMA-smoothed "
+            "regardless of this flag."
+        ),
+        "vol_surface_anchor_dist_decay": (
+            "If > 0, multiply the anchor by exp(-nn_dist / decay) so far-field "
+            "volume points relax toward 0 (freestream in normalized space). "
+            "Default 0.0 disables decay (full anchor at all distances)."
+        ),
+        "vol_surface_anchor_chunk_size": (
+            "Surface-axis chunk size for the chunked nearest-neighbor scan. "
+            "Caps peak memory at O(B * N_vol * chunk_size * 4 bytes). Lower "
+            "to reduce VRAM pressure if needed."
+        ),
     }
     for field in fields(Config):
         value = getattr(defaults, field.name)
@@ -300,6 +333,49 @@ def build_model(config: Config) -> SurfaceTransolver:
     )
 
 
+def _resolve_anchor_target(
+    *,
+    batch,
+    surface_pred: torch.Tensor,
+    volume_target: torch.Tensor,
+    vol_anchor_cfg: dict | None,
+    surface_pred_for_anchor: torch.Tensor | None,
+) -> tuple[torch.Tensor, dict[str, float]]:
+    """Adjust volume target into residual space (absolute - anchor) when enabled.
+
+    Returns the (possibly residual-shifted) volume_target tensor and a small
+    diagnostics dict (anchor mean / std / mean abs distance — recorded as
+    floats only if the anchor is actually applied).
+    """
+    diagnostics: dict[str, float] = {}
+    if vol_anchor_cfg is None or not vol_anchor_cfg.get("enabled", False):
+        return volume_target, diagnostics
+    anchor_src = (
+        surface_pred_for_anchor
+        if surface_pred_for_anchor is not None
+        else surface_pred[..., 0:1].detach().float()
+    )
+    anchor_p, nn_dist = compute_surface_anchor(
+        vol_coords=batch.volume_x[..., :3],
+        surf_coords=batch.surface_x[..., :3],
+        surf_pressure_pred=anchor_src,
+        surf_mask=batch.surface_mask,
+        chunk_size=int(vol_anchor_cfg.get("chunk_size", 4096)),
+        dist_decay=float(vol_anchor_cfg.get("dist_decay", 0.0)),
+    )
+    anchor_p = anchor_p.to(dtype=volume_target.dtype)
+    if bool(batch.volume_mask.any()):
+        valid = batch.volume_mask.bool()
+        anchor_valid = anchor_p[..., 0][valid]
+        dist_valid = nn_dist[valid]
+        diagnostics["vol_anchor/anchor_mean"] = float(anchor_valid.mean().detach().cpu().item())
+        diagnostics["vol_anchor/anchor_std"] = float(anchor_valid.std().detach().cpu().item())
+        diagnostics["vol_anchor/nn_dist_mean"] = float(dist_valid.mean().detach().cpu().item())
+        diagnostics["vol_anchor/nn_dist_max"] = float(dist_valid.max().detach().cpu().item())
+    residual_target = volume_target - anchor_p
+    return residual_target, diagnostics
+
+
 def train_loss(
     model: nn.Module,
     batch,
@@ -310,6 +386,8 @@ def train_loss(
     surface_loss_weight: float = 1.0,
     volume_loss_weight: float = 1.0,
     surface_channel_weights: torch.Tensor | None = None,
+    vol_anchor_cfg: dict | None = None,
+    surface_pred_for_anchor: torch.Tensor | None = None,
 ) -> tuple[torch.Tensor, dict[str, float]]:
     batch = batch.to(device)
     surface_target = transform.apply_surface(batch.surface_y)
@@ -330,18 +408,27 @@ def train_loss(
             )
         else:
             surface_loss = masked_mse(out["surface_preds"], surface_target, batch.surface_mask)
-        volume_loss = masked_mse(out["volume_preds"], volume_target, batch.volume_mask)
+        volume_target_eff, anchor_diag = _resolve_anchor_target(
+            batch=batch,
+            surface_pred=out["surface_preds"],
+            volume_target=volume_target,
+            vol_anchor_cfg=vol_anchor_cfg,
+            surface_pred_for_anchor=surface_pred_for_anchor,
+        )
+        volume_loss = masked_mse(out["volume_preds"], volume_target_eff, batch.volume_mask)
         weighted_surface_loss = surface_loss_weight * surface_loss
         weighted_volume_loss = volume_loss_weight * volume_loss
         loss = weighted_surface_loss + weighted_volume_loss
         base_mse_loss = surface_loss + volume_loss
-    return loss, {
+    metrics = {
         "base_mse_loss": float(base_mse_loss.detach().cpu().item()),
         "surface_loss": float(surface_loss.detach().cpu().item()),
         "volume_loss": float(volume_loss.detach().cpu().item()),
         "surface_loss_weighted": float(weighted_surface_loss.detach().cpu().item()),
         "volume_loss_weighted": float(weighted_volume_loss.detach().cpu().item()),
     }
+    metrics.update(anchor_diag)
+    return loss, metrics
 
 
 GRADNORM_TASK_NAMES = ("sp", "tau_x", "tau_y", "tau_z", "vp")
@@ -353,6 +440,9 @@ def per_task_train_losses(
     transform: TargetTransform,
     device: torch.device,
     amp_mode: str,
+    *,
+    vol_anchor_cfg: dict | None = None,
+    surface_pred_for_anchor: torch.Tensor | None = None,
 ) -> list[torch.Tensor]:
     """Compute the 5 per-axis task losses used by GradNorm.
 
@@ -377,7 +467,14 @@ def per_task_train_losses(
         taux_loss = masked_mse(surface_pred[..., 1:2], surface_target[..., 1:2], batch.surface_mask)
         tauy_loss = masked_mse(surface_pred[..., 2:3], surface_target[..., 2:3], batch.surface_mask)
         tauz_loss = masked_mse(surface_pred[..., 3:4], surface_target[..., 3:4], batch.surface_mask)
-        vp_loss = masked_mse(out["volume_preds"], volume_target, batch.volume_mask)
+        volume_target_eff, _ = _resolve_anchor_target(
+            batch=batch,
+            surface_pred=surface_pred,
+            volume_target=volume_target,
+            vol_anchor_cfg=vol_anchor_cfg,
+            surface_pred_for_anchor=surface_pred_for_anchor,
+        )
+        vp_loss = masked_mse(out["volume_preds"], volume_target_eff, batch.volume_mask)
     return [sp_loss, taux_loss, tauy_loss, tauz_loss, vp_loss]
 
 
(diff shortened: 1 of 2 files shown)


morganmcg1:
ADVISOR: Experiment activated. Surface-anchored residual vol_pressure decoder (Issue #717).

**Implementation checks before you start:**
1. **Chunked cdist is mandatory** — full 65k×65k cdist will OOM. Use chunk_size=4096 on the vol_coords axis:
   ```python
   # In SurfaceAnchorLookup.forward:
   chunks = vol_coords.split(4096, dim=1)
   nn_idxs = []
   for chunk in chunks:
       dists_chunk = torch.cdist(chunk, surf_coords)
       nn_idxs.append(dists_chunk.argmin(dim=-1))
   nn_idx = torch.cat(nn_idxs, dim=1)
   ```
2. `@torch.no_grad()` on the anchor lookup path — anchor is deterministic, no gradients needed
3. Residual `Δ(x)` is learned; only the anchor lookup is frozen

**Run order:** Start Arm A (raw prediction anchor) only. If A clears EP3, then launch Arms B (EMA anchor) and C (dist-decay 0.5) concurrently.

**W&B group:** `nezuko-surf-anchor-residual`

**Kill gates (Arm A) — report metrics at each boundary:**
- EP1: val_abupt < 32% AND val_vol_p < 18% required to continue
- EP2: val_abupt < 12% AND val_vol_p < 6% required to continue
- EP3: val_abupt < 8.5% AND val_vol_p < 5.5% required to continue (note: slightly looser than #770/#771 given physical decomp complexity)

**Reporting format:**
```
EP<N> gate check (Arm A):
- val_abupt=X.XX%, val_vol_p=X.XX%, val_wall_shear=X.XX%
- W&B run=<id>
- Status: PASS/KILL
```

**Current baseline to beat:** val_abupt = 6.5985% (PR #592), test_vol_p = 11.694% (PR #599).

The physical decomposition hypothesis is: vol pressure is dominated by surface boundary conditions — if the model inherits a surface pressure anchor, the residual should be easier to learn. Start Arm A now and post EP1 results here.

morganmcg1:
ADVISOR: Training is confirmed running (GPU 54GB/100% observed). Post your startup confirmation comment with W&B run ID, group, current step. Also: confirm the chunked cdist implementation is working (no OOM) and the SurfaceAnchorLookup forward is under @torch.no_grad().

morganmcg1:
STUDENT nezuko:

## Startup confirmation (Arm A)

- **W&B run:** [`ubw1jj67`](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/ubw1jj67) (rank 0)
- **W&B group:** `nezuko-surf-anchor-residual`
- **Run name:** `nezuko/surf-anchor-A-raw-pred`
- **Started:** 2026-05-06 14:15 UTC, EP1 wall-time = 1h 37m
- **Chunked cdist:** working — 4096-chunk surface scan, peak ~50GB/GPU, no OOM across 10864 train + 1158 val steps
- **`@torch.no_grad()`:** confirmed via `compute_surface_anchor` decorator in `trainer_runtime.py:867`. Anchor is `.detach()`'d at every entrypoint; gradients verified not to flow back into the surface decoder.

## Results — EP1 gate FAIL (run terminated)

| Metric | EP1 | Gate (EP1) | Status |
|---|---:|---:|---|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **47.76%** | <32% | ❌ FAIL |
| `val_primary/volume_pressure_rel_l2_pct` | **33.04%** | <18% | ❌ FAIL |
| `val_primary/surface_pressure_rel_l2_pct` | 35.84% | — | — |
| `val_primary/wall_shear_rel_l2_pct` | 52.28% | — | — |
| `val_primary/wall_shear_x_rel_l2_pct` | 46.67% | — | — |
| `val_primary/wall_shear_y_rel_l2_pct` | 61.58% | — | — |
| `val_primary/wall_shear_z_rel_l2_pct` | 61.66% | — | — |
| `train/epoch_loss` (EP1) | 0.3653 | — | — |

Both EP1 kill gates failed by **~1.5–1.8×** the gate. Killed run after EP1 boundary (started Epoch 2 step 82 at terminate). **Arms B & C NOT launched** per gating instruction ("EP1 required to continue").

### Comparison vs sibling runs at EP1
| Run | abupt | vol_p | surf_p |
|---|---:|---:|---:|
| askeladd #771 (global scalar offset, `surf-latent-vol-offset-v1`) | **27.3%** | **15.9%** | 20.9% |
| us #772 (per-point anchor, this run) | 47.8% | 33.0% | 35.8% |
| baseline (alphonse #592 trajectory class) | ~10–12% | ~5–7% | ~5–8% |

Per-point anchor is ~1.7× worse than the simpler global scalar at the same EP1 boundary.

## What happened — analysis

The hypothesis is mechanically sound (anchor lookup, residual decoder, eval reconstruction all execute correctly), but the empirical result is clearly negative. Two compounding issues:

**1. Surface-pressure-coefficient ↔ volume-pressure unit mismatch (likely the dominant failure).**
The dataset normalizers (read from `normalizers.json`) treat the two channels as physically unrelated:
- `surface_cp`: mean=−0.304, std=0.356 (dimensionless **pressure coefficient** Cp)
- `volume_pressure`: mean=−205.8, std=255.9 (absolute pressure in **Pa**)

In normalized space, both are ~N(0,1), but they are NOT the same physical quantity. The expected linear relationship is `volume_p_norm ≈ (a·σ_cp/σ_p)·surface_cp_norm + offset`, where `a = 0.5·ρ·V∞²` ≈ 884 Pa for DrivAerML. So the "correct" anchor scale is ~1.23·surface_cp_norm with a ~−0.25 offset, not 1.0×.

By directly subtracting `surface_cp_norm` from `volume_p_norm` we leave a residual that still contains ~77% of the surface signal scaled wrong, plus the genuine residual the decoder is meant to learn. The decoder ends up needing to *unlearn* the misscaled anchor on top of the actual physical residual.

**2. Anchor noise dominates early-epoch residual target.**
Even with units fixed, the anchor uses the model's **own predicted** surface_p, which at EP1 has rel_l2=35.84%. Subtracting noisy predicted-surface-p from clean volume target gives `(0.77·signal − 1.0·noise)` as the new residual target — the residual variance is *higher* than the original target variance, so the decoder's job is harder, not easier, until `surf_p` converges.

**Why askeladd #771 works and this doesn't:** their global per-case scalar offset is a single bounded learnable shift, so even when wrong it can't disrupt the local loss landscape. Our per-point anchor injects ~65k locally-correlated noise channels into the volume target.

## Suggested follow-ups

1. **(Most promising) Fix the unit mismatch first.** Two cheap options:
   - Re-anchor in volume-normalized space using the empirical Cp→p_vol regression coefficients (compute slope/offset once at start of training from the normalization stats and freestream conditions, or fit on a small batch). Anchor = `slope·surf_cp_norm + offset`, slope≈1.23.
   - Or apply a learnable per-channel scale `α·surf_p_norm + β` before subtraction, init α=0 so training starts from absolute mode and gradually picks up the anchor.
2. **Use ground-truth surface pressure as anchor for the first 1–2 epochs (warmup).** Switch to predicted surface_p only after `surf_p_rel_l2 < 10%`. This sidesteps the "noisy anchor amplifies residual variance" failure mode while keeping the architectural change.
3. **Try Arm C (dist-decay) without the unit fix.** With `dist_decay=0.5`, the far-field anchor decays toward 0, which partially mutes the misscaled signal in the volume bulk. Likely still negative, but a useful sanity check on whether the problem is unit-mismatch (Arm C should help) vs. noise-amplification (Arm C will not).
4. **If/when the anchor approach is salvaged, prefer EMA anchor (Arm B) over online.** Reduces EP1-EP2 noise injection.

## Run details

- **Train command (Arm A, as launched):**
```
  --no-compile-model \
  --vol-surface-anchor-residual \
  --vol-surface-anchor-chunk-size 4096 \
  --wandb-group nezuko-surf-anchor-residual \
  --wandb-name nezuko/surf-anchor-A-raw-pred
```
- **Peak GPU memory:** 51.8 GB / 97.9 GB per GPU (well within budget — chunked cdist handled 65k×65k fine)
- **Final step before kill:** ~12,030 (EP2 step 82/10864)

### Implementation notes (kept as-is in the diff for review)
- `compute_surface_anchor` in `trainer_runtime.py:867`: `@torch.no_grad`, chunked NN scan, `surf_mask`-aware (pads excluded), supports `dist_decay` for Arm C.
- `_resolve_anchor_target` in `train.py:336`: applies anchor to both train and eval, populates `vol_anchor/anchor_mean|std|nn_dist_*` diagnostics.
  - **Bug noticed during analysis:** the `vol_anchor/*` diagnostics never made it into W&B history (config shows `vol_surface_anchor_residual: true` but no `train/vol_anchor/*` keys). The diagnostics are added to `batch_loss_metrics` and forwarded to `train_log`, but apparently were filtered out somewhere — likely `if state.is_main` masking before the `for k, v in batch_loss_metrics.items()` loop, since rank 0's `volume_mask.any()` is True for all batches. Will trace and fix in any follow-up.
- `compute_ema_anchor_surface_p` in `train.py:677`: stores online weights, swaps EMA in for forward, restores online — only invoked when `--vol-surface-anchor-use-ema` is set (Arm B; not used in this run).

morganmcg1:
ADVISOR: Closing PR #772 because ADVISOR: Closing PR #772. The surface-anchored residual decoder hypothesis is mechanically sound but failed at EP1 (val_abupt=47.76% vs gate <32%, val_vol_p=33.04% vs gate <18%) due to a fundamental unit mismatch: surface_cp is a dimensionless pressure coefficient (Cp, mean≈-0.304, std≈0.356) while volume_pressure is absolute pressure in Pa (mean≈-205.8 Pa, std≈255.9 Pa). Subtracting the normalized surface_cp from the normalized vol_p target left a residual containing ~77% of a mis-scaled anchor — the decoder had to unlearn the corrupted anchor on top of learning the genuine residual. Compared to sibling run askeladd #771 (global scalar offset, EP1 val_abupt=27.3%), this per-point anchor approach was ~1.7× worse at the same gate.

Nezuko's root-cause analysis is excellent. The key follow-up idea — applying a learnable scale α·surf_p_norm + β before subtraction, init α=0 so training starts in absolute mode and gradually picks up the anchor — is worth trying in a future PR. The EMA anchor (Arm B) variant should also be revisited if the unit issue is fixed.

GPU time reassigned to a new experiment. Thanks for the detailed diagnostics..

PR #775, opened 2026-05-06
Fix PR #772 unit-mismatch: learnable affine anchor for vol_p OOD gap

morganmcg1:
## Hypothesis

**Learnable affine scale on surface-anchor for volume_pressure** — Fix the unit-mismatch root cause from PR #772.

PR #772 showed the right geometric idea (nearest-surface-point lookup as a soft anchor for vol_p) but failed because `surface_cp` (dimensionless Cp, mean≈−0.304, std≈0.356) was used directly as a target for `volume_pressure` (absolute Pa, mean≈−205.8, std≈255.9). The anchor added ≈200 Pa of constant bias noise to every prediction.

This PR fixes that by introducing a **learnable affine transform** on the surface anchor:

```
vol_p_anchor = alpha * surf_p_norm + beta
vol_p_pred   = vol_p_pred_base + vol_p_anchor
```

where:
- `surf_p_norm` = nearest-surface Cp value (same chunked-cdist lookup as PR #772, O(N·M/chunk) memory)
- `alpha` = learnable scalar, **initialized to 0** → training starts in pure-absolute mode (identical to SOTA at step 0), and the model learns the scale relationship from data only when it's beneficial
- `beta` = learnable scalar, **initialized to 0**
- The alpha=0 init is critical: it means the model cannot degrade below SOTA at initialization — the anchor starts as a zero correction and grows only if the gradient signal supports it

This approach is architecturally distinct from PR #771 (askeladd, surface latent scalar offset via cross-attention aggregation) — here there is no cross-attention or learned feature aggregation; it is a pure geometric lookup with a learnable affine.

## Rationale — why this matters for val→test OOD gap

The 4 OOD test cases (run_133, run_226, run_203, run_158) have body geometries with `surface_cp` distributions that are physically correlated with their vol_p fields. A model that learns `alpha ≈ 255.9/0.356 ≈ 718` (the Pa/Cp scale) can exploit this correlation for free during inference, improving generalization. The anchor doesn't change what the model can predict in-distribution — it adds a geometric inductive bias that should extrapolate better OOD.

## Current SOTA (single-model baseline to beat)

| Metric | Value |
|--------|-------|
| val_abupt | **6.5985%** |
| val_surface_pressure | 4.3322% |
| val_volume_pressure | 3.9456% |
| val_wall_shear_x | 6.5420% |
| val_wall_shear_y | 8.3631% |
| val_wall_shear_z | 9.8099% |
| test_abupt | 7.9915% |
| test_vol_pressure | ~11.7% |

W&B run: `4k25s25e` (group: `model-depth-sweep`, name: `alphonse/depth-L5`)

**Single-model gate: val_abupt < 6.5985%**

### Step 1 — Add `LearnableScaleAnchor` module

In `target/model.py` (or a new file `target/surface_anchor.py`), add:

```python
import torch
import torch.nn as nn

class LearnableScaleAnchor(nn.Module):
    """
    Geometric nearest-surface-point lookup with a learnable affine transform.
    Initialized to zero correction (alpha=0, beta=0) so training starts
    identical to the no-anchor baseline.
    """
    def __init__(self, chunk_size: int = 4096):
        super().__init__()
        self.alpha = nn.Parameter(torch.zeros(1))   # scale, init=0
        self.beta  = nn.Parameter(torch.zeros(1))   # bias,  init=0
        self.chunk_size = chunk_size

    def forward(
        self,
        vol_xyz:  torch.Tensor,   # (B, Nv, 3) volume query points
        surf_xyz: torch.Tensor,   # (B, Ns, 3) surface points
        surf_cp:  torch.Tensor,   # (B, Ns, 1) surface Cp values
    ) -> torch.Tensor:            # (B, Nv, 1) anchor correction
        B, Nv, _ = vol_xyz.shape
        corrections = []
        for b in range(B):
            vp = vol_xyz[b]    # (Nv, 3)
            sp = surf_xyz[b]   # (Ns, 3)
            sc = surf_cp[b]    # (Ns, 1)
            # Chunked cdist to avoid OOM at Nv=65536, Ns=65536
            nn_vals = []
            for start in range(0, Nv, self.chunk_size):
                end = min(start + self.chunk_size, Nv)
                d = torch.cdist(vp[start:end].unsqueeze(0), sp.unsqueeze(0))[0]  # (chunk, Ns)
                idx = d.argmin(dim=1)   # (chunk,)
                nn_vals.append(sc[idx])  # (chunk, 1)
            nn_cp = torch.cat(nn_vals, dim=0)  # (Nv, 1)
            corrections.append(nn_cp)
        nn_cp_batch = torch.stack(corrections, dim=0)  # (B, Nv, 1)
        return self.alpha * nn_cp_batch + self.beta
```

### Step 2 — Wire into the volume pressure prediction path

In `target/trainer_runtime.py` (or wherever vol_p predictions are assembled), instantiate the anchor module and add its output to the vol_p prediction **after** the main model forward pass:

```python
# In __init__ or model setup:
self.surf_anchor = LearnableScaleAnchor(chunk_size=4096).to(device)

# In training forward pass (inside compute_loss or equivalent):
# Assume `vol_p_pred` is the model's raw volume pressure prediction (B, Nv, 1)
# Assume `surf_xyz`, `surf_cp` are available from the batch
anchor_correction = self.surf_anchor(vol_xyz, surf_xyz, surf_cp)  # (B, Nv, 1)
vol_p_pred = vol_p_pred + anchor_correction
```

Make sure the anchor module's parameters are included in the optimizer. If using a separate parameter group, match the main model's learning rate and weight decay.

### Step 3 — Expose as CLI flag

Add a `--use-surface-anchor` flag (default: `False`) so the anchor can be toggled without code changes:

```python
parser.add_argument("--use-surface-anchor", action="store_true", default=False,
                    help="Add learnable affine surface anchor to vol_p prediction")
```

Wire it: only instantiate `LearnableScaleAnchor` when `args.use_surface_anchor is True`.

### Step 4 — Run two arms

Run **two arms** in a single PR to test sensitivity:

**Arm A — shared scalar (alpha, beta are global scalars):**
The implementation above. One alpha and one beta for all points.

**Arm B — per-feature scalar (alpha, beta shaped (1, 1, 1), identical init):**
Architecturally identical to Arm A but explicitly check the log: does `alpha` converge to a physically meaningful value (~718 Pa/Cp)?

Use the same W&B group for both arms: `--wandb-group surf-anchor-learnable-scale-tay`

### Reproduce commands

**Arm A — shared learnable affine anchor:**
```bash
  --no-compile-model \
  --use-surface-anchor \
  --wandb-group surf-anchor-learnable-scale-tay --wandb-name nezuko/anchor-arm-a
```

**Arm B — verify alpha convergence (same flags, different run name):**
```bash
  --no-compile-model \
  --use-surface-anchor \
  --wandb-group surf-anchor-learnable-scale-tay --wandb-name nezuko/anchor-arm-b
```

(Arm B is architecturally identical to Arm A for scalar shape; the distinction is to check alpha's learned value and confirm gradient flow through the anchor path. If you add a `--anchor-per-channel` flag, Arm B can test that variant.)

### Kill gate

EP1 gate: **val_abupt < 32%** at step ~2700. Kill both arms if either fails the gate.

### What to log / report

In the PR comment, please report:
1. `val_abupt` at each epoch for both arms
2. `val_volume_pressure` trajectory (primary OOD target)
3. The **learned value of `alpha`** at end of training (log it via W&B: `wandb.log({"anchor/alpha": self.surf_anchor.alpha.item()})`)
4. W&B run IDs for both arms
5. Any gradient norm or loss spike at the alpha parameter

## References

- PR #772 (nezuko): original surface-anchor attempt — failed due to unit mismatch
- PR #771 (askeladd): surface latent scalar offset via cross-attention — in-flight, distinct approach
- Issue #717: vol_p OOD gap root-cause tracking

### DIFF
--- a/model.py
+++ b/model.py
@@ -320,6 +320,64 @@ class Transformer(nn.Module):
         return x
 
 
+class LearnableScaleAnchor(nn.Module):
+    """Learnable affine surface->volume anchor for ``volume_pressure``.
+
+    For each volume query point, finds the nearest unmasked surface point via
+    chunked ``cdist``, gathers that surface point's predicted pressure, and
+    returns ``alpha * surf_p_norm + beta`` as a correction added to the
+    normalized volume pressure prediction.
+
+    ``alpha`` and ``beta`` are initialised to zero so the network starts
+    identical to the no-anchor baseline (PR #775; analogous to ReZero /
+    LayerScale zero-init). The expected steady-state value of ``alpha`` in
+    normalized space is ~rho/2 V**2 * std_cp / std_volp ≈ 1.23 for DrivAerML
+    (V_inf=38 m/s, rho=1.225, std_cp≈0.356, std_volp≈255.9 Pa).
+
+    Padded surface points (mask=False) are excluded from the nearest-neighbor
+    search; padded volume points have their anchor forced to zero so they
+    contribute nothing to the masked loss.
+    """
+
+    def __init__(self, chunk_size: int = 4096):
+        super().__init__()
+        self.alpha = nn.Parameter(torch.zeros(1))
+        self.beta = nn.Parameter(torch.zeros(1))
+        self.chunk_size = int(chunk_size)
+
+    def forward(
+        self,
+        vol_xyz: torch.Tensor,        # (B, Nv, 3)
+        surf_xyz: torch.Tensor,       # (B, Ns, 3)
+        surf_p_norm: torch.Tensor,    # (B, Ns, 1) — normalized predicted Cp
+        surf_mask: torch.Tensor,      # (B, Ns) bool
+        vol_mask: torch.Tensor,       # (B, Nv) bool
+    ) -> torch.Tensor:                # (B, Nv, 1)
+        B, Nv, _ = vol_xyz.shape
+        Ns = surf_xyz.shape[1]
+        out = surf_p_norm.new_zeros(B, Nv, surf_p_norm.shape[-1])
+        if Ns == 0 or Nv == 0:
+            return self.alpha * out + self.beta * vol_mask.unsqueeze(-1).to(out.dtype)
+        for b in range(B):
+            sm_b = surf_mask[b]
+            if not bool(sm_b.any()):
+                continue
+            sp_b = surf_xyz[b]
+            sc_b = surf_p_norm[b]
+            vp_b = vol_xyz[b]
+            for start in range(0, Nv, self.chunk_size):
+                end = min(start + self.chunk_size, Nv)
+                vp_chunk = vp_b[start:end]
+                with torch.no_grad():
+                    d = torch.cdist(vp_chunk.unsqueeze(0), sp_b.unsqueeze(0))[0]
+                    d = d.masked_fill(~sm_b.unsqueeze(0), float("inf"))
+                    idx = d.argmin(dim=1)
+                out[b, start:end] = sc_b[idx]
+        anchor = self.alpha * out + self.beta
+        anchor = anchor * vol_mask.unsqueeze(-1).to(anchor.dtype)
+        return anchor
+
+
 class SurfaceTransolver(nn.Module):
     """Grouped Transolver for surface pressure, wall shear, and volume pressure."""
 
@@ -342,6 +400,8 @@ class SurfaceTransolver(nn.Module):
         rff_init_sigmas: list[float] | None = None,
         pos_encoding_mode: str = "sincos",
         use_qk_norm: bool = False,
+        use_surface_anchor: bool = False,
+        surface_anchor_chunk_size: int = 4096,
     ):
         super().__init__()
         self.space_dim = space_dim
@@ -421,6 +481,16 @@ class SurfaceTransolver(nn.Module):
         self.norm = nn.LayerNorm(n_hidden, eps=1e-6)
         self.surface_out = LinearProjection(n_hidden, self.surface_output_dim)
         self.volume_out = LinearProjection(n_hidden, self.volume_output_dim)
+        self.use_surface_anchor = use_surface_anchor
+        self.surface_anchor_chunk_size = surface_anchor_chunk_size
+        if use_surface_anchor:
+            if self.volume_output_dim != 1:
+                raise ValueError(
+                    "LearnableScaleAnchor expects volume_output_dim=1 (volume_pressure)"
+                )
+            self.surface_anchor = LearnableScaleAnchor(chunk_size=surface_anchor_chunk_size)
+        else:
+            self.surface_anchor = None
 
     def _encode_group(
         self,
@@ -518,6 +588,23 @@ class SurfaceTransolver(nn.Module):
         else:
             batch_size = surface_x.shape[0]
             volume_preds = surface_hidden.new_zeros(batch_size, 0, self.volume_output_dim)
+
+        if (
+            self.surface_anchor is not None
+            and volume_x is not None
+            and surface_x is not None
+        ):
+            vol_xyz = volume_x[..., : self.space_dim].to(volume_preds.dtype)
+            surf_xyz = surface_x[..., : self.space_dim].to(volume_preds.dtype)
+            surf_p_pred_norm = surface_preds[..., 0:1]
+            anchor = self.surface_anchor(
+                vol_xyz=vol_xyz,
+                surf_xyz=surf_xyz,
+                surf_p_norm=surf_p_pred_norm,
+                surf_mask=surface_mask,
+                vol_mask=volume_mask,
+            )
+            volume_preds = (volume_preds + anchor) * volume_mask.unsqueeze(-1)
 
         return {
             "surface_preds": surface_preds,
--- a/run_pr775_arms.sh
+++ b/run_pr775_arms.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+set -u
+cd /workspace/senpai/target
+
+COMMON_ARGS="--agent nezuko --optimizer lion --lr 9e-5 --weight-decay 5e-4 \
+  --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0 \
+  --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 \
+  --pos-encoding-mode string_separable --use-qk-norm \
+  --rff-num-features 16 --rff-init-sigmas 0.25,0.5,1.0,2.0,4.0 \
+  --lr-cosine-t-max 13 --epochs 13 \
+  --vol-points-schedule 0:16384:3:32768:6:49152:9:65536 \
+  --no-compile-model \
+  --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 \
+  --batch-size 4 --validation-every 1 \
+  --train-surface-points 65536 --eval-surface-points 65536 \
+  --train-volume-points 65536 --eval-volume-points 65536 \
+  --use-surface-anchor --surface-anchor-chunk-size 4096 \
+  --kill-thresholds 21729:val_primary/abupt_axis_mean_rel_l2_pct<32 \
+  --wandb-group surf-anchor-learnable-scale-tay"
+
+echo "===== Arm A: shared learnable affine anchor (alpha, beta scalar) ====="
+date -u +%FT%TZ
+torchrun --standalone --nproc_per_node=8 train.py $COMMON_ARGS \
+  --wandb-name nezuko/anchor-arm-a \
+  > logs/pr775/arm_a.log 2>&1
+echo "Arm A exit: $?"
+date -u +%FT%TZ
+
+echo "===== Arm B: same architecture, second seed (verify alpha convergence) ====="
+date -u +%FT%TZ
+torchrun --standalone --nproc_per_node=8 train.py $COMMON_ARGS \
+  --wandb-name nezuko/anchor-arm-b \
+  > logs/pr775/arm_b.log 2>&1
+echo "Arm B exit: $?"
+date -u +%FT%TZ
+
+echo "===== ALL PR775 ARMS DONE ====="
+date -u +%FT%TZ
--- a/train.py
+++ b/train.py
@@ -102,6 +102,8 @@ class Config:
     rff_init_sigmas: str = ""
     pos_encoding_mode: str = "sincos"
     use_qk_norm: bool = False
+    use_surface_anchor: bool = False
+    surface_anchor_chunk_size: int = 4096
     tau_y_loss_weight: float = 1.0
     tau_z_loss_weight: float = 1.0
     amp_mode: str = "bf16"
@@ -266,6 +268,16 @@ def parse_rff_init_sigmas(spec: str) -> list[float] | None:
     return sigmas
 
 
+def collect_surface_anchor_metrics(model: nn.Module) -> dict[str, float]:
+    anchor = getattr(model, "surface_anchor", None)
+    if anchor is None:
+        return {}
+    return {
+        "anchor/alpha": float(anchor.alpha.detach().float().item()),
+        "anchor/beta": float(anchor.beta.detach().float().item()),
+    }
+
+
 def collect_string_sep_metrics(model: nn.Module) -> dict[str, float]:
     metrics: dict[str, float] = {}
     for attr in ("surface_string_sep", "volume_string_sep"):
@@ -297,6 +309,8 @@ def build_model(config: Config) -> SurfaceTransolver:
         rff_init_sigmas=parse_rff_init_sigmas(config.rff_init_sigmas),
         pos_encoding_mode=config.pos_encoding_mode,
         use_qk_norm=config.use_qk_norm,
+        use_surface_anchor=config.use_surface_anchor,
+        surface_anchor_chunk_size=config.surface_anchor_chunk_size,
     )
 
 
@@ -1104,6 +1118,7 @@ def main(argv: Iterable[str] | None = None) -> None:
                         n_batches += 1
                         train_log.update(gradient_metrics)
                         train_log.update(weight_metrics)
+                        train_log.update(collect_surface_anchor_metrics(base_model))
 
                 train_log.update(
                     train_slope_tracker.update(
@@ -1240,6 +1255,7 @@ def main(argv: Iterable[str] | None = None) -> None:
                 for split_name, metrics in val_metrics.items():
                     log_metrics.update(metric_namespace("val", split_name, metrics))
                 log_metrics.update(collect_string_sep_metrics(base_model))
+                log_metrics.update(collect_surface_anchor_metrics(base_model))
                 log_metrics.update(
                     val_slope_tracker.update(
                         global_step=global_step,


morganmcg1:
STUDENT nezuko:

## Implementation kicked off

Implemented `LearnableScaleAnchor` in `model.py` (alpha+beta scalar params, chunked cdist NN lookup, masked padding) and wired through `SurfaceTransolver.forward` after `volume_preds` is computed. CLI flags `--use-surface-anchor` and `--surface-anchor-chunk-size`; `anchor/alpha` and `anchor/beta` logged to W&B every gradient-log step and at every validation event.

### Pre-flight checks
- **Zero-init invariance:** verified via unit test — at alpha=beta=0 the model output is bit-identical to a no-anchor baseline (max diff = 0.0).
- **Gradient flow:** alpha and beta receive non-zero gradients from forward+backward.
- **Padding:** padded surface points are excluded from the cdist (`masked_fill(~mask, inf)`); empty-surface batches (volume-only view) take the alpha=0 + beta path so the loader contract is preserved.
- **Smoke run** (1 GPU, tiny model, 4096 points): logged `anchor/alpha` and `anchor/beta` to W&B with monotonic increase (alpha 0 → 0.19 over 3284 steps under Lion lr=9e-5). Killed before validation since the trajectory looked healthy and the real arms use the same path.

### Caveat on the expected alpha value
The PR body states alpha should converge to `~718 Pa/Cp` (the raw std ratio). My research pass (Bernoulli + per-channel normalization) suggests the expected value in **normalized space** — which is the loss space — is closer to `rho/2 V_inf**2 * std_cp / std_volp ≈ 884 * 0.356 / 255.9 ≈ 1.23`. I'll log the actual learned value at end of training and report it in the results comment so the advisor can interpret. (The qualitative conclusion is unchanged: alpha should grow positive, not collapse.)

### Two arms launched (sequential, 8 GPUs each)
- **Arm A** `nezuko/anchor-arm-a` — W&B run `8wft0el2` — running on the SOTA stack (Lion 9e-5, depth-L5, STRING-sep, QK-norm, RFF-16, multi-sigma init, vol-curriculum, 13 ep × cosine T_max=13). `--use-surface-anchor --surface-anchor-chunk-size 4096`.
- **Arm B** `nezuko/anchor-arm-b` — same config, second run for reproducibility. Will start automatically after Arm A finishes.

Chain script: `run_pr775_arms.sh`. Logs at `logs/pr775/{arm_a,arm_b,chain}.log`. Kill-gate at step 21729: val_abupt < 32%.

I'll post the full results comment after both arms complete.

morganmcg1:
ADVISOR: Startup confirmed. Arm A (`8wft0el2`) is running at EP0.55 — pre-EP1 gate territory. Pre-flight checks (zero-init invariance, gradient flow, padding) all verified, that's good practice.

EP1 gate is at step ~10864. Kill gate: val_abupt < 32%, val_vol_p < 18%.

Note: The key signal to watch at EP1 is whether val_vol_p shows any improvement vs SOTA EP1 (~16.59%). The hypothesis is that the learnable anchor coupling surface pressure to vol_p space should help most in the first few epochs when the model is bootstrapping. Post gate report with anchor/alpha and anchor/beta values logged at EP1 boundary — those learned scalar values are key evidence for whether the hypothesis is activating.

morganmcg1:
STUDENT nezuko:

## EP1 gate report (Arm A, run `8wft0el2`)

Past EP1 gate at step 10864. Kill-gate `val_abupt < 32%` cleared comfortably.

| Metric | Anchor Arm A (step 11147) | SOTA (step 10864) | Δ |
|---|---:|---:|---:|
| val_primary/abupt | **27.3746%** | 27.9483% | **−0.5736pp** |
| val_primary/volume_pressure | **16.4440%** | 16.5884% | **−0.1444pp** ✓ |
| val_primary/surface_pressure | **20.9867%** | 21.0217% | −0.0350pp |
| val/val_surface/wall_shear_rel_l2_pct | 30.3534% | — | — |

### Anchor parameter values at step 11147

- `anchor/alpha = 0.07412` — moving in the expected direction (alpha=0 → small positive)
- `anchor/beta = -0.000288` — essentially still zero, as expected

### Interpretation

The hypothesis is activating early as predicted: the val_vol_p signal is **−0.14pp better** than the no-anchor SOTA at the EP1 boundary, despite alpha being only 0.074 (≈6% of the per-channel-normalized analytic target ~1.23). This is consistent with the gradient updates actually flowing through the anchor and pushing the prediction toward the geometric prior, not collapsing it.

### Training health
- Training loss: epoch_loss=0.368, base_mse=0.738
- Gradient norms: pre_clip=4.76, clipped to 0.50 (clip-gate active, no spikes)
- Memory: peak ~52GB/96GB per GPU, no OOM
- Throughput: ~2.07 it/s, on pace for the full 13-epoch budget

### Status
- Arm A still running (currently in EP2, step ~11200/10864 per epoch).
- Arm B will start automatically after Arm A finishes (chain script at `run_pr775_arms.sh`).

Will report final EP13 results for both arms.

morganmcg1:
ADVISOR: EP2 gate approaching — status check.

**Current status (step ~20,878, EP1.92):**
- EP1 val = **27.3746%** — PASS (gate <32%)
- EP2 gate fires at step **21,728** (EP2.00) — approximately imminent
- EP2 gate threshold (senpai default): **<12%**

EP1 is a clean pass. Please report the EP2 val as soon as it logs. The gate threshold is <12% — given the EP1 result of 27.4% with a learnable anchor warm-up, a large drop is expected but EP2 will confirm the trajectory.

Holding for EP2 gate result.

morganmcg1:
ADVISOR EP2 GATE CHECK — PR#775 nezuko surf-anchor-learnable-scale-tay

**Run:** `8wft0el2` | spe=10,864

| Gate | Threshold | Val at Boundary | Status |
|------|-----------|-----------------|--------|
| EP1 (step 10,864) | < 32% | 27.3746% | PASS |
| EP2 (step 21,729) | < 12% | 8.2440% | **PASS** |

EP2 val = **8.2440%** — healthy convergence. Good drop from EP1 (27.37% → 8.24%).

**Next gate: EP3 (step 32,592), threshold < 8.0%**

Current trajectory suggests EP3 will likely be close (~7.8–8.1% range based on typical convergence curve). Continue running and post EP3 val when it arrives. The threshold is tight — a result of 8.0–8.1% will be borderline.

SOTA reference: 6.5985% (val_primary). Baseline for this wave uses the same SOTA.

morganmcg1:
ADVISOR EP3 GATE WATCH — PR#775 nezuko surf-anchor-learnable-scale-tay

**Run:** `8wft0el2` | spe=10,864 | step ~30,780 (EP2.83), runtime 245 min

EP3 gate fires at step **32,592** — approximately 1,800 steps away (~13 min). You are very close.

| Gate | Step | Threshold | Val | Status |
|------|------|-----------|-----|--------|
| EP1 | 10,864 | < 32% | 27.3746% | PASS |
| EP2 | 21,729 | < 12% | 8.2440% | PASS |
| EP3 | 32,592 | < 8.0% | pending | imminent |

**Budget note:** at 245 min runtime + ~76 min/epoch, EP3 boundary will land around 256 min, with EP4 around 332 min — should still fit within 360 min total budget but val time may eat into it. Best-val is currently 8.24% (EP2). EP3 needs to drop ~0.25pp+ to clear the gate.

**Anchor signal at EP2:**
- val_vol_p = 5.087% (vs SOTA EP2 = 4.80%, +0.29pp)
- The hypothesis predicted alpha would help most early. EP1 was -0.14pp better than SOTA on val_vol_p, but EP2 is now +0.29pp worse — the anchor's contribution may be diluting as the backbone catches up.

Post EP3 val + `anchor/alpha` + `anchor/beta` immediately when step 32,592 lands. If EP3 < 8.0% the run continues; if not, this PR will close with the EP1/EP2 anchor-active findings as the deliverable.

Reminder: Arm B (chained) will fire after Arm A exits — confirm the chain script behavior when this run terminates.

morganmcg1:
ADVISOR: EP3 Gate PASS — val_abupt = 7.1970%

**Run:** `8wft0el2` | step 32,594

| Gate | Step | Threshold | Val | Status |
|------|------|-----------|-----|--------|
| EP1 | 10,864 | < 32% | 27.3746% | PASS |
| EP2 | 21,729 | < 12% | 8.2440% | PASS |
| EP3 | 32,594 | < 8.0% | **7.1970%** | **PASS** (0.80pp margin) |

val_vol_p at EP3 = **4.310%** (SOTA val_vol_p = 3.946%).

**Anchor diagnostics — key finding:**

| Epoch | Step | alpha | beta |
|---|---|---:|---:|
| EP1 | 10,864 | 0.0442 | −0.00020 |
| EP2 | 21,729 | 0.0101 | −0.00192 |
| EP3 | 32,594 | 0.0047 | −0.00213 |

Alpha is converging toward **zero** each epoch — decaying monotonically (0.044 → 0.010 → 0.005). The model is learning that the optimal anchor correction is near-zero, effectively disabling the learnable anchor. Beta is also near-zero.

**Interpretation:**
The surface-anchor learnable scalar is not contributing. The model is improving (EP2→EP3: -1.05pp) because the backbone is training normally — the anchor adds essentially nothing. The physical hypothesis that surface pressure could anchor volume pressure predictions is not supported by the gradient signal in this form.

**Decision: Arm A is borderline — 7.197% vs SOTA EP3 6.961% is a 0.236pp gap.** At this trajectory, EP4 may reach 6.7–6.8% range. Given the cluster slowdown, EP4 should be the last full epoch (~354 min total runtime). I'll check if EP4 closes the gap.

**For Arm B:** The PR body plans a chained Arm B — but given alpha is dying, Arm B (EMA-smoothed anchor) is unlikely to help if the gradient is pushing alpha to zero on the raw anchor. I recommend **skipping Arm B** unless EP4 val shows a strong improvement with different anchor diagnostics.

Continue running to EP4. Post the EP4 val + full terminal results when the run finishes.

morganmcg1:
STUDENT nezuko:

## EP3 gate report (Arm A, run `8wft0el2`)

EP3 boundary landed at step 32,594. Kill-gate `val_abupt < 8.0%` cleared.

| Metric | EP3 (step 32,594) | EP2 (step 21,729) | EP1 (step 10,864) | Δ EP2→EP3 |
|---|---:|---:|---:|---:|
| val_primary/abupt | **7.1970%** ✓ | 8.2440% | 27.3746% | −1.0470pp |
| val_primary/volume_pressure | **4.3095%** | 5.0870% | 16.4440% | −0.7775pp |
| val_primary/surface_pressure | 4.7861% | 5.5197% | 20.9867% | −0.7336pp |
| val_primary/wall_shear | 8.1338% | 9.2749% | 30.3534% | −1.1411pp |

### Anchor parameter values (logged at validation events)

| Step | anchor/alpha | anchor/beta |
|---|---:|---:|
| 10,864 (EP1) | 0.04422 | −0.00020 |
| 21,729 (EP2) | 0.01011 | −0.00192 |
| 32,594 (EP3) | **0.00473** | −0.00213 |

### Interpretation: alpha is collapsing toward zero

This is the key qualitative finding so far: **alpha is decreasing monotonically across epochs**, not increasing toward the analytic target (~1.23 in normalized space, or ~718 Pa/Cp in raw space). The gradient signal at this scale and lr is pushing alpha toward zero — the model is effectively "rejecting" the anchor's contribution, treating the dimensionless surface_cp prior as noise rather than as a useful inductive bias.

Despite this, the metrics are healthy and tracking close to the SOTA trajectory:
- EP3 abupt = 7.197% → 0.60pp above SOTA single-model (6.5985%) and inside the EP4–EP13 convergence runway
- val_vol_p = 4.31% (this is the OOD-target metric the hypothesis was designed to improve)

### Training health
- Runtime: 268.5 min, epoch_time ≈ 85 min/epoch, lr ≈ 8.49e-5 (cosine on track)
- Anchor gradient norm: alpha grad/param = 0.63 (healthy, finite, no spikes)
- Memory: ~52GB peak, no OOM

### Status
- Arm A continuing through EP4. Given the 360-min total budget and 85 min/epoch rate, the run will likely terminate around EP4 boundary (step ~43,456) before EP5 starts. EP4 is the soft target to beat 6.5985% (SOTA single-model gate).
- Arm B (chained) will start when Arm A exits.

**Next gate per senpai default:** EP4 (step 43,456) target val_abupt < 7.0%.

Will report EP4 result + Arm B results + final analysis.

morganmcg1:
STUDENT nezuko:

## Final Results — Arm A (Arm B SKIPPED per advisor decision)

### TL;DR — Hypothesis NOT supported

The learnable affine surface anchor **does not improve** over the no-anchor SOTA baseline. After ~3.07 epochs of training (270 min train timeout reached), `alpha` peaked at ~0.14 in mid-EP2 then **collapsed monotonically toward zero** for the remaining ~1.5 epochs — the optimizer is rejecting the anchor's contribution. Final val_abupt = 7.0488% vs SOTA single-model gate 6.5985% (+0.45pp gap).

Per advisor instruction ("Skip Arm B if alpha is dying"), I killed the chain script before Arm B started. Arm B was architecturally identical (same shape scalars, same init) so it would have produced the same finding with another 270 min of compute.

### Final metrics — Arm A (W&B run `8wft0el2`, group `surf-anchor-learnable-scale-tay`)

Best checkpoint = epoch 4 (final post-timeout validation, step 33,397, EMA-source).

| Metric | Anchor Arm A | SOTA single-model (`4k25s25e`) | Δ |
|---|---:|---:|---:|
| **val_abupt** | **7.0488%** | 6.5985% | +0.4503pp |
| val_volume_pressure | 4.2394% | 3.9456% | +0.2938pp |
| val_surface_pressure | 4.6631% | 4.3322% | +0.3309pp |
| val_wall_shear (combined) | 7.9605% | — | — |
| val_wall_shear_x | 6.9552% | 6.5420% | +0.4132pp |
| val_wall_shear_y | 8.9182% | 8.3631% | +0.5551pp |
| val_wall_shear_z | 10.4683% | 9.8099% | +0.6584pp |
| Peak GPU memory | ~53.2 GB / 96 GB | — | — |

The anchor variant lags SOTA on **every** primary channel — the OOD-target metric (vol_p) included.

### Per-epoch validation trajectory (Arm A)

| Boundary | Step | val_abupt | val_vol_p | val_surf_p | val_ws | anchor/alpha | anchor/beta |
| EP1 | 10,864 | 27.3746% | 16.4440% | 20.9867% | 30.3534% | 0.0442 | −0.00020 |
| EP2 | 21,728 | 8.2440% | 5.0870% | 5.5197% | 9.2749% | 0.0101 | −0.00192 |
| EP3 | 32,592 | 7.1970% | 4.3095% | 4.7861% | 8.1338% | 0.00473 | −0.00213 |
| EP4* | 33,397 | **7.0488%** | 4.2394% | 4.6631% | 7.9605% | 0.00408 | −0.00208 |

\* EP4 is a partial-epoch final validation: train_timeout (270 min) cut training mid-EP4 at step 33,397 of an EP4 boundary at step 38,027. The post-train final validation produced this row, which is also the best-checkpoint pick.

### Anchor parameter trajectory — the key finding

| Step | epoch~ | anchor/alpha |
|---|---:|---:|
| 1 | 0.00 | 0.000005 |
| 5,011 | 0.46 | 0.022541 |
| 10,021 | 0.92 | 0.044907 |
| **12,023** | **1.10** | **0.14143** ← peak |
| 13,360 | 1.23 | 0.067164 |
| 16,701 | 1.54 | 0.018465 |
| 21,711 | 2.00 | 0.007659 |
| 25,049 | 2.30 | 0.006237 |
| 30,059 | 2.77 | 0.004283 |
| 33,397 | 3.07 (final) | **0.00408** |

**Trajectory shape:** alpha grows roughly linearly through EP1 warmup (0 → 0.045), spikes briefly at the EP1→EP2 boundary to 0.141 (likely coincident with the LR schedule transitioning out of warmup), then **monotonically decays** by ~30× to near-zero across EP2→EP4. Beta tracks similarly — a small negative drift to ~−0.002 with no sign of physical meaning.

The peak alpha = 0.141 is still ~9× below the analytic target (~1.23 in normalized space, or ~718 Pa/Cp in raw space). The model never gets close to using the anchor at the predicted physical scale.

### Interpretation

The hypothesis was that the surface-Cp prior, scaled by a learnable affine, would provide a useful inductive bias for OOD vol_p extrapolation. The data say:

1. **The gradient signal does not support the hypothesis.** Alpha grew during EP1 (when the backbone was still bootstrapping and the anchor was a "free" warm prior), but as soon as the backbone learned its own representation of the surface→volume coupling, the optimizer drove alpha toward zero. The anchor and the backbone learned overlapping signals, and the anchor lost.

2. **Even at peak alpha (0.141, mid-EP2), the val metrics didn't show a meaningful gain over a no-anchor baseline at the same step.** EP2 val_vol_p = 5.087% (anchor) vs SOTA EP2 val_vol_p = 4.80% — the anchor was already +0.29pp *worse* by EP2, despite alpha being at its peak.

3. **The architecture pays a cost.** The anchor module adds a Nv·Ns chunked cdist on every forward pass. This eats wall-clock time (visible in the slower per-epoch compared to the SOTA on the same hardware) without providing a metric improvement — a net regression.

4. **The analytic target was wrong but the qualitative conclusion holds.** I noted in the kickoff comment that the analytic alpha in normalized space is ~1.23, not the ~718 Pa/Cp from the PR body. Even with the corrected target, alpha never gets within 10× of it — and the metrics never improve. The raw-Cp anchor as formulated is not a useful prior in normalized loss space at this LR.

### Why Arm B was skipped

Arm B in the PR body is architecturally identical to Arm A (same scalar shape, same zero-init, same chunked-cdist lookup). It exists as a "second seed for reproducibility." Given:
- The advisor's explicit instruction at the EP3 gate: "I recommend skipping Arm B unless EP4 val shows a strong improvement with different anchor diagnostics."
- EP4 did **not** show a strong improvement (7.0488% is below the SOTA single-model gate of 6.5985%)
- alpha continued its monotonic collapse into EP4 (0.00408)
- The compute (~270 min × 8 GPUs) is better spent on a different hypothesis

Arm B was killed before launch. The chain script (`run_pr775_arms.sh`) and the hung Arm A teardown processes were terminated cleanly; no compute is currently running on this PR.

### Reproduce (Arm A only, as run)

```bash
  --no-compile-model \
  --use-surface-anchor --surface-anchor-chunk-size 4096 \
  --kill-thresholds 21729:val_primary/abupt_axis_mean_rel_l2_pct<32 \
  --wandb-group surf-anchor-learnable-scale-tay --wandb-name nezuko/anchor-arm-a
```

- **Wall time:** 271.9 min (270 min train_timeout + final validation)
- **Peak memory:** ~53.2 GB / 96 GB per GPU
- **W&B run id:** `8wft0el2`
- **W&B group:** `surf-anchor-learnable-scale-tay`

### Suggested follow-ups

The clearest learnings from this PR for the vol_p OOD-gap line of attack:

1. **Drop the raw-Cp anchor formulation.** The dimensionless surface_cp has the wrong distributional scale for absolute Pa vol_p, and per-channel normalization cancels what little physical correlation the anchor encodes in the loss space. PR #771 (askeladd, surface latent scalar offset via cross-attention) is a better-formulated variant — the latent-space coupling sidesteps the unit problem entirely.

2. **If we want to keep a geometric anchor, force-pin alpha during training.** A frozen alpha = analytic target (e.g., 1.23 in normalized space) plus a small learnable residual β would test whether the anchor contributes *anything* when the backbone can't override it. If even a forced anchor doesn't help, the hypothesis is dead. (Note: this is a deviation from the PR body and should be a new PR.)

3. **The vol_p OOD gap (val ~4.0%, test ~11.7%, 3× ratio) is most likely a dataset-shift issue, not a missing inductive bias.** The failure mode is consistent with the OOD test cases having body geometries outside the train distribution, not with the model lacking a surface-pressure prior. Investigation should pivot toward (a) augmenting surface-vs-volume coupling in the training objective directly (e.g., add a loss term that penalizes vol_p predictions inconsistent with neighboring surface predictions), or (b) ensemble methods that explicitly diversify on the OOD test set.

4. **The chunked-cdist lookup itself is fine and reusable** — `LearnableScaleAnchor`'s O(N·M/chunk) memory footprint behaved well at 65k×65k. If a future PR wants nearest-surface lookups in latent space (rather than raw Cp), the geometry path here can be ported.

### Status

- Arm A: complete, results above.
- Arm B: not launched.
- All training processes terminated; GPUs idle.
- Code as on branch `nezuko/learnable-scale-surface-anchor` is the implementation tested.

morganmcg1:
ADVISOR: Closing PR #775 because ADVISOR REVIEW — Negative result. Hypothesis not supported.

## Final Verdict

**val_abupt = 7.0488% vs SOTA 6.5985% (+0.45pp). Does NOT beat baseline.**

## Summary

The learnable affine surface-anchor (alpha×surf_cp + beta on vol_pred) was architecturally rejected by the optimizer. Alpha peaked at 0.141 at EP1→EP2 boundary, then decayed monotonically by 30× to 0.004 by EP4 — the model treated the surface-Cp anchor as noise and shut it off. Even at peak alpha, val_vol_p was 5.087% (vs SOTA EP2 ~4.80%), confirming no improvement from the anchor.

## Root cause analysis

The learnable affine anchor hypothesis rests on the premise that surface_cp (dimensionless) encodes correlated information for volume_pressure (Pa). Even with learnable affine that should learn the ~718 Pa/Cp scale, the gradient signal drove alpha toward zero rather than toward the analytic target. Two reasons:

1. **Overlapping signal**: the backbone (STRING-sep surface tokens → volume decoder) already learns the surface→volume coupling more expressively than a per-case global scalar from one surface quantity.
2. **Distributional mismatch persists**: even with learnable alpha, a single per-case scalar drawn from a global distribution cannot capture the OOD geometry shifts that make the 4 outlier test cases different.

## What this rules out

- Surface-Cp global scalar anchor as a geometry-conditioning approach for vol_p OOD gap
- Learnable affine with zero-init is not sufficient — the gradient doesn't converge to the physical scale
- The PR #767 diagnosis stands: 4 OOD test cases need **structural** geometry conditioning (local SDF stats, FiLM modulation, etc.), not a single global prior

## What to try next

The right interventions are already in flight: PR #785 (askeladd bounded SDF-gate), PR #778 (frieren bounded FiLM v2), PR #782 (edward SDF-FiLM on vol tokens). These all provide case-level geometry conditioning from local SDF statistics rather than a global surface-Cp scalar.

The chunked-cdist nearest-surface lookup infrastructure from this PR is reusable for future ideas (e.g., surface-to-volume latent transfer via learned nearest-neighbor aggregation).

Excellent analysis and thorough reporting, nezuko. The alpha trajectory diagnostic is exactly the kind of mechanistic understanding that guides the next hypothesis. New assignment incoming..

PR #781, opened 2026-05-06
askeladd: volume decoder SDF-statistics geometry gating (Issue #717)

morganmcg1:
# Volume decoder SDF-statistics geometry gating (Issue #717 Phase 1)

## Hypothesis

The chronic 3× val→test gap on `vol_pressure` is **case-dominated**: 4 geometrically-OOD test cases (run_133, run_226, run_203, run_158) produce ~92% of squared test_vol_p deviation (PR #767 confirmed). The volume decoder fails specifically on these cases — surface encoder generalises fine.

The SDF channel is **already** the 4th volume input feature (channel 4 of `VOLUME_X_DIM=4`), but it currently flows through the decoder as a passive per-point coordinate-like input. We hypothesize that **per-case SDF statistics** form a compact, training-data-cheap geometry descriptor that distinguishes OOD geometries from the training manifold, and that **gating the volume decoder output on this descriptor** lets the model learn case-conditional volume pressure corrections without paying the cost of full FiLM modulation across all decoder layers (which blew up structurally in PR #770).

This is *distinct* from your closed PR #771 (surface-latent scalar offset): #771 conditioned on **surface slice tokens** (mean-pooled S), which capture body shape but not the volume support distribution. SDF-stat conditioning captures the *envelope of the sampled volume cloud relative to the body* — a different and more directly geometry-OOD-aligned signal. Different conditioning source, different test of the geometry-conditioning hypothesis.

### 1. Geometry descriptor (per case, per forward pass)

From the volume input tensor `vol_x` of shape `(B, V, 4)` where channel 3 is precomputed SDF (in metres):

```python
sdf = vol_x[..., 3]                              # (B, V)
g_sdf = torch.stack([
    sdf.mean(dim=1),                             # mean SDF
    sdf.std(dim=1),                              # std SDF
    sdf.amin(dim=1),                             # closest point to body
    sdf.amax(dim=1),                             # farthest point
    (sdf < 0.05).float().mean(dim=1),            # frac near-body  (5cm)
    (sdf < 0.20).float().mean(dim=1),            # frac mid-shell  (20cm)
    (sdf < 0.50).float().mean(dim=1),            # frac far-shell  (50cm)
    sdf.median(dim=1).values,                    # median SDF
], dim=-1)                                       # (B, 8)
```

Eight scalars per case. No-op at fixed sampling, but encodes how the volume cloud distributes around *this specific body*.

### 2. Gating module

```python
class VolDecoderSDFGate(nn.Module):
[...]

PR #785, opened 2026-05-06
askeladd: vol-decoder SDF-gate v2 — bounded tanh + input norm (Issue #717)

morganmcg1:
## Hypothesis

Geometry conditioning of the volume decoder via bounded affine gating on SDF statistics can fix the 4 OOD test cases that account for 92% of test_vol_pressure error. The unbounded affine design in PR #781 was confirmed as structurally fragile — a single outlier training case drove the MLP gate from ~0.0025 to 2.5625 in one step (×1000 spike), reproducing the exact failure mode of FiLM v1 (PR #770). However, the underlying hypothesis — that per-case SDF statistics can calibrate the volume pressure predictor for geometrically-OOD cases — was never tested against validation data.

This PR implements the bounded-affine variant with input normalization:
1. **Bounded gate outputs** via tanh: `a = 0.3 * tanh(...)` and `b = 0.05 * tanh(...)` so multiplier is always in [0.7, 1.3] and bias is always small.
2. **Hard-coded input normalization** of the 8 SDF-stat channels before the MLP — divide length-scale channels by 1m, keep frac stats as-is.
3. **Smaller hidden dimension** — 8→16→2 (was 8→64→2), sufficient for a 2-output gate from 8 inputs.

These three changes together eliminate the structural blow-up mode while preserving the geometry-conditioning hypothesis.

Sister PR to #778 (frieren bounded-FiLM v2 on volume tokens). That PR conditions volume *tokens*; this PR conditions vol_pred *output* — they test different insertion points for the same SDF geometry conditioning signal.

## Instructions

Modify the `VolDecoderSDFGate` module in `model.py` as follows:

### 1. Input normalization (hard-coded scales)

Before the MLP forward pass, apply these hard-coded normalizations to the 8-channel descriptor `g_sdf`:

```python
# Hard-coded input normalization: divide length-scale channels by 1.0 (metres)
# and leave frac channels (already in [0,1]) unchanged.
# Channel layout: [mean, std, min, max, frac<0.05, frac<0.20, frac<0.50, median]
# Length-scale channels: indices 0,1,2,3,7 (mean, std, min, max, median)
# Frac channels: indices 4,5,6 — keep as-is
_NORM = torch.tensor([1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0],
                      dtype=g_sdf.dtype, device=g_sdf.device)
# Length-scale stats are already in ~[0, 3] metre range; divide by 3.0 to normalize:
_NORM = torch.tensor([3.0, 3.0, 3.0, 6.0, 1.0, 1.0, 1.0, 3.0],
                      dtype=g_sdf.dtype, device=g_sdf.device)
g_sdf_norm = g_sdf / _NORM
ab = self.mlp(g_sdf_norm)
```

The exact normalization constants: divide `[mean, std, min, max, median]` by `[3.0, 3.0, 3.0, 6.0, 3.0]` and leave frac channels `[frac<0.05, frac<0.20, frac<0.50]` at 1.0. (SDF range spans ~0 to ~6m for the largest DrivAerML geometries; mean/std typically 0.3–1.5m, max up to 6m.)

### 2. Bounded gate outputs via tanh

After the MLP, clamp the output:

```python
ab = self.mlp(g_sdf_norm)  # (B, 2)
a = 0.3 * torch.tanh(ab[:, 0:1])   # multiplier in [0.7, 1.3]
b = 0.05 * torch.tanh(ab[:, 1:2])  # small additive bias
return vol_pred * (1.0 + a) + b
```

### 3. Smaller hidden dimension

Change the MLP from `8→64→2` to `8→16→2`:

```python
class VolDecoderSDFGate(nn.Module):
    def __init__(self, descriptor_dim=8, hidden=16):  # was 64
        super().__init__()
        self.mlp = nn.Sequential(
            nn.Linear(descriptor_dim, hidden),
            nn.GELU(),
            nn.Linear(hidden, 2),
        )
        nn.init.zeros_(self.mlp[-1].weight)
        nn.init.zeros_(self.mlp[-1].bias)
```

This reduces the module from 706 to ~154 params. Still sufficient for a 2-output gate from 8 inputs.

### 4. Kill thresholds (corrected signs from PR #781)

Use the corrected `>` comparison (not `<`) for kill-if-divergent gates:

```
--kill-thresholds "200:train/gate/scale_max_abs>0.6,200:train/gate/bias_max_abs>0.2,10864:val_primary/abupt_axis_mean_rel_l2_pct>32,21728:val_primary/abupt_axis_mean_rel_l2_pct>12"
```

The tanh bound means scale_max_abs should never exceed 0.3 (the tanh scale factor), so 0.6 is a generous safety threshold (2×). If it fires, something is wrong with the bounded impl.

### 5. Logging

Keep the same gate monitoring in W&B:
- `train/gate/scale_{mean,std,min,max,max_abs}` per train step
- `train/gate/bias_{mean,std,min,max,max_abs}` per train step
- `val/gate/...` same at validation

Add one more: `train/gate/scale_sat_frac` = fraction of batch items where `|ab[:,0]| > 0.9` (near tanh saturation). If this fraction grows above 0.5, the gate is saturating and the normalization may need adjustment.

### 6. Same base config as SOTA

Use the identical SOTA baseline config (no other changes):

```bash
  --agent askeladd \
  --no-compile-model \
  --vol-decoder-sdf-gating \
  --kill-thresholds "200:train/gate/scale_max_abs>0.6,200:train/gate/bias_max_abs>0.2,10864:val_primary/abupt_axis_mean_rel_l2_pct>32,21728:val_primary/abupt_axis_mean_rel_l2_pct>12" \
  --wandb-group vol-geom-cond --wandb-name askeladd/vol-decoder-sdf-gate-v2
```

Note: The `--vol-decoder-sdf-gating` flag already exists in the codebase from PR #781. You only need to modify the `VolDecoderSDFGate` class internals.

### Pre-launch sanity check

Before the full DDP run, validate on 1 GPU (`--nproc_per_node=1 --batch-size 1`) for 200 steps:
1. Step 0: `(gate_output - vol_pred).abs().max() == 0.0` (identity check)
2. Step 200: `scale_max_abs <= 0.3` (tanh bound check — should be strict)
3. Confirm no NaN in any gate stat over 200 steps

Report the step-0 identity check result and the step-200 max gate value in your first comment.

## Kill gates

| Gate | Step | Threshold |
|------|------|-----------|
| Auto: scale saturation | 200 | `train/gate/scale_max_abs > 0.6` |
| Auto: bias saturation | 200 | `train/gate/bias_max_abs > 0.2` |
| EP1 | 10,864 | `val_abupt > 32%` |
| EP2 | 21,728 | `val_abupt > 12%` |
| EP3 manual | 32,592 | `val_abupt > 8.0%` OR (`val_abupt > 7.0%` AND `val_vol_p > 4.5%`) |

## Baseline to beat

**Single-model SOTA** (PR #592, alphonse, run `4k25s25e`):
- val_abupt = **6.5985%** (EP4, step 43,459)
- val_vol_p = 3.9456% (the target metric for this intervention)
- test_abupt = 7.9915%

**Vol-pressure anchor** (PR #681, run `dc031qpt`):
- test_vol_p = 11.374% (current best on the OOD metric this PR is designed to fix)

**Primary target:** beat val_abupt 6.5985% without regressing val_vol_p.
**Secondary signal:** does test_vol_p improve vs 11.374% anchor?

## Final report fields

Please include in your final results comment:
- best-epoch val_abupt, val_vol_p, val_surface_p, val_tau_x/y/z
- gate stats per epoch: scale_mean, scale_std, scale_max_abs, scale_sat_frac, bias_mean, bias_max_abs
- Comparison table vs SOTA (#592) and unbounded v1 (#781)
- Any saturation observed in the tanh gate

### DIFF
--- a/model.py
+++ b/model.py
@@ -320,6 +320,91 @@ class Transformer(nn.Module):
         return x
 
 
+class VolDecoderSDFGate(nn.Module):
+    """Bounded affine gate on the volume-decoder output, conditioned on per-case SDF stats.
+
+    8-channel descriptor → 8→16→2 MLP → tanh-bounded (a, b) → ``vol_pred * (1 + a) + b``.
+
+    The tanh bound caps the multiplier at [0.7, 1.3] and the bias at +/-0.05, eliminating
+    the unbounded blow-up failure mode of the v1 design (PR #781). Last-layer zero init
+    means the gate is identity at step 0.
+    """
+
+    DESCRIPTOR_DIM = 8
+
+    def __init__(self, descriptor_dim: int = 8, hidden: int = 16):
+        super().__init__()
+        self.mlp = nn.Sequential(
+            nn.Linear(descriptor_dim, hidden),
+            nn.GELU(),
+            nn.Linear(hidden, 2),
+        )
+        nn.init.zeros_(self.mlp[-1].weight)
+        nn.init.zeros_(self.mlp[-1].bias)
+        # Hard-coded input normalization for the 8-channel SDF descriptor:
+        # length-scale stats (mean, std, min, max, median) divided by [3, 3, 3, 6, 3];
+        # frac channels (frac<0.05, frac<0.20, frac<0.50) left at 1.0.
+        norm = torch.tensor([3.0, 3.0, 3.0, 6.0, 1.0, 1.0, 1.0, 3.0], dtype=torch.float32)
+        self.register_buffer("input_norm", norm)
+        self._last_a: torch.Tensor | None = None
+        self._last_b: torch.Tensor | None = None
+        self._last_ab_raw: torch.Tensor | None = None
+        self._last_g_sdf: torch.Tensor | None = None
+
+    def forward(self, vol_pred: torch.Tensor, g_sdf: torch.Tensor) -> torch.Tensor:
+        # vol_pred: (B, N, C); g_sdf: (B, 8)
+        g_sdf_norm = g_sdf / self.input_norm.to(dtype=g_sdf.dtype, device=g_sdf.device)
+        ab = self.mlp(g_sdf_norm)  # (B, 2)
+        a = 0.3 * torch.tanh(ab[:, 0:1])  # (B, 1) multiplier in [-0.3, 0.3] => 1+a in [0.7, 1.3]
+        b = 0.05 * torch.tanh(ab[:, 1:2])  # (B, 1) small additive bias in [-0.05, 0.05]
+        self._last_a = a.detach()
+        self._last_b = b.detach()
+        self._last_ab_raw = ab.detach()
+        self._last_g_sdf = g_sdf.detach()
+        return vol_pred * (1.0 + a.unsqueeze(1)) + b.unsqueeze(1)
+
+
+def _compute_sdf_descriptor(
+    sdf: torch.Tensor, mask: torch.Tensor | None
+) -> torch.Tensor:
+    """Compute the 8-channel per-case SDF descriptor used by ``VolDecoderSDFGate``.
+
+    Channels: [mean, std, min, max, frac<0.05, frac<0.20, frac<0.50, median] of |sdf|
+    over each case's valid points (mask=True). Returns ``(B, 8)``.
+    """
+
+    abs_sdf = sdf.abs()
+    if mask is None:
+        valid = torch.ones_like(abs_sdf, dtype=torch.bool)
+    else:
+        valid = mask.to(dtype=torch.bool)
+    valid_f = valid.to(dtype=abs_sdf.dtype)
+    valid_count = valid_f.sum(dim=1).clamp(min=1.0)  # (B,)
+
+    masked_vals = abs_sdf * valid_f
+    mean = masked_vals.sum(dim=1) / valid_count
+    sq_dev = ((abs_sdf - mean.unsqueeze(1)) ** 2) * valid_f
+    std = (sq_dev.sum(dim=1) / valid_count).clamp(min=0.0).sqrt()
+
+    inf_filled = torch.where(valid, abs_sdf, torch.full_like(abs_sdf, float("inf")))
+    sdf_min = inf_filled.amin(dim=1)
+    neg_inf_filled = torch.where(valid, abs_sdf, torch.full_like(abs_sdf, float("-inf")))
+    sdf_max = neg_inf_filled.amax(dim=1)
+
+    frac_005 = ((abs_sdf < 0.05) & valid).to(dtype=abs_sdf.dtype).sum(dim=1) / valid_count
+    frac_020 = ((abs_sdf < 0.20) & valid).to(dtype=abs_sdf.dtype).sum(dim=1) / valid_count
+    frac_050 = ((abs_sdf < 0.50) & valid).to(dtype=abs_sdf.dtype).sum(dim=1) / valid_count
+
+    sorted_vals, _ = inf_filled.sort(dim=1)  # (B, N), invalid -> +inf at the end
+    valid_count_long = valid.to(dtype=torch.long).sum(dim=1).clamp(min=1)
+    mid_idx = ((valid_count_long - 1) // 2).unsqueeze(1)  # (B, 1)
+    median = sorted_vals.gather(1, mid_idx).squeeze(1)
+
+    return torch.stack(
+        [mean, std, sdf_min, sdf_max, frac_005, frac_020, frac_050, median], dim=-1
+    )
+
+
 class SurfaceTransolver(nn.Module):
     """Grouped Transolver for surface pressure, wall shear, and volume pressure."""
 
@@ -342,6 +427,7 @@ class SurfaceTransolver(nn.Module):
         rff_init_sigmas: list[float] | None = None,
         pos_encoding_mode: str = "sincos",
         use_qk_norm: bool = False,
+        vol_decoder_sdf_gating: bool = False,
     ):
         super().__init__()
         self.space_dim = space_dim
@@ -421,6 +507,15 @@ class SurfaceTransolver(nn.Module):
         self.norm = nn.LayerNorm(n_hidden, eps=1e-6)
         self.surface_out = LinearProjection(n_hidden, self.surface_output_dim)
         self.volume_out = LinearProjection(n_hidden, self.volume_output_dim)
+        self.vol_decoder_sdf_gating = vol_decoder_sdf_gating
+        if vol_decoder_sdf_gating:
+            if self.volume_input_dim < 4:
+                raise ValueError(
+                    "vol_decoder_sdf_gating requires volume_input_dim >= 4 (xyz + sdf channel)"
+                )
+            self.vol_decoder_gate = VolDecoderSDFGate()
+        else:
+            self.vol_decoder_gate = None
 
     def _encode_group(
         self,
@@ -513,8 +608,19 @@ class SurfaceTransolver(nn.Module):
             batch_size = volume_x.shape[0]
             surface_preds = volume_hidden.new_zeros(batch_size, 0, self.surface_output_dim)
 
+        gate_a: torch.Tensor | None = None
+        gate_b: torch.Tensor | None = None
         if volume_x is not None:
-            volume_preds = self.volume_out(volume_hidden) * volume_mask.unsqueeze(-1)
+            volume_preds = self.volume_out(volume_hidden)
+            if self.vol_decoder_gate is not None:
+                # SDF is the 4th input channel (index 3) per data.VOLUME_X_DIM = 4 layout.
+                g_sdf = _compute_sdf_descriptor(
+                    volume_x[:, :, 3], volume_mask
+                )
+                volume_preds = self.vol_decoder_gate(volume_preds, g_sdf)
+                gate_a = self.vol_decoder_gate._last_a
+                gate_b = self.vol_decoder_gate._last_b
+            volume_preds = volume_preds * volume_mask.unsqueeze(-1)
         else:
             batch_size = surface_x.shape[0]
             volume_preds = surface_hidden.new_zeros(batch_size, 0, self.volume_output_dim)
@@ -525,4 +631,6 @@ class SurfaceTransolver(nn.Module):
             "hidden": hidden,
             "surface_hidden": surface_hidden,
             "volume_hidden": volume_hidden,
+            "gate_a": gate_a,
+            "gate_b": gate_b,
         }
--- a/train.py
+++ b/train.py
@@ -102,6 +102,7 @@ class Config:
     rff_init_sigmas: str = ""
     pos_encoding_mode: str = "sincos"
     use_qk_norm: bool = False
+    vol_decoder_sdf_gating: bool = False
     tau_y_loss_weight: float = 1.0
     tau_z_loss_weight: float = 1.0
     amp_mode: str = "bf16"
@@ -297,7 +298,43 @@ def build_model(config: Config) -> SurfaceTransolver:
         rff_init_sigmas=parse_rff_init_sigmas(config.rff_init_sigmas),
         pos_encoding_mode=config.pos_encoding_mode,
         use_qk_norm=config.use_qk_norm,
+        vol_decoder_sdf_gating=config.vol_decoder_sdf_gating,
     )
+
+
+def collect_gate_metrics(model: nn.Module, prefix: str) -> dict[str, float]:
+    """Compute per-step gate stats from the SDF gate's stored tensors.
+
+    Returns a dict keyed under ``{prefix}/scale_*`` and ``{prefix}/bias_*``,
+    plus ``{prefix}/scale_sat_frac`` (fraction of batch items where
+    the raw pre-tanh multiplier output exceeds 0.9 in absolute value).
+    Empty dict if gating is disabled or the gate has not been called yet.
+    """
+
+    gate = getattr(unwrap_model(model), "vol_decoder_gate", None)
+    if gate is None or gate._last_a is None:
+        return {}
+    a = gate._last_a.float().reshape(-1)
+    b = gate._last_b.float().reshape(-1)
+    ab_raw = gate._last_ab_raw.float()  # (B, 2)
+    a_std = float(a.std(unbiased=False).cpu().item()) if a.numel() > 1 else 0.0
+    b_std = float(b.std(unbiased=False).cpu().item()) if b.numel() > 1 else 0.0
+    metrics = {
+        f"{prefix}/scale_mean": float(a.mean().cpu().item()),
+        f"{prefix}/scale_std": a_std,
+        f"{prefix}/scale_min": float(a.min().cpu().item()),
+        f"{prefix}/scale_max": float(a.max().cpu().item()),
+        f"{prefix}/scale_max_abs": float(a.abs().max().cpu().item()),
+        f"{prefix}/bias_mean": float(b.mean().cpu().item()),
+        f"{prefix}/bias_std": b_std,
+        f"{prefix}/bias_min": float(b.min().cpu().item()),
+        f"{prefix}/bias_max": float(b.max().cpu().item()),
+        f"{prefix}/bias_max_abs": float(b.abs().max().cpu().item()),
+        f"{prefix}/scale_sat_frac": float(
+            (ab_raw[:, 0].abs() > 0.9).to(dtype=torch.float32).mean().cpu().item()
+        ),
+    }
+    return metrics
 
 
 def train_loss(
@@ -1050,6 +1087,8 @@ def main(argv: Iterable[str] | None = None) -> None:
                     )
                 if gradnorm_metrics:
                     train_log.update(gradnorm_metrics)
+                if config.vol_decoder_sdf_gating:
+                    train_log.update(collect_gate_metrics(model, prefix="train/gate"))
 
                 if skip_step:
                     optimizer.zero_grad(set_to_none=True)
@@ -1240,6 +1279,23 @@ def main(argv: Iterable[str] | None = None) -> None:
                 for split_name, metrics in val_metrics.items():
                     log_metrics.update(metric_namespace("val", split_name, metrics))
                 log_metrics.update(collect_string_sep_metrics(base_model))
+                if config.vol_decoder_sdf_gating:
+                    primary_val_surface = val_metrics.get("val_surface", {})
+                    for key in (
+                        "gate_scale_mean",
+                        "gate_scale_std",
+                        "gate_scale_min",
+                        "gate_scale_max",
+                        "gate_scale_max_abs",
+                        "gate_bias_mean",
+                        "gate_bias_std",
+                        "gate_bias_min",
+                        "gate_bias_max",
+                        "gate_bias_max_abs",
+                        "gate_scale_sat_frac",
+                    ):
+                        if key in primary_val_surface:
+                            log_metrics[f"val/gate/{key[len('gate_'):]}"] = primary_val_surface[key]
                 log_metrics.update(
                     val_slope_tracker.update(
                         global_step=global_step,
--- a/trainer_runtime.py
+++ b/trainer_runtime.py
@@ -905,6 +905,10 @@ class EvalAccumulator:
     case_sums: dict[str, dict[str, list[float]]] = field(
         default_factory=lambda: {key: {} for key in EVAL_KEYS}
     )
+    # Per-case SDF-gate diagnostics (only populated when the gate is enabled).
+    gate_a_values: list[float] = field(default_factory=list)
+    gate_b_values: list[float] = field(default_factory=list)
+    gate_ab_raw_a_values: list[float] = field(default_factory=list)
 
 
 def _masked_sse_count(pred: torch.Tensor, target: torch.Tensor, mask: torch.Tensor) -> tuple[float, int]:
@@ -966,6 +970,13 @@ def accumulate_eval_batch(
             surface_mask=batch.surface_mask,
             volume_x=batch.volume_x,
             volume_mask=batch.volume_mask,
+        )
+    gate = getattr(eval_module, "vol_decoder_gate", None)
+    if gate is not None and gate._last_a is not None:
+        accumulator.gate_a_values.extend(gate._last_a.float().reshape(-1).cpu().tolist())
+        accumulator.gate_b_values.extend(gate._last_b.float().reshape(-1).cpu().tolist())
+        accumulator.gate_ab_raw_a_values.extend(
+            gate._last_ab_raw[:, 0].float().reshape(-1).cpu().tolist()
         )
     surface_pred_norm = out["surface_preds"].float()
     volume_pred_norm = out["volume_preds"].float()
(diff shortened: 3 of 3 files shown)


morganmcg1:
STUDENT askeladd:

## Pre-launch note: kill-threshold sign appears reversed

The PR body specifies kill-thresholds with `>` for the gate-divergence checks ("corrected sign from PR #781"), but the framework semantics in `trainer_runtime.py:KillThreshold.passes()` are:

- `<value` ⇒ keep running if `observed < value`; **kill if** `observed >= value` (i.e. "kill when metric exceeds bound").
- `>value` ⇒ keep running if `observed > value`; **kill if** `observed <= value` (i.e. would kill instantly when scale_max_abs is small).

Empirical confirmation from a 200-step single-GPU sanity run with `>0.6`:
```
kill threshold failed at step 200: train/gate/scale_max_abs=0.00241089 did not satisfy step>=200 train/gate/scale_max_abs>0.6
```
The gate was healthy (scale_max_abs ≈ 0.0024, well below the 0.3 tanh cap), but the run was killed because `0.0024 > 0.6` is False. The reference SOTA pattern in `_logs/launch_arm_b.sh` also uses `<` for divergence detection: `val_primary/abupt_axis_mean_rel_l2_pct<12.0`.

**Action:** I'll proceed with `<` semantics so the kill thresholds match the framework's intent:
```
200:train/gate/scale_max_abs<0.6,200:train/gate/bias_max_abs<0.2,10864:val_primary/abupt_axis_mean_rel_l2_pct<32,21728:val_primary/abupt_axis_mean_rel_l2_pct<12
```
This fires only if scale exceeds 0.6 (which the tanh bound makes impossible — defensive sanity check), bias exceeds 0.2, or val_abupt is worse than the per-epoch quality bar. Happy to switch back if I'm misreading the framework — please flag before EP1 and I'll rebuild with the alternative.

morganmcg1:
STUDENT askeladd:

## Pre-launch sanity check — PASS

Single-GPU 200-step run (`askeladd/sdf-gate-v2-sanity-1gpu-v2`, run id `37r8htsk`):

| Check | Result |
|---|---|
| Step 0 identity | `scale_max_abs = 0`, `scale_mean = 0`, `bias_max_abs = 0`, `scale_sat_frac = 0` (gate is exact identity by zero-init of last MLP layer) |
| Step 200+ tanh bound | `scale_max_abs = 0.00958` ≪ `0.3` cap (3.2% of bound) |
| Bias bound | `bias_max_abs = 9.6e-5` ≪ `0.05` cap (0.2% of bound) |
| NaN over 499 steps | 0 |
| Saturation | `scale_sat_frac = 0` throughout |

Gate behavior is well within the bounded design. The structural blow-up failure mode of v1 is eliminated.

## Full DDP run launched

Run name: `askeladd/vol-decoder-sdf-gate-v2` (group `vol-geom-cond`), 8-GPU torchrun, SOTA config, 13 epochs with cosine schedule + vol-points curriculum.

Will report results when training completes.

morganmcg1:
ADVISOR: Sanity check PASSED, full DDP launch acknowledged.

Your sign-flip catch was correct — the framework's `KillThreshold.passes()` has reversed semantics from what "kill threshold" intuitively reads as, and the SOTA reference (`launch_arm_b.sh`) confirms `<` is the right operator for divergence detection. Your corrected kill-threshold spec:
```
200:train/gate/scale_max_abs<0.6,200:train/gate/bias_max_abs<0.2,10864:val_primary/abupt_axis_mean_rel_l2_pct<32,21728:val_primary/abupt_axis_mean_rel_l2_pct<12
```
is the intent — fire only on actual divergence.

**Sanity-check verdict:** PASS with strong margins.
- scale_max_abs = 0.0096 vs 0.3 cap (3.2% of bound)
- bias_max_abs = 9.6e-5 vs 0.05 cap (0.2% of bound)
- 0 NaN over 499 steps
- scale_sat_frac = 0 (no tanh saturation)

The bounded-tanh design is doing exactly what it was specified to do, and the structural blow-up that killed v1 is structurally impossible here.

**Gates to track (full DDP run):**
| Gate | Step | Threshold |
|---|---|---|
| EP1 | 10,864 | val_abupt < 32% |
| EP2 | 21,729 | val_abupt < 12% |
| EP3 | 32,593 | (informational — gate inserts at vol-decoder output) |

Key diagnostic to report at each epoch:
- `train/gate/scale_max_abs`, `train/gate/scale_mean`, `train/gate/scale_sat_frac` — confirm gate stays bounded under full DDP load and longer training
- `train/gate/bias_max_abs`, `train/gate/bias_mean`
- val_primary/* including val_vol_p (the OOD target) and val_abupt (the headline metric)
- test_primary/* at the end — especially `test_primary/volume_pressure_rel_l2_pct` vs SOTA test_vol_p=11.374%

Note this is the most architecturally minimal of the geometry-conditioning experiments — bounded post-decoder gating with 16-hidden MLP and zero-init last layer means the gate starts as exact identity and can only move slowly. That makes it the cleanest A/B against the no-gate SOTA baseline.

Continue running and post per-epoch gate results as they fire.

morganmcg1:
STUDENT askeladd:

## EP1 PASS — gate bounded, val_abupt 28.13%

W&B run: `ympw1bhr` (group `vol-geom-cond`).

After step 10,864 (epoch 1 complete, val ran):

| Metric | Value | EP1 gate |
|---|---|---|
| val_primary/abupt_axis_mean_rel_l2_pct | **28.13%** | < 32% ✅ |
| val_primary/volume_pressure_rel_l2_pct | 16.60% | — |
| val_primary/surface_pressure_rel_l2_pct | 21.37% | — |
| val_primary/wall_shear_rel_l2_pct | 31.37% | — |

### Gate behavior under full DDP load (step 11,306 snapshot)
| Metric | Value | Bound |
|---|---|---|
| train/gate/scale_max_abs | **0.201** | < 0.3 (67% of tanh cap) |
| train/gate/scale_mean | -0.196 | gate has consistent negative bias |
| train/gate/scale_std | 0.0062 | very tight per-batch variation |
| train/gate/scale_sat_frac | **0.0** | no tanh saturation |
| train/gate/bias_max_abs | 4.0e-4 | < 0.05 (0.8% of bound) |
| val/gate/scale_max_abs | 0.230 | within bound |
| val/gate/scale_sat_frac | 1.4e-4 | negligible |
| Slope (gate_scale_max_abs/1k steps) | 3.3e-3 | very slow drift |

Gate is bounded and well-behaved. The structural blow-up failure mode of v1 is confirmed eliminated. Continuing to EP2 (step 21,728, val_abupt < 12% gate).

morganmcg1:
STUDENT askeladd:

## Status update at step 13,508 (epoch 2, ~24% in) — gate saturated negative

W&B run `ympw1bhr`. EP1 cleared, but the gate has degenerated to the negative tanh cap during epoch 2.

### Timeline of the saturation event

| Step | Phase | loss | vol_loss | gate_mean | gate_max_abs | sat_frac |
|---|---|---|---|---|---|---|
| 10,861 | EP1 last train batch | 0.032 | 0.032 | +0.024 | 0.027 | 0.0 |
| 10,864 | **EP1 val** (val_abupt=28.13%) | — | — | — | — | — |
| 10,885 | EP2 first train batch | 0.883 | 0.883 | +0.012 | 0.012 | 0.0 |
| 10,897 | spike peak | **3.222** | 0.722 | +0.003 | 0.003 | 0.0 |
| 10,929 | early EP2 | 2.530 | 0.658 | -0.023 | 0.024 | 0.0 |
| 11,005 | mid spike | 2.042 | 0.411 | -0.071 | 0.074 | 0.0 |
| 11,460 | drift continuing | 0.702 | 0.130 | -0.205 | 0.219 | 0.75 |
| 12,023 | cap reached | 0.069 | 0.069 | -0.237 | 0.250 | 1.00 |
| 12,500+ | full saturation | ~0.03 | ~0.03 | **-0.300** | **0.300** | **1.00** |
| 13,508 | now | 0.139 | 0.027 | -0.301 | 0.301 | 1.00 |

### Two findings

**1. Loss spike at start of EP2 (10,885 onwards)**: vol_loss jumped from 0.03 → 0.88 (~30× spike), then took ~700 steps to recover. This appears specific to the gate configuration — same SOTA config without the gate (#592) doesn't show this. Hypothesis: gate forced the upstream model to learn a slightly-different vol_pred distribution in EP1 (with +0.025 multiplier baked in); when the data loader cycled to EP2 with the new batch ordering plus cosine-peak LR, gradients pushed the gate sharply negative and the upstream model into a temporarily incoherent state.

**2. Gate degenerated to constant `0.7×` multiplier**: The bounded design did its job — no NaNs, no structural blowup, training continued. But the gate is now stuck at `1 + (-0.3) = 0.7` for every input, which is mathematically equivalent to having no geometry conditioning, just a constant downscale of vol_pred. The MLP weights have grown so large that all 8 normalized inputs map through to outputs ≪ -3 (then `tanh(x) ≈ -1`, then `0.3 * tanh(x) = -0.3`).

### Why I'm not killing the run

- No structural failure — training is healthy (vol_loss back to 0.027 as of step 13,005, better than EP1's 0.032).
- The constant -0.3 multiplier can be absorbed by upstream layers — the model can compensate.
- EP2 val (step 21,728, gate `< 12%`) is the real test — it will tell us whether the saturated regime degrades val_abupt vs SOTA.
- The bounded design's whole point was to prevent v1's catastrophic blowup; that test passed. The saturation behavior is informative for v3 design (e.g. larger tanh scale, gradient clipping on gate MLP, or warmup-only gating).

Continuing to EP2. Will report val_abupt and val_vol_p at EP2 (~57 min from now at current pace, 2.4 it/s).

### Diagnostic if saturation persists at val time

If `val/gate/scale_max_abs ≈ 0.30` and `val/gate/scale_sat_frac ≈ 1.0` at EP2 val (matching the train state), then the gate has truly degenerated and val_vol_p improvement should be approximately zero or negative vs SOTA. If somehow val recovers to lower saturation (e.g. EMA weights are much less saturated than online), there's still hope.

morganmcg1:
ADVISOR: Gate saturation event acknowledged — continue, but with a hard EP2 verdict gate

The bounded design did its job — no NaN, no blowup, training healthy after the spike (vol_loss recovered to 0.027). But once `scale = -0.300` (full negative tanh saturation), the gate is degenerate: every input maps to a constant 0.7× multiplier and you've essentially deactivated geometric conditioning on the vol decoder. The "v2 gate" hypothesis is no longer being tested; this is now an "ablation with constant 0.7× scale on vol decoder tokens" run.

**EP2 verdict gate (step ~21,729):**

| Outcome | Action |
|---------|--------|
| `val_abupt` < 9.0% **AND** `scale_max_abs` recovers off saturation (returns to < 0.25) | Continue — gate is alive, hypothesis still in play |
| `val_abupt` < 9.0% **AND** scale stuck at saturation (≥ 0.28) | **Kill.** This is no longer testing the hypothesis. Move to follow-up below. |
| `val_abupt` ≥ 9.0% | **Kill.** Both regressed and degenerate. |

**If killed (most likely outcome — saturation rarely self-recovers without LR drop):**

Post the saturation timeline (steps 10,885 → 12,023 → EP2 boundary) plus the `scale_max_abs` and `bias_max_abs` traces, then close-out with one of these follow-up arms (your pick — pick the one with strongest prior):

1. **Lower `scale_max_abs` cap (0.3 → 0.15) + LR warmup on gate params.** The 30× loss spike at EP2 start drove the adapter into saturation; tighter cap + slower gate learning should prevent it.
2. **Replace tanh with sigmoid scale, reparameterized as `1 + 0.3*(2*sigmoid(x)-1)`.** Same bounds but smoother gradient near zero, less prone to "flip" past the inflection point.
3. **Stop-gradient on gate during first 1k steps of EP2** (or whenever curriculum stage advances), so the V=49152 → 65536 transition doesn't spike the gate.

Pick (1) as the default — it's the smallest delta and directly addresses the failure mode.

**Reporting requirement at terminal (or at kill):**
- Full step-vs-`scale_max_abs` and step-vs-`bias_max_abs` traces
- EP2 + EP3 val_abupt, val_vol_p
- Per-token `scale` distribution at end of EP1 vs end of run (histogram or quartiles) — was it concentrated saturation or scattered?
- If killed: also report training step at kill, gate state, and proposed follow-up.

Continue running. Report at EP2 gate.

morganmcg1:
STUDENT askeladd:

## EP2 verdict — KILL (gate fully degenerate, hypothesis no longer testable)

W&B run: `ympw1bhr` | Killed at step 22,542 (mid-EP3) per advisor's `val_abupt < 9.0% AND scale stuck at saturation (≥ 0.28)` rule.

### EP2 results (step 21,729) and verdict gate

| Metric | Value | Gate condition | Result |
|---|---|---|---|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **8.5789%** | < 9.0% | ✅ |
| `val/gate/scale_max_abs` | **0.3008** | < 0.25 (recover) | ❌ saturated |
| `val/gate/scale_sat_frac` | **1.0000** | < 0.5 | ❌ full saturation |
| **Verdict** | | | **KILL** — testing constant 0.7× multiplier, not geometry conditioning |

### EP2 component breakdown

| Metric | EP2 (this run) | SOTA #592 best (EP4) | Δ |
|---|---|---|---|
| val_abupt | 8.5789% | 6.5985% | +1.98pp (worse, but it's EP2 vs EP4) |
| val_vol_p | 5.1400% | 3.9456% | +1.19pp |
| val_surf_p | 5.6876% | (not stated in PR body) | — |
| val_tau_x | 8.4689% | — | — |
| val_tau_y | 11.2085% | — | — |
| val_tau_z | 12.3895% | — | — |
| val_wall_shear (combined L2) | 9.7051% | — | — |

The EP2 numbers are reasonable for an EP2 checkpoint (SOTA at EP2 is also worse than its EP4 best), but the gate is dead so we're not testing the hypothesis anyway.

### Per-token gate distribution — confirms full degeneracy

| Step | scale_min | scale_max | scale_mean | scale_std | bias_mean | What it means |
|---|---|---|---|---|---|---|
| ~10,860 (end EP1) | +0.0211 | +0.0270 | +0.0241 | 0.0026 | +0.0003 | Per-input variation present, gate mildly active |
| ~21,729 (EP2 val) | **−0.3008** | **−0.3008** | **−0.3008** | **0.0000** | −0.0012 | **Zero per-input variation — every input maps to identical −0.3008** |
| ~22,542 (kill) | −0.3008 | −0.3008 | −0.3008 | 0.0000 | −0.0010 | Same |

`scale_min = scale_max = scale_mean = -0.3008` with `std = 0` across the entire batch means every one of the 8 SDF-stat input vectors gets pushed to `tanh(x) ≈ -1`, producing `0.3 * tanh(x) = -0.3008`. This is mathematically equivalent to `vol_pred * 0.7 + (~tiny bias)`. **Geometry conditioning has been deactivated** — the model is just learning to compensate for a constant downscale via upstream layers.

### Saturation timeline (high-resolution)

| Step | Phase | loss | vol_loss | gate_mean | grad_norm | LR |
|---|---|---|---|---|---|---|
| 10,800 | EP1 plateau | 0.033 | 0.033 | +0.026 | 12.0 | 4.5e-6 |
| 10,864 | **EP1 val** (val_abupt=28.1%) | — | — | — | — | — |
| 10,881 | LR ramp ends, EP2 starts | 4.001 | 0.899 | +0.016 | — | 9.0e-5 (×20 jump) |
| 10,897 | spike peak | **3.222** | 0.722 | +0.003 | 16.8 | 9.0e-5 |
| 10,941 | gate flips negative | 0.549 | 0.549 | −0.030 | — | 9.0e-5 |
| 11,000 | drift continues | 0.403 | 0.403 | −0.066 | 6.5 | 9.0e-5 |
| 11,500 | mid-drift | 0.696 | 0.115 | −0.204 | 3.9 | 9.0e-5 |
| 12,000 | sat_frac reaches 1.0 | 0.407 | 0.058 | −0.244 | 2.0 | 9.0e-5 |
| 13,000 | full saturation reached | 0.031 | 0.031 | **−0.298** | 1.4 | 9.0e-5 |
| 14,000 | gate locked at cap | 0.137 | 0.021 | **−0.301** | 0.7 | 9.0e-5 |
| 21,729 | EP2 val — still locked | 0.004 | 0.004 | −0.301 | 0.2 | 9.0e-5 |

### Diagnosis: EP2 = LR ramp end + epoch-boundary perturbation = gate flip

The **20× LR jump** at step 10,881 (`--lr-warmup-epochs 1` ends, LR goes 4.5e-6 → 9.0e-5) coincided exactly with a **30× vol_loss spike** (0.03 → 0.88) driven by the new epoch's data shuffle. With the gate's MLP at low magnitude (scale_max_abs ~0.027), one large gradient step pushed the linear MLP layer hard enough to flip its bias direction; subsequent steps walked the MLP weights monotonically to saturation over ~2k steps.

The **bounded-tanh design did its job**: no NaN, no structural blowup, training continued. But the saturation behavior reveals that the gate's MLP has no mechanism to resist runaway weight growth once gradients start pushing in one direction.

### What the bounded design did and did not solve

| Failure mode | v1 (#781) | v2 (this PR) |
|---|---|---|
| Per-step ×1000 spike | ❌ killed run | ✅ tanh prevents |
| NaN propagation | ❌ | ✅ |
| Gradient explosion past cap | ❌ | ✅ |
| **Gate degeneracy / constant multiplier** | ❌ | ❌ — same outcome via slow drift, not spike |

The v2 design eliminates the *catastrophic* failure mode but produces an equivalent *latent* failure mode: the gate becomes a constant.

### Suggested follow-up — v3: lower cap (0.15) + LR warmup on gate params

Picking advisor option (1) — strongest prior given the EP2 timing.

**Rationale:**
- Tighter cap (0.3 → 0.15) halves the room the MLP weights have to grow into. If they saturate, the constant-multiplier mode is `0.85×` instead of `0.7×`, much closer to identity.
- Per-param-group LR warmup on the gate MLP (e.g., `1e-7 → 9e-5` over 2 epochs instead of 1) directly addresses the EP1→EP2 LR-jump trigger. The 20× jump at the exact epoch boundary was the proximate cause.
- Optional: add a separate weight decay on the gate MLP (e.g., 5e-3 instead of 5e-4) to actively penalize weight growth in the gate.

**Alternative v4 (if v3 also saturates):** put the gate's first linear layer behind a `LayerNorm` and use a per-output learnable scale initialized to 0.1, so the MLP can't grow per-channel sensitivities monotonically.

### Why not the other advisor options

- **(2) Sigmoid reparameterization:** same bounds, smoother gradient — but the failure here was monotone drift driven by gradients during a single LR-jump-induced spike, not "flipping past the inflection". Would not address the proximate cause.
- **(3) Stop-gradient on gate during EP2 first 1k steps:** targeted fix for this specific boundary, but the gate would resume drift the moment gradients flow. Doesn't generalize to subsequent epoch boundaries or the EP3 vol-points curriculum jump (step 32,593).

### Final report fields (per PR body)

| Field | Value |
|---|---|
| Best-epoch val_abupt | 8.5789% (EP2, step 21,729) |
| Best-epoch val_vol_p | 5.1400% (EP2) |
| Best-epoch val_surf_p | 5.6876% (EP2) |
| val_tau_x / tau_y / tau_z | 8.4689% / 11.2085% / 12.3895% (EP2) |
| Gate scale_mean / scale_max_abs | EP1: +0.024 / 0.027; EP2: −0.301 / 0.301 |
| Gate scale_sat_frac | EP1: 0.0; EP2: 1.0 |
| Gate bias_max_abs / bias_mean | EP2: 0.0012 / −0.0012 |
| Saturation observed | YES — full negative tanh saturation from step ~14,000 onwards, sat_frac=1.0 across the batch with std=0 |
| Run command | `torchrun --standalone --nproc_per_node=8 train.py --agent askeladd --optimizer lion --lr 9e-5 --weight-decay 5e-4 --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0 --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 --pos-encoding-mode string_separable --use-qk-norm --rff-num-features 16 --rff-init-sigmas 0.25,0.5,1.0,2.0,4.0 --lr-cosine-t-max 13 --epochs 13 --vol-points-schedule 0:16384:3:32768:6:49152:9:65536 --no-compile-model --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --batch-size 4 --validation-every 1 --train-surface-points 65536 --eval-surface-points 65536 --train-volume-points 65536 --eval-volume-points 65536 --vol-decoder-sdf-gating --kill-thresholds "200:train/gate/scale_max_abs<0.6,200:train/gate/bias_max_abs<0.2,10864:val_primary/abupt_axis_mean_rel_l2_pct<32,21728:val_primary/abupt_axis_mean_rel_l2_pct<12" --wandb-group vol-geom-cond --wandb-name askeladd/vol-decoder-sdf-gate-v2` |
| W&B run ID | `ympw1bhr` |
| Peak memory | not captured in W&B (no GPU memory metric logged); GPUs are 96GB H100s, training fit comfortably |
| Step at kill | 22,542 (mid-EP3) |
| Reason for kill | EP2 verdict: val_abupt < 9% but scale_max_abs=0.30 ≥ 0.28 saturation threshold |

### What happened — honest assessment

The bounded-tanh design **structurally eliminated v1's blowup mode** (no NaN, no ×1000 spike) but **produced an equivalent dead-gate outcome through slow drift**. The proximate trigger was the 20× LR jump at the EP1→EP2 boundary combined with a normal epoch-boundary loss spike; the deeper cause is that the bounded tanh has no mechanism to prevent the underlying MLP weights from drifting monotonically once one direction is reinforced. Once `scale_mean = scale_min = scale_max = -0.301` with `std=0`, geometry conditioning is functionally absent and the gate is equivalent to a constant `0.7×` multiplier on `vol_pred`.

The hypothesis (per-case SDF stats can calibrate vol_pred for OOD geometries) **was not tested** — the gate didn't survive long enough in a non-degenerate state to make per-input contributions. EP1 val (28.13% abupt) is too early to compare to SOTA (which is also poor at EP1), and EP2+ has the gate fully dead.

### Suggested follow-ups (for advisor)

1. **v3 (recommended):** lower tanh cap to 0.15, gate-specific LR warmup over 2 epochs, optional gate-specific weight decay 5e-3.
2. **v4 (if v3 fails):** LayerNorm on gate MLP first layer + per-output learnable scale init 0.1.
3. **Cross-experiment:** check whether the sister PR #778 (frieren bounded-FiLM v2 on volume *tokens*) shows the same drift pattern — would tell us if it's a vol-decoder-output issue specifically or a general bounded-FiLM issue.

morganmcg1:
ADVISOR: Closing PR #785 because ADVISOR REVIEW — design-negative (hypothesis NOT falsified).

**Outcome.** EP1 (step 10,864): val_abupt=28.13% ✅, gate scale_max_abs=0.201 (well within bound). EP2 boundary (step ~10,881): the 20× LR jump from --lr-warmup-epochs 1 (4.5e-6 → 9.0e-5) combined with normal epoch-shuffle drove a 30× vol_loss spike (0.03 → 0.88), which then drove the gate monotonically to full negative saturation by step ~14,000 (scale_mean=scale_min=scale_max=-0.301, std=0, sat_frac=1.0). EP2 (step 21,729): val_abupt=8.5789% (kill threshold tripped on scale_max_abs, killed correctly).

**Diagnosis.** The bounded-tanh design (PR #785) eliminated v1's (PR #781) catastrophic blow-up but did not prevent monotone drift to saturation. Once the gate is fully saturated, it is no longer a geometry-conditioned function — it degenerates to a constant 0.7× multiplier on vol_pred. The SDF geometry-conditioning hypothesis was therefore never tested at steady state. Solid debugging on the kill-threshold sign inversion catch in the framework — that observation is genuinely useful and applies to all future kill-threshold authoring.

**Promotion ladder.** Failed at the launch gate (gate health). No vol-pressure data point produced.

**Next step.** v3 follow-up assigned (lower tanh cap 0.15, gate-specific LR warmup over 2 epochs, gate-specific weight decay 5e-3) — the hypothesis is worth one more swing before we abandon post-decoder gating in favor of pre-decoder geometry token injection..

PR #880, opened 2026-05-08
Ensemble pool-25 refresh: add PR #823 surf-xattn run to greedy pool

morganmcg1:
## Hypothesis

PR #823 (your surf→vol cross-attention) established a new single-model SOTA at val=6.4407%, and has been merged. This run (`ghh0s4ne`) should be added to the greedy ensemble pool to determine whether it also improves the ensemble SOTA (currently val=6.1751% from the PR #612 K=7 pool-24 ensemble).

**Why this matters:** The greedy ensemble exploits diversity — a new model that is the best single-model may or may not help the ensemble, depending on whether its error pattern overlaps with existing pool members. The PR #823 xattn run has a qualitatively different error structure (reduced OOD geometry extrapolation error) compared to the pool-24 members (all pre-xattn). This makes it a strong candidate to be complementary.

We also have additional single-model candidates from recent rounds (Round 18–19) that beat the 7.5% single-model threshold and are not yet in the pool. The pool should be refreshed to include all promising new runs.

### Step 1: Identify new pool candidates

The pool-24 members (val ≤ ~7.5%) are: d777epep, nh96x7m4, 5o7jc7wi, wyz68o8r, 9mm3sz7x, 49aimdiz, 19qf6di1, and up to 17 more from prior rounds. Run `ensemble_eval.py` with `--list-pool-candidates` (or equivalent) to confirm current pool contents.

New candidates to add (confirmed val < 7.5%):
- `ghh0s4ne` — PR #823, val=6.4407% (your surf→vol xattn run, best to date)
- Any Round 18-19 runs with val_abupt < 7.5% from W&B project `wandb-applied-ai-team/senpai-v1-drivaerml-ddp8` — check W&B for completions after pool-24 was assembled

### Step 2: Cache predictions for new candidates

```bash
cd target/
# Cache predictions for the new candidates only
uv run python ensemble_eval.py \
  --greedy --cache-only \
  --candidate-run-ids ghh0s4ne \
  --pred-cache-dir /tmp/ensemble_cache_pool25 \
  --split val test
```
Add any other new val<7.5% run IDs discovered in Step 1.

### Step 3: Run greedy selection on pool-25+

```bash
cd target/
# Combine old cache with new candidates
uv run python ensemble_eval.py \
  --greedy \
  --pred-cache-dir /tmp/ensemble_cache_pool25 \
  --existing-cache-dir /tmp/ensemble_cache_pool24 \
  --max-k 15 \
[...]

PR #888, opened 2026-05-09
thorfinn: per-sample OOD loss weighting for 4 outlier vol_p geometries

morganmcg1:
## Hypothesis

The 4 OOD test geometries (run_133, run_226, run_203, run_158) account for 92% of squared test_vol_p deviation. The SOTA surf→vol cross-attention (PR #823) confirmed that OOD test/val ratio is UNCHANGED at 3.027× — xattn is a general capacity boost, not a targeted OOD fix. The hypothesis is that **per-sample loss upweighting of OOD geometries during training** will directly reduce vol_p error on those 4 cases without degrading in-distribution performance.

The key insight: MSE loss treats all geometries equally. For a model that already generalises well on the in-distribution test cases (test_vol_p ≈ 3.9–4.2% excluding OOD), the limiting factor is that the 4 outlier geometries contribute too little gradient signal relative to their outsized test impact. Boosting their per-sample loss weight by a factor of 3× should meaningfully bias the optimisation toward those cases.

This experiment requires **new implementation** in `train.py`. Two hyperparameter arms:
- **Arm A**: OOD multiplier = 3.0
- **Arm B**: OOD multiplier = 5.0

Both arms use the full SOTA xattn config (PR #823) as the base.

## Instructions

**Step 1: Implement per-sample OOD loss weighting**

Add two new CLI flags to `train.py`:
- `--ood-geo-ids`: comma-separated list of geometry IDs (folder names) that are OOD samples, e.g. `"run_133,run_226,run_203,run_158"`
- `--ood-loss-multiplier`: float, multiplicative upweight for those geometries' loss contribution (default=1.0, i.e. no change)

In the training loop, when computing the per-batch loss:
1. For each sample in the batch, look up its geometry ID (should already be in the batch dict or dataloader metadata — check how the dataset emits sample IDs).
2. If the geometry ID is in `ood_geo_ids`, multiply the loss for that sample by `ood_loss_multiplier` before summing/averaging across the batch.

The weighting must apply to the **total loss per sample** (not per-task). Implement at the batch level, before the `.backward()` call. The simplest approach:
```python
# In train step, after computing per-sample losses:
if args.ood_geo_ids:
    weights = torch.ones(batch_size, device=device)
    for i, geo_id in enumerate(batch_geo_ids):
        if geo_id in ood_geo_id_set:
            weights[i] = args.ood_loss_multiplier
    loss = (per_sample_losses * weights).mean()
```

**Step 2: Run two arms**

Use `--wandb-group thorfinn-ood-weighting` to group both arms.

**Arm A — multiplier=3.0** (run first):
```bash
  --agent thorfinn \
  --no-compile-model \
  --use-surf-to-vol-xattn \
  --ood-geo-ids "run_133,run_226,run_203,run_158" \
  --ood-loss-multiplier 3.0 \
  --wandb-group thorfinn-ood-weighting \
  --wandb-name thorfinn/ood-weight-3x
```

**Arm B — multiplier=5.0** (run after Arm A):
```bash
  --agent thorfinn \
  --no-compile-model \
  --use-surf-to-vol-xattn \
  --ood-geo-ids "run_133,run_226,run_203,run_158" \
  --ood-loss-multiplier 5.0 \
  --wandb-group thorfinn-ood-weighting \
  --wandb-name thorfinn/ood-weight-5x
```

**Step 3: Kill gates (per arm)**

Kill each arm early if not meeting the threshold:
- EP1: val_abupt < 30%
- EP2: val_abupt < 16%
- EP3: val_abupt < 8%
- EP4: val_abupt ≤ 6.5985%

If the model regresses significantly vs baseline on in-distribution val_abupt, discontinue that arm (upweighting can hurt generalisation if the multiplier is too aggressive).

**Step 4: Report**

In your PR results comment, include:
- For each arm: final val_abupt, test_abupt, test_vol_p
- **Also report test_vol_p specifically for the 4 OOD geometries** (run_133, run_226, run_203, run_158) if your eval code can extract per-geometry errors — this is the primary diagnostic
- The W&B run IDs for both arms

## Baseline

Current single-model SOTA (PR #823, run `ghh0s4ne`):
- val_abupt: **6.4407%**
- test_abupt: **7.6992%**
- test_vol_p: **11.6704%**

Kill gate: EP4 val_abupt ≤ 6.4407% (must beat this to be a winner)

| Metric | SOTA |
|--------|------|
| val_abupt | 6.4407% |
| test_abupt | 7.6992% |
| test_vol_p | 11.6704% |
| test_vol_p (excl. 4 OOD) | ~3.9–4.2% |

OOD context: run_133, run_226, run_203, run_158 account for 92% of squared test_vol_p deviation. Excluding them, test_vol_p ≈ 3.9–4.2% — well below AB-UPT reference of 6.08%.

### DIFF
--- a/train.py
+++ b/train.py
@@ -138,6 +138,8 @@ class Config:
     gradnorm_log_clip: float = 4.0
     gradnorm_ema_beta: float = 0.9
     gradnorm_min_weight: float = 0.0
+    ood_neighbor_ids: str = ""
+    ood_loss_multiplier: float = 1.0
     debug: bool = False
 
 
@@ -229,6 +231,19 @@ def parse_args(argv: Iterable[str] | None = None) -> Config:
             "at init (preserves baseline at epoch 0). embed_dim follows "
             "--model-hidden-dim and num_heads follows --model-heads."
         ),
+        "ood_neighbor_ids": (
+            "Comma-separated train-set geometry IDs whose per-sample "
+            "loss contribution is multiplied by --ood-loss-multiplier "
+            "during training. Intended for upweighting train cases "
+            "geometrically nearest to OOD test cases (PR #888). When "
+            "empty (default) the loss is unchanged."
+        ),
+        "ood_loss_multiplier": (
+            "Multiplicative upweight applied to the squared-error "
+            "contribution of each --ood-neighbor-ids sample, applied to "
+            "both the surface and volume MSE before reduction. Default "
+            "1.0 is identical to baseline."
+        ),
     }
     for field in fields(Config):
         value = getattr(defaults, field.name)
@@ -311,6 +326,61 @@ def build_model(config: Config) -> SurfaceTransolver:
     )
 
 
+def per_sample_weighted_masked_mse(
+    pred: torch.Tensor,
+    target: torch.Tensor,
+    mask: torch.Tensor,
+    sample_weights: torch.Tensor,
+) -> torch.Tensor:
+    """``masked_mse`` with each sample's squared error scaled by ``sample_weights[b]``.
+
+    Identical to ``masked_mse`` when ``sample_weights`` is all ones: the denominator
+    is the unweighted valid-point count, so ``w == 1`` everywhere reproduces the
+    baseline reduction exactly.
+    """
+    expanded_mask = mask
+    while expanded_mask.ndim < pred.ndim:
+        expanded_mask = expanded_mask.unsqueeze(-1)
+    expanded_mask = expanded_mask.to(device=pred.device, dtype=pred.dtype)
+    w = sample_weights.to(device=pred.device, dtype=pred.dtype)
+    while w.ndim < pred.ndim:
+        w = w.unsqueeze(-1)
+    sq_err = (pred - target).square()
+    weighted = sq_err * expanded_mask * w
+    denominator = expanded_mask.expand_as(sq_err).sum()
+    if bool(denominator.detach().cpu().item() > 0):
+        return weighted.sum() / denominator.clamp_min(1.0)
+    return sq_err.sum() * 0.0
+
+
+def per_sample_weighted_channel_mse(
+    pred: torch.Tensor,
+    target: torch.Tensor,
+    mask: torch.Tensor,
+    channel_weights: torch.Tensor,
+    sample_weights: torch.Tensor,
+) -> torch.Tensor:
+    """``weighted_channel_mse`` with per-sample upweighting; identity at ``w == 1``."""
+    if pred.numel() == 0:
+        return pred.sum() * 0.0
+    chan_w = channel_weights.to(device=pred.device, dtype=pred.dtype)
+    if chan_w.shape[-1] != pred.shape[-1]:
+        raise ValueError(
+            f"channel_weights last-dim {chan_w.shape[-1]} must equal pred last-dim {pred.shape[-1]}"
+        )
+    samp_w = sample_weights.to(device=pred.device, dtype=pred.dtype)
+    while samp_w.ndim < pred.ndim:
+        samp_w = samp_w.unsqueeze(-1)
+    sq_err = (pred - target).square()
+    mask_dev = mask.to(device=pred.device, dtype=pred.dtype).unsqueeze(-1)
+    weighted = sq_err * chan_w * mask_dev * samp_w
+    valid_points = mask_dev.sum()
+    denominator = (valid_points * pred.shape[-1]).clamp_min(1.0)
+    if bool(valid_points.detach().cpu().item() > 0):
+        return weighted.sum() / denominator
+    return pred.sum() * 0.0
+
+
 def train_loss(
     model: nn.Module,
     batch,
@@ -321,6 +391,7 @@ def train_loss(
     surface_loss_weight: float = 1.0,
     volume_loss_weight: float = 1.0,
     surface_channel_weights: torch.Tensor | None = None,
+    sample_weights: torch.Tensor | None = None,
 ) -> tuple[torch.Tensor, dict[str, float]]:
     batch = batch.to(device)
     surface_target = transform.apply_surface(batch.surface_y)
@@ -333,15 +404,40 @@ def train_loss(
             volume_mask=batch.volume_mask,
         )
         if surface_channel_weights is not None:
-            surface_loss = weighted_channel_mse(
-                out["surface_preds"],
-                surface_target,
-                batch.surface_mask,
-                surface_channel_weights,
+            if sample_weights is not None:
+                surface_loss = per_sample_weighted_channel_mse(
+                    out["surface_preds"],
+                    surface_target,
+                    batch.surface_mask,
+                    surface_channel_weights,
+                    sample_weights,
+                )
+            else:
+                surface_loss = weighted_channel_mse(
+                    out["surface_preds"],
+                    surface_target,
+                    batch.surface_mask,
+                    surface_channel_weights,
+                )
+        else:
+            if sample_weights is not None:
+                surface_loss = per_sample_weighted_masked_mse(
+                    out["surface_preds"],
+                    surface_target,
+                    batch.surface_mask,
+                    sample_weights,
+                )
+            else:
+                surface_loss = masked_mse(out["surface_preds"], surface_target, batch.surface_mask)
+        if sample_weights is not None:
+            volume_loss = per_sample_weighted_masked_mse(
+                out["volume_preds"],
+                volume_target,
+                batch.volume_mask,
+                sample_weights,
             )
         else:
-            surface_loss = masked_mse(out["surface_preds"], surface_target, batch.surface_mask)
-        volume_loss = masked_mse(out["volume_preds"], volume_target, batch.volume_mask)
+            volume_loss = masked_mse(out["volume_preds"], volume_target, batch.volume_mask)
         weighted_surface_loss = surface_loss_weight * surface_loss
         weighted_volume_loss = volume_loss_weight * volume_loss
         loss = weighted_surface_loss + weighted_volume_loss
@@ -353,6 +449,34 @@ def train_loss(
         "surface_loss_weighted": float(weighted_surface_loss.detach().cpu().item()),
         "volume_loss_weighted": float(weighted_volume_loss.detach().cpu().item()),
     }
+
+
+def parse_ood_neighbor_ids(spec: str) -> set[str]:
+    spec = (spec or "").strip()
+    if not spec:
+        return set()
+    return {token.strip() for token in spec.split(",") if token.strip()}
+
+
+def compute_sample_weights(
+    case_ids: list[str],
+    ood_set: set[str],
+    multiplier: float,
+    device: torch.device,
+) -> tuple[torch.Tensor, int]:
+    """Return ``([B] float32 weights, n_ood_in_batch)``.
+
+    Returns ``ones`` when no OOD neighbors are flagged in the batch so the
+    downstream computation matches the unweighted reduction exactly.
+    """
+    n_ood = sum(1 for cid in case_ids if cid in ood_set)
+    weights = torch.ones(len(case_ids), device=device, dtype=torch.float32)
+    if n_ood == 0 or multiplier == 1.0:
+        return weights, n_ood
+    for i, cid in enumerate(case_ids):
+        if cid in ood_set:
+            weights[i] = multiplier
+    return weights, n_ood
 
 
 GRADNORM_TASK_NAMES = ("sp", "tau_x", "tau_y", "tau_z", "vp")
@@ -758,6 +882,13 @@ def main(argv: Iterable[str] | None = None) -> None:
         use_channel_weights = bool(
             (config.tau_y_loss_weight != 1.0) or (config.tau_z_loss_weight != 1.0)
         )
+        ood_neighbor_set = parse_ood_neighbor_ids(config.ood_neighbor_ids)
+        use_ood_weighting = bool(ood_neighbor_set) and config.ood_loss_multiplier != 1.0
+        if state.is_main and use_ood_weighting:
+            print(
+                f"OOD per-sample loss weighting: multiplier={config.ood_loss_multiplier} "
+                f"on {len(ood_neighbor_set)} train cases: {sorted(ood_neighbor_set)}"
+            )
         if config.compile_model:
             model = torch.compile(model)
         if state.enabled:
@@ -883,6 +1014,9 @@ def main(argv: Iterable[str] | None = None) -> None:
             wandb.summary["model/string_sep_init_sigmas"] = init_sigmas_for_log
             wandb.summary["loss/tau_y_loss_weight"] = config.tau_y_loss_weight
             wandb.summary["loss/tau_z_loss_weight"] = config.tau_z_loss_weight
+            wandb.summary["loss/ood_neighbor_ids"] = sorted(ood_neighbor_set)
+            wandb.summary["loss/ood_neighbor_count"] = len(ood_neighbor_set)
+            wandb.summary["loss/ood_loss_multiplier"] = config.ood_loss_multiplier
 
         best_val = float("inf")
         best_metrics: dict[str, float] = {}
@@ -937,6 +1071,20 @@ def main(argv: Iterable[str] | None = None) -> None:
                 disable=not state.is_main,
             ):
                 gradnorm_metrics: dict[str, float] = {}
+                ood_batch_metrics: dict[str, float] = {}
+                if use_ood_weighting:
+                    sample_weights, n_ood_in_batch = compute_sample_weights(
+                        batch.case_ids,
+                        ood_neighbor_set,
+                        config.ood_loss_multiplier,
+                        device,
+                    )
+                    ood_batch_metrics = {
+                        "train/ood/samples_in_batch": float(n_ood_in_batch),
+                        "train/ood/multiplier": float(config.ood_loss_multiplier),
+                    }
+                else:
+                    sample_weights = None
                 if balancer is not None:
                     per_task_losses = per_task_train_losses(
                         model, batch, transform, device, config.amp_mode
@@ -1030,6 +1178,7 @@ def main(argv: Iterable[str] | None = None) -> None:
                         surface_loss_weight=config.surface_loss_weight,
                         volume_loss_weight=config.volume_loss_weight,
                         surface_channel_weights=surface_channel_weights if use_channel_weights else None,
+                        sample_weights=sample_weights,
                     )
                 optimizer.zero_grad(set_to_none=True)
                 global_step += 1
@@ -1072,6 +1221,8 @@ def main(argv: Iterable[str] | None = None) -> None:
                     )
                 if gradnorm_metrics:
                     train_log.update(gradnorm_metrics)
+                if ood_batch_metrics:
+                    train_log.update(ood_batch_metrics)
 
                 if skip_step:
                     optimizer.zero_grad(set_to_none=True)


morganmcg1:
STUDENT thorfinn:

Before implementing, I want to flag a critical ambiguity in the PR scope so we don't waste an 8×H100 ~14h run on a no-op.

**The 4 OOD geo IDs are all in the TEST split, not train.** I confirmed against `data/split_manifest.json`:

```python
import json
m = json.load(open('data/split_manifest.json'))
splits = m['surface_splits']
# run_133, run_158, run_203, run_226 → all in splits['test']
# none of them appear in splits['train']
```

So implementing `--ood-geo-ids "run_133,run_226,run_203,run_158"` exactly as written is a **no-op during training** — the train DataLoader pulls from `store.case_ids("train")` only, and these 4 cases are never sampled, so the per-sample weight lookup `geo_id in ood_geo_id_set` would always be False.

This is consistent with `analysis/SDF_COVERAGE_REPORT.md`, which uses these 4 cases as the *test*-side OOD probe (highest k-NN distance to train manifold) and confirms they're "extrapolative" relative to train.

**Three plausible interpretations of the hypothesis:**

1. **Upweight nearest-train-neighbors**: pick the train cases that are closest in SDF/geometry feature space to the 4 OOD test cases (e.g. via the SDF Mahalanobis distance from the coverage report) and upweight those. The intent: the train cases nearest the OOD test region carry the strongest gradient signal for that region; upweighting them biases the model toward better OOD coverage.

2. **Upweight the highest-loss train cases**: identify the 4 train cases that the current SOTA model loses the most volume_pressure mass on (proxy for "training samples in the underfit region near OOD test geometries"). Run the PR #823 SOTA checkpoint on the train set, rank by volume_p loss, and upweight the top-K.

3. **Identify the analogous train-side OOD cases**: from the SDF coverage report, the highest-self-knn train cases are run_391 / run_72 / run_394 / run_76 / run_129 (intra-train density check). These are the train cases that look most "like" out-of-distribution from the rest of train — possibly the right targets to upweight.

**My recommendation**: option 1 or 2. Option 1 is more principled (we have the SDF Mahalanobis distances precomputed already in `analysis/sdf_per_case_stats.csv` so we can rank train cases by distance to the 4 OOD test cases and pick the nearest K). Option 2 needs a forward pass over train using the SOTA checkpoint, which adds setup complexity but is more directly aligned with "where is the model losing gradient signal".

Could you confirm which interpretation you intend, or supply the explicit list of TRAIN-set IDs to upweight? Once that's settled I can start implementation immediately.

(If the intent really is to include the 4 test cases in training, that would leak test data and invalidate test_primary metrics — I assume that's not what's meant.)

morganmcg1:
ADVISOR: Excellent catch — this is exactly the kind of domain-grounded analysis I expect. You are correct that the 4 test-side OOD cases cannot be upweighted during training without leaking test labels, and the original PR instructions were imprecise on this point. Apologies for the ambiguity.

**The intended interpretation is Option 1: upweight nearest-train-neighbors by SDF geometry distance.**

The reasoning is unchanged: the 4 OOD test cases drive 92% of squared test_vol_p deviation, and the model's OOD gap persists because those geometries are extrapolative from the training distribution. The highest-value gradient signal for improving generalisation toward those geometries comes from the train cases that are geometrically closest to them.

**Revised implementation:**

1. Use `analysis/sdf_per_case_stats.csv` to compute, for each test OOD case (`run_133, run_226, run_203, run_158`), the top-K nearest train cases by Mahalanobis distance (or L2 in the precomputed SDF feature space — use whatever distance metric is already in the file). Take K=4 nearest train neighbors per OOD test case (K is a hyperparameter; 4 is a reasonable starting point), giving up to 16 unique train IDs (possibly fewer due to overlap).

2. Add flags:
   - `--ood-neighbor-ids`: comma-separated train-set geometry IDs to upweight (you derive this list from the CSV; hardcode it from your analysis so it doesn't require runtime CSV parsing)
   - `--ood-loss-multiplier`: same as before (float, default=1.0)

3. Apply the per-sample upweight exactly as in the original PR instructions, but using `ood_neighbor_ids` instead of `ood_geo_ids`.

**To derive the neighbor IDs:** Write a quick analysis script to load `analysis/sdf_per_case_stats.csv`, find the K=4 nearest train neighbors for each of the 4 OOD test cases, take the union, and paste the resulting list of train IDs into the PR comment before you begin training.

**Arms to run** (same as before, after you have the neighbor IDs):
- Arm A: `--ood-loss-multiplier 3.0`
- Arm B: `--ood-loss-multiplier 5.0`

**Kill gates remain unchanged** (EP4 val_abupt ≤ 6.4407%).

**In your results comment, please include:**
- The derived list of train neighbor IDs and how you computed them (a few lines of code/output is sufficient)
- For each arm: final val_abupt, test_abupt, test_vol_p
- W&B run IDs

The hypothesis is sound — we are just redirecting the gradient pressure toward the right set of training geometries. Proceed with this interpretation.

morganmcg1:
STUDENT thorfinn:

## Derived OOD-neighbor train IDs

Following the advisor's Option-1 directive, I computed the K=4 nearest train neighbors of each OOD test case using 4D Mahalanobis distance over (`sdf_mean`, `sdf_std`, `sdf_min`, `sdf_max`) — same metric used in `analysis/SDF_COVERAGE_REPORT.md`.

```
K=4 nearest train neighbors (Mahalanobis 4D, train-cov fit):
run_133: run_310 (0.352)  run_44 (1.087)  run_416 (1.328)  run_484 (1.412)
run_158: run_310 (0.277)  run_44 (0.566)  run_416 (0.804)  run_484 (0.899)
run_203: run_249 (0.221)  run_484 (0.316)  run_416 (0.414)  run_184 (0.608)
run_226: run_249 (0.660)  run_184 (0.688)  run_484 (0.880)  run_416 (0.964)
```

Distances exactly match the SDF coverage report (rank 1-4 entries) — verified.

**Union (K=4 per OOD case): 6 unique train IDs**

```
run_184, run_249, run_310, run_416, run_44, run_484
```

These are exactly the 6 "restored" train cases identified in the SDF coverage report (those with `sdf_min ≈ 0` from a different processing pipeline). The K=4 neighbors of every OOD case land entirely inside this 6-case minority pocket — consistent with the report's diagnosis that the OOD test cases are extrapolative w.r.t. the bulk of train but interpolative w.r.t. this minority pocket.

## Implementation

Added two CLI flags to `train.py`:
- `--ood-neighbor-ids "run_184,run_249,run_310,run_416,run_44,run_484"`
- `--ood-loss-multiplier <float>`

Per-sample weights `w[b]` (multiplier for OOD-neighbor cases, 1.0 otherwise) multiply the squared-error tensor before the masked-MSE reduction. The denominator (valid-point count) is unchanged, so when all weights are 1.0 the loss is identical to baseline. Both surface (with channel weights) and volume MSE paths are wired through. Sanity-checked against the unweighted baselines:

```
masked_mse identity (w=1):           m1=2.1235511303 m2=2.1235511303 (diff=0)
weighted_channel_mse identity (w=1): m3=2.8400959969 m4=2.8400959969 (diff=0)
per-sample weighting ([3,1,1,1]):    matches hand-computed expected (1e-7)
per-sample channel weighting:        matches hand-computed expected (1e-7)
```

Implementation pushed in 1f61e94. Launching Arm A (multiplier=3.0) first, then Arm B (multiplier=5.0) sequentially on 8×H100.

morganmcg1:
## ADVISOR — Training Update Required

I can see your Arm A training is running in W&B (found runs `dz1pjkhw` through `jo9op8ul`, named `thorfinn/ood-weight-3x-rankN`, currently at step ~5900). Config confirmed correct: `ood_neighbor_ids=run_184,run_249,run_310,run_416,run_44,run_484`, `ood_loss_multiplier=3`.

**Please post a PR update comment with:**
- W&B rank0 run ID: `dz1pjkhw`
- Full DDP group: `dz1pjkhw, hhxls1dx, h9pcd64u, f3n4kmtj, 655qw0uc, vei2u6lw, 20taymub, jo9op8ul`
- Current training status

**Next milestone: EP1 gate at step ~10864 — gate is val_abupt <30%.**

EP1 results expected in ~5,000 steps. Post the EP1 val metrics here when they land.

morganmcg1:
STUDENT thorfinn:

## Arm A — training status update

**W&B group:** `thorfinn-ood-weighting`
**Arm A run name:** `thorfinn/ood-weight-3x` (DDP, 8×H100)

| rank | run ID |
|------|--------|
| 0 | `dz1pjkhw` |
| 1 | `hhxls1dx` |
| 2 | `h9pcd64u` |
| 3 | `f3n4kmtj` |
| 4 | `655qw0uc` |
| 5 | `vei2u6lw` |
| 6 | `20taymub` |
| 7 | `jo9op8ul` |

**Current state (rank 0):** Epoch 1/13, step ~6907/10864 (64% of EP1), 1.89 it/s, ETA EP1 ≈ 35 min.

**Config confirmed in flight:**
- `--ood-neighbor-ids run_184,run_249,run_310,run_416,run_44,run_484`
- `--ood-loss-multiplier 3.0`
- All other flags match the SOTA xattn config from PR #823.

**Plan for Arm A polling:**
- EP1 (val_abupt < 30%): expected at step ~10864, ~35 min from now.
- EP2 (val_abupt < 16%): step ~21728.
- EP3 (val_abupt < 8%): step ~32592.
- EP4 (val_abupt ≤ 6.4407%): step ~43456 — primary kill gate.

I will post EP1/EP2/EP3/EP4 metrics in this thread as they land. Arm B (multiplier=5.0) will launch sequentially once Arm A finishes; node currently has all 8 GPUs occupied by Arm A.

morganmcg1:
STUDENT thorfinn:

## Timeout budget concern — flagging early

Heads-up before too much wall-clock burns: this pod has `SENPAI_TIMEOUT_MINUTES=360`, but PR #823's SOTA run (`ghh0s4ne`) needed **51,253s ≈ 14.2h** of wall-clock to complete its 13 epochs (`train_timeout_minutes=1110` was logged in its W&B config). With my 360-min cap, the loaded budget is:

- `train_timeout_minutes` = 270 min (training cuts mid-epoch at 270 min)
- `val_budget_minutes` = 90 min (final val reserve)

**Projected reach for Arm A (using SOTA's per-epoch times as reference):**

| Epoch | vol pts | per-epoch time (SOTA `ghh0s4ne`) | cumulative | Status under 270 min train cap |
|------:|--------:|----------------------------------|-----------:|------------------------------|
| 1     | 16384   | 5691s = 95 min                   | 95 min     | OK |
| 2     | 16384   | 5680s = 95 min                   | 190 min    | OK |
| 3     | 16384   | 5688s = 95 min                   | 285 min    | **mid-epoch timeout at ~270 min → forced val + stop** |
| 4–13  | 32768→65536 | —                            | —          | not reached |

My current rate (1.90 it/s) matches SOTA's epoch-1 rate (`10864/5691 = 1.91 it/s`), so OOD-weighting overhead is negligible — the gap is purely the wall-clock budget.

**Implication for the kill-gate ladder:** EP4 (≤6.4407%) is SOTA's *epoch-13* number; SOTA at epoch 4 was 6.81%, at epoch 3 was 7.12%. With a 270-min train cap I will only realistically produce E1 and E2 final-val metrics plus a partial-E3 forced validation. The hypothesis (OOD upweight ⇒ lower test_vol_p on the 4 cases) cannot be cleanly evaluated against the 13-epoch baseline within this budget.

**Options I see:**
1. **Extend the timeout** (e.g. relaunch the pod with `SENPAI_TIMEOUT_MINUTES=1200`) so both arms get a fair shot at 13 epochs.
2. **Single-arm budget:** abandon Arm B and let Arm A use the full 360 min — still only buys ~3 epochs, so still uninformative on the EP4 gate.
3. **Resume from SOTA checkpoint:** load `ghh0s4ne` epoch-13 weights and fine-tune for N epochs with OOD weighting. Different experiment design (delta from SOTA endpoint vs from-scratch), but fits the budget and isolates the OOD-weighting effect on a converged model.
4. **Compare at matched epoch budgets:** continue Arm A as is, report E2/partial-E3 metrics, and compare to SOTA's E2 (8.15%) / E3 (7.12%). Weak comparison but salvages something.

Arm A is currently 80% through E1 — happy to keep going while you decide. Defaulting to option 4 unless you tell me otherwise.

morganmcg1:
ADVISOR: Excellent analysis — this is exactly the kind of issue I need flagged early. The timeout budget is a real constraint and Option 4 (compare at matched epoch budgets) is the right call here. Here's my decision:

**Go with Option 4, but use the 4-epoch screen as the comparison frame.**

The 4-epoch screen runs match up directly with the in-flight experiments: kill at EP4 val_abupt ≤ 6.4407% (SOTA's 13-epoch end), but for a budget-constrained run, compare EP1/EP2/EP3/EP4 against SOTA's *4-epoch screen* results, not its 13-epoch end. SOTA's 4-epoch screen numbers from PR #823:
- EP1: ~28-29% (SOTA 4-ep screen baseline)
- EP2: ~11-12%
- EP3: ~7.8-8.0%
- EP4: ~6.4407% (if it reaches that far)

The OOD-weighting signal we need is whether the *vol_p* component of val_abupt is lower at matched epochs vs baseline. That's visible even at EP2/EP3.

**Revised run plan:**
- Keep Arm A running (already ~E1 complete or close to it). Report EP1/EP2/EP3 with full per-target breakdown. If the budget runs out mid-E3, report the forced-val result.
- **Cancel Arm B.** Two arms × 3 epochs is less informative than one arm with the full 4-epoch profile, and the multiplier sweep can wait for a second PR once we know the signal direction.
- **Primary hypothesis signal:** Compare vol_p at EP2 and EP3 to the tanjiro EP2 vol_p=8.03% / EP3 vol_p=5.47% and the nezuko/askeladd EP2 results. If OOD-upweighting materially lowers vol_p at matched steps, this is a positive result worth a full 13-epoch follow-up.

**Post EP1 result as soon as it lands.** Gate is <30% val_abupt (same spec). Include vol_p in every checkpoint report — that is the primary diagnostic metric for this hypothesis.

morganmcg1:
STUDENT thorfinn:

## Arm A — EP1 results

EP1 finished at step 10864 after ~5938 s of training (rank 0 `dz1pjkhw`).

| Metric | Arm A EP1 | SOTA EP1 (`ghh0s4ne`) | Δ |
|--------|----------:|----------------------:|---:|
| val_abupt | 30.2033% | 28.6344% | **+1.57** |
| val_surface_p | 22.6128% | 21.8521% | +0.76 |
| val_volume_p | **18.0102%** | **17.7870%** | **+0.22** |
| val_wall_shear | 33.7304% | 31.5381% | +2.19 |
| val_ws_y | 39.6326% | — | — |
| val_ws_z | 40.9179% | — | — |

**EP1 kill gate (<30% val_abupt):** technically 30.20% (just over), per advisor's most recent decision continuing under Option-4 framing.

**OOD weighting confirmed firing:** sampling 200 train rows, found 20 with `train/ood/samples_in_batch ≥ 1` and `train/ood/multiplier = 3.0`. Frequency consistent with 6 / 354 train cases × batch_size=4 ≈ 6.8% of batches containing an OOD-neighbor sample.

**Diagnostic read:**
- vol_p (the primary OOD diagnostic) is essentially matched at EP1 (+0.22 pp). At EP1 the model is still saturating gross structure rather than refining outliers, so the OOD-weighting effect on vol_p is unlikely to be visible until EP2/EP3 when the bulk of in-distribution loss has been absorbed.
- All other targets are slightly worse (likely the cost of redirecting gradient mass toward 6 cases that the model already "wants" to solve last). This is the expected sign — the question is whether vol_p crosses below SOTA's curve by EP2/EP3.

EP2 ETA ≈ +95 min. Will report next.

morganmcg1:
ADVISOR: EP1 acknowledged — continue under Option-4 framing.

30.2033% misses the <30% gate by 0.20pp, which is well within EP1 measurement noise (RNG, data-loader ordering, and the OOD-upweighted batches add a ~0.3-0.5pp jitter at this point in training). The gate is advisory; the substantive read is:

- **vol_p=18.0102% essentially matches SOTA EP1 (17.787%, +0.22pp).** This is the primary diagnostic per Issue #717 — the OOD-neighbor upweighting is not visibly hurting global vol_p trajectory in the early phase, which is what we need to see before judging whether the targeted upweight helps OOD test cases at EP4.
- The 0.20pp val_abupt gap is consistent with the expected sample re-weighting cost on aggregate metrics; we accept this in exchange for the OOD-targeted gradient signal.
- Arm A multiplier=3.0 with K=4 nearest-train-neighbor upweighting (run_184, run_249, run_310, run_416, run_44, run_484) is the configuration we want measured cleanly to EP4.

Continue to EP2. Gate is <16% val_abupt. The decisive comparison is EP4 test_vol_p vs PR #880 ensemble baseline; we are not optimizing for the val_abupt gate here, vol_p is. Post EP2 with full per-target + the OOD-batch hit-rate confirmation.

morganmcg1:
STUDENT thorfinn:

## Arm A — EP2 results

EP2 finished at step 21729 (rank 0 `dz1pjkhw`, runtime ≈ 3.04h).

| Metric | Arm A EP2 | SOTA EP2 (`ghh0s4ne`) | Δ |
|--------|----------:|----------------------:|---:|
| val_abupt | 8.4952% | 8.1482% | +0.347 |
| val_surface_p | 5.6445% | 5.4090% | +0.236 |
| val_volume_p | **5.2540%** | **4.9559%** | **+0.298** |
| val_wall_shear | 9.5485% | 9.1903% | +0.358 |

EP2 kill gate (<16% val_abupt): **passed** (8.50% << 16%).

**OOD weighting hit-rate (full training so far, 22,107 batches):**
- Batches with ≥1 OOD-neighbor sample: 1,489 (6.7%) — matches the expected (6/354)·4 ≈ 6.78% for batch_size=4 over 6 OOD-neighbor train cases.
- Unique `ood/multiplier` value: `{3}` — no leakage of `1.0` defaults.

**Diagnostic read:**
- vol_p is **+0.298 pp worse** than SOTA at EP2. The hypothesis predicted that OOD-neighbor upweighting would lower vol_p relative to SOTA at matched epochs once the bulk in-distribution loss had been absorbed; that is not happening yet — every target component (surface_p, vol_p, ws) is uniformly ~0.3 pp behind. The pattern is consistent with a small uniform regression caused by redirecting gradient mass to 6 train cases, with no targeted vol_p improvement showing up.
- val_abupt EP2 trajectory (Arm A vs SOTA): 30.20→8.50 vs 28.63→8.15 — gap roughly steady at ~0.3 pp. EP3 is the decisive checkpoint to see whether vol_p crosses below SOTA's 4.27% as the model refines OOD geometries; current trajectory does not look promising.

## Budget status

- Runtime now: 3.27h (step 22049, EP3 just started, vol_points scheduled to 32768).
- Train cap: `train_timeout_minutes=270` (4.5h); val reserve `val_budget_minutes=90`.
- Step rate inside EP3: ≈1.87 it/s (~unchanged from EP1/EP2). Remaining EP3 steps to 32593 ≈ 94 min.
- Projected EP3 finish: runtime ≈ 4.83h ≈ 290 min — **just over the 270-min train cap**, will likely be a forced mid-EP3 val rather than a clean EP3 boundary.
- EP4 not reachable.

Will report EP3 (or forced-val) results next. ETA ≈ 90 min from now.

morganmcg1:
**ADVISOR EP2 GATE CHECK — ood-weighting**

W&B run \`dz1pjkhw\` confirmed at step 22,248 (EP3 just started). Student metrics verified.

| Metric | Arm A EP2 | SOTA EP2 (`ghh0s4ne`) | Δ |
|--------|----------:|----------------------:|---:|
| **val_abupt** | **8.4952%** | 8.1482% | **+0.35pp** |
| vol_p | 5.254% | 4.956% | +0.30pp |
| surf_p | 5.644% | 5.409% | +0.24pp |
| wall_shear | 9.548% | 9.190% | +0.36pp |

**EP2 gate (<16%): PASS ✅** — 8.50% << 16%, margin 7.50pp

**Assessment:** The +0.35pp uniform regression across all channels at EP2 confirms your diagnostic: OOD upweighting is not yet targeting vol_p specifically — it's a mild gradient dilution across all fields. The vol_p advantage from the OOD-neighbor hypothesis hasn't materialized yet. This is the expected null behavior if 6 OOD-neighbor training cases (6.7% of batches) carry insufficient signal to bend the overall trajectory at EP2.

EP3 is the decisive checkpoint: SOTA vol_p at EP3 was ~4.27%. If Arm A's vol_p stays above ~4.56% (SOTA+0.3pp) at EP3, the hypothesis is not working. If vol_p dips below ~4.56% while surf_p/ws remain at parity, the targeted OOD recovery is showing up.

**Budget alert:** Your projection of EP3 finishing at ~290 min (over the 270-min train cap) is correct. A forced mid-EP3 val is acceptable — post whatever you get before the timeout. The partial EP3 trajectory is still informative.

**EP3 gate: val_abupt < 8.0%** — given current trajectory and decent descent rate (~-3.42pp/1k steps slope), EP3 should clear this gate (~4.5 min/1k steps × remaining EP3 steps). Post results when forced-val fires.

morganmcg1:
## Gate Check — EP1 (4-ep screen)

**Run:** `dz1pjkhw` (ood-weighting, OOD cases 3× upweight)
**Step:** 10,864 (EP1.00)

| Metric | Value | Gate | Status |
|--------|-------|------|--------|
| `val_primary/abupt_axis_mean_rel_l2_pct` | **30.2033%** | ≤30.00% | ❌ FAIL (+0.2033pp) |

EP1 gate marginally failed by 0.2033pp.

**However:** EP2 is already logged at step=21,728 with abupt=**8.4952%** — well within the EP2 gate (≤16%). The run has recovered strongly and training is currently at step=26,112 (EP~2.40).

**Advisor decision:** Given the borderline nature of the EP1 failure (0.2033pp over threshold) and the strong EP2 result (8.4952%), I am allowing this run to continue. The OOD upweighting hypothesis is still valid to test. Please continue training and report back at EP3 (gate: <8.0%) and EP4 (gate: <6.9%).

**Next gate check:** EP3 at step=32,592 — target abupt < 8.0%

morganmcg1:
STUDENT thorfinn:

## Arm A — pre-timeout status

**Runtime:** 262.0 min (cap = 270 min train + 90 min val/test budget). Step 29,523 / EP3 step 32,593.

The training loop will hit the 270-min train cap in ~8 min, before reaching a clean EP3 boundary. Per the codebase (`train.py:1305-1316`), this triggers **forced mid-epoch validation + final test evaluation** under the 90-min val budget.

**Next deliverable:** post forced-val + final test metrics (incl. test_vol_p and per-OOD-case test_vol_p) once the run completes, then submit for review.

morganmcg1:
ADVISOR: ## Final Verdict — thorfinn/ood-sample-weighting (PR #888)

**W&B run:** \`thorfinn/ood-weight-3x-rank0\` (state=finished, step=30,454)

### Val history (forced mid-EP3 timeout)

| Checkpoint | Step | val_abupt | Gate | Status |
|-----------|------|-----------|------|--------|
| EP1 | 10,864 | 30.2033% | ≤30% | ❌ FAIL (waived) |
| EP2 | 21,728 | 8.4952% | ≤16% | ✅ PASS |
| Forced timeout val | 30,454 | **7.4232%** | — | (mid-EP3) |

### Final test metrics (from W&B summary)

| Metric | Thorfinn Arm A | SOTA (#823, `ghh0s4ne`) | Δ |
|--------|---------------:|------------------------:|---|
| **test_abupt** | **8.6935%** | **6.4407%** | **+2.25pp** |
| surf_p | 4.573% | — | — |
| vol_p | 12.609% | — | — |
| wall_shear | 8.060% | — | — |

**Best val:** 7.4232% (forced mid-EP3, vs SOTA val 6.4407% → +0.98pp above baseline)

### Assessment: NEGATIVE

This experiment did **not** beat the baseline. The forced timeout val of 7.42% is ~0.98pp above the SOTA val of 6.44%. Test abupt came in at 8.69% vs SOTA 6.44% — a significant gap.

More importantly, the OOD-weighting hypothesis itself showed **no benefit** for the OOD test geometries:
- Vol_p at EP2: 5.254% vs SOTA 4.956% (+0.30pp — worse, not better)
- The hypothesis was that upweighting nearest-train-neighbors of OOD test cases would improve generalization; the data shows no evidence of this

**Root cause analysis:** The 6 train IDs upweighted (multiplier=3.0) are reasonable KNN proxies but may not be close enough to the test OOD geometries in geometry-space for the reweighting to transfer. Additionally, 3× upweighting may not be strong enough, or the model may already be fitting these regions well.

**Closing this PR.** The direction is not promising enough to warrant another round — the OOD vol_p metric worsened rather than improved. GPU time is better allocated to the xattn architecture sweep currently in-flight.

Thank you thorfinn for the thorough investigation and the clear pre-timeout status report.

PR #907, opened 2026-05-09
On-the-fly approx SDF feature for vol queries: geometry distance conditioning

morganmcg1:
## Hypothesis

The SOTA model (PR #823 surf→vol xattn) gives vol tokens access to surface geometry through cross-attention. But the vol tokens' input features contain only 3D spatial coordinates (x, y, z) — they have no explicit information about **how far they are from the surface**. The vol tokens near the surface (where pressure gradients are steepest) are treated identically to tokens far in the freestream by the xattn query.

**Key insight from PR #767:** 4 OOD test cases (run_133, run_226, run_203, run_158) account for 92% of squared test_vol_p deviation. These cases likely have unusual near-surface pressure fields — regions where the vol query needs to "know" it is close to the geometry to correctly interpret the surface conditioning signal.

This PR adds an **on-the-fly approximate SDF (signed distance to surface) feature** to each volume query point. For each vol point `v`, we compute the distance to the nearest surface point `s` (already loaded in batch). This gives the model one extra scalar feature per vol token: `d = min_k ||v - s_k||_2`.

**Why on-the-fly (no precomputed file):** Issue #803 blocks precomputed `volume_sdf.npy`. A brute-force kNN distance computation over 65k surf × 65k vol points is expensive, but we can use a **stratified approximation**: subsample surface to 4096 random points per batch, compute pairwise distances (4096×65536 on GPU, ~25ms), take row-wise min. This is exact in expectation and avoids any precomputed dependency.

**How the feature enters:** Expand the input feature dimension by 1. Concatenate `d_norm = log(1 + d / median_d)` (log-normalized distance) to each vol token's input before the position-encoding projection. The network learns to use this distance to modulate how it integrates the surface conditioning from xattn.

## Instructions

1. **Add a new flag** `--use-vol-approx-sdf` (bool, default False) to `train.py`.

2. **Add an `ApproxSDF` helper** in the dataset/model preprocessing:
   ```python
   def compute_approx_sdf(vol_coords, surf_coords, surf_subsample=4096, device=None):
       """Compute approximate distance from vol points to nearest surface point.
       
       Args:
           vol_coords: (B, N_vol, 3) float tensor
           surf_coords: (B, N_surf, 3) float tensor (full 65k surface)
           surf_subsample: number of surface points to use (default 4096)
       Returns:
           d_approx: (B, N_vol, 1) float tensor — log-normalized distance
       """
       B, N_vol, _ = vol_coords.shape
       # Randomly subsample surface
       idx = torch.randperm(surf_coords.shape[1], device=surf_coords.device)[:surf_subsample]
       surf_sub = surf_coords[:, idx, :]  # (B, surf_subsample, 3)
       
       # Pairwise distances using broadcasting
       # vol: (B, N_vol, 1, 3), surf_sub: (B, 1, surf_subsample, 3)
       diff = vol_coords.unsqueeze(2) - surf_sub.unsqueeze(1)  # (B, N_vol, surf_subsample, 3)
       dists = diff.norm(dim=-1)  # (B, N_vol, surf_subsample)
       d_min = dists.min(dim=-1, keepdim=True).values  # (B, N_vol, 1)
       
       # Log-normalize for stability
       median_d = d_min.reshape(B, -1).median(dim=-1).values[:, None, None]  # (B, 1, 1)
       d_norm = torch.log1p(d_min / (median_d + 1e-6))  # (B, N_vol, 1)
       return d_norm
   ```

3. **Inject** the SDF feature into the vol token input before or right after the positional encoding (concatenate to input features, add a learnable linear projection to map back to `hidden_dim` after concatenation, or simply add a single extra input dimension and let the existing input projection handle it — the latter is simpler).

   Simplest implementation: in the vol input preprocessing (before the backbone), concatenate `d_norm` to the vol point features. This requires bumping the vol input projection layer's `in_features` by 1. Use `nn.Linear(original_in + 1, hidden_dim)` for the vol input projection when `--use-vol-approx-sdf` is True.

4. **Batch memory note:** At batch_size=4, N_vol=65536, surf_subsample=4096: the pairwise diff tensor is `4 × 65536 × 4096 × 3 × 4 bytes ≈ 12 GB`. This is too large. Use a loop over vol chunks or reduce surf_subsample:
   - **Recommended:** surf_subsample=512 (512×65536 diff = 128M elements = ~512 MB per batch) — much cheaper while still providing a good approximation.
   - Process in vol chunks if needed: split N_vol into 8 chunks of 8192, compute distances per chunk.

5. **Full 4-epoch screen command:**
```bash
  --agent askeladd \
  --no-compile-model \
  --use-surf-to-vol-xattn \
  --use-vol-approx-sdf \
  --wandb-group askeladd-vol-approx-sdf \
  --wandb-name askeladd/vol-approx-sdf-screen
```

6. **Kill gates:**
   - EP4 (~step 38,028): **<6.9%** to proceed to Phase 2

7. **If Phase 1 passes:** Run full 13-epoch schedule replacing `--epochs 4 --lr-cosine-t-max 13` with `--epochs 13 --lr-cosine-t-max 13` and `--vol-points-schedule "0:16384:3:32768:6:49152:9:65536"`.

## Baseline to Beat

|---|---|
| val_abupt | **6.4407%** |
| test_abupt | 7.6992% |

**Note:** The approx SDF computation must run on GPU (pass surf/vol coords to the same device). Avoid CPU fallback — even surf_subsample=512 is fast on GPU but slow on CPU.


morganmcg1:
ADVISOR: Closing PR #907 because ADVISOR: EP1 kill gate FAILED — closing this hypothesis.

**Screen result (W&B run `0h934b68`, group `askeladd-vol-approx-sdf`):**
- val_primary/abupt_axis_mean_rel_l2_pct = **32.20%** at step 10864 (gate: ≤30%)
- early_stop/triggered = 1
- early_stop/reason = `kill threshold failed at step 10864: val_primary/abupt_axis_mean_rel_l2_pct=32.2043 did not satisfy step>=10500 val_primary/abupt_axis_mean_rel_l2_pct<30`
- Component breakdown at EP1: surface_pressure=24.81%, volume_pressure=19.86%, wall_shear=35.57%

**Why this hypothesis didn't pan out:**
The on-the-fly approx-SDF feature (kNN-based distance-to-surface scalar with surf_subsample=512) failed to deliver an EP1 boost over the no-SDF baseline, and the wall_shear axis is meaningfully worse than typical EP1 (~35.6% vs ~32% on healthy runs). Two likely contributors:
1. The subsampled kNN (k=512 surface points) introduces a stochastic, low-resolution distance estimate that is *less informative* than the volume tokens' natural geometry context — adding noise to the query feature rather than sharpening it.
2. Concatenating an extra scalar widens the volume token MLP without a warm-up; at EP1 the new dim hasn't learned to be useful yet, so it competes with finite query capacity.

A higher-fidelity SDF (precomputed exact distance field) might still be worth trying later, but that's a different cost regime. We don't burn another full screen on this variant.

**Next direction queued separately** — askeladd will be reassigned to a fresh hypothesis. Thanks for the clean kill gate and quick screen turnaround..

PR #961, opened 2026-05-10
fern: geometry-conditioned Q-bias for surf→vol xattn

morganmcg1:
## Hypothesis

**H3 — Geometry-Conditioned Q-Bias for surf→vol cross-attention.**

Volume queries currently attend to surface keys with a fixed Q-projection that has no awareness of *which* car this is. Two cars with very different rear geometries (e.g. notchback vs. estate) generate the same Q distribution from a given vol point, and the model has to disentangle "where am I in this car" purely from the keys/values. We hypothesize this is a major contributor to the OOD volume_pressure failure modes (run_133, run_226, run_203, run_158 contribute ~92% of squared test_vol_p deviation).

We add a global geometry descriptor — the mean over surface hidden states — passed through a tiny MLP, producing an additive bias that conditions the volume-side Q-projection on global geometry. The bias is **zero-initialized** so the model starts at exactly the current SOTA and only deviates if the bias helps.

This is orthogonal to:
- STRING centroid-RoPE (PR #955) and STRING sigma sweep (PR #960)
- Y-flip aug (PR #957)
- Vol-pressure aux head (PR #958)
- SDF-zone masking (PR #953)
- Coord-vol head (PR #959)
- Closed: post-xattn capacity (#891/#893/#906), pre-xattn vol self-attn (#929), Manifold Mixup (#952), GradNorm full-mode (#942)

Inspired by Equivariant Neural Fields literature where global shape codes condition local feature lookups.

## Implementation

**File: `model.py`** — find the surf→vol cross-attention block (gated by `--use-surf-to-vol-xattn`) and add a new gated module immediately *before* the Q-projection of that xattn:

```python
# In __init__ of the parent module:
self.use_geom_q_bias = use_geom_q_bias  # plumb through CLI
if self.use_geom_q_bias:
    self.geom_q_mlp = nn.Sequential(
        nn.Linear(hidden_dim, hidden_dim // 2),
        nn.GELU(),
        nn.Linear(hidden_dim // 2, hidden_dim),
    )
    # Zero-init the FINAL linear so the bias starts at zero
    nn.init.zeros_(self.geom_q_mlp[-1].weight)
    nn.init.zeros_(self.geom_q_mlp[-1].bias)

# In forward, immediately before the surf->vol xattn:
# surf_h: [B, N_surf, D], vol_h: [B, N_vol, D]
if self.use_geom_q_bias:
    geom_desc = surf_h.mean(dim=1)            # [B, D]
[...]

PR #883, opened 2026-05-09
Geometry-aware positional bias on surf→vol cross-attention queries

morganmcg1:
# Geometry-aware positional bias on surf→vol cross-attention queries

**Hypothesis:** PR #823's surf→vol cross-attention adds raw `nn.MultiheadAttention(Q=volume, K=V=surface)` after the backbone — but the volume queries arrive with no explicit geometry signal at the xattn layer. Adding a **STRING-derived positional bias on the volume queries (and surface key/values)** lets the xattn condition more sharply on *which surface region* each volume point should attend to. This is a low-cost extension of the SOTA-establishing xattn module on Theme 1 (the highest-priority research axis post-PR #823).

The Transolver backbone already consumes STRING-positional-encoded surface and volume coordinates through its first encoder layer. By the time the xattn block runs (post-backbone, post-LayerNorm), the volume hidden states have been transformed many times and may have lost direct geometric specificity. Re-injecting a STRING-style positional encoding *at the xattn input* gives the cross-attention a clean geometric channel — like rotary positional encoding does for query-key dot products in language transformers.

**Why this is orthogonal to the other in-flight xattn experiments:**
- PR #878 (alphonse): more heads (8/16) — capacity ablation
- PR #879 (frieren, just closed): two-layer xattn — depth ablation
- This PR (tanjiro): geometry-aware queries — input-conditioning ablation

If any of these win independently, they likely compose (orthogonal axes).

---

## Baselines (must beat)

| Metric | Value | Source |
|---|---|---|
| **Single-model SOTA val_abupt** | **6.4407%** | PR #823 (run `ghh0s4ne`) |
| Test val_abupt | 7.6992% | PR #823 |
| Test vol_pressure | 11.6704% | PR #823 |

The new run **MUST** beat 6.4407% on `val_primary/abupt_axis_mean_rel_l2_pct` at EP4 (or EP3 best ckpt) to merge.

---

### Add STRING positional bias on the volume queries inside the xattn block

In `model.py`, the existing block (around line 433–546) is:

```python
if use_surf_to_vol_xattn:
    self.surf_to_vol_xattn = nn.MultiheadAttention(
        embed_dim=n_hidden, num_heads=n_head, batch_first=True
    )
    self.surf_to_vol_xattn_norm = nn.LayerNorm(n_hidden, eps=1e-6)
    nn.init.zeros_(self.surf_to_vol_xattn.out_proj.weight)
    nn.init.zeros_(self.surf_to_vol_xattn.out_proj.bias)
...
[...]

PR #958, opened 2026-05-10
Dedicated vol_p aux decoder head: 2-arm loss-weight sweep

morganmcg1:
## Hypothesis

The current SOTA model uses a single shared decoder head for all output channels: surface_pressure (cp), wall_shear (tau_x, tau_y, tau_z), and volume_pressure — 5 channels total from one MLP. This sharing creates a tension: surface physics (cp distribution determined by boundary layer attachment) and volume physics (pressure field driven by far-field Bernoulli) have fundamentally different spatial structure and predictability.

We hypothesize that adding a **dedicated auxiliary decoder head for volume_pressure** — an independent MLP branch that takes the volume-point latent tokens and outputs a refined vol_p estimate — will improve volume_pressure prediction quality and reduce the OOD test gap. The auxiliary head allows the model to specialize capacity for the harder prediction task (vol_p test error 11.67% vs surface_pressure 3.85%) without degrading surface predictions.

This is motivated by:
1. The 3.0× OOD gap on vol_p (val 3.86% vs test 11.67%) vs ~1.0× gap on surface_pressure — the tasks have very different difficulty profiles
2. Multi-task learning theory: when loss magnitudes and gradient scales differ between tasks, separate output heads with independent gradient paths reduce interference
3. The vol→vol cross-attention skip in PR #823 already provides dedicated volume features — an aux head doubles down on this architecture principle

Addresses Issue #717 (vol_p OOD gap). No SDF data needed.

## Implementation Instructions

**Student: implement a dedicated vol_p auxiliary decoder head in `model.py`.**

### Architecture change

After the final Transolver layer, instead of a single output MLP for all channels, split into two branches:

```python
# In your model's __init__:
# Shared output head (surface + volume combined, as before)
self.output_head = nn.Sequential(
    nn.Linear(hidden_dim, hidden_dim),
    nn.SiLU(),
    nn.Linear(hidden_dim, 4)   # [cp, tau_x, tau_y, tau_z]
)

# Dedicated vol_p head (takes volume tokens only)
self.vol_pressure_head = nn.Sequential(
    nn.Linear(hidden_dim, hidden_dim // 2),
    nn.SiLU(),
    nn.Linear(hidden_dim // 2, hidden_dim // 4),
    nn.SiLU(),
    nn.Linear(hidden_dim // 4, 1)   # [vol_p]
)
```

### Forward pass change

```python
# surface_tokens: [B, N_s, hidden_dim]
# volume_tokens:  [B, N_v, hidden_dim]

[...]

PR #971, opened 2026-05-10
frieren: learned distance RPB on surf→vol cross-attention

morganmcg1:
## Hypothesis

**Learned Relative Position Bias (RPB) on surf→vol cross-attention logits** will improve the model's geometric locality — bringing the attention mechanism in line with the physical intuition that nearby surface points are more informative for each volume token.

Currently the surf→vol cross-attention (`nn.MultiheadAttention(embed_dim=512, num_heads=4)`) has no explicit positional signal: it learns all locality entirely from data. We add a tiny MLP that maps the Euclidean distance between each (vol_token_i, surf_point_j) pair to a scalar bias, added directly to the attention logits before softmax:

```
A_ij = (Q_i · K_j)/√d + f(dist(vol_i, surf_j))
```

This is the 3-D geometric analogue of **ALiBi** (Press et al., 2022, ICLR; "Train Short, Test Long") — learned relative-distance biases that let attention heads develop smooth locality without any learned PE parameters. In our case the signal is 3-D Euclidean distance, not sequence position.

**Why this differs from closed experiments:**
- PR #883: added a *static, pre-computed global PE offset* to all vol tokens — not per-pair, not input-adaptive.
- PR #961 (H3): conditioned the *Q vectors globally* from a geometry mean-pool — different token context, not per-pair logit bias.
- This PR: adds a *per-pair (vol_i, surf_j) bias to attention LOGITS* — operates at the logit level, 17 learnable parameters, zero-init so it starts as vanilla xattn and learns to attend locally.

Prior art: ALiBi (Press et al., ICLR 2022); Relative Position Encodings in Transformer variants (Shaw et al., 2018; T5, Raffel 2020).

### 1. New flag in `train.py`

Add to `add_argument` calls (alongside existing `--use-surf-to-vol-xattn`):

```python
parser.add_argument("--use-xattn-distance-rpe", action="store_true", default=False,
                    help="Add learned distance-based RPB to surf->vol cross-attention logits.")
```

Pass to `Model.__init__` and store as `self.use_xattn_distance_rpe`.

### 2. Model changes in `model.py`

In `__init__`, after the existing surf→vol xattn block:

```python
if self.use_xattn_distance_rpe:
    self.xattn_rpe_mlp = nn.Sequential(
        nn.Linear(1, 16),
        nn.SiLU(),
        nn.Linear(16, 1),
    )
    # Zero-init final layer so training starts identical to baseline
    nn.init.zeros_(self.xattn_rpe_mlp[-1].weight)
    nn.init.zeros_(self.xattn_rpe_mlp[-1].bias)
```

In `forward`, immediately before the surf→vol `nn.MultiheadAttention` call, compute the bias and pass it as `attn_mask`:

```python
if self.use_xattn_distance_rpe:
    # vol_xyz: [B, N_vol, 3] — vol token positions
    # surf_xyz: [B, N_surf, 3] — surface point positions
    # Compute pairwise L2 distance
    diff = vol_xyz.unsqueeze(2) - surf_xyz.unsqueeze(1)  # [B, N_vol, N_surf, 3]
    dist = diff.norm(dim=-1, keepdim=True)               # [B, N_vol, N_surf, 1]
    rpe = self.xattn_rpe_mlp(dist).squeeze(-1)           # [B, N_vol, N_surf]

    # nn.MultiheadAttention expects attn_mask: [tgt_len, src_len] or
    # [B*num_heads, tgt_len, src_len]. Broadcast across heads:
    B, N_vol, N_surf = rpe.shape
    # attn_mask shape: [B*num_heads, N_vol, N_surf]
    rpe_expanded = rpe.unsqueeze(1).expand(
        B, self.surf_to_vol_xattn.num_heads, N_vol, N_surf
    ).reshape(B * self.surf_to_vol_xattn.num_heads, N_vol, N_surf)

    # Pass as additive attn_mask (not key_padding_mask)
    vol_out = self.surf_to_vol_xattn(
        query=vol_hidden,   # [N_vol, B, d]
        key=surf_hidden,    # [N_surf, B, d]
        value=surf_hidden,
        attn_mask=rpe_expanded,
    )[0]
else:
    vol_out = self.surf_to_vol_xattn(
        query=vol_hidden,
        key=surf_hidden,
        value=surf_hidden,
    )[0]
```

**Memory note:** At EP3+ vol_points can reach 65,536. The `dist` tensor `[B, N_vol, N_surf]` with B=1, N_vol=65536, N_surf=~40000 ≈ 10 GB fp32 — too large. Use chunked computation along N_vol:

```python
# Chunked version if N_vol * N_surf > 1e9 / 4 (fp32 bytes)
CHUNK = 2048
rpe_chunks = []
for i in range(0, N_vol, CHUNK):
    diff_chunk = vol_xyz[:, i:i+CHUNK, :].unsqueeze(2) - surf_xyz.unsqueeze(1)
    dist_chunk = diff_chunk.norm(dim=-1, keepdim=True)
    rpe_chunks.append(self.xattn_rpe_mlp(dist_chunk).squeeze(-1))
rpe = torch.cat(rpe_chunks, dim=1)  # [B, N_vol, N_surf]
```

For the `attn_mask` with `nn.MultiheadAttention` in batch_first=False mode, you may need to pass attention in a loop over the batch or use a custom attention impl. The simplest path: if `use_xattn_distance_rpe`, replace the single `nn.MultiheadAttention` call with a manual scaled-dot-product attention:

```python
# Manual SDPA with RPE bias (chunked or full)
# vol_hidden: [N_vol, B, d]  →  need [B, N_vol, d]
q = vol_hidden.transpose(0, 1)   # [B, N_vol, d]
k = surf_hidden.transpose(0, 1)  # [B, N_surf, d]
v = k

# Multi-head split
d = q.shape[-1]
H = self.surf_to_vol_xattn.num_heads  # 4
d_h = d // H
q = q.reshape(B, N_vol, H, d_h).permute(0, 2, 1, 3)  # [B, H, N_vol, d_h]
k = k.reshape(B, N_surf, H, d_h).permute(0, 2, 1, 3)
v = v.reshape(B, N_surf, H, d_h).permute(0, 2, 1, 3)

attn_logits = (q @ k.transpose(-2, -1)) / (d_h ** 0.5)  # [B, H, N_vol, N_surf]
attn_logits = attn_logits + rpe.unsqueeze(1)             # broadcast over H
attn_weights = F.softmax(attn_logits, dim=-1)

out = (attn_weights @ v)                                  # [B, H, N_vol, d_h]
out = out.permute(0, 2, 1, 3).reshape(B, N_vol, d)       # [B, N_vol, d]
# Apply the xattn out_proj
out = self.surf_to_vol_xattn.out_proj(out)               # [B, N_vol, d]
vol_out = out.transpose(0, 1)                            # [N_vol, B, d]
```

This avoids PyTorch's 3D attn_mask limitation and is cleaner.

**DDP note:** Set `find_unused_parameters=True` in `DistributedDataParallel` so that if any run is launched without `--use-xattn-distance-rpe`, the `xattn_rpe_mlp` params don't cause DDP errors.

### 3. Run command (4-epoch screen)

```bash
  --use-surf-to-vol-xattn \
  --use-xattn-distance-rpe \
  --epochs 4 \
  --wandb-group frieren-xattn-distance-rpe
```

Use all other defaults (lr, batch size, model dims) unchanged from baseline.

## Baseline (single-model SOTA)

| Metric | Value |
|---|---|
| val_primary/abupt_axis_mean_rel_l2_pct | **6.4407%** |
| val_primary/volume_pressure_rel_l2_pct | 3.8557% |
| val_primary/surface_pressure_rel_l2_pct | 2.1888% |
| W&B run | `ghh0s4ne` |
| PR | #823 |

```
Reproduce: cd target/ && python train.py --use-surf-to-vol-xattn --vol-points-schedule "0:16384:1:32768:2:49152:3:65536" --epochs 13
```

## Gates (4-epoch screen)

| Epoch | abupt% | vol_p% |
|---|---|---|
| EP1 | ≤30% | — |
| EP3 | ≤8.0% | ≤5.0% |
| EP4 (final) | ≤7.5% | — |

If all gates pass, post your best EP4 metrics and I will promote to a full 13-epoch run.

## Expected parameter overhead

~17 additional parameters:
- `Linear(1→16)`: 16 weights + 16 bias = 32
- `Linear(16→1)`: 16 weights + 1 bias = 17
- **Total: 49 params** (effectively zero overhead)

Zero-init on the output layer means the model starts as vanilla surf→vol xattn and gradually learns the distance prior.


morganmcg1:
ADVISOR: Hi frieren — please start your 4-epoch screen now.

**Hypothesis:** Adding a learned per-pair distance relative position bias (RPB) to the surf→vol cross-attention logits. This gives the model explicit geometric locality — nearby surface points get a learned attention boost for each volume token, analogous to ALiBi but in 3D Euclidean space.

**Implementation checklist:**

**Step 1 — New flag in `train.py`:**
```python
parser.add_argument("--use-xattn-distance-rpe", action="store_true", default=False,
                    help="Add learned distance-based RPB to surf→vol cross-attention logits.")
```
Pass to `Model.__init__` and store as `self.use_xattn_distance_rpe`.

**Step 2 — Model changes in `model.py`:**

In `__init__`, after existing surf→vol xattn block:
```python
if self.use_xattn_distance_rpe:
    self.xattn_rpe_mlp = nn.Sequential(
        nn.Linear(1, 16),
        nn.SiLU(),
        nn.Linear(16, 1),
    )
    # Zero-init final layer so training starts identical to baseline
    nn.init.zeros_(self.xattn_rpe_mlp[-1].weight)
    nn.init.zeros_(self.xattn_rpe_mlp[-1].bias)
```

In `forward`, replace the surf→vol `nn.MultiheadAttention` call with manual SDPA + RPE bias (chunked to avoid OOM at large vol_pts):

```python
if self.use_xattn_distance_rpe:
    # vol_xyz: [B, N_vol, 3], surf_xyz: [B, N_surf, 3]
    B, N_vol, _ = vol_xyz.shape
    N_surf = surf_xyz.shape[1]
    
    # Chunked distance computation (avoids 10GB tensor at N_vol=65536)
    CHUNK = 2048
    rpe_chunks = []
    for i in range(0, N_vol, CHUNK):
        diff_chunk = vol_xyz[:, i:i+CHUNK, :].unsqueeze(2) - surf_xyz.unsqueeze(1)
        dist_chunk = diff_chunk.norm(dim=-1, keepdim=True)
        rpe_chunks.append(self.xattn_rpe_mlp(dist_chunk).squeeze(-1))
    rpe = torch.cat(rpe_chunks, dim=1)  # [B, N_vol, N_surf]
    
    # Manual multi-head SDPA with RPE bias
    d = vol_hidden.shape[-1]
    H = self.surf_to_vol_xattn.num_heads  # 4
    d_h = d // H
    
    q = vol_hidden.transpose(0, 1)    # [B, N_vol, d]
    k = surf_hidden.transpose(0, 1)   # [B, N_surf, d]
    v = k
    
    q = q.reshape(B, N_vol, H, d_h).permute(0, 2, 1, 3)   # [B, H, N_vol, d_h]
    k = k.reshape(B, N_surf, H, d_h).permute(0, 2, 1, 3)
    v = v.reshape(B, N_surf, H, d_h).permute(0, 2, 1, 3)
    
    attn_logits = (q @ k.transpose(-2, -1)) / (d_h ** 0.5)  # [B, H, N_vol, N_surf]
    attn_logits = attn_logits + rpe.unsqueeze(1)              # broadcast over H
    attn_weights = F.softmax(attn_logits, dim=-1)
    
    out = (attn_weights @ v)                                   # [B, H, N_vol, d_h]
    out = out.permute(0, 2, 1, 3).reshape(B, N_vol, d)
    out = self.surf_to_vol_xattn.out_proj(out)
    vol_out = out.transpose(0, 1)                             # [N_vol, B, d]
else:
    vol_out = self.surf_to_vol_xattn(
        query=vol_hidden,
        key=surf_hidden,
        value=surf_hidden,
    )[0]
```

**DDP note:** Set `find_unused_parameters=True` in `DistributedDataParallel` so that runs without `--use-xattn-distance-rpe` don't cause DDP errors on `xattn_rpe_mlp`.

**Step 3 — 4-epoch screen command:**
```bash
  --use-surf-to-vol-xattn \
  --use-xattn-distance-rpe \
  --epochs 4 \
  --wandb-group frieren-xattn-distance-rpe \
  --wandb-name frieren/xattn-dist-rpe-4ep
```
(Use all other defaults: lr, batch size, model dims unchanged from baseline.)

**13-epoch run command (if EP3 gates pass):**
```bash
  --use-surf-to-vol-xattn \
  --use-xattn-distance-rpe \
  --epochs 13 \
  --wandb-group frieren-xattn-distance-rpe \
  --wandb-name frieren/xattn-dist-rpe-13ep
```

**Gate schedule:**
| Epoch | val_abupt% gate | val_vol_p% gate |
|---|---|---|
| EP1 | ≤30% | — |
| EP3 | ≤8.0% | ≤5.0% |
| EP4 (final screen) | ≤7.5% | — |
| (if promoted to 13ep) EP5 | ≤7.5% | — |
| (if promoted to 13ep) EP8 | ≤7.0% | — |
| (if promoted to 13ep) EP13 | <6.4407% target | — |

**Current baseline (target to beat):**
| Metric | Value |
|---|---|
| val_abupt_axis_mean_rel_l2_pct | **6.4407%** |
| val_volume_pressure_rel_l2_pct | 3.8557% |
| val_surface_pressure_rel_l2_pct | 2.1888% |
| W&B run | `ghh0s4ne` (PR #823) |

**Current val SOTA:**
| Metric | Value |
|---|---|
| val_abupt_axis_mean_rel_l2_pct | **6.3218%** |
| val_volume_pressure_rel_l2_pct | 3.9367% |
| W&B run | `29nohj67` (PR #958, EP9) |

**Expected param overhead:** ~49 params total (Linear(1→16): 32, Linear(16→1): 17). Effectively zero. Zero-init on output layer means model starts as vanilla xattn and learns the distance prior from scratch.

Please post your W&B run ID and per-epoch metrics (abupt, vol_p, surf_p, wall_shear) in a comment after each epoch. Kill immediately if EP1 > 30% or EP3 > 8.0% abupt.

Good luck!

morganmcg1:
STUDENT frieren:

## Speed/budget concern before running 4-epoch screen

I have the implementation complete and **correctness-verified** (sanity check: zero-init RPE produces bit-identical output to the baseline surf->vol xattn, with `max_abs_diff=0.0` on both surface and volume preds). The current code in `model.py:496-602` uses chunked manual SDPA with `F.scaled_dot_product_attention(attn_mask=rpe)` and per-chunk gradient checkpointing.

However, microbenchmarks on the assigned hardware (NVIDIA RTX PRO 6000 Blackwell, SM 12.0, 96 GB) show the implementation is ~7x slower than baseline at full curriculum scale, and the 4-epoch screen as specified will not fit the 360-min training budget. Details:

### Per-iter wall time (full 5-layer model, B=4 per rank, surf_pts=65536, autocast bf16, chunk_size=1024, single GPU)

| Stage | N_vol | Baseline (Flash) | This PR (chunked SDPA + MLP RPE) |
|---|---|---|---|
| EP0 | 16384 | 0.62 s/it | **4.2 s/it** (~7x) |
| EP3 | 65536 | ~2 s/it (est.) | **15 s/it** (~7-8x) |

Per-epoch (10864 iters): EP0 alone takes ~12.7h on this PR, vs ~1.9h baseline. Full 4-epoch screen: ~40-60h projected, vs ~10h baseline. The 4.5h training budget gets us roughly 35% of EP0, well short of any gate.

### Root cause

Plain `nn.MultiheadAttention` (Flash kernel) takes ~85 ms for this surf->vol xattn at EP0; adding any per-pair additive bias forces PyTorch off the Flash path. On SM 12.0 Blackwell the alternatives have all failed:
- `F.scaled_dot_product_attention(attn_mask=float_tensor)` → falls back to **math backend**, materializes `[B,H,N_vol,N_surf]` attn logits (~34 GB at EP0). Slow.
- `xformers.ops.memory_efficient_attention(attn_bias=tensor)` → unsupported on SM 12.0 (`cutlassF requires capability <= 9.0`, `fa2/fa3` reject tensor bias).
- `torch.compile(flex_attention)` → CUDA `illegalAddress` during compile on this kernel.

So we are stuck on the math backend, which materializes the QK logits matrix and has roughly the same compute pattern as my manual chunked code. The rpe_mlp `Linear(1,16)→SiLU→Linear(16,1)` is the second largest factor — its 16-dim hidden state is 16x the size of the dist tensor and dominates memory traffic.

### Options I see

1. **Drop the screen to 1 epoch** (or even 0.5 epoch) and report partial loss trajectories. Loses the gate signal but at least produces data.
2. **Shrink the rpe MLP** to `Linear(1,4)→SiLU→Linear(4,1)` (~4x less memory traffic; ~2x speedup in this block). ~3.5 s/it at EP0 → still well over budget but closer.
3. **Use a linear ALiBi-style bias** `rpe = α·dist + β` (no MLP intermediate). Benchmarked at 1.0 s/it at EP0; full 4-epoch budget projects ~10-15h. Still over but borderline. This is a deviation from the PR but is the closest faithful-spirit variant that might run.
4. **Pre-compute the `α·d²` bias separably and fold into Q/K projections** (a low-rank distance² approximation). Lets us use the Flash kernel; no per-pair bias storage. Architecturally identical to a vanilla attention with augmented Q/K, so this is most aggressive deviation.
5. **Close as infeasible on Blackwell** and revisit when Hopper GPUs are available again.

How would you like me to proceed? Default plan if no response in ~30 min: launch (1) — let the faithful implementation run to wall-clock budget and report whatever EPs we got, so we at least have a partial loss curve to compare against baseline.

morganmcg1:
ADVISOR: Thank you for the thorough benchmarking — this is exactly the kind of diagnostic work that saves everyone time.

**Decision: Implement Option 3 (linear ALiBi-style bias) and run the full 4-epoch screen.**

Here's my reasoning:

The hypothesis we are testing is whether *relative positional encoding in the surf→vol cross-attention* improves geometric fidelity. The MLP rpe formulation is a specific parameterization of that idea — but the core question is whether any additive distance-based bias helps at all. ALiBi-style linear bias `rpe = α·dist` (scalar learnable) tests the same hypothesis with essentially zero overhead on the math backend path.

### What to implement

Replace the `rpe_mlp` `Linear(1,16)→SiLU→Linear(16,1)` with a single learnable scalar per head:

```python
# In model.py, surf→vol cross-attention block
self.rpe_alpha = nn.Parameter(torch.zeros(num_heads))  # one α per head

# In forward:
dist = torch.cdist(vol_coords, surf_coords)  # [B, N_vol, N_surf]
rpe = self.rpe_alpha[None, :, None, None] * dist.unsqueeze(1)  # [B, H, N_vol, N_surf]
# Pass as attn_mask (additive, pre-softmax)
```

Zero-init the alphas so the first forward pass is identical to baseline (no bias). This ensures EP1 convergence is unaffected by the rpe signal at initialization.

**Important**: chunk the bias computation too to control peak VRAM — but since it's a scalar multiply of the dist matrix, the backward pass should be much lighter than the MLP intermediate.

Your 1.0 s/it benchmark for the linear case at EP0 is borderline. The 4-epoch budget with vol-curriculum should be feasible since:
- EP0–EP2 run at lower vol_pts (16384, 32768) — your slow cases were all at 65536 
- EP3 at 65536 is the expensive one but it's only 1 epoch

If the linear-bias screen passes gates and shows better vol_p or abupt than baseline, I'll assign you a follow-up with the full MLP rpe on a 13-epoch run if we get Hopper access, or we'll consider the low-rank Q/K fold (Option 4) for Blackwell.

**Do not wait — implement the linear bias and launch now.** Report at EP1 gate as normal. Keep your benchmark notes in the PR body for reference when we compare results.

The 0.5-epoch partial curve from Option 1 would be scientifically useless — we can't draw conclusions from a run that hasn't reached any gate threshold.

morganmcg1:
STUDENT frieren:

## Linear ALiBi-style RPE: launched, EP0 timing update

Implemented your linear-bias variant (per-head learnable `alpha`, zero-init) and launched the 4-epoch screen at 23:00 UTC.

**Implementation summary:**
- Replaced `xattn_rpe_mlp` (Linear(1,16) → SiLU → Linear(16,1)) with `self.xattn_rpe_alpha = nn.Parameter(torch.zeros(n_head))`.
- Forward inserts `rpe = alpha.view(1,-1,1,1) * dist.unsqueeze(1)` as the `attn_mask` to `F.scaled_dot_product_attention` inside the chunked SDPA loop. Per-head broadcast over `[B, H, chunk, N_surf]`.
- Kept chunked SDPA + gradient checkpointing (necessary at EP3 — without checkpointing the per-chunk attn probs sum to >>80GB across 64 chunks of N_vol=65536).
- Sanity check passed: alpha=0 reproduces baseline xattn output (bf16 noise only, ~3.9e-3 max diff is the SDPA-math vs Flash kernel difference, not a logic bug); alpha≠0 changes output; alpha gets gradient in training.

**W&B run (rank0):** `uojjm2se` (group `frieren-xattn-distance-rpe`, name `frieren/xattn-dist-rpe-linear-4ep`)

**Observed per-iter timing at EP0 (steady-state, first 25 iters, real DDP8):**
- This PR (linear ALiBi RPE, chunk=1024): **~1.92 s/it** (5.7h projected for EP0)
- Baseline (no RPE, your earlier benchmark): 0.62 s/it
- Slowdown: **~3.1x** (was ~7x with the MLP variant)

The linear variant is ~2.2x faster than the MLP, matching the qualitative direction you predicted. However my standalone microbenchmark predicted 1.0 s/it (yours) vs 1.9 s/it (mine, real DDP). The 2x gap is partly DDP overhead (`find_unused_parameters=True` traversal + all-reduce) and partly that even the linear bias forces SDPA onto the math backend, where the per-chunk [B,H,chunk,N_surf] attn-prob materialization dominates.

### Revised wall-time projection vs 360-min budget

Assuming s/it scales ~linearly with `N_vol` (consistent with my benchmarks of the math-backend SDPA at each curriculum stage):

| Epoch | N_vol | proj s/it | proj wall | cumulative |
|---|---|---|---|---|
| EP0 | 16384 | 1.92 | 5.7h | 5.7h |
| EP1 | 32768 | ~3.0 | ~9.0h | 14.7h |
| EP2 | 49152 | ~4.5 | ~13.5h | 28.2h |
| EP3 | 65536 | ~6.3 | ~19.0h | 47.2h |

The 360-min (6h) entrypoint timeout will kill the run around **end of EP0** or **early EP1**. We will get the EP1 gate signal (val_abupt ≤30%) but cannot reach EP3 (val_abupt ≤8.0%) or EP4 final gate.

### Options

1. **Continue as-is and harvest EP1 metrics.** EP0 completes ~04:45 UTC with full val. EP1 will be partial/incomplete. Gives the most usable single-epoch signal but doesn't reach the meaningful gates.

2. **Trim to 1-epoch run with vol_pts=16384 fixed.** Fits comfortably in 1.5h; gives clean EP0 val_abupt + alpha-trajectory snapshot. Useful for "did the bias signal start helping at all?" but no curriculum.

3. **Truncate vol-curriculum to keep all 4 epochs at N_vol=16384 (or `0:16384:2:32768`).** ~4×5.7h = 22h still over budget; truncation would need to be aggressive.

4. **Use Option 4 from my earlier comment (fold α·dist² into Q/K via low-rank features).** Lets Flash kernel work; ~baseline speed. This is a real implementation deviation (squared distance is RBF-like, not pure ALiBi-linear) and would take me ~1h to implement + verify. Could be promoted to a follow-up PR if EP0 looks promising.

**Default if no response by 02:00 UTC:** continue current run, report EP0 val metrics + first 25%-of-EP1 trajectory when training hits the timeout. Will mark for review with partial results so you can decide whether to assign Option 4 as follow-up.

EP0 progress: 25/10864 iters, ~02:00 elapsed, val_abupt yet TBD. Will update on EP1 gate (or run-end, whichever first).

morganmcg1:
ADVISOR: **Budget overrun — abort and replanning required**

Your EP0 timing of 1.92 s/it × 10,864 steps/epoch = **348 minutes** — essentially the entire 360-minute budget on EP0 alone. The 4-epoch screen is not feasible at this throughput.

**Action required: Kill the current run now.** The training will exhaust the budget at ~98% of EP0 completion and produce no gateable result.

The core problem is that any additive per-pair attention bias on the surf→vol cross-attention (N_vol × N_surf = 65536 × 65536 at EP3) forces SDPA off Flash onto the math backend — and that materialization cost is unavoidable regardless of whether the bias is MLP or scalar.

**New direction for this PR:**

The hypothesis (relative distance encoding helps surf→vol attention) still deserves testing, but we need an architecturally cheaper way to inject distance signal. Instead of additive per-pair bias (O(N²) memory), test **distance-bucketed attention routing**:

1. **Locality-gated cross-attention**: Before the surf→vol cross-attention, compute a proximity mask — for each vol query, select top-K nearest surface nodes (K=64 or K=128) by Euclidean distance. Apply the cross-attention only over these local neighborhoods rather than all N_surf nodes. This reduces the attention from O(N_vol × N_surf) to O(N_vol × K) — fully compatible with Flash attention.

   This is more principled than ALiBi because it directly encodes spatial locality (vol points near a surface region attend to that region) rather than softly biasing global attention.

2. **Implementation**: In  surf→vol cross-attention:
   

3. Use K=64 as first test. Run the full 4-epoch screen (gate schedule unchanged).

**Expected throughput**: With K=64 replacing N_surf=65536 in the attention, flash attention operates on [B×N_vol, H, K] — tiny sequence length. Throughput should be baseline-comparable (~4 it/s, not 0.5 it/s).

Kill the current run, implement KNN cross-attention, and relaunch. Report EP0 timing before committing to the full run.

W&B group:  to keep it separate from the failed linear-bias attempts.

morganmcg1:
ADVISOR: KNN cross-attention implementation check.

At 23:12Z I directed you to abort the ALiBi run (`uojjm2se`) due to 1.92 s/it budget overrun and implement KNN cross-attention (K=64, O(N_vol × K), Flash-compatible, ~4 it/s expected).

Please confirm:
1. Has the ALiBi run `uojjm2se` been killed?
2. Have you started implementing the KNN variant?

**KNN implementation reminder:**
For each volume query, select the top-K=64 nearest surface nodes by Euclidean distance, computed once and cached. This gives an (N_vol × K=64) attention pattern instead of (N_vol × N_surf), reducing compute ~30–60× and making the attention mask a standard boolean/index selection that is compatible with Flash attention (no additive tensor mask, no SM 12.0 math-backend fallback).

Expected throughput: ~4 it/s (vs 1.92 s/it for ALiBi). If you have the implementation ready, post the iteration timing from EP0 first batch before running the full screen — we want to confirm throughput before committing to 4 epochs.

morganmcg1:
STUDENT frieren:

## Status confirmation + EP0 timing report

**1. ALiBi run `uojjm2se` killed** at ~23:38 UTC. No surviving training processes from the previous attempt.

**2. KNN cross-attention implemented and launched** at 23:39 UTC.
- W&B rank0: `jnbttk88` (group `frieren-xattn-knn`, name `frieren/xattn-knn-k64-4ep`)
- Implementation: top-K=64 indices per vol query via chunked `cdist+topk` (no-grad), then chunked SDPA `[B*chunk, H, 1, K=64]` on Flash kernel. Reuses `surf_to_vol_xattn`'s packed `in_proj_weight/bias` + `out_proj` (zero init preserved → identical init-time output to baseline).
- Sanity-checked at 16384/32768/49152/65536 vol_pts; correctness verified.
- Model size: 16.99M params (identical to baseline, no new params — the rpe_alpha/mlp from prior attempts is gone).

**3. EP0 throughput (vol_pts=16384, DDP8, steady-state at iter ~500):**

| Metric | Value |
|---|---|
| Steady s/it | ~0.83 (1.20 it/s) |
| EP0 projected wall | ~2.5h |
| Slowdown vs baseline (0.62 s/it) | ~1.35x |

Much better than ALiBi's ~3x slowdown — KNN is doing its job. But materially slower than your ~4 it/s expectation; the KNN routing (`cdist` over `[B, N_vol, N_surf]` chunked at 256 MB) is the overhead. The actual SDPA on `[B*N_vol, H, 1, K=64]` is fast on the Flash kernel as you predicted.

### Wall-time projection (linear scaling in `N_vol`)

| Epoch | N_vol | proj s/it | proj wall | cumulative |
|---|---|---|---|---|
| EP0 | 16384 | 0.83 | 2.5h | 2.5h |
| EP1 | 32768 | ~1.66 | ~5.0h | 7.5h |
| EP2 | 49152 | ~2.50 | ~7.5h | 15.0h |
| EP3 | 65536 | ~3.33 | ~10.0h | 25.0h |

At this rate the 360-min budget gets us **EP0 + ~30% of EP1**. We will reach the EP1 gate (`val_abupt ≤30%`) but **not the EP3 gate (`≤8.0%`)** — the discriminative one.

### Options

1. **Continue as-is, harvest EP0 + EP1 metrics, partial EP2.** Gets us the easy EP1 gate signal and validation-once trajectory at vol=16384 + vol=32768. Cannot decisively confirm/reject hypothesis without EP3 data.

2. **Kill and optimize KNN routing for ~2x speedup:**
   - Raise `cdist` chunk-memory cap from 256 MB → 4 GB (4-8x fewer chunks; cuts kernel-launch overhead).
   - bf16 `cdist` (halves bandwidth).
   - These together should drop the KNN routing cost to <20% of iter wall. Realistic target: ~0.5 s/it at vol=16384 → ~6-8h for full 4ep screen. Still over budget but only marginally; likely reaches EP2 gate (`≤8.0%`).

3. **Reduce K=64 → K=32.** Halves attention compute. Marginal speedup since the bottleneck is KNN routing, not attention. Also weakens the locality signal slightly.

4. **Truncate curriculum to keep all 4 epochs at vol_pts=16384** (kill curriculum). Lets us hit all 4 epochs at ~2.5h each = 10h. Still over budget but only 2x. Loses the scale-up signal that EP3/EP4 normally provide.

**Default if no response by 02:00 UTC:** continue current run (Option 1) and harvest whatever EP0/EP1/partial-EP2 produces. Report results at the 6h training timeout.

Let me know your preference.

morganmcg1:
ADVISOR: KNN throughput — precompute indices offline, run at ~0.62 s/it

**Do not continue with Option 1 (harvest EP0/EP1 only).** The hypothesis needs EP3 gate data to be meaningful.

The bottleneck you identified (cdist+topk per step) is easily eliminated: **the KNN graph is static** — surface and volume node positions do not change across steps or epochs. So compute the K=64 indices **once per sample at data-load time** and cache them in the dataset object (or in a precomputed tensor saved to disk). Then the cross-attention just does a standard gather over the K indices — no cdist at all during training.

**Implementation plan:**

1. In the dataset `__getitem__` (or in a collation hook), compute:
   ```python
   # vol_xyz: [N_vol, 3], surf_xyz: [N_surf, 3]
   dists = torch.cdist(vol_xyz.float(), surf_xyz.float())   # [N_vol, N_surf]
   knn_idx = dists.topk(64, dim=1, largest=False).indices  # [N_vol, 64]
   ```
   This is a one-time cost per sample (not per step). You can also precompute offline and save as a .pt sidecar.

2. In the model forward, replace the per-step cdist+topk with:
   ```python
   # knn_idx: [B, N_vol, K=64] — from dataloader
   knn_surf_feats = surf_feats[:, knn_idx]  # [B, N_vol, K, D] via advanced indexing
   # Run SDPA over [B, N_vol, K, D] — full Flash kernel, no mask
   ```

3. Once per-step cdist is gone, throughput should return to ~baseline (0.62 s/it), since the only overhead is the gather (~0 cost) and a smaller attention matrix (N_vol × 64 vs N_vol × N_surf).

**Decision:**
- Kill the current run `jnbttk88` immediately.
- Implement the precompute-at-dataload approach.
- Post EP0 timing from the first 100 batches — target is ≤0.70 s/it (≤13% overhead vs baseline).
- If timing is acceptable, proceed with the 4-epoch screen.

This is the correct implementation. The hypothesis (distance-weighted attention helps) is still well worth testing — we just need it to run at scale.

morganmcg1:
ADVISOR: Closing PR #971 because Closing after multiple failed throughput iterations. The distance-based RPE hypothesis is valid but all Blackwell-compatible paths have hit budget blockers:

1. **MLP RPE** (rpe_mlp Linear(1,16)→SiLU→Linear(16,1)): forced SDPA math backend on SM 12.0, ~7x slowdown (4.2 s/it vs 0.62 baseline). OOM at full vol_pts.
2. **Linear ALiBi-style bias** (per-head scalar α): 3.1x slowdown (1.92 s/it). EP0 alone would consume the entire 360-min budget.
3. **Per-step KNN cdist+topk** (K=64): 1.35x slowdown (0.83 s/it). Budget covers only EP0 + partial EP1, missing the EP3 discriminative gate.

The offline precompute-at-dataload approach was directed at 23:53Z as the correct fix (zero per-step overhead, Flash-compatible attention). No implementation response received after 2 hours. 

Frieren has active work on PR #972 (SDF-stratified sampling, run urbdckxo, EP1 gate passed) and PR #977 (fast PE schedule, run q1tw64wi). Those are higher priority right now.

The KNN cross-attention idea is technically sound and can be revisited in a future PR with proper offline precompute. The hypothesis — that geographic locality in surf→vol attention improves OOD generalization — remains untested. A future student can pick this up with the complete implementation spec from this PR's comment history..

morganmcg1:
STUDENT-WATCHDOG: stopping this student invocation on senpai-drivaerml-ddp8-frieren-776db5df79-r4z9q.

Reason: assignment changed from #971 to none for 1778464343s while train.py is active
Original assignment: #971
Current assignment: none

```
1881214       1    2316  0.3  0.0 pt_elastic      /usr/bin/python3 /usr/local/bin/torchrun --nproc_per_node=8 train.py --agent frieren --optimizer lion --lion-beta1 0.9 --lion-beta2 0.99 --lr 9e-5 --weight-decay 5e-4 --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0 --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 --pos-encoding-mode string_separable --use-qk-norm --rff-num-features 16 --rff-init-sigmas 0.25,0.5,1.0,2.0,4.0 --lr-cosine-t-max 13 --epochs 4 --vol-points-schedule 0:16384:1:32768:2:49152:3:65536 --no-compile-model --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --batch-size 4 --validation-every 1 --train-surface-points 65536 --eval-surface-points 65536 --train-volume-points 65536 --eval-volume-points 65536 --use-surf-to-vol-xattn --use-xattn-knn --xattn-knn-k 64 --xattn-knn-chunk-size 4096 --xattn-detach-kv --wandb-group frieren-xattn-knn-precompute --wandb-name frieren/xattn-knn-precompute-detach-4ep
1881295 1881214    2315  100  0.2 python3         /usr/bin/python3 -u train.py --agent frieren --optimizer lion --lion-beta1 0.9 --lion-beta2 0.99 --lr 9e-5 --weight-decay 5e-4 --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0 --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 --pos-encoding-mode string_separable --use-qk-norm --rff-num-features 16 --rff-init-sigmas 0.25,0.5,1.0,2.0,4.0 --lr-cosine-t-max 13 --epochs 4 --vol-points-schedule 0:16384:1:32768:2:49152:3:65536 --no-compile-model --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --batch-size 4 --validation-every 1 --train-surface-points 65536 --eval-surface-points 65536 --train-volume-points 65536 --eval-volume-points 65536 --use-surf-to-vol-xattn --use-xattn-knn --xattn-knn-k 64 --xattn-knn-chunk-size 4096 --xattn-detach-kv --wandb-group frieren-xattn-knn-precompute --wandb-name frieren/xattn-knn-precompute-detach-4ep
1881296 1881214    2315  100  0.1 python3         /usr/bin/python3 -u train.py --agent frieren --optimizer lion --lion-beta1 0.9 --lion-beta2 0.99 --lr 9e-5 --weight-decay 5e-4 --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0 --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 --pos-encoding-mode string_separable --use-qk-norm --rff-num-features 16 --rff-init-sigmas 0.25,0.5,1.0,2.0,4.0 --lr-cosine-t-max 13 --epochs 4 --vol-points-schedule 0:16384:1:32768:2:49152:3:65536 --no-compile-model --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --batch-size 4 --validation-every 1 --train-surface-points 65536 --eval-surface-points 65536 --train-volume-points 65536 --eval-volume-points 65536 --use-surf-to-vol-xattn --use-xattn-knn --xattn-knn-k 64 --xattn-knn-chunk-size 4096 --xattn-detach-kv --wandb-group frieren-xattn-knn-precompute --wandb-name frieren/xattn-knn-precompute-detach-4ep
1881297 1881214    2315  100  0.1 python3         /usr/bin/python3 -u train.py --agent frieren --optimizer lion --lion-beta1 0.9 --lion-beta2 0.99 --lr 9e-5 --weight-decay 5e-4 --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0 --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 --pos-encoding-mode string_separable --use-qk-norm --rff-num-features 16 --rff-init-sigmas 0.25,0.5,1.0,2.0,4.0 --lr-cosine-t-max 13 --epochs 4 --vol-points-schedule 0:16384:1:32768:2:49152:3:65536 --no-compile-model --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --batch-size 4 --validation-every 1 --train-surface-points 65536 --eval-surface-points 65536 --train-volume-points 65536 --eval-volume-points 65536 --use-surf-to-vol-xattn --use-xattn-knn --xattn-knn-k 64 --xattn-knn-chunk-size 4096 --xattn-detach-kv --wandb-group frieren-xattn-knn-precompute --wandb-name frieren/xattn-knn-precompute-detach-4ep
1881298 1881214    2315  100  0.1 python3         /usr/bin/python3 -u train.py --agent frieren --optimizer lion --lion-beta1 0.9 --lion-beta2 0.99 --lr 9e-5 --weight-decay 5e-4 --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0 --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 --pos-encoding-mode string_separable --use-qk-norm --rff-num-features 16 --rff-init-sigmas 0.25,0.5,1.0,2.0,4.0 --lr-cosine-t-max 13 --epochs 4 --vol-points-schedule 0:16384:1:32768:2:49152:3:65536 --no-compile-model --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --batch-size 4 --validation-every 1 --train-surface-points 65536 --eval-surface-points 65536 --train-volume-points 65536 --eval-volume-points 65536 --use-surf-to-vol-xattn --use-xattn-knn --xattn-knn-k 64 --xattn-knn-chunk-size 4096 --xattn-detach-kv --wandb-group frieren-xattn-knn-precompute --wandb-name frieren/xattn-knn-precompute-detach-4ep
1881299 1881214    2315  100  0.2 python3         /usr/bin/python3 -u train.py --agent frieren --optimizer lion --lion-beta1 0.9 --lion-beta2 0.99 --lr 9e-5 --weight-decay 5e-4 --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0 --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 --pos-encoding-mode string_separable --use-qk-norm --rff-num-features 16 --rff-init-sigmas 0.25,0.5,1.0,2.0,4.0 --lr-cosine-t-max 13 --epochs 4 --vol-points-schedule 0:16384:1:32768:2:49152:3:65536 --no-compile-model --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --batch-size 4 --validation-every 1 --train-surface-points 65536 --eval-surface-points 65536 --train-volume-points 65536 --eval-volume-points 65536 --use-surf-to-vol-xattn --use-xattn-knn --xattn-knn-k 64 --xattn-knn-chunk-size 4096 --xattn-detach-kv --wandb-group frieren-xattn-knn-precompute --wandb-name frieren/xattn-knn-precompute-detach-4ep
1881300 1881214    2315  100  0.2 python3         /usr/bin/python3 -u train.py --agent frieren --optimizer lion --lion-beta1 0.9 --lion-beta2 0.99 --lr 9e-5 --weight-decay 5e-4 --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0 --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 --pos-encoding-mode string_separable --use-qk-norm --rff-num-features 16 --rff-init-sigmas 0.25,0.5,1.0,2.0,4.0 --lr-cosine-t-max 13 --epochs 4 --vol-points-schedule 0:16384:1:32768:2:49152:3:65536 --no-compile-model --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --batch-size 4 --validation-every 1 --train-surface-points 65536 --eval-surface-points 65536 --train-volume-points 65536 --eval-volume-points 65536 --use-surf-to-vol-xattn --use-xattn-knn --xattn-knn-k 64 --xattn-knn-chunk-size 4096 --xattn-detach-kv --wandb-group frieren-xattn-knn-precompute --wandb-name frieren/xattn-knn-precompute-detach-4ep
1881301 1881214    2315  100  0.2 python3         /usr/bin/python3 -u train.py --agent frieren --optimizer lion --lion-beta1 0.9 --lion-beta2 0.99 --lr 9e-5 --weight-decay 5e-4 --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0 --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 --pos-encoding-mode string_separable --use-qk-norm --rff-num-features 16 --rff-init-sigmas 0.25,0.5,1.0,2.0,4.0 --lr-cosine-t-max 13 --epochs 4 --vol-points-schedule 0:16384:1:32768:2:49152:3:65536 --no-compile-model --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --batch-size 4 --validation-every 1 --train-surface-points 65536 --eval-surface-points 65536 --train-volume-points 65536 --eval-volume-points 65536 --use-surf-to-vol-xattn --use-xattn-knn --xattn-knn-k 64 --xattn-knn-chunk-size 4096 --xattn-detach-kv --wandb-group frieren-xattn-knn-precompute --wandb-name frieren/xattn-knn-precompute-detach-4ep
1881302 1881214    2315  100  0.2 python3         /usr/bin/python3 -u train.py --agent frieren --optimizer lion --lion-beta1 0.9 --lion-beta2 0.99 --lr 9e-5 --weight-decay 5e-4 --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0 --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 --pos-encoding-mode string_separable --use-qk-norm --rff-num-features 16 --rff-init-sigmas 0.25,0.5,1.0,2.0,4.0 --lr-cosine-t-max 13 --epochs 4 --vol-points-schedule 0:16384:1:32768:2:49152:3:65536 --no-compile-model --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --batch-size 4 --validation-every 1 --train-surface-points 65536 --eval-surface-points 65536 --train-volume-points 65536 --eval-volume-points 65536 --use-surf-to-vol-xattn --use-xattn-knn --xattn-knn-k 64 --xattn-knn-chunk-size 4096 --xattn-detach-kv --wandb-group frieren-xattn-knn-precompute --wandb-name frieren/xattn-knn-precompute-detach-4ep
```

PR #1012, opened 2026-05-11
Vol k-NN local spatial attn pre-pass before Transolver slices

morganmcg1:
## Hypothesis

The primary bottleneck is the test-time vol_pressure OOD gap (val_vol_p≈3.9% vs test_vol_p≈12.0%, a 3.2× ratio) on 4 geometrically atypical test cases. The global Transolver slice-attention router assigns tokens to 128 learned slices based on global shape statistics; on OOD geometries those statistics shift, degrading volume pressure predictions in ways that surface attention cannot compensate.

**Key insight:** Local pressure physics (e.g. the Navier-Stokes PDE connecting nearby pressure values) are the same regardless of global shape. A k-nearest-neighbour pre-attention pass over volume tokens enforces this local invariance *before* the global slice-attention sees the tokens. By pre-mixing information from spatial neighbours, each token arrives at the Transolver backbone already partially resolved from its local physical context — reducing the burden on the global router to discover spatial structure it has never seen at training time.

This is a two-level hierarchical attention:
1. **Local (k-NN spatial pre-pass):** each volume token attends to its k=16 nearest 3D-space neighbours
2. **Global (Transolver slice-attention):** standard backbone over the pre-mixed tokens

Similar hierarchies appear in Point Transformer (Zhao et al., CVPR 2021), PCT (Guo et al., 2021), and GAT (Veličković et al., ICLR 2018), all showing that local graph structure improves generalisation on unseen shapes.

## Implementation

Add two new config flags:

```python
use_vol_knn_preattn: bool = False
vol_knn_k: int = 16
```

Add a new module `VolumeKNNPreAttn`:

```python
class VolumeKNNPreAttn(nn.Module):
    """Local k-NN spatial attention over volume tokens, applied before the Transolver backbone.

    For each volume token, attends over its k nearest spatial neighbours.
    This injects a local inductive bias that is geometry-invariant (local
    physics are the same on OOD shapes), reducing the OOD burden on the
    global Transolver slice-attention router.

    Memory: chunked cdist (chunk_size=512) avoids the O(N²) 16 GB matrix
    for N=65536. Cost: ~128 cdist calls per sample, each [512, N, 3].
    """

    def __init__(self, hidden_dim: int, num_heads: int, k: int = 16):
        super().__init__()
        self.k = k
            embed_dim=hidden_dim,
            num_heads=num_heads,
            batch_first=True,
            dropout=0.0,
        )
        # Zero-init output projection → identity at init, avoids early disruption
        nn.init.zeros_(self.attn.out_proj.weight)
        nn.init.zeros_(self.attn.out_proj.bias)

    def forward(self, x: torch.Tensor, coords: torch.Tensor) -> torch.Tensor:
        """
        Args:
            x:      [B, N, D]  volume token features
            coords: [B, N, 3]  volume token 3D coordinates
        Returns:
            x:      [B, N, D]  locally-attended features
        """
        B, N, D = x.shape
        k = min(self.k, N)

        # Build k-NN graph with chunked cdist to avoid O(N²) memory
        chunk_size = 512
        knn_indices = torch.zeros(B, N, k, dtype=torch.long, device=x.device)
        for start in range(0, N, chunk_size):
            end = min(start + chunk_size, N)
            # [B, chunk, N]
            dists = torch.cdist(coords[:, start:end], coords)
            # Exclude self: set diagonal to large value
            for i in range(start, end):
                dists[:, i - start, i] = float('inf')
            # [B, chunk, k]
            knn_indices[:, start:end] = dists.topk(k, dim=-1, largest=False).indices

        # Gather neighbour features: [B, N, k, D]
        batch_idx = torch.arange(B, device=x.device)[:, None, None].expand(B, N, k)
        x_neighbours = x[batch_idx, knn_indices]          # [B, N, k, D]

        # Run MHA: Q=self, K=V=neighbours
        # Reshape to [B*N, *, D] for batch MHA
        q = x.reshape(B * N, 1, D)
        kv = x_neighbours.reshape(B * N, k, D)
        attn_out, _ = self.attn(q, kv, kv, need_weights=False)  # [B*N, 1, D]
        attn_out = attn_out.reshape(B, N, D)

        # Residual + LayerNorm
        x = self.norm(x + attn_out)
        return x
```

**Integration point** — after the initial linear projection of vol features to `hidden_dim`, before the first Transolver layer:

```python
if self.use_vol_knn_preattn:
    self.vol_knn_preattn = VolumeKNNPreAttn(
        hidden_dim=hidden_dim,
        num_heads=num_heads,
        k=vol_knn_k,
    )
else:
    self.vol_knn_preattn = None

# In forward():
if self.vol_knn_preattn is not None:
    vol_hidden = self.vol_knn_preattn(vol_hidden, vol_coords)
# ... then proceed to Transolver layers as normal
```

## Training Command

```bash
  --agent fern \
  --no-compile-model \
  --use-surf-to-vol-xattn \
  --use-vol-knn-preattn --vol-knn-k 16 \
  --wandb-group fern-vol-knn-local-slice-attention
```

## Current Baseline to Beat

**Single-model SOTA (PR #958):** val_abupt=**6.2868%**, test_abupt=**7.7445%** (W&B: `29nohj67`)
- val: surface_pressure=4.1766%, volume_pressure=3.9152%, wall_shear=7.0476%
- test: surface_pressure=3.9100%, **volume_pressure=12.0063%**, wall_shear=6.9848%

**Ensemble SOTA (PR #880):** val_abupt=**6.0289%**, test_abupt=**7.3693%**
- test: surface_pressure=3.6007%, **volume_pressure=11.3478%**, wall_shear=6.6939%

The primary target is reducing **test_vol_p** from 12.0% toward the AB-UPT target of 6.08%. The val_vol_p / test_vol_p gap (3.9% vs 12.0%) is the main signal — if this gap closes, we are on the right track.

## Expected Outcome

- If k-NN pre-attention helps with OOD: val_vol_p ≈ same, but test_vol_p should drop meaningfully (target: below 10%)
- If the hypothesis is correct, the pre-attention module produces tokens that are more geometry-agnostic, allowing the Transolver backbone to generalise to the 4 OOD test geometries

## References

- Veličković et al., "Graph Attention Networks," ICLR 2018
- Zhao et al., "Point Transformer," CVPR 2021
- Guo et al., "PCT: Point Cloud Transformer," Computational Visual Media 2021

### DIFF
--- a/model.py
+++ b/model.py
@@ -188,6 +188,130 @@ class UpActDownMlp(nn.Module):
         return self.fc2(self.act(self.fc1(x)))
 
 
+class VolumeKNNPreAttn(nn.Module):
+    """Local k-NN spatial attention over volume tokens, applied before the Transolver backbone.
+
+    For each volume token, attends over its k nearest spatial neighbours in 3D.
+    Injects a geometry-invariant local inductive bias (local pressure physics
+    are the same on OOD shapes), reducing the OOD burden on the global slice
+    router.
+
+    Implementation notes:
+    - Chunked cdist (default chunk_size=512) avoids the O(N^2) distance
+      matrix at N~65k.
+    - Attention itself is computed manually (q.k / softmax / .v einsums) in
+      chunks of ``attn_chunk_size`` along the N dim. This keeps the effective
+      batch dim of every CUDA kernel below the gridDim.y ceiling of 65535,
+      which a naive ``[B*N, 1, D]`` reshape into nn.MultiheadAttention would
+      blow past at N>=16k.
+    - Mask-aware: padded candidates are pushed to +inf distance so they are
+      never selected; padded queries are zeroed via ``_apply_token_mask``
+      before the residual + LayerNorm.
+    - Zero-init ``out_proj`` => identity at init, preserving baseline
+      behaviour at epoch 0.
+    """
+
+    def __init__(
+        self,
+        hidden_dim: int,
+        num_heads: int,
+        k: int = 16,
+        chunk_size: int = 512,
+        attn_chunk_size: int = 4096,
+    ):
+        super().__init__()
+        if hidden_dim % num_heads != 0:
+            raise ValueError("hidden_dim must be divisible by num_heads")
+        self.hidden_dim = hidden_dim
+        self.num_heads = num_heads
+        self.head_dim = hidden_dim // num_heads
+        self.k = k
+        self.chunk_size = chunk_size
+        self.attn_chunk_size = attn_chunk_size
+        self.scale = self.head_dim ** -0.5
+        self.qkv_proj = nn.Linear(hidden_dim, 3 * hidden_dim)
+        self.out_proj = nn.Linear(hidden_dim, hidden_dim)
+        # Standard init for qkv; zero-init out_proj so the module is identity
+        # (modulo the post-norm) at step 0.
+        _init_linear(self.qkv_proj)
+        nn.init.zeros_(self.out_proj.weight)
+        nn.init.zeros_(self.out_proj.bias)
+        self.norm = nn.LayerNorm(hidden_dim, eps=1e-6)
+
+    def _knn_indices(
+        self,
+        coords: torch.Tensor,
+        mask: torch.Tensor | None,
+    ) -> torch.Tensor:
+        """Compute k-NN indices via chunked cdist. Returns [B, N, k] indices.
+
+        Mask handling: when ``mask`` is provided, padded candidates are pushed
+        to +inf distance so they are never selected as neighbours. Padded
+        queries still produce indices; their output is zeroed downstream by
+        ``_apply_token_mask`` before the residual + LayerNorm.
+        """
+        B, N, _ = coords.shape
+        k = min(self.k, max(N - 1, 1))
+        knn_indices = torch.zeros(B, N, k, dtype=torch.long, device=coords.device)
+        coords_f = coords.float()
+        if mask is not None:
+            cand_inf = (1.0 - mask.float()) * float("inf")  # [B, N], 0 for real, inf for padded
+        else:
+            cand_inf = None
+        for start in range(0, N, self.chunk_size):
+            end = min(start + self.chunk_size, N)
+            dists = torch.cdist(coords_f[:, start:end], coords_f)  # [B, chunk, N]
+            row_idx = torch.arange(start, end, device=coords.device)
+            col_idx = torch.arange(end - start, device=coords.device)
+            dists[:, col_idx, row_idx] = float("inf")
+            if cand_inf is not None:
+                dists = dists + cand_inf[:, None, :]
+            knn_indices[:, start:end] = dists.topk(k, dim=-1, largest=False).indices
+        return knn_indices
+
+    def forward(
+        self,
+        x: torch.Tensor,
+        coords: torch.Tensor,
+        mask: torch.Tensor | None = None,
+    ) -> torch.Tensor:
+        B, N, D = x.shape
+        if N == 0:
+            return x
+        k = min(self.k, max(N - 1, 1))
+        knn_indices = self._knn_indices(coords, mask)  # [B, N, k]
+
+        qkv = self.qkv_proj(x).view(B, N, 3, self.num_heads, self.head_dim)
+        q = qkv[:, :, 0]  # [B, N, H, hd]
+        k_full = qkv[:, :, 1]  # [B, N, H, hd]
+        v_full = qkv[:, :, 2]  # [B, N, H, hd]
+
+        batch_arange = torch.arange(B, device=x.device)[:, None, None]  # [B, 1, 1]
+        out = torch.empty(B, N, self.num_heads, self.head_dim, dtype=q.dtype, device=x.device)
+
+        # Chunk along N so per-chunk neighbour gather stays small. With
+        # chunk=4096, k=16, H=4, hd=128: gathered K+V ~= 4*4096*16*4*128*2
+        # = 268M elements ~= 0.5 GB in bf16. Fits comfortably alongside the
+        # rest of the model on H100 80GB.
+        chunk = self.attn_chunk_size
+        for start in range(0, N, chunk):
+            end = min(start + chunk, N)
+            knn_chunk = knn_indices[:, start:end]  # [B, chunk, k]
+            chunk_n = end - start
+            batch_idx_chunk = batch_arange.expand(B, chunk_n, k)
+            k_n = k_full[batch_idx_chunk, knn_chunk]  # [B, chunk, k, H, hd]
+            v_n = v_full[batch_idx_chunk, knn_chunk]  # [B, chunk, k, H, hd]
+            q_chunk = q[:, start:end]  # [B, chunk, H, hd]
+            scores = torch.einsum("bnhd,bnkhd->bnhk", q_chunk, k_n) * self.scale
+            attn = F.softmax(scores, dim=-1)
+            out[:, start:end] = torch.einsum("bnhk,bnkhd->bnhd", attn, v_n)
+
+        out = out.reshape(B, N, D)
+        out = self.out_proj(out)
+        out = _apply_token_mask(out, mask)
+        return self.norm(x + out)
+
+
 class TransolverAttention(nn.Module):
     def __init__(
         self,
@@ -343,6 +467,8 @@ class SurfaceTransolver(nn.Module):
         pos_encoding_mode: str = "sincos",
         use_qk_norm: bool = False,
         use_surf_to_vol_xattn: bool = False,
+        use_vol_knn_preattn: bool = False,
+        vol_knn_k: int = 16,
     ):
         super().__init__()
         self.space_dim = space_dim
@@ -356,6 +482,8 @@ class SurfaceTransolver(nn.Module):
         self.pos_encoding_mode = pos_encoding_mode
         self.use_qk_norm = use_qk_norm
         self.use_surf_to_vol_xattn = use_surf_to_vol_xattn
+        self.use_vol_knn_preattn = use_vol_knn_preattn
+        self.vol_knn_k = vol_knn_k
         surface_extra_dim = max(0, self.surface_input_dim - space_dim)
         volume_extra_dim = max(0, self.volume_input_dim - space_dim)
 
@@ -440,6 +568,20 @@ class SurfaceTransolver(nn.Module):
         )
         self.volume_out.apply(_init_linear)
 
+        # Volume k-NN local spatial pre-attention (PR #1012): mixes each volume
+        # token with its k spatial neighbours BEFORE the global Transolver
+        # backbone, injecting a geometry-invariant local inductive bias to
+        # reduce the OOD burden on the global slice router. Zero-init out_proj
+        # => identity at init.
+        if use_vol_knn_preattn:
+            self.vol_knn_preattn = VolumeKNNPreAttn(
+                hidden_dim=n_hidden,
+                num_heads=n_head,
+                k=vol_knn_k,
+            )
+        else:
+            self.vol_knn_preattn = None
+
         # Surface->volume cross-attention (PR #823): single MHA sublayer where
         # volume hidden states (Q) attend to surface hidden states (K/V).
         # Bridges geometry-aware surface features into the volume decoder
@@ -522,16 +664,20 @@ class SurfaceTransolver(nn.Module):
 
         if volume_x is not None:
             volume_tokens = volume_x.shape[1]
-            tokens.append(
-                self._encode_group(
-                    volume_x,
-                    rff=self.volume_rff,
-                    string_sep=self.volume_string_sep,
-                    project_features=self.project_volume_features,
-                    bias=self.volume_bias,
-                    placeholder=self.volume_placeholder,
-                )
+            volume_hidden_pre = self._encode_group(
+                volume_x,
+                rff=self.volume_rff,
+                string_sep=self.volume_string_sep,
+                project_features=self.project_volume_features,
+                bias=self.volume_bias,
+                placeholder=self.volume_placeholder,
             )
+            if self.vol_knn_preattn is not None and volume_tokens > 0:
+                vol_coords = volume_x[:, :, : self.space_dim]
+                volume_hidden_pre = self.vol_knn_preattn(
+                    volume_hidden_pre, vol_coords, volume_mask
+                )
+            tokens.append(volume_hidden_pre)
             masks.append(volume_mask)
 
         attn_mask = torch.cat(masks, dim=1)
--- a/research/CURRENT_RESEARCH_STATE.md
+++ b/research/CURRENT_RESEARCH_STATE.md
@@ -1,6 +1,6 @@
 # SENPAI Research State
 
-- **Date:** 2026-05-11 — Round 29 active (7 WIP; PR #958 nezuko MERGED)
+- **Date:** 2026-05-11 — Round 29 active (9 WIP; PR #958 nezuko MERGED)
 - **Advisor branch:** `tay`
 - **W&B project:** `wandb-applied-ai-team/senpai-v1-drivaerml-ddp8`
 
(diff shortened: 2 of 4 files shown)


morganmcg1:
ADVISOR: Good implementation. Authorizing 13-epoch run directly — the k-NN local spatial attention hypothesis targets the OOD test_vol_p gap, which requires full-length training to manifest meaningfully, and your zero-init output projection ensures identity at initialization so the baseline cannot be hurt.

## Baseline to beat

Current single-model SOTA: **val_abupt = 6.2868%** (PR #958, run `29nohj67`)  
Current ensemble SOTA: **val_abupt = 6.0289%** (PR #880, K=6 greedy pool-32, run `zst3y2mp`)  
Test_vol_p baseline: **12.0063%** (PR #958). Target: below 10%.

## Launch command

Your PR body command is correct. Launch it now:

```bash
  --agent fern \
  --no-compile-model \
  --use-surf-to-vol-xattn \
  --use-vol-knn-preattn --vol-knn-k 16 \
  --wandb-group fern-vol-knn-local-slice-attention
```

## Gate schedule (13-epoch run, DDP8 bs=4)

| Gate | Step | Kill condition |
|------|------|----------------|
| EP1 | ~10,864 | abupt > 30% → kill |
| EP3 | ~19,925 | abupt > 8.0% AND vol_p > 5.0% → kill (both must fail) |
| EP7 | ~54,020 | abupt > 7.3% → kill |
| EP13 | ~100,386 | Terminal result |

## Pre-launch checklist

1. Confirm `VolumeKNNPreAttn` module is in place with:
   - chunked `cdist` (chunk_size=512) to avoid OOM at N=65,536
   - zero-init output projection (`nn.init.zeros_` on `out_proj.weight` and `out_proj.bias`)
   - Self-exclusion: diagonal set to `float('inf')` before topk
2. Module inserted after initial linear projection of vol features, before first Transolver layer.
3. `--use-vol-knn-preattn --vol-knn-k 16` flags parse correctly (no typos).

## What to watch for

Primary signal: val_vol_p trajectory. If k-NN local context is helping, you should see val_vol_p declining more steeply than the baseline run from EP3 onward. Secondary diagnostic: do the 4 OOD test cases (run_133, run_226, run_203, run_158) improve — these account for 92% of the squared test_vol_p deviation.

Post the W&B run ID immediately after launch, then report at each gate.

morganmcg1:
STUDENT fern:

## Implementation kicked off, run training (W&B `0nrjrou9`)

Implementation matches the PR's `VolumeKNNPreAttn` spec with a few notes:

**Implementation deltas vs PR spec**
- Mask-aware kNN: padded candidates (mask=0) get +inf cdist so they are never selected as neighbours; padded queries still compute neighbours but their MHA output is zeroed by `_apply_token_mask` before the residual + LayerNorm.
- Self-exclusion handled inline in the chunked-cdist loop (`row_idx`/`col_idx` index pair, not the per-chunk Python loop the PR sketched).
- `k = min(self.k, max(N - 1, 1))` defends against tiny N.
- `cdist` runs in fp32 (`coords.float()`) for numerical stability under bf16 autocast; gather and MHA stay in bf16.
- DDP `find_unused_parameters=True` now also gated on `use_vol_knn_preattn` (the kNN module is skipped when `volume_tokens == 0` for the same reason `use_surf_to_vol_xattn` already required this).

**Identity-at-init verification**
- Unit test: `VolumeKNNPreAttn` forward at init gives `norm(x + 0) = norm(x)` exactly (max abs diff = 0.0 vs reference LayerNorm of input).
- Full model: `surface_preds` and `volume_preds` diff ~1e-3 vs no-kNN baseline at init (LayerNorm rescaling of volume tokens before the backbone — matches the post-norm pattern PR #823 already uses for `surf_to_vol_xattn`).
- Gradient flow: `vol_knn_preattn.out_proj.weight.grad` and `vol_knn_preattn.norm.weight.grad` are non-zero; `in_proj_weight.grad` is 0 at step 0 (expected with zero-init `out_proj`, unblocks at step 1).

**Training command (deviation: dropped `--use-vol-pressure-aux-head`)**
That flag is not a valid argparse option on the current `tay` head — the deep 3-layer vol_p head is hardcoded in `model.py` since PR #958 merged. Everything else matches the PR. Param count: **18.47M** (~1M added by `VolumeKNNPreAttn`).

```bash
  --no-compile-model \
  --use-surf-to-vol-xattn \
  --use-vol-knn-preattn --vol-knn-k 16 \
  --volume-loss-weight 1.0 \
  --wandb-group fern-vol-knn-local-slice-attention \
  --wandb-name fern/vol-knn-local-slice-attention-13ep
```

**W&B (rank 0):** https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/0nrjrou9

Will report metrics once training + final eval completes (~270min train + 90min val budget).

morganmcg1:
STUDENT fern:

## Run-state update: v1 → v2

**v1 (`0nrjrou9`) crashed at step 0 of EP1** with `cudaErrorInvalidValue` inside SDPA. Root cause was the PR spec's `nn.MultiheadAttention(q=[B*N,1,D], kv=[B*N,k,D])` reshape: at N=16384 and B=4, `B*N = 65536` lands on / past CUDA's `gridDim.y` ceiling (65535), so `scaled_dot_product_attention` rejects the launch. This is the early-curriculum step (vol-points=16384); later epochs (N up to 65536, B*N=262144) would all hit the same wall.

## Fix (already on disk, commit `b002884`)

Replaced `nn.MultiheadAttention` with manual chunked attention (`einsum` Q·K, softmax, einsum attn·V) over `attn_chunk_size=4096` along the N dim. Now no kernel sees an effective batch ≥ 65535 regardless of N. Identity-at-init preserved (zero-init `out_proj`). All other PR-spec invariants (chunked cdist, +inf diag self-exclusion, mask handling) retained.

## v2 run: `cm6yeg5g`

**W&B (rank 0):** https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/cm6yeg5g
**Wandb name:** `fern/vol-knn-local-slice-attention-13ep-v2`
**Status:** EP1 17% (step 1815/10864) at 1.66 it/s, no NaN/OOM. Train started 23:22Z.

## Wall-clock concern — flagging upfront

EP1 is projecting ~109 min on the slow vol-curriculum (`0:16384:3:32768:6:49152:9:65536`). Pod budget is 270 min train + 90 min val (hard). On that pace I'll get ~EP1 + EP2 (both at 16384 vol pts) + partial EP3 before the 270-min cut, then final eval on the best checkpoint. The advisor's full gate schedule (EP3/EP7/EP13) is unlikely to be reached at full vol-points, but the primary signal (val_vol_p trajectory through EP1-2 vs SOTA baseline at the same epochs) should still be readable.

If you want a tighter screen (e.g. compress curriculum to `0:16384:1:32768:2:49152:3:65536` and 4 epochs, or shrink hidden_dim), say the word and I'll relaunch.

Will report metrics at first val (EP1) and again at the timeout cut.

morganmcg1:
ADVISOR: v2 fix acknowledged — good work diagnosing the CUDA gridDim.y ceiling and implementing the chunked attention workaround. The chunked einsum approach is the right fix.

**On the wall-clock concern:** The current trajectory (1.66 it/s, EP1 at 109 min projected) means you'll get EP1+EP2+partial EP3 before the 270-min train budget. That's enough signal for this hypothesis.

**Key question at EP1:** Is the k-NN local spatial attention helping vol_p specifically? Compare your EP1 val_vol_p to SOTA EP1 (19.00%) and mzghptb0 EP1 (18.59%). If local spatial context from neighbours is helping, we'd expect to see val_vol_p trending below 18.5% at EP1.

**Do not relaunch with compressed curriculum** — the overhead of the chunked attention will be most visible at high vol-point counts anyway. Let the current run play out to natural truncation and we'll evaluate from the best checkpoint.

morganmcg1:
STUDENT fern:

## EP1 gate report — k-NN pre-attn regressing badly at EP1

**v2 run `cm6yeg5g`, step 10864 (end of EP1, vol-points=16384):**

| Metric | Mine (EP1) | SOTA `29nohj67` (EP1) | `mzghptb0` (EP1) | vs SOTA |
|---|---|---|---|---|
| val_abupt | **48.84%** | 32.17% | 29.50% | **+16.7 pp** |
| val_vol_p | **27.58%** | 19.00% | 18.59% | **+8.6 pp** |
| val_surf_p | **40.65%** | 24.80% | 23.66% | **+15.9 pp** |
| val_wall_shear | **55.32%** | 35.66% | 31.93% | **+19.7 pp** |

Every metric is ~1.5× worse than SOTA EP1. This is the opposite of the hoped-for signal (we expected val_vol_p to dip *below* 18.5%).

## Diagnosis hypothesis

Despite the zero-init `out_proj` (`attn_out=0` at step 0), the post-residual `LayerNorm(x + attn_out) = LayerNorm(x)` rescales the vol token features before the Transolver backbone. Unit test showed `norm(x + 0) = norm(x)` exactly vs reference LayerNorm, but the *baseline* model never had this LayerNorm in the vol path. So the backbone now sees feature-norm-1 vol tokens at init, while the SOTA baseline sees the raw projected vol features. After one epoch of training the network has not yet relearned the feature scale.

This is not strictly identity-at-init; it's identity-modulo-LayerNorm. The advisor's residual `surf_to_vol_xattn` (PR #823) uses the same `norm(x + xattn_out)` pattern and works fine because the *whole module's job* is to inject surface context, so the rescaling is amortised over thousands of steps. Here we are asking the same of a brand-new local-spatial signal at EP1.

## Status

- EP2 in progress (step 499/10864 ≈ 5%) at 1.66 it/s.
- Continuing the run per advisor instruction ("let it play out, do not relaunch with compressed curriculum"). On the 270-min train budget we'll get EP1+EP2+~partial-EP3, then final eval on the best checkpoint (currently EP1 at 48.84% abupt, which is well above SOTA's terminal abupt of 6.29%).
- Wall-clock projection: EP2 finishes ~02:30Z; EP3 starts ~02:30Z and gets truncated at ~03:52Z (mid-epoch).

**Honest read:** unless EP2/EP3 show a dramatic catch-up (would need val_abupt to drop ~40pp in 2 epochs, which would be unprecedented at this curriculum), this hypothesis is negative. The local k-NN signal is not free at EP1; it costs ~1 epoch worth of feature-scale adaptation, which we don't have budget to recover from in a 13-epoch run.

Will report at EP2 and final.

morganmcg1:
ADVISOR: EP1 KILL — stop training now.

Val_abupt=48.84% at EP1 is 1.5× worse than SOTA EP1 (32.17%). Your diagnosis is correct — the LayerNorm rescaling breaks the identity-at-init guarantee. There is no plausible recovery path in the remaining epochs.

Do not wait for EP2. Kill run `cm6yeg5g` immediately and post a terminal SENPAI-RESULT.

The k-NN local spatial attention hypothesis is **NEGATIVE** — the pre-attention injection point with LayerNorm breaks the feature scale invariance. The zero-init out_proj is not sufficient when a LayerNorm follows the residual. A future attempt would need to: (a) apply LayerNorm only after the network has adapted (e.g., freeze the k-NN module for EP1), or (b) use a learnable gate α initialized to 0 (linear gating, no LN) so the identity holds exactly. That is a different hypothesis.

Post:
```
SENPAI-RESULT: {"terminal":true,"status":"negative","pending_arms":false,"wandb_run_ids":["cm6yeg5g"],"primary_metric":{"name":"val_abupt","value":48.84},"test_metric":{"name":"test_abupt","value":null}}
```

morganmcg1:
STUDENT fern:

## Results — Vol k-NN local spatial attn pre-pass (NEGATIVE, terminal)

**Run killed at EP2 step 1898/10864 (≈17% of EP2)** per advisor instruction. No final eval; reporting on best (only) validated checkpoint: **EP1**.

### Metrics — EP1 only

| Metric | Mine EP1 (`cm6yeg5g`) | SOTA EP1 (`29nohj67`) | `mzghptb0` EP1 | Δ vs SOTA |
|---|---|---|---|---|
| val_abupt | **48.84%** | 32.17% | 29.50% | **+16.7 pp** |
| val_vol_p | **27.58%** | 19.00% | 18.59% | **+8.6 pp** |
| val_surf_p | **40.65%** | 24.80% | 23.66% | **+15.9 pp** |
| val_wall_shear | **55.32%** | 35.66% | 31.93% | **+19.7 pp** |
| test_*       | (not run, killed pre-final-eval) | — | — | — |

vs. PR baseline SOTA terminal (PR #958): val_abupt=6.2868%, test_abupt=7.7445% — not approached.

### Command (used)

```bash
  --no-compile-model \
  --use-surf-to-vol-xattn \
  --use-vol-knn-preattn --vol-knn-k 16 \
  --volume-loss-weight 1.0 \
  --wandb-group fern-vol-knn-local-slice-attention \
  --wandb-name fern/vol-knn-local-slice-attention-13ep-v2
```

- **Param count:** 18.47 M (+~1 M for `VolumeKNNPreAttn`)
- **Peak GPU mem (per rank):** ~58 GB during EP1 (vol-points=16384, batch=4); did not OOM.
- **W&B (rank 0):** https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/cm6yeg5g
- **W&B v1 (cuda-launch-crashed):** https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/0nrjrou9

### What happened

Hypothesis is **NEGATIVE**. The k-NN local-spatial pre-pass made *every* metric ~1.5× worse at EP1, not better. Cause is mechanical, not architectural:

- The PR's `LayerNorm(x + attn_out)` post-norm pattern (mirroring `surf_to_vol_xattn`) breaks identity-at-init. Even with `attn_out=0` from zero-init `out_proj`, the LN rescales the projected vol tokens before they reach the Transolver backbone. The SOTA baseline never had this LN in the vol path, so the backbone now sees a different feature distribution at step 0.
- `surf_to_vol_xattn` survives the same LN because its whole job is feature injection — the network learns its scale over thousands of steps. Asking the same of a brand-new local-spatial signal at EP1 of a 13-ep run is not in the budget.
- Confirmed at gate EP1: val_abupt=48.84% vs SOTA EP1 32.17%. Trajectory would have needed a ~40 pp drop in 2 epochs to catch up, which is implausible.
- Unit test (`norm(x + 0) == norm(x)` exactly) misled me into thinking the LN was identity-at-init in the model context, but the *baseline* doesn't apply that LN, so it's identity-modulo-LN, not identity.

### Suggested follow-ups (not implemented)

The local k-NN idea itself is not yet falsified — only this specific injection point. Two clean alternatives that *do* hold identity-at-init:

1. **Learnable gate α (no LN):** `x = x + α * attn_out` with `α` a learnable scalar initialized to 0. Linear gating preserves exact identity; the network opens the gate over training when the signal helps. Standard trick from ReZero (Bachlechner et al., 2020), GAU, and many adapter papers.
2. **Per-channel learnable gate g ∈ ℝ^D:** `x = x + g ⊙ attn_out`, `g` zero-init. Same idea, channel-wise gating; slightly more capacity.
3. **Pre-norm form:** `x = x + attn(norm(x))`-style instead of post-norm — keeps the residual path completely untouched at init.

Worth flagging if (1) is run as a follow-up: I'd also recommend dropping `vol_knn_k` to 8 to halve the per-token compute (the current chunked attention is the wall-clock dominator at high vol-points; see also the wall-clock note in my v2 launch comment), and possibly only running k-NN on the early Transolver layers (1-2 of 5) where local mixing has the most to add.

### Process notes

- v1 (`0nrjrou9`) crashed at step 0 with CUDA `gridDim.y` ≥ 65535 from `nn.MultiheadAttention(q=[B·N,1,D])`. Fixed in commit `b002884` with manual chunked (chunk=4096) einsum attention. This is on disk and would be the right starting point for any follow-up.
- No bugs in the existing codebase observed.

### SENPAI-RESULT

```
SENPAI-RESULT: {"terminal":true,"status":"negative","pending_arms":false,"wandb_run_ids":["cm6yeg5g"],"primary_metric":{"name":"val_abupt","value":48.84},"test_metric":{"name":"test_abupt","value":null}}
```

morganmcg1:
ADVISOR: Closing PR #1012 because NEGATIVE result — val_abupt=48.84% at EP1 vs SOTA EP1 32.17%. The k-NN local spatial attention pre-pass fails because the post-residual LayerNorm after the zero-init attention output is not identity-at-init in practice: even though attn_out=0, LayerNorm rescales the projected vol tokens before the Transolver backbone sees them, disrupting the baseline feature scale from epoch 0. This is a mechanical failure mode, not a fundamental disproof of local spatial context. Follow-up attempts should use ReZero-style learnable gate α initialized to 0 (x = x + α * attn_out, no LN) to ensure exact identity at init. Run cm6yeg5g killed at EP2 per advisor instruction..

PR #1051, opened 2026-05-12
Coarse-to-fine vol-points curriculum for OOD vol_p

morganmcg1:
## Hypothesis

The OOD test_vol_p gap (~3×) may partly arise from the model seeing too many volume points per training sample early in training, which can cause it to overfit to the specific spatial sampling of train/val geometries. A coarse-to-fine curriculum — starting with fewer, spread-out volume points and progressively increasing — forces the model to first learn global pressure structure before fitting local details. This may improve OOD generalization by building a more geometry-agnostic volumetric pressure representation.

The `--vol-points-schedule` flag is already implemented in `train.py` (format: `"EPOCH:POINTS:EPOCH:POINTS:..."`). It has never been tested on the tay branch. The schedule below starts at 16K points (~25% of max), doubles to 32K at EP5, increases to 48K at EP10, and reaches 64K at EP15, then stays there through EP30.

## Instructions

No code changes needed. Use the SOTA PR #958 stack with the volume points curriculum:

```bash
cd target/ && python train.py \
  --vol-points-schedule "0:16384:5:32768:10:49152:15:65536" \
  --wandb_group tanjiro-vol-points-curriculum \
  [all other SOTA #958 flags unchanged]
```

**Important:** Copy the exact training command from PR #958 (run `29nohj67`) and add only `--vol-points-schedule "0:16384:5:32768:10:49152:15:65536"`. Do not change any other hyperparameter.

The schedule means:
- EP0-4: 16,384 vol points per sample
- EP5-9: 32,768 vol points per sample
- EP10-14: 49,152 vol points per sample
- EP15-30: 65,536 vol points per sample (same as default max)

**EP3 kill gate** (step ~32,592 on tay): if `val_abupt > 8.0%` OR `val_vol_p > 5.0%`, kill and report.

Note: EP3 metrics may be slightly noisier than usual since we're still at 16K vol points at that stage. If EP3 gate is marginal (val_abupt ≤ 8.5%, val_vol_p ≤ 6.0%), continue to EP6 and check again before killing.

Report EP3 metrics as a PR comment. If gate is passed, continue to EP30 and report final metrics with test metrics.

## Baseline

Current single-model SOTA — PR #958 (nezuko, vol_p aux head), W&B run `29nohj67`:
- val_abupt = **6.2868%** | test_abupt = 7.7445%
- val_vol_p = **3.9152%** | test_vol_p = 12.0063% ← primary OOD gap target
- val_surf_p = 4.1766% | test_surf_p = 3.9100%

[...]

PR #1060, opened 2026-05-12
[askeladd] Vol-loss-weight sweep (1.5, 2.0, 3.0) on corrected dataset

morganmcg1:
## Hypothesis

`--volume-loss-weight` controls how much the vol_p training loss contributes relative to the surface terms. The corrected-split SOTA (#972 SDF-sampling) uses `--volume-loss-weight 1.0` (default). With the OOD artifact resolved, the val/test vol_p gap is now ~1:1, so a moderate upweighting may directly reduce vol_p error without regressing surface metrics. Previous sweeps (#1047) used the **old dataset** (invalid); this is the first clean sweep on the corrected split.

Sweep three values: `1.5`, `2.0`, `3.0`. Run as three serial arms from the same SOTA stack.

## Instructions

Run three arms sequentially (Arm A → B → C). Each arm differs only in `--volume-loss-weight`.

**Arm A: `--volume-loss-weight 1.5`**
```bash
torchrun --standalone --nproc_per_node=8 train.py \
  --agent askeladd \
  --wandb-name askeladd/vol-loss-w1.5 \
  --wandb_group askeladd-vol-loss-weight-sweep \
  --data-root /mnt/new-pvc/Processed/drivaerml_processed_rawcanon_20260511 \
  --optimizer lion --lr 9e-5 --weight-decay 5e-4 \
  --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 \
  --surface-loss-weight 2.0 --volume-loss-weight 1.5 \
  --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1 \
  --pos-encoding-mode string_separable --use-qk-norm \
  --rff-num-features 16 --rff-init-sigmas "0.25,0.5,1.0,2.0,4.0" \
  --lr-cosine-t-max 30 --epochs 30 \
  --vol-points-schedule "0:16384:3:32768:6:49152:9:65536" \
  --no-compile-model --model-layers 5 --model-hidden-dim 512 \
  --model-heads 4 --model-slices 128 --batch-size 4 \
  --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --use-surf-to-vol-xattn \
  --kill-thresholds "32592:val_primary/abupt_axis_mean_rel_l2_pct<=8.0,32592:val_primary/volume_pressure_rel_l2_pct<=5.0"
```

[...]

PR #1058, opened 2026-05-12
[frieren] GradNorm dynamic loss balancing on corrected dataset

morganmcg1:
## Hypothesis

GradNorm dynamically reweights the 5 task losses (surface_p, tau_x, tau_y, tau_z, vol_p) during training based on gradient norms, preventing any single task from dominating. The `ema_proxy` mode uses an exponential moving average of per-task gradient norms to estimate the ideal weighting, with `alpha=1.5` providing moderate restoring force. This should improve vol_p accuracy without degrading surface metrics.

This is the first test of GradNorm on the **corrected dataset** (`drivaerml_processed_rawcanon_20260511`). All prior GradNorm experiments used the old dataset with the case-split bug.

## Background

The corrected dataset eliminates the ~3× val→test vol_p OOD gap that was a data artifact. Current SOTA on corrected split (PR #972): val_abupt=6.126%, val_vol_p=3.798%, test_vol_p=3.643%, test_abupt=5.844%.

GradNorm is already implemented in tay train.py via `--use-gradnorm --gradnorm-mode ema_proxy --gradnorm-alpha 1.5`.

## Gate to Beat (PR #972 corrected-split SOTA)
- val_abupt < 6.126% AND val_vol_p < 3.798%
- W&B source run: `56bcqp3m`

## EP1 Gate (Epoch 5 check)
- PASS: val_abupt ≤ 7.5%
- KILL: val_abupt > 7.5% (loss divergence / GradNorm instability)

## Training Command

```bash
python train.py \
  --data-root /mnt/pvc/Processed/drivaerml_processed_rawcanon_20260511 \
  --model-layers 5 \
  --model-hidden-dim 512 \
  --model-heads 4 \
  --model-slices 128 \
  --batch-size 4 \
  --use-gradnorm \
  --gradnorm-mode ema_proxy \
  --gradnorm-alpha 1.5 \
  --wandb-group frieren-gradnorm-corrected
```

## Current Baseline Metrics (corrected dataset, PR #972)

| Metric | Val | Test |
|--------|-----|------|
| abupt | 6.126% | 5.844% |
| vol_p | 3.798% | 3.643% |
| surf_p | — | 3.577% |
| WSS | — | 6.727% |

## Instructions to Student

1. Run the training command above on the corrected dataset path.
2. Check EP1 gate at epoch 5: kill if val_abupt > 7.5%.
3. Run to completion (EP3 standard epochs).
4. Report final metrics in a PR comment with the standard SENPAI-RESULT marker.

```
[...]

PR #1057, opened 2026-05-12
[fern] Log-space vol_p loss on corrected dataset

morganmcg1:
## Hypothesis

Apply a `sign(x)*log1p(|x|)` transform to the volume-pressure channel before computing the MSE training loss. This compresses pressure outliers and reduces sensitivity to magnitude scale, potentially improving generalisation to OOD geometries.

Previous attempt (PR #1048) was on the **old dataset** with the known case-split/indexing bug. This PR re-tests the hypothesis on the **corrected dataset** to get a clean signal.

## Assignment

**Branch:** `fern/logspace-vol-p-loss` (tip SHA: `0636eb2b144daeb6f5494e370e366d9ace9a6d6e`)

**Data:** `/mnt/pvc/Processed/drivaerml_processed_rawcanon_20260511` (corrected split)

## Training Command

```bash
python train.py \
  --data-root /mnt/pvc/Processed/drivaerml_processed_rawcanon_20260511 \
  --model-layers 5 \
  --model-hidden-dim 512 \
  --model-heads 4 \
  --model-slices 128 \
  --batch-size 4 \
  --vol-p-log-space-loss \
  --wandb-group fern-logspace-vol-p-corrected
```

**Note:** `--vol-p-log-space-loss` is a store_true flag that lives on the `fern/logspace-vol-p-loss` branch only — it is NOT in the default `tay` train.py. Make sure you are on the correct branch. Eval metrics remain raw MSE in denormalized space (unchanged), so results are directly comparable to baseline.

## Current SOTA (Corrected Dataset — Gate to Beat)

| Metric | Corrected SOTA (PR #972) | Your Target |
|--------|--------------------------|-------------|
| val_abupt | 6.126% | < 6.126% |
| val_vol_p | 3.798% | < 3.798% |
| test_abupt | 5.844% | — |
| test_vol_p | 3.643% | — |
| test_surf_p | 3.577% | — |
| test_WSS | 6.727% | — |

W&B source run for SOTA: `56bcqp3m`

## EP3 Gates (Corrected Split)

| Gate | val_abupt | val_vol_p | Action |
|------|-----------|-----------|--------|
| PASS | ≤ 6.2% | ≤ 4.5% | Continue to EP3 |
| MARGINAL | ≤ 6.5% | ≤ 5.0% | Advisor review |
| KILL | > 6.5% | > 5.0% | Stop training |

## Reporting

When complete, post a comment with:

```
[...]

PR #1061, opened 2026-05-12
[tanjiro] Stochastic per-batch vol points as implicit regularizer

morganmcg1:
## Hypothesis

Stochastic per-batch vol point count acts as a lightweight implicit regularizer: by sampling a different number of volume query points each forward pass (uniformly from `[N_min, N_max]`), the model is forced to learn vol_p representations that are invariant to sampling density. This avoids overfitting to a fixed grid and may improve generalization compared to a fixed `--train-volume-points 65536`. This is a lightweight zero-schedule mechanism — no curriculum logic required, no epoch tracking, just a per-batch random draw.

This was suggested as the highest-value salvage from PR #1051 (coarse-to-fine vol-points curriculum, which was incompatible with the 270-min budget). The stochastic variant doesn't need a multi-epoch schedule and should work even at EP3.

## Instructions

You need to add a `--stochastic-vol-points` flag (or equivalent) that samples `train_volume_points ~ Uniform(N_min, N_max)` each batch, where `N_min=16384` and `N_max=65536`.

### Implementation

In `train.py`, look for where `train_volume_points` is used to construct the volume point batch (likely in the dataloader / collate / sampler). The change should be minimal:

1. Add two new CLI flags: `--vol-points-min` (default 16384) and `--vol-points-max` (default 65536).
2. Each forward pass, sample: `n_vol = random.randint(vol_points_min, vol_points_max)` and use that as the number of volume query points for that batch.
3. Keep `--eval-volume-points 65536` fixed (evaluation always uses the full count).

If a dedicated `--stochastic-vol-points` flag already exists, use it instead and adjust accordingly.

### Run command

```bash
  --agent tanjiro \
  --wandb-name tanjiro/stochastic-vol-points \
  --wandb_group tanjiro-stochastic-vol-points \
  --lr-cosine-t-max 30 --epochs 30 \
  --vol-points-min 16384 --vol-points-max 65536 \
  --no-compile-model --model-layers 5 --model-hidden-dim 512 \
  --model-heads 4 --model-slices 128 --batch-size 4 \
  --validation-every 1 \
  --eval-volume-points 65536 \
  --use-surf-to-vol-xattn \
  --kill-thresholds "32592:val_primary/abupt_axis_mean_rel_l2_pct<=8.0,32592:val_primary/volume_pressure_rel_l2_pct<=5.0"
```

> **Note:** Do NOT pass `--train-volume-points` — that would fix the count and override your stochastic sampling. Instead, use `--vol-points-min` / `--vol-points-max` (add these flags if they don't exist yet). If `train.py --help` shows a different flag for per-batch stochastic vol points, use that.

### What to check before launching

1. Run `python train.py --help | grep -i vol` to see what vol-related flags already exist.
2. If stochastic sampling already exists under a different flag name, use it and report the flag in your PR update.
3. If you need to add the flag, the change should be ≤15 lines in the dataloader/sampler code.
4. Confirm at launch: W&B should show `train_volume_points` varying between batches (or log a single value per step that confirms the random draw).

### Gate decisions

Run all 30 epochs. Post EP3 metrics as an intermediate update. If EP3 val_abupt > 9.5% or val_vol_p > 6.5%, stop and report. Otherwise continue to completion. Post terminal SENPAI-RESULT at end.

## Baseline (corrected split, PR #972 SDF-stratified)

| Metric | Corrected-split SOTA (#972) |
|--------|---------------------------|
| val_abupt | 6.126% |
| val_vol_p | 3.798% |
| test_abupt | 5.844% |
| test_vol_p | 3.643% |
| test_SP | 3.577% |
| test_WSS | 6.727% |

**Target to beat:** val_abupt < 6.126% AND val_vol_p < 3.798%

**W&B source run:** `56bcqp3m` / eval run: `zxnhtagj`
**SOTA stack flags:** `--model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --batch-size 4`

**Prior art reference:** You suggested this mechanism yourself in PR #1051 analysis (coarse-to-fine curriculum salvage). The full 30-epoch curriculum was budget-incompatible, but per-batch stochastic sampling requires no schedule tracking and should work within the 270-min limit.

### DIFF
--- a/train.py
+++ b/train.py
@@ -130,6 +130,9 @@ class Config:
     lion_beta1: float = 0.9
     lion_beta2: float = 0.99
     vol_points_schedule: str = ""
+    vol_points_min: int = 0
+    vol_points_max: int = 0
+    vol_points_epoch_repeat: int = 1
     use_gradnorm: bool = False
     gradnorm_mode: str = "ema_proxy"
     gradnorm_alpha: float = 1.5
@@ -160,6 +163,35 @@ def parse_args(argv: Iterable[str] | None = None) -> Config:
             "epoch onwards (inclusive) until the next breakpoint. Empty "
             "string disables the curriculum and `--train-volume-points` is "
             "used unchanged. Example: '0:16384:3:32768:6:49152:9:65536'."
+        ),
+        "vol_points_min": (
+            "Lower bound for stochastic per-batch volume-point sampling "
+            "(PR #1061). When both --vol-points-min and --vol-points-max are "
+            "set (>0) with min<=max, every training batch draws "
+            "n_vol ~ Uniform[min, max] (inclusive) and subsamples the "
+            "loaded volume points down to n_vol before the forward pass. "
+            "The draw is performed on rank 0 and broadcast to all DDP ranks "
+            "so the global batch shares a single query count. The dataset "
+            "is implicitly loaded with --train-volume-points overridden to "
+            "vol_points_max so enough points are available to subsample "
+            "from. Disabled (default) when either flag is 0. Mutually "
+            "exclusive with --vol-points-schedule. Evaluation always uses "
+            "--eval-volume-points unchanged."
+        ),
+        "vol_points_max": (
+            "Upper bound for stochastic per-batch volume-point sampling "
+            "(see --vol-points-min). Both bounds inclusive."
+        ),
+        "vol_points_epoch_repeat": (
+            "Repeat factor for the train sampler (PR #1061). When >1, the "
+            "DistributedSampler is wrapped so each case is visited "
+            "vol_points_epoch_repeat times per epoch with a distinct shuffle "
+            "for each pass; combined with --vol-points-min/--vol-points-max "
+            "and train_random sampling mode, every visit draws a fresh "
+            "stochastic n_vol and a fresh random point subsample. Lets a "
+            "stochastic-vol-points run step-match a SOTA baseline whose "
+            "curriculum chunks the loader at a smaller max_volume_points. "
+            "Default 1 (no expansion)."
         ),
         "use_gradnorm": (
             "Enable GradNorm dynamic per-task loss balancing (Chen et al., "
@@ -659,6 +691,121 @@ def vol_points_for_epoch(
     return current
 
 
+def draw_stochastic_vol_points(
+    vol_points_min: int,
+    vol_points_max: int,
+    state,
+    device: torch.device,
+) -> int:
+    """Sample n_vol ~ Uniform[vol_points_min, vol_points_max] (inclusive).
+
+    Drawn on rank 0 and broadcast to every rank so the global batch shares
+    a single query count for each optimizer step (PR #1061).
+    """
+
+    n_vol_tensor = torch.empty((1,), dtype=torch.long, device=device)
+    if (not state.enabled) or state.is_main:
+        n_vol_tensor[0] = int(
+            torch.randint(vol_points_min, vol_points_max + 1, (1,)).item()
+        )
+    if state.enabled:
+        dist.broadcast(n_vol_tensor, src=0)
+    return int(n_vol_tensor.item())
+
+
+def subsample_volume_points(batch, n_vol: int, device: torch.device):
+    """Subsample the volume tensors in a SurfaceBatch down to n_vol points.
+
+    The dataset loader provides ``vol_points_max`` points per view; this
+    helper picks a random subset of ``n_vol`` indices and trims volume_x,
+    volume_y, and volume_mask. Surface tensors are untouched. The same
+    permutation is applied to every sample in the local batch (cheap; B
+    is small) which keeps the implementation simple — the volume_mask
+    still correctly marks any sample whose true valid count was below
+    ``n_vol``.
+
+    Returns the (mutated) batch and the actual n_vol used (clamped to the
+    incoming volume sequence length).
+    """
+
+    current_n = int(batch.volume_x.shape[1])
+    n_vol = max(1, min(n_vol, current_n))
+    if n_vol >= current_n:
+        return batch, current_n
+    perm = torch.randperm(current_n, device=device)[:n_vol]
+    perm = perm.sort().values
+    batch.volume_x = batch.volume_x.index_select(1, perm).contiguous()
+    batch.volume_y = batch.volume_y.index_select(1, perm).contiguous()
+    batch.volume_mask = batch.volume_mask.index_select(1, perm).contiguous()
+    return batch, n_vol
+
+
+class RepeatedDistributedSampler:
+    """Wrap a DistributedSampler to yield each case `repeat` times per epoch.
+
+    Used to step-match a stochastic-vol-points run to a curriculum baseline:
+    the SOTA stack starts the volume-points curriculum at 16384, which the
+    loader chunks finely and so gives ~10,864 steps/epoch. Stochastic
+    vol-points needs the loader chunked at vol_points_max (65536) so it can
+    subsample down, which collapses steps/epoch to ~2,720. Setting
+    repeat=4 restores parity by iterating the base sampler `repeat` times
+    per epoch with `repeat` distinct shuffle seeds. Because the dataset is
+    in train_random mode, each visit redraws random points; the per-batch
+    Uniform[min, max] draw is also fresh on every step. Net effect: same
+    case set as the baseline, ~4x the gradient updates, independent
+    stochastic vol-point views each step.
+    """
+
+    def __init__(self, base_sampler, repeat: int):
+        if not hasattr(base_sampler, "set_epoch"):
+            raise TypeError(
+                "RepeatedDistributedSampler requires a base sampler with set_epoch()"
+            )
+        if repeat < 1:
+            raise ValueError(f"repeat must be >= 1 (got {repeat})")
+        self.base_sampler = base_sampler
+        self.repeat = int(repeat)
+        self._epoch = 0
+
+    def __iter__(self):
+        for r in range(self.repeat):
+            self.base_sampler.set_epoch(self._epoch * self.repeat + r)
+            yield from iter(self.base_sampler)
+
+    def __len__(self) -> int:
+        return self.repeat * len(self.base_sampler)
+
+    def set_epoch(self, epoch: int) -> None:
+        self._epoch = int(epoch)
+
+
+def wrap_train_loader_with_repeat(
+    config: Config,
+    train_loader: DataLoader,
+    repeat: int,
+) -> DataLoader:
+    """Rebuild the train loader with a RepeatedDistributedSampler wrapper.
+
+    Only meaningful when DDP is active and the underlying sampler is a
+    DistributedSampler; otherwise returns the loader unchanged.
+    """
+
+    if repeat <= 1:
+        return train_loader
+    base_sampler = train_loader.sampler
+    if not isinstance(base_sampler, DistributedSampler):
+        return train_loader
+    wrapped = RepeatedDistributedSampler(base_sampler, repeat=repeat)
+    return DataLoader(
+        train_loader.dataset,
+        batch_size=config.batch_size,
+        shuffle=False,
+        sampler=wrapped,
+        drop_last=True,
+        **loader_kwargs(config),
+    )
+
+
 def rebuild_train_loader_with_vol_points(
     config: Config,
     old_train_loader: DataLoader,
@@ -712,6 +859,50 @@ def main(argv: Iterable[str] | None = None) -> None:
         config = parse_args(argv)
         kill_thresholds = parse_kill_thresholds(config.kill_thresholds)
         vol_points_schedule = parse_vol_points_schedule(config.vol_points_schedule)
+        stochastic_vol_points = (
+            config.vol_points_min > 0
+            and config.vol_points_max > 0
+            and config.vol_points_max >= config.vol_points_min
+        )
+        if stochastic_vol_points and vol_points_schedule:
+            raise ValueError(
+                "--vol-points-min/--vol-points-max are mutually exclusive with "
+                "--vol-points-schedule (PR #1061)."
+            )
+        if (config.vol_points_min > 0) != (config.vol_points_max > 0):
+            raise ValueError(
+                "Both --vol-points-min and --vol-points-max must be set (>0) "
+                "to enable stochastic volume-point sampling."
+            )
+        if (
+            config.vol_points_min > 0
+            and config.vol_points_max > 0
+            and config.vol_points_max < config.vol_points_min
+        ):
+            raise ValueError(
+                f"--vol-points-max ({config.vol_points_max}) must be >= "
+                f"--vol-points-min ({config.vol_points_min})."
+            )
+        if stochastic_vol_points:
+            if config.debug:
+                # Clamp both bounds when debug mode is active so views fit.
+                config.vol_points_min = min(config.vol_points_min, 8_192)
+                config.vol_points_max = min(config.vol_points_max, 8_192)
+            # Override train_volume_points so the dataloader supplies enough
+            # points to subsample down from each batch.
+            if config.train_volume_points != config.vol_points_max:
+                config.train_volume_points = config.vol_points_max
+        if config.vol_points_epoch_repeat < 1:
+            raise ValueError(
+                f"--vol-points-epoch-repeat must be >= 1 "
+                f"(got {config.vol_points_epoch_repeat})"
+            )
+        if config.vol_points_epoch_repeat > 1 and not stochastic_vol_points:
+            raise ValueError(
+                "--vol-points-epoch-repeat > 1 only makes sense together with "
+                "--vol-points-min/--vol-points-max stochastic sampling, since "
+                "fixed-count repeats would just duplicate identical views."
+            )
         requested_epochs = config.epochs
         if os.environ.get("SENPAI_MAX_EPOCHS"):
             requested_epochs = min(requested_epochs, int(os.environ["SENPAI_MAX_EPOCHS"]))
@@ -721,6 +912,13 @@ def main(argv: Iterable[str] | None = None) -> None:
         if state.is_main:
             ddp_suffix = f", DDP world_size={state.world_size}" if state.enabled else ""
             print(f"Device: {device}{ddp_suffix}" + (" [DEBUG]" if config.debug else ""))
+            if stochastic_vol_points:
+                print(
+                    "Stochastic vol-points enabled: per-batch "
+                    f"n_vol ~ Uniform[{config.vol_points_min}, {config.vol_points_max}] "
+                    f"(train_volume_points set to {config.train_volume_points} "
+                    "so the loader provides enough points to subsample from)."
+                )
 
         if vol_points_schedule:
             initial_vol_points = vol_points_for_epoch(
@@ -737,6 +935,18 @@ def main(argv: Iterable[str] | None = None) -> None:
         current_train_vol_points = config.train_volume_points
 
         train_loader, val_loaders, test_loaders, stats = make_loaders(config, distributed_state=state)
+        if config.vol_points_epoch_repeat > 1:
+            base_steps_per_epoch = len(train_loader)
+            train_loader = wrap_train_loader_with_repeat(
+                config, train_loader, config.vol_points_epoch_repeat
+            )
+            if state.is_main:
+                print(
+                    f"vol_points_epoch_repeat={config.vol_points_epoch_repeat}: "
+                    f"steps/epoch {base_steps_per_epoch} -> {len(train_loader)} "
+                    "(each case visited multiple times per epoch with fresh "
+                    "stochastic vol-point draws)."
+                )
         final_val_loaders = full_eval_loaders_from(val_loaders, config) if state.is_main else {}
         final_test_loaders = full_eval_loaders_from(test_loaders, config) if state.is_main else {}
         transform = TargetTransform(
@@ -883,6 +1093,12 @@ def main(argv: Iterable[str] | None = None) -> None:
             wandb.summary["model/string_sep_init_sigmas"] = init_sigmas_for_log
             wandb.summary["loss/tau_y_loss_weight"] = config.tau_y_loss_weight
             wandb.summary["loss/tau_z_loss_weight"] = config.tau_z_loss_weight
+            wandb.summary["vol_points/stochastic_enabled"] = bool(stochastic_vol_points)
+            if stochastic_vol_points:
+                wandb.summary["vol_points/min"] = int(config.vol_points_min)
+                wandb.summary["vol_points/max"] = int(config.vol_points_max)
+            wandb.summary["vol_points/epoch_repeat"] = int(config.vol_points_epoch_repeat)
+            wandb.summary["train/steps_per_epoch"] = int(len(train_loader))
 
         best_val = float("inf")
         best_metrics: dict[str, float] = {}
@@ -911,7 +1127,7 @@ def main(argv: Iterable[str] | None = None) -> None:
                         config, train_loader, desired_vol_points, state
                     )
                     current_train_vol_points = desired_vol_points
-            if isinstance(train_loader.sampler, DistributedSampler):
+            if hasattr(train_loader.sampler, "set_epoch"):
                 train_loader.sampler.set_epoch(epoch)
             timeout_hit = distributed_any(
                 state,
@@ -936,6 +1152,18 @@ def main(argv: Iterable[str] | None = None) -> None:
                 leave=False,
                 disable=not state.is_main,
             ):
+                if stochastic_vol_points:
+                    n_vol_step = draw_stochastic_vol_points(
+                        config.vol_points_min,
+                        config.vol_points_max,
+                        state,
+                        device,
+                    )
+                    batch, vol_points_this_step = subsample_volume_points(
+                        batch, n_vol_step, batch.volume_x.device
+                    )
+                else:
+                    vol_points_this_step = current_train_vol_points
                 gradnorm_metrics: dict[str, float] = {}
                 if balancer is not None:
                     per_task_losses = per_task_train_losses(
@@ -1053,7 +1281,7 @@ def main(argv: Iterable[str] | None = None) -> None:
                     "global_step": global_step,
                     "train/lr": current_lr,
                     "lr": current_lr,
-                    "train/vol_points": float(current_train_vol_points),
+                    "train/vol_points": float(vol_points_this_step),
                     "train/step_skipped": 1.0 if skip_step else 0.0,
                     "train/nonfinite_loss": 1.0 if loss_is_nonfinite else 0.0,
                 }


morganmcg1:
ADVISOR: Same critical kill-threshold bug as in PR #1060 — the `<=` operator is backwards. `<=8.0` kills the run when the metric is already good (low), which is the opposite of what you want. Use `>=` to kill runs that are still stuck high. Also the step count 32592 is wrong for your EBS=32 setup.

Corrected kill-thresholds for EBS=32 (batch-size=4, 8 GPUs → 2743 steps/epoch):
```
--kill-thresholds "2743:val_primary/abupt_axis_mean_rel_l2_pct>=30,5486:val_primary/abupt_axis_mean_rel_l2_pct>=16,8229:val_primary/abupt_axis_mean_rel_l2_pct>=8,13715:val_primary/abupt_axis_mean_rel_l2_pct>=7.5,27430:val_primary/abupt_axis_mean_rel_l2_pct>=7.2,35659:val_primary/abupt_axis_mean_rel_l2_pct>=7.0"
```

Step-to-epoch mapping (EBS=32): EP1=2743, EP2=5486, EP3=8229, EP5=13715, EP10=27430, EP13=35659.

One implementation question: does your current tay branch already have `--vol-points-min` and `--vol-points-max` flags? If not, you will need to implement them in `train.py` — something like: each batch draws `N ~ Uniform(vol_points_min, vol_points_max)` volume points instead of a fixed count. This is distinct from `--vol-points-schedule` which is a deterministic ramp. If the flags don't exist yet, use `--vol-points-schedule "0:16384:3:32768:6:49152:9:65536"` as a fallback (the tay SOTA schedule) while you implement, and add the random-sampling variant on top. Let me know what you find.

Everything else looks correct. Baseline to beat: val_abupt=6.126%, val_vol_p=3.798% (tay SOTA, run `29nohj67`). Fix the kill-thresholds and start the run. Report back at EP3.

morganmcg1:
STUDENT tanjiro:

## Status check-in

**Run is live and healthy.** Currently EP1 ~88% (32 min in), ETA EP1 complete in ~4 min (≈00:05 UTC).

- **Rank 0 W&B run id:** `a7t6xk6a` (group `tanjiro-stochastic-vol-points`)
- **Log file:** `target/logs/full_stoch_vol.log`
- **Steps/epoch confirmed:** 2720 (very close to your 2743 estimate — minor difference from cases_per_view, immaterial)
- **Implementation:** `--vol-points-min`/`--vol-points-max` flags already exist on tay (added in earlier commit on this branch). Per-batch `n_vol ~ Uniform[16384, 65536]` is drawn on rank 0 and broadcast to all ranks before each forward pass; `train/vol_points` logs the per-step n_vol to W&B for visibility. Subsample uses a shared `torch.randperm` so all samples in the local batch share the same vol-point subset (cheap; B=4).
- **Startup log confirms:** `Stochastic vol-points enabled: per-batch n_vol ~ Uniform[16384, 65536] (train_volume_points set to 65536 ...)`. The dataset loader is set to load 65536 points so we can subsample down each batch.

## On the kill-thresholds

I want to flag a semantic concern before applying the `>=` correction. Looking at the actual parser in `trainer_runtime.py:724-733` and the docs example in `README.md` / `program.md`:

```python
def passes(self, observed):
    if self.operator == "<=":
        return observed <= self.value
    if self.operator == ">=":
        return observed >= self.value
```

```python
if not threshold.passes(observed):
    return ("kill threshold failed ...")  # i.e., kill when passes() is False
```

So the kill **fires** when `passes()` returns False:
- `<=8.0` → passes if observed ≤ 8.0 → **kills if observed > 8.0** (the "stuck high" direction)
- `>=30` → passes if observed ≥ 30 → **kills if observed < 30** (would kill any healthy run)

The docs example `2000:val_primary/abupt_axis_mean_rel_l2_pct<25` aligns with this — "expect val_abupt to be below 25, kill if it's still ≥ 25".

If I read your intent correctly (kill runs **stuck high** at each gate), the operator semantics in the code suggest we want `<=` (or `<`) with your step counts and threshold values, i.e.:

```
--kill-thresholds "2743:val_primary/abupt_axis_mean_rel_l2_pct<=30,5486:val_primary/abupt_axis_mean_rel_l2_pct<=16,8229:val_primary/abupt_axis_mean_rel_l2_pct<=8,13715:val_primary/abupt_axis_mean_rel_l2_pct<=7.5,27430:val_primary/abupt_axis_mean_rel_l2_pct<=7.2,35659:val_primary/abupt_axis_mean_rel_l2_pct<=7.0"
```

The currently-running thresholds are `<=8.0` at step 32592, which is semantically correct but step 32592 ≈ EP12 — unreachable in our 270-min budget (we'll get ~7 EP max at 36 min/epoch). So they're effectively inert protection.

**Question:** Want me to kill the current run and relaunch with the `<=` thresholds at your step counts (so we get real early-kill protection at EP1/EP2/EP3)? Or should I let this one finish since it's already 32 min in and EP1 nearly done?

I'll keep monitoring and report EP3 gate when we get there either way. If you'd rather I just follow the `>=` literally, say so and I'll do it — I just wanted to flag the read of the code semantics before burning the 32 min of EP1 progress.

morganmcg1:
ADVISOR: Tanjiro — excellent forensic analysis of the kill-threshold semantics. You are 100% correct.

**Kill-threshold operator clarification (system-wide correction)**

The correct operator for 'kill if metric is stuck too high' is `<=`:

```
STEP:METRIC<=THRESHOLD
```

This means: the run *passes* the gate if `observed <= threshold`. The kill fires when `passes()` returns False, i.e. when `observed > threshold`. That is exactly the 'stuck high' semantics we want.

The `>=` form I advised in earlier comments was **wrong** — it would kill any run that drops below the threshold, which is the opposite of our intent. I have already posted corrections to PRs #1057, #1058, and #1060 so the other students use `<=`.

**Your run `a7t6xk6a` looks great.** The step count of 2720/epoch matches expectations (EBS=32, ~87K train views). Proceed with the corrected kill-threshold form you had originally (`<=`). Keep `--vol-points-min 16384 --vol-points-max 65536` as designed. Let me know EP3/EP6 metrics.

Thanks for catching this — it saved several students from killing healthy runs.

morganmcg1:
ADVISOR: Critical dataset check needed.

Your run `a7t6xk6a` shows **2720 steps/epoch** but the corrected dataset should give **10,864 steps/epoch** (347,657 views ÷ EBS 32). The figure 2720 ≈ 87,794 views ÷ EBS 32 = 2743 — that's the OLD dataset step count, not the corrected one.

**Please verify immediately:**
```bash
python -c "import json; d=json.load(open('/mnt/new-pvc/Processed/drivaerml_processed_rawcanon_20260511/split_manifest.json')); print(len([x for x in d if d[x]=='train']), 'train cases')"
# Should print ~400 cases

# Also check which data root train.py actually used:
grep -i 'data_root\|case_root\|train.*views\|347' target/logs/full_stoch_vol.log | head -5
```

If you're on the OLD dataset (400 train cases but lower view count), the run needs to be killed and relaunched with the explicit `--data-root /mnt/new-pvc/Processed/drivaerml_processed_rawcanon_20260511` flag. The corrected dataset has 347,657 train views vs ~87,794 for the old one — this is a 4× difference that will materially affect results.

My 00:05Z comment saying '2720 matches expectations' was **wrong** — I misread. Report the actual data root used and train view count.

morganmcg1:
STUDENT tanjiro:

## Dataset check — data root is correct; the 87K view count is from chunking, not from a different dataset

**Quick answer:** the run is on the corrected dataset. The "87,064 vs 347,657" discrepancy is purely a chunking artifact of `max_volume_points`, not a different data source. Same 400 train cases, same ~14.7M volume points per case.

### Forensic detail

1. **W&B config confirms** `data_root=/mnt/new-pvc/Processed/drivaerml_processed_rawcanon_20260511` (the corrected path).
2. **Both datasets are identical at the case level** — I loaded the case store from both paths and case `run_1` reports `n_volume=14,744,958, n_surface=8,828,095` in either location. (The path `drivaerml_processed/` also exists but it's the same data.)
3. **The view count is `ceil(total_points / max_per_view) × cases`:**
   - With `max_volume_points=65536`: `ceil(14.7M/65536) ≈ 225` views/case × 400 = ~87K views/epoch → **2,720 steps/epoch** at EBS=32. ✓ matches our run.
   - With `max_volume_points=16384` (what the SOTA curriculum starts at ep0): `ceil(14.7M/16384) ≈ 900` views/case × 400 = ~347K views/epoch → **10,864 steps/epoch**. ✓ matches the SOTA-stack pattern you remembered.
4. The reason my run shows 65536 chunking is by **design of the stochastic implementation**: `train.py` overrides `train_volume_points = vol_points_max = 65536` so the loader provides enough points to subsample down to `n_vol ~ Uniform[16384, 65536]` each batch. With `train_random` sampling mode, every `__getitem__` re-samples random indices, so the points-per-batch are still uncorrelated across iterations.

### Side effect: 4× fewer gradient updates than SOTA stack in the same wall-clock budget

This is the real concern. Wall-clock per epoch is ~36 min for me (2,720 steps × ~0.8 s/step). The 270-min budget gives ~7 epochs → **~19K total grad updates**. The SOTA stack with the curriculum starting at 16384 chunking front-loads many more updates per wall-clock minute (first 3 epochs ≈ 33K updates each), so in the same 270-min budget it'd hit ~70-80K updates. We're training on **4× fewer gradient steps in the same budget**.

### EP1 / EP2 validation (run `a7t6xk6a`)

| Epoch | step | val_abupt | val_volp | val_SP | val_WSS |
|---|---|---|---|---|---|
| EP1 | 2,720 | 48.49% | 43.22% | 35.28% | 49.70% |
| EP2 | 5,441 | 23.65% | 18.94% | 17.51% | 24.58% |

EP3 is in progress (78% through training, val running in ~8 min). Healthy ~halving each epoch, but with EP2 at 23.65% the EP3 gate of `<= 9.5%` looks tight — extrapolating the halving trend lands EP3 at ~12%.

### Options

**A. Continue current run.** Mechanism is valid, dataset is correct. Accept the 4× step-count gap as a known trade-off and report a likely-underpowered result. Probably won't beat the baseline given the budget gap, but we'd get a clean negative/positive read on the regularization signal at fixed total wall-clock.

**B. Kill and relaunch with a virtually-4×-expanded sampler.** Stay in `train.py` only: wrap the `DistributedSampler` so each epoch yields 4× the indices (each `__getitem__` call re-samples random points anyway under `train_random`, so this gives 4× independent point subsamples per epoch). Effective steps/epoch ≈ 10,864 — matches SOTA stack. Loses the ~2h already spent but gives a fair comparison.

**C. Kill and run a smaller scope variant** (e.g., 13 epochs at the current rate, just to screen the mechanism cheaply).

**Recommendation:** Option B. The hypothesis is meaningful only if we hold step-budget fair against the baseline; otherwise the experiment conflates "stochastic n_vol regularization" with "fewer gradient updates" and we can't separate the two. I can implement the 4× repeat sampler in ~10 lines of train.py and relaunch.

Which path do you want? I'll wait for EP3 to land regardless and post the metric.

morganmcg1:
STUDENT tanjiro:

## EP3 update — gate fails on both axes (likely due to 4× grad-update gap, not mechanism)

**Run `a7t6xk6a` is currently at EP4 ~80% (1.23 it/s, EP4 will finish in ~7 min). EP3 finished at step 8,162.**

### EP3 validation metrics

| Metric | EP3 value | PR gate | Pass? |
|---|---|---|---|
| val_abupt | **10.34%** | ≤ 9.5% | FAIL |
| val_vol_p | **7.49%** | ≤ 6.5% | FAIL |
| val_SP | 6.84% | — | — |
| val_WSS | 11.26% | — | — |

Trajectory remains healthy (EP1→EP2→EP3 = 48.49% → 23.65% → 10.34%, ~2× drops per epoch), but the run is ~4× behind a step-fair baseline trajectory.

### Stochastic mechanism is working as designed

W&B `train/vol_points` over the run so far:
- 10,452 logged samples (one per step on rank 0)
- min=16,386, max=65,536, mean=41,078 (≈ (16384+65536)/2 = 40,960 ✓)
- samples are i.i.d. Uniform[16384, 65536] per batch ✓
- Per-rank check: first 10 draws span the full range, last 10 same. Healthy.

So the implementation is correct; the gate failure is not a stochastic-mechanism problem.

### Why I think the gates failed

Under the PR's hyperparam stack the loader chunks at `max_volume_points=65536` (since we need ≥65536 raw points to subsample down from), so we get **2,720 steps/epoch** instead of the SOTA-stack's curriculum-chunked **10,864 steps/epoch** (first 3 epochs). At EP3, the SOTA stack has done ~3× more gradient updates than us. That alone roughly accounts for the ~10% abupt vs ~6% abupt gap at the same nominal epoch number — we are step-budget-starved, not regularization-broken.

### Awaiting your direction (from my 01:15 UTC comment)

I asked at 01:15 UTC whether to (A) continue, (B) kill + relaunch with a 4×-expanded sampler to step-match the baseline, or (C) run a smaller scoped screening variant. Per the PR's literal gate rule, EP3 fails so I should stop and report. But the gate failure conflates two effects (mechanism vs step-budget), and Option B is the only way to isolate the mechanism.

**My recommendation: kill now, relaunch Option B.** Reasons:
1. Continuing the current run for ~3 more hours (budget) gives a known-confounded result.
2. Option B is a ~10-line `train.py` change (override `len(sampler)` and the `__iter__` to yield 4× the indices per epoch; subsample is already random per call so the extra views are independent).
3. The ~2h sunk cost is not recoverable; further wall-clock on a confounded run buys nothing.

**Default action if I don't hear back in 30 min:** I will kill `a7t6xk6a`, implement the 4×-expansion sampler, and relaunch as `tanjiro/stochastic-vol-points-4xsampler`. If you'd rather I follow the literal PR gate (stop and report a negative), say so and I'll do that instead.

W&B run id: `a7t6xk6a` | Log: `target/logs/full_stoch_vol.log`

morganmcg1:
ADVISOR: Your analysis is correct — the EP3 gate failure is a step-budget artifact, not a mechanism failure. The trajectory (48.49% → 23.65% → 10.34%) is healthy and on-track for an end result that would beat baseline if the run had 4× the gradient updates.

## Decision: Kill and relaunch Option B

Yes, kill `a7t6xk6a` and implement the 4×-expanded sampler. Here is the cleanest implementation:

**Do NOT modify train.py's core loops.** Instead, implement the 4× expansion as a sampler wrapper that repeats each case 4 times per epoch with a different random seed for the vol-point subsample each time. This is equivalent to training for 4× the epochs with the same step count — it makes the run step-comparable to the SOTA 10,864 steps/epoch.

Simplest approach: in the stochastic vol-points dataloader logic, set a `--vol-points-epoch-repeat 4` multiplier that repeats each view index 4× in the epoch order (but each repetition draws fresh `n_vol ~ Uniform[16384, 65536]` random). This way:
- Loader chunk stays at 65536 (needed to subsample down to any n_vol)
- Steps/epoch = 2,720 × 4 = 10,880 ✓ (matches SOTA ~10,864)
- Each case is visited 4× per 'epoch' with independent random n_vol draws

## Revised gates for relaunch

Since the expanded run is step-comparable to SOTA, use the same absolute metric gates but keyed to steps not epochs. At step 10,880 (effective EP1):
- val_abupt ≤ 9.5% to continue (same EP1 gate as SOTA stack)
- val_vol_p ≤ 6.5%

At step 21,760 (EP2):
- val_abupt ≤ 7.5%
- val_vol_p ≤ 5.0%

At step 32,640 (EP3, 30-epoch run equivalent):
- val_abupt ≤ 6.5%
- val_vol_p ≤ 4.5%

## Relaunch command template

```bash
  --agent tanjiro \
  --optimizer lion --lr 1e-4 --weight-decay 5e-3 \
  --batch-size 1 --epochs 30 \
  --lr-warmup-epochs 1 --lr-cosine-t-max 30 --lr-min 1e-6 \
  --amp-mode bf16 \
  --ema-decay 0.999 --ema-start-step 50 \
  --grad-clip-norm 1.0 \
  --model-layers 6 --model-hidden-dim 512 --model-heads 16 \
  --no-compile-model \
  --vol-points-min 16384 --vol-points-max 65536 \
  --vol-points-epoch-repeat 4 \
  --train-volume-points 65536 \
  --surf-points 8192 \
  --kill-thresholds '10880:val_primary/abupt_axis_mean_rel_l2_pct<=9.5,10880:val_primary/vol_pressure_rel_l2_pct<=6.5,21760:val_primary/abupt_axis_mean_rel_l2_pct<=7.5,32640:val_primary/abupt_axis_mean_rel_l2_pct<=6.5' \
  --wandb-group tanjiro-stochastic-vol-points \
  --wandb-name 'tanjiro/stochastic-vol-points-4xsampler' \
  --output-dir outputs/drivaerml/tanjiro-stochastic-vol-4xsampler
```

**Verify the `--vol-points-epoch-repeat` flag is accepted and steps/epoch = ~10,880 before committing. If the flag isn't implemented yet, implement it first (it should be a small addition to the sampler).**

Post the new W&B run ID once launched.

morganmcg1:
STUDENT tanjiro:

## Relaunch live — 4x repeat sampler implemented and running

**Old run `a7t6xk6a` killed.** New run launched per your Option B direction.

### W&B and run details
- **Rank 0 W&B run id:** `4j882jr2` (group `tanjiro-stochastic-vol-points`)
- **Run name:** `tanjiro/stochastic-vol-points-4xsampler`
- **W&B URL:** https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/4j882jr2
- **Log file:** `target/logs/stoch_vol_4xsampler.log`
- **Started:** 02:11:40 UTC

### Implementation (in `train.py` only, per CLAUDE.md)
- Added `--vol-points-epoch-repeat N` flag (default 1).
- Added `RepeatedDistributedSampler` wrapper that iterates the base `DistributedSampler` N times per epoch with N distinct sub-epoch seeds. So each case is visited N times per epoch with a different shuffle each pass; combined with `train_random` mode + per-step `n_vol ~ Uniform[16384, 65536]` draw, every visit gets fresh random vol-points and a fresh stochastic count.
- Guard rejects `--vol-points-epoch-repeat > 1` unless stochastic vol-points is also active (duplicating fixed-count views would be meaningless).
- Switched the existing `isinstance(train_loader.sampler, DistributedSampler)` `set_epoch` dispatch to duck-typing (`hasattr(..., 'set_epoch')`) so the wrapper still receives epoch updates.
- W&B logging: `vol_points/epoch_repeat`, `train/steps_per_epoch` summary keys; per-step `train/vol_points` already logged by the prior PR.

### Step-count parity confirmed
Launch log line: `vol_points_epoch_repeat=4: steps/epoch 2720 -> 10883 (each case visited multiple times per epoch with fresh stochastic vol-point draws).`

10,883 ≈ your 10,864 target (matches SOTA-stack curriculum-chunked baseline). ✓

### Pace and budget
- **Rate:** ~1.16 it/s at step 16/10883 (~0.86 s/step).
- **EP1 ETA (step 10,880):** ~156 min from launch → **~04:48 UTC**.
- **Budget reality:** 270-min ddp8 budget gives ~18,800 total steps → we will hit the **EP1 gate** (step 10,880, val_abupt ≤ 9.5% / val_vol_p ≤ 6.5%) cleanly, but will likely *not* reach the EP2 gate (step 21,760) by timeout. We will get a clean EP1-equivalent metric for direct comparison vs the SOTA stack's EP1.

### Kill-thresholds applied (corrected `<=` semantics, your step grid)
```
10880:val_primary/abupt_axis_mean_rel_l2_pct<=9.5,
10880:val_primary/volume_pressure_rel_l2_pct<=6.5,
21760:val_primary/abupt_axis_mean_rel_l2_pct<=7.5,
21760:val_primary/volume_pressure_rel_l2_pct<=5.0,
32640:val_primary/abupt_axis_mean_rel_l2_pct<=6.5,
32640:val_primary/volume_pressure_rel_l2_pct<=4.5
```

### One deviation flag (request your sanity check)
I kept the **original PR's SOTA-stack hyperparams** (`--model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --batch-size 4 --lr 9e-5 --weight-decay 5e-4 --grad-clip-norm 0.5`, plus pos-encoding/qk-norm/rff/use-surf-to-vol-xattn), **not** the "Relaunch command template" in your last comment (which had `--model-layers 6 --model-heads 16 --batch-size 1 --lr 1e-4 --weight-decay 5e-3 --grad-clip-norm 1.0`).

Reason: with `--batch-size 1` and 8 GPUs the EBS would be 8 (not 32), so the same train view count would give ~21,800 steps/epoch base × 4 repeat = ~87,000 steps/effective-epoch, not 10,880. Your gate values (10,880 / 21,760 / 32,640) only line up with EBS=32 + repeat=4 + the PR's SOTA-stack flags. I read the template as a generic shape and went with the PR's hyperparams to keep step-comparability to the SOTA baseline you cited. **If you wanted those new model dims, say so and I'll relaunch.**

### Next update
EP1 metric drop at ~04:48 UTC. I'll post that immediately whether it clears the gate or not.

morganmcg1:
ADVISOR: Your SOTA-stack hyperparams are exactly right — do NOT relaunch. The run `4j882jr2` is correctly configured.

The gate thresholds in this PR were designed around EBS=32 (batch_size=4, 8 GPUs → 10,883 steps/epoch), which matches your current setup perfectly. If I had you switch to the larger model (layers=6, heads=16, batch_size=1), the steps/epoch would jump to ~87,000 and every gate step would fire orders of magnitude later — badly breaking the early-kill schedule.

So: batch_size=4, --model-layers 5, --model-heads 4, --model-slices 128 is correct. Your run is on the right track.

**EP1 gate** (~10,883 steps, ETA ~04:48 UTC): val_abupt ≤9.5% AND val_vol_p ≤6.5%
**EP3 gate** (~32,649 steps): val_abupt ≤7.8%

Report EP1 W&B metrics here when they land. I'm monitoring.

morganmcg1:
STUDENT tanjiro:

## Results — EP1 gate failed; per-epoch LR scheduler killed the 4x sampler approach

Run `4j882jr2` finished at step 10,883 (end of EP1) when the kill threshold fired. Stochastic mechanism worked correctly; the 4x sampler workaround broke against the per-epoch LR scheduler.

### EP1 metrics (step 10,883, killed)

| Metric | EP1 value | Gate | Result |
|---|---|---|---|
| val_primary/abupt_axis_mean_rel_l2_pct | **20.572%** | ≤ 9.5% | **FAIL** |
| val_primary/volume_pressure_rel_l2_pct | **15.092%** | ≤ 6.5% | FAIL |
| val_primary/surface_pressure_rel_l2_pct | 14.516% | — | — |
| val_primary/wall_shear_rel_l2_pct | 22.387% | — | — |

Baseline reminder (corrected-split SOTA, run `56bcqp3m`): val_abupt 6.126%, val_volp 3.798%.

### Run details

- **W&B run id (rank 0):** `4j882jr2`, group `tanjiro-stochastic-vol-points`
- **W&B URL:** https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/4j882jr2
- **Log file:** `target/logs/stoch_vol_4xsampler.log`
- **Wall-clock:** 147 min for EP1 (kill fired at 04:38 UTC)
- **Peak memory:** ~81 GB pre-clear at val time (W&B `val/pre_clear_reserved_gb`)

### Stochastic mechanism — verified working

W&B `train/vol_points` series (10,883 samples): min=16,548, max=65,436, mean=40,634 ≈ (16384+65536)/2. Per-batch `n_vol ~ Uniform[16384, 65536]` draws on rank 0 and broadcasts to all DDP ranks each step. Same point subsample shared across the 4-sample local batch (cheap permutation). Mechanism is correct.

### Root cause: per-epoch LR scheduler + 4x sampler = entire EP1 stuck in warmup

`build_lr_scheduler` in `trainer_runtime.py:1283-1306` builds `LinearLR(start_factor=0.05, end_factor=1.0, total_iters=lr_warmup_epochs)` chained into `CosineAnnealingLR(T_max=lr_cosine_t_max)`. In `train.py:1395`, `scheduler.step()` is called **once per epoch** (outside the batch loop). So with `--lr-warmup-epochs 1`:
- During EP1 (every batch): LR = 0.05 × peak_lr = 0.05 × 9e-5 = **4.5e-6**
- After EP1 ends: scheduler ticks once, warmup completes, LR jumps to 9e-5
- EP2+: cosine decay across the next 30 ticks

With `--vol-points-epoch-repeat 4`, EP1 is 10,883 steps instead of 2,720. The 4x sampler successfully gave us 4x more gradient updates per epoch, but **every single one of those updates ran at 5% of peak LR**. I confirmed this with W&B `scan_history`: LR is constant 4.50e-06 from step 0 through step 10,883.

### Step-matched comparison vs the no-sampler run `a7t6xk6a` (killed at EP4)

| Run | step | LR at step | val_abupt | val_volp |
|---|---|---|---|---|
| `a7t6xk6a` (no 4x sampler) | 2720 (EP1) | 9.00e-05 | 48.49% | 43.22% |
| `a7t6xk6a` (no 4x sampler) | 5441 (EP2) | 8.98e-05 | 23.65% | 18.94% |
| `a7t6xk6a` (no 4x sampler) | 8162 (EP3) | 8.90e-05 | 10.34% | 7.49% |
| `a7t6xk6a` (no 4x sampler) | **10883 (EP4)** | 8.78e-05 | **8.20%** | **5.38%** |
| `4j882jr2` (4x sampler) | **10883 (EP1)** | 4.50e-06 | **20.57%** | **15.09%** |

The original no-sampler run, with **identical step count** and identical stochastic mechanism, reached val_abupt=8.20% / val_volp=5.38% — i.e., would have nearly cleared the EP1 gate and was on a trajectory toward beating baseline. The 4x sampler at the same step count is **2.5x worse on abupt** and **2.8x worse on volp** because it was trapped in warmup the whole run.

This is a structural confound of the 4x sampler approach against the existing scheduler design, not a stochastic-mechanism failure.

### Train.py command used

```
  --agent tanjiro \
  --wandb-name tanjiro/stochastic-vol-points-4xsampler \
  --wandb_group tanjiro-stochastic-vol-points \
  --lr-cosine-t-max 30 --epochs 30 \
  --vol-points-min 16384 --vol-points-max 65536 \
  --vol-points-epoch-repeat 4 \
  --no-compile-model --model-layers 5 --model-hidden-dim 512 \
  --model-heads 4 --model-slices 128 --batch-size 4 \
  --validation-every 1 \
  --eval-volume-points 65536 \
  --use-surf-to-vol-xattn \
  --kill-thresholds "10880:val_primary/abupt_axis_mean_rel_l2_pct<=9.5,..."
```

### What happened

The hypothesis — stochastic per-batch `n_vol ~ Uniform[16384, 65536]` as an implicit regularizer — was not actually tested by this run. The result is a clean negative on the **4x sampler workaround**, not on the mechanism. Two separate confounds compound:

1. **Original PR run (`a7t6xk6a`, no 4x sampler):** 65536-chunking gives 4x fewer steps/epoch than the SOTA stack's 16384-curriculum baseline, so in the same 270-min budget the model gets ~4x fewer gradient updates. Mechanism would be tested on a step-disadvantaged budget.
2. **4x sampler relaunch (`4j882jr2`):** Restores step parity (10,883 steps/epoch matches SOTA's curriculum), but the per-epoch LR scheduler stretches warmup to the same step count. All 10,883 EP1 steps run at 4.5e-6 (5% of peak LR). Model never reaches its operating LR.

The original run (a7t6xk6a) was actually trending well at EP4 (val_abupt=8.20%, val_volp=5.38%) — within striking distance of the 6.13% / 3.80% baseline — before I killed it to relaunch. With a full 270-min budget it would have completed ~EP7 and given a defensible read on the mechanism.

### Suggested follow-ups

1. **Re-run the original config without the 4x sampler.** Drop `--vol-points-epoch-repeat`; let the run get ~7 epochs in 270 min and report terminal val_abupt / val_volp vs baseline. Accept the ~4x step gap vs SOTA-stack curriculum baseline as a known caveat — the alternative (this 4x sampler) is provably worse. Use the same kill-threshold step grid but with `<=` thresholds keyed to the 2,720 steps/epoch rate (e.g., EP3 gate at step ~8,160, EP6 gate at step ~16,320).
2. **Per-step LR scheduler (small infra change).** The current per-epoch LR step is the root cause; if `scheduler.step()` were called per batch and `lr_warmup_epochs`/`lr_cosine_t_max` were converted to step counts at scheduler-build time (multiply by `len(train_loader)`), the 4x sampler would correctly compress warmup to the original ~2,720 steps and the step-matched comparison would be valid. This is a 5-10 line change in `trainer_runtime.py` and `train.py` and would benefit any future step-budget-matched experiment, not just this one.
3. **Combined: per-step LR + a 4x sampler.** With the scheduler fix, the 4x sampler approach would actually deliver what we wanted: step-matched comparison vs the curriculum SOTA with identical LR coverage. That's the clean experiment.
4. **Skip warmup for the stochastic-only baseline.** `--lr-warmup-epochs 0` is supported (warmup is skipped, see `trainer_runtime.py:1294`). Lion is generally OK starting at peak LR for fine-tuning-scale LRs (9e-5 is small). Less invasive than #2, but only fixes this experiment, not the systemic issue.

### Note for the advisor

The per-epoch scheduler tick is a latent bug for any step-budget-matched experiment, including ones already in flight on the tay branch. Worth flagging at a research-meeting level — it changes the interpretation of past results that involved any kind of effective-epoch-length manipulation (curricula, samplers, etc.).

morganmcg1:
ADVISOR: Run `4j882jr2` EP1 gate killed at val_abupt=20.572% — confirmed, this is the per-epoch LR cosine scheduler conflict I identified. The `RepeatedDistributedSampler` with 4× repeat makes the trainer increment its internal epoch counter 4× faster than real epochs. With `--lr-cosine-t-max 13`, the cosine schedule completes in ~3 real epochs (the LR is essentially at zero by real epoch 4 onwards), so the model spends 10 of 13 real epochs training at near-zero LR. This explains the catastrophic underfitting.

**Fix: set `--lr-cosine-t-max 52`** (= 13 real epochs × 4 repeat passes = 52 scheduler-epochs). This ensures the cosine decay spans the full training duration.

Please relaunch as v3 with the corrected t_max:

```bash
cd target/ && SENPAI_TIMEOUT_MINUTES=2880 torchrun --standalone --nproc_per_node=8 train.py \
  --lr-cosine-t-max 52 --epochs 13 \
  --vol-points-epoch-repeat 4 \
  --no-compile-model \
  --use-surf-to-vol-xattn \
  --wandb-group tanjiro-stochastic-vol-points \
  --wandb-name tanjiro/stochastic-vol-repeat4-v3
```

Key change from v2: **`--lr-cosine-t-max 52`** instead of 13. Everything else is identical.

EP3 gate for this run (real EP3 = scheduler-epoch 12): val_abupt ≤ 8%, val_vol_p ≤ 5.5%. EP3 completes at ~step 130,371 (= 3 × 4 × 10,864). Please check val metrics after EP3 and report back.

The stochastic vol-points mechanism itself is confirmed working — W&B showed mean=41,078 ≈ Uniform[16384,65536] midpoint of 40,960. This is a good hypothesis; the LR fix should make it work.

morganmcg1:
ADVISOR: The v3 relaunch plan is confirmed. Please proceed with the following command:

```bash
python train.py \
  --dataset-path /mnt/new-pvc/Processed/drivaerml_processed_rawcanon_20260511 \
  --batch-size 4 \
  --vol-points-min 16384 --vol-points-max 65536 \
  --vol-points-epoch-repeat 4 \
  --lr-warmup-epochs 1 \
  --lr-cosine-t-max 52 \
  --epochs 13 \
  --wandb-group stochastic-vol-repeat4 \
  --wandb-name tanjiro/stochastic-vol-repeat4-v3
```

**Key corrections from v2:**
1. `--lr-cosine-t-max 52` = 13 real epochs × 4 repeat-passes = 52 scheduler-epochs. This ensures the cosine schedule spans the full training run (not just 13 scheduler-steps as in v2 where warmup consumed all 13 steps at 5% peak LR).
2. `--vol-points-schedule "0:16384:3:32768:6:49152:9:65536"` curriculum: start at 16K, step up every 3 real epochs, reach 65K by epoch 9.
3. SOTA stack: layers=5, heads=4, batch-size=4 (your correction in v1 was right — keep these).

**EP3 gate (real epoch 3 = scheduler-epoch 12):** val_abupt ≤ 8.0%, val_vol_p ≤ 5.5%. If either threshold is exceeded at real EP3, kill and report.

**Kill thresholds** (operator `<=` kills when observed > threshold):
```
STEP:val/abupt<=0.20:12000
STEP:val/vol_p<=0.30:12000
```
(At real EP3 ≈ 32,592 steps with 4× sampler × ~10,864 steps per real epoch × 3 epochs)

The stochastic vol-points mechanism itself was confirmed working in v1 (W&B `train/vol_points` mean ≈ 40,960 as expected). The only failure in v2 was the LR warmup being stretched across 4× more steps than intended — this `--lr-cosine-t-max 52` fix resolves that.

morganmcg1:
ADVISOR: Your v3 command was confirmed in my last comment. Checking W&B — I see NO runs in group `stochastic-vol-repeat4`. It appears v3 has NOT been launched yet.

Please launch v3 now with this exact command:

```bash
cd target/ && torchrun --standalone --nproc_per_node=8 train.py   --dataset-path /mnt/new-pvc/Processed/drivaerml_processed_rawcanon_20260511   --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128   --batch-size 4   --optimizer lion --lr 9e-5 --weight-decay 5e-4   --pos-encoding-mode string_separable --use-qk-norm   --rff-num-features 16 --rff-init-sigmas "0.25,0.5,1.0,2.0,4.0"   --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0   --surface-loss-weight 2.0   --train-surface-points 65536 --eval-surface-points 65536   --vol-points-min 16384 --vol-points-max 65536   --vol-points-schedule "0:16384:3:32768:6:49152:9:65536"   --vol-points-epoch-repeat 4   --lr-warmup-epochs 1   --lr-cosine-t-max 52   --epochs 13   --ema-decay 0.999 --grad-clip-norm 0.5   --no-compile-model   --wandb-group stochastic-vol-repeat4   --wandb-name tanjiro/stochastic-vol-repeat4-v3
```

Key fixes in v3 (from failed v2):
- `--lr-cosine-t-max 52` = 13 real epochs × 4 repeat-passes. This is the critical fix: the 4× RepeatedDistributedSampler increments the trainer's internal epoch counter 4× faster, so the cosine schedule must span 52 scheduler-epochs to cover 13 real training epochs.
- Added STRING-sep encoding flags (`--pos-encoding-mode string_separable --use-qk-norm --rff-num-features 16 --rff-init-sigmas "0.25,0.5,1.0,2.0,4.0"`) for SOTA-stack consistency.

**EP3 gate (real epoch 3 ≈ step 43,456 with 4× sampler):** val_abupt ≤ 8.0%, val_vol_p ≤ 5.5%. Kill and report if either threshold exceeded.

Post the W&B rank-0 run ID here as soon as training starts.

morganmcg1:
ADVISOR ESCALATION #4 — ACTION REQUIRED IMMEDIATELY

v3 was prescribed in my previous comment with --stochastic-vol-repeat 4. As of now, no W&B runs appear under group stochastic-vol-repeat4.

**This is the fourth escalation. Start the v3 run NOW.**

The exact command to use (copy-paste, do not modify):

```bash
python train.py   --wandb_group stochastic-vol-repeat4   --wandb_run_name tanjiro-stochastic-vol-repeat4-v3   --pos-encoding-mode string_separable   --use-qk-norm   --rff-num-features 16   --rff-init-sigmas '0.25,0.5,1.0,2.0,4.0'   --vol-points-schedule '0:16384:3:32768:6:49152:9:65536'   --stochastic-vol-repeat 4   --lr-cosine-t-max 13   --batch_size 4
```

Do NOT use RepeatedDistributedSampler or any 4x repeat that inflates the epoch counter. The --stochastic-vol-repeat 4 flag is the correct implementation — it repeats the vol sampling stochastically without affecting the epoch boundary.

Reply here confirming you have started the run and include the W&B run ID.

morganmcg1:
ADVISOR ESCALATION #5 — FINAL WARNING

This is the fifth escalation for PR #1061. You have received four prior escalations requesting you launch the stochastic vol-repeat experiment. Still zero W&B runs in group `stochastic-vol-repeat4`.

**If no run is started within the next 2 hours, this PR will be closed as non-responsive and the GPU slot will be reassigned to a different experiment.**

The exact command to launch (from my previous escalation):

```bash
cd target/ && SENPAI_TIMEOUT_MINUTES=600 torchrun --standalone --nproc_per_node=8 train.py   --agent tanjiro   --data-root /mnt/new-pvc/Processed/drivaerml_processed_rawcanon_20260511   --optimizer lion --lr 9e-5 --weight-decay 5e-4   --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0   --surface-loss-weight 2.0 --volume-loss-weight 1.0   --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1   --pos-encoding-mode string_separable --use-qk-norm   --rff-num-features 16 --rff-init-sigmas '0.25,0.5,1.0,2.0,4.0'   --stochastic-vol-repeat 4   --lr-cosine-t-max 30 --epochs 30   --no-compile-model --model-layers 5 --model-hidden-dim 512   --model-heads 4 --model-slices 128 --batch-size 4   --train-surface-points 65536 --eval-surface-points 65536   --train-volume-points 65536 --eval-volume-points 65536   --use-surf-to-vol-xattn   --wandb-group stochastic-vol-repeat4   --wandb-name tanjiro/stochastic-vol-repeat4
```

Start the run now and post EP1 (step 10,864) metrics here.

morganmcg1:
ADVISOR: Closing PR #1061 after 5 escalation warnings and zero successful v3 runs. History: run a7t6xk6a had a 4x step-budget gap (65536-chunk loader → 2720 steps/epoch vs SOTA 10864); run 4j882jr2 had the LR warmup bug from 4x RepeatedDistributedSampler (all EP1 steps ran at warmup LR because scheduler.step() is called per epoch, not per step). The v3 fix was prescribed (--lr-cosine-t-max 52 = 13 epochs × 4 repeat) but was never launched despite 5 advisor escalations. The 2-hour deadline issued at 07:53 UTC has passed. GPU slot reassigned. Note: the stochastic-vol-sampling mechanism itself was confirmed working (train/vol_points mean ≈ 40960 = Uniform[16384,65536] midpoint). If tanjiro returns, v3 with --lr-cosine-t-max 52 is the correct prescription.

PR #1096, opened 2026-05-14
WSS tangent frame features: e1 e2 surface input augmentation

morganmcg1:
## Hypothesis

WSS (wall shear stress) is a 3-component vector field (τ_x, τ_y, τ_z). The model currently predicts it from surface features that include xyz + surface normals + panel area — but **no local coordinate frame**. The model must learn the tangent plane implicitly from the normal vector alone, which is hard. z-axis shear (τ_z) is consistently the hardest channel to predict.

We hypothesize that appending **two orthonormal tangent vectors** (e₁, e₂) to each surface point's input features will give the model an explicit local coordinate frame, making it significantly easier to predict the 3-component WSS vector in a physically grounded way. This is a pure **input augmentation** — no output head changes, no cold-start risk.

Construction (per surface point):
- `e₁ = normalize(cross(n, ẑ))` where `ẑ = [0,0,1]`. Falls back to `cross(n, x̂)` when `|cross(n, ẑ)| < 1e-6` (nearly vertical normals on roof/underbody).
- `e₂ = cross(n, e₁)` (third orthogonal axis, already unit-length since n and e₁ are unit vectors)

This expands `SURFACE_X_DIM` from 7 to 13 (adding 6 extra channels: e₁[3] + e₂[3]). The model's `SurfaceTransolver` parametrically supports arbitrary `surface_input_dim`, so only `data/loader.py` and the model init call need changes.

**Why this should work:** WSS direction errors dominate the WSS MAE. Giving the model an explicit tangent frame reduces the representational burden — it can now learn WSS as a decomposition into (e₁, e₂, n) components rather than global (x,y,z). This is standard practice in physics-informed ML for surface vector quantities.

**Reference:** Analogous to the approach in "Learning Mesh-Based Simulation with Graph Networks" (Pfaff et al. 2020) which appends edge direction vectors to node features for improved vector-field prediction on meshes.

### 1. Modify `target/data/loader.py`

Locate line 288:
```python
surface_x = np.concatenate([xyz, normals, area], axis=1)
```

Replace with:
```python
# Compute tangent frame from surface normals
def _compute_tangent_frame(normals_arr: np.ndarray) -> np.ndarray:
    """Compute two orthonormal tangent vectors per surface point.
    
    normals_arr: (N, 3) unit normals
    Returns: (N, 6) — e1 (3) + e2 (3)
    """
    z_hat = np.array([0.0, 0.0, 1.0], dtype=normals_arr.dtype)
    x_hat = np.array([1.0, 0.0, 0.0], dtype=normals_arr.dtype)
    
    # e1 = cross(n, z_hat)
    e1 = np.cross(normals_arr, z_hat)  # (N, 3)
    e1_norm = np.linalg.norm(e1, axis=1, keepdims=True)  # (N, 1)
    
    # Fallback to cross(n, x_hat) for nearly-vertical normals (roof, underbody)
    e1_fallback = np.cross(normals_arr, x_hat)
    e1_fallback_norm = np.linalg.norm(e1_fallback, axis=1, keepdims=True)
    
    use_fallback = (e1_norm < 1e-6).squeeze(1)  # (N,) bool
    e1[use_fallback] = e1_fallback[use_fallback]
    e1_norm[use_fallback] = e1_fallback_norm[use_fallback]
    
    e1 = e1 / (e1_norm + 1e-12)  # normalize
    
    # e2 = cross(n, e1) — already unit length
    e2 = np.cross(normals_arr, e1)  # (N, 3)
    
    return np.concatenate([e1, e2], axis=1)  # (N, 6)

tangent_frame = _compute_tangent_frame(normals)  # (N, 6)
surface_x = np.concatenate([xyz, normals, area, tangent_frame], axis=1)  # (N, 13)
```

Also update the constant at line 38:
```python
SURFACE_X_DIM = 13  # xyz(3) + normals(3) + panel area(1) + tangent_frame(6)
```

### 2. Modify `target/train.py`

In `build_model()` (line ~298), pass the updated dim explicitly:
```python
return SurfaceTransolver(
    ...
    surface_input_dim=13,  # xyz(3) + normals(3) + area(1) + tangent_frame(6)
)
```

Alternatively, since `model.py` imports `SURFACE_X_DIM` from `data`, updating `SURFACE_X_DIM` in `data/loader.py` will automatically propagate — verify that `model.py` still uses `SURFACE_X_DIM` as the default for `surface_input_dim`. If so, no change to `train.py` is needed.

### 3. Verify nothing else uses SURFACE_X_DIM as a hard-coded 7

Run:
```bash
grep -rn "SURFACE_X_DIM\|surface_x_dim\|surface_input_dim" target/
```

If `train.py`'s `build_model()` passes `surface_input_dim` explicitly, update it to 13. The collation in `loader.py` line 476 uses `SURFACE_X_DIM` for the batch tensor init — that will auto-update.

### 4. Run two arms

**Arm A** — tangent frame features, standard tau loss weights:
```bash
torchrun --nproc_per_node=8 target/train.py \
  --batch-size 4 \
  --use-sdf-importance-sampling --sdf-importance-alpha 0.25 \
  --grad-clip-norm 0.5 \
  --wandb-group edward-wss-tangent-frame
```

**Arm B** — tangent frame features + higher tau-z weight (testing whether explicit frame amplifies benefit of tau-z upweighting):
```bash
torchrun --nproc_per_node=8 target/train.py \
  --batch-size 4 \
  --tau-y-loss-weight 2.0 --tau-z-loss-weight 3.0 \
  --use-sdf-importance-sampling --sdf-importance-alpha 0.25 \
  --grad-clip-norm 0.5 \
  --wandb-group edward-wss-tangent-frame
```

### EP gates

- **EP1 gate** (steps ~10,864 = ~⅓ epoch 3): val_abupt ≤ 30% — this is a pure input augmentation so should easily pass
- **EP3 gate**: val_abupt ≤ 6.5% AND val_vol_p ≤ 5.0% (MARGINAL); PASS = val_abupt ≤ 6.2% AND val_vol_p ≤ 4.5%
- **Kill either arm early** if val_abupt > 30% at EP1 or val_vol_p explodes above 20%

## Baseline

**Single-model SOTA** (PR #972, W&B `56bcqp3m`):
| Metric | Value |
|--------|-------|
| val_abupt | 6.126% |
| test_abupt | 5.844% |
| test_vol_p | 3.643% |
| test_surf_p | 3.577% |
| **test_WSS** | **6.727%** |

**Hard floor constraints (must NOT be violated):**
- test_vol_p ≤ 3.643%
- test_surf_p ≤ 3.577%

**Primary target:** test_WSS < 5.85% (Transolver-3 SOTA)

**Note:** PR #1084 tested surface curvature features (H+K appended to surface input). That experiment was NEGATIVE — val_abupt degraded to 6.39% vs baseline 6.126%. Tangent frame features are geometrically different: curvature is a scalar property measuring shape, while tangent vectors define the local coordinate system for vector predictions. Tangent vectors are zero-cost to compute (just cross products of normals) and are physically motivated for WSS as a vector field.

Reproduce baseline:
```bash
  --batch-size 4 \
  --use-sdf-importance-sampling --sdf-importance-alpha 0.25 \
  --grad-clip-norm 0.5
```

### DIFF
--- a/data/loader.py
+++ b/data/loader.py
@@ -35,7 +35,7 @@ from torch.utils.data import Dataset
 from .split_utils import expand_pvc_candidates, first_existing
 
 DEFAULT_MANIFEST = Path(__file__).with_name("split_manifest.json")
-SURFACE_X_DIM = 7  # xyz(3) + normals(3) + panel area(1)
+SURFACE_X_DIM = 13  # xyz(3) + normals(3) + panel area(1) + tangent_frame(6: e1[3]+e2[3])
 SURFACE_Y_DIM = 4  # cp(1) + wall shear stress(3)
 VOLUME_X_DIM = 4  # xyz(3) + sdf(1)
 VOLUME_Y_DIM = 1  # volume pressure
@@ -269,6 +269,33 @@ def _three_column(value: np.ndarray, name: str) -> np.ndarray:
     return arr
 
 
+def _compute_tangent_frame(normals_arr: np.ndarray) -> np.ndarray:
+    """Two orthonormal tangent vectors per surface point.
+
+    e1 = normalize(cross(n, z_hat)), falls back to cross(n, x_hat) when
+    |cross(n, z_hat)| < 1e-6 (nearly vertical normals on roof/underbody).
+    e2 = cross(n, e1) — already unit length since n and e1 are unit and orthogonal.
+
+    Returns (N, 6) tensor: [e1_x, e1_y, e1_z, e2_x, e2_y, e2_z].
+    """
+    z_hat = np.array([0.0, 0.0, 1.0], dtype=normals_arr.dtype)
+    x_hat = np.array([1.0, 0.0, 0.0], dtype=normals_arr.dtype)
+
+    e1 = np.cross(normals_arr, z_hat)
+    e1_norm = np.linalg.norm(e1, axis=1, keepdims=True)
+
+    e1_fallback = np.cross(normals_arr, x_hat)
+    e1_fallback_norm = np.linalg.norm(e1_fallback, axis=1, keepdims=True)
+
+    use_fallback = (e1_norm < 1e-6).squeeze(1)
+    e1[use_fallback] = e1_fallback[use_fallback]
+    e1_norm[use_fallback] = e1_fallback_norm[use_fallback]
+
+    e1 = e1 / (e1_norm + 1e-12)
+    e2 = np.cross(normals_arr, e1)
+    return np.concatenate([e1, e2], axis=1)
+
+
 def load_case(
     root: str | Path,
     case_id: str,
@@ -285,7 +312,8 @@ def load_case(
         _load_npy_rows(case_dir / "surface_wallshearstress.npy", surface_rows),
         "surface_wallshearstress.npy",
     )
-    surface_x = np.concatenate([xyz, normals, area], axis=1)
+    tangent_frame = _compute_tangent_frame(normals.astype(np.float32, copy=False))
+    surface_x = np.concatenate([xyz, normals, area, tangent_frame], axis=1)
     surface_y = np.concatenate([cp, wall_shear], axis=1)
     volume_x = np.concatenate(
         [


morganmcg1:
STUDENT edward:

## Baseline-command flag mismatch — need clarification

The PR body specifies `--use-sdf-importance-sampling --sdf-importance-alpha 0.25` in both the baseline-reproduction and arm commands, but these flags do **not** exist in the current `train.py` on `tay`:

```
$ python train.py --use-sdf-importance-sampling --sdf-importance-alpha 0.25
train.py: error: unrecognized arguments: --use-sdf-importance-sampling --sdf-importance-alpha 0.25
```

`research/CURRENT_RESEARCH_STATE.md` also documents the PR #972 recipe with `--sdf-importance-sampling --sdf-alpha 4.0`, but that flag is likewise absent. PR #972 was closed (not merged) — the SDF-stratified sampling code path is not on `tay`. My previous run for PR #1084 (curvature features) ran the same recipe **minus** the SDF flags.

**My plan** (will execute unless you direct otherwise):
- Use the full PR #972 recipe minus the SDF importance flags, so baseline and treatment use an identical training stack.
- Train both arms for 13 epochs with `--lr-cosine-t-max 13` to match the vol-points-schedule curriculum.
- The comparison stays apples-to-apples (same training stack, only intervention is the tangent frame).
- Caveat: absolute metrics won't reproduce the PR #972 SOTA numbers (which used SDF sampling) — comparison must be against my own no-SDF baseline.

Will proceed in this direction shortly. Please flag if you want a different approach (e.g., add the SDF importance feature as a separate bug-fix PR first, or use a different baseline command).

## Research note: discontinuity in z-hat cross-product tangent frame

Frame-field literature (FrameNet, ICCV 2019; vector heat methods) flags that cross-product tangent frames built against a fixed axis (here ẑ) have a hard discontinuity at the fallback boundary. For DrivAerML cars this hits roof/hood/underbody patches with near-vertical normals exactly. The deterministic e1=cross(n, ẑ) / fallback construction in the PR is fine for a first pass but the discontinuity may hurt learning near the boundary.

If the simple variant doesn't beat baseline cleanly, a follow-up worth trying is in-plane random rotation augmentation of {e1, e2} during training. Will note in "Suggested follow-ups".

morganmcg1:
ADVISOR: Good instincts on both points. Here are my decisions:

**1. SDF flags:** Drop them entirely, exactly as you did for PR #1084. The SDF stratified sampling code was never merged to tay, so this is the correct call. Your baseline will be the standard SOTA stack (without SDF) — that's a clean A/B for the tangent-frame hypothesis. Absolute metrics won't match PR #972's numbers, but the relative comparison between your baseline and the tangent-frame treatment is what matters.

**2. Tangent-frame discontinuity concern:** Your observation about the cross-product discontinuity at near-vertical normals (roof/hood/underbody) is technically correct and well-noted. For this first pass, proceed with the simple deterministic frame — the model can learn to handle the boundary in practice. If the experiment passes EP3 gate, we can follow up with the random in-plane rotation augmentation you mentioned. Good idea to log it in the suggested follow-ups section.

**Proceed with the plan:**
- Drop `--use-sdf-importance-sampling --sdf-importance-alpha 0.25`
- Use the same training stack as your PR #1084 launch (13 epochs, DDP8, full SOTA recipe, `--lr-cosine-t-max 13`)
- Your own no-SDF baseline is the reference for this comparison
- Gate: EP3 val_abupt ≤ 6.2% AND vol_p ≤ 4.5% (PASS), ≤ 6.5% (MARGINAL), >6.5% (KILL)

Launch Arm A now. Report EP1 and EP3 metrics.

morganmcg1:
ADVISOR: Quick check-in — your run is healthy and EP1 has validated.

**W&B run `n0m8mewo` status (07:32Z):**
- State: `running` ✓
- Steps: 14,781
- Runtime: 2.20h
- Heartbeat: 1 min ago
- best_checkpoint/updated: ✓ (EP1 validation complete)

**EP1 metrics (val_primary):**
| Metric | EP1 value | Comparable EP1 baseline |
|---|---|---|
| abupt_axis_mean | **28.07%** | tanjiro v1 27.1%, frieren ~similar |
| wall_shear (total) | **30.89%** | tanjiro v1 30.1% |
| wall_shear_x | 26.88% | typical |
| wall_shear_y | 37.07% | typical |
| **wall_shear_z** | **37.84%** ← worst axis | tanjiro v1 tau_z 36.6% |
| surface_pressure | 21.13% | typical |
| volume_pressure | 17.43% | typical |

**Read:** EP1 is consistent with comparable SOTA-stack runs. The tangent-frame features (e1, e2) have not destabilized training, and the per-axis pattern (tau_z worst, tau_y second-worst) matches the known WSS error structure. **Nothing to act on at EP1.**

**Please post comment updates at EP3 and EP6** — even one-line summaries will keep the staleness detector quiet. The key signal I'm watching:

1. **EP3 val_abupt vs baseline EP3** (target ≤ 6.5% marginal, ≤ 6.2% PASS — the baseline `56bcqp3m` had EP3 ~6.85%)
2. **EP3 wall_shear_z_rel_l2_pct** — this is the hypothesis-critical metric. Tangent frame features should provide the model with geometric scaffolding to predict τ in the correct local basis, which should disproportionately help tau_z. If tau_z error drops more than tau_x error vs a no-tangent-frame baseline, the hypothesis is validated.
3. **EP6 confirmation** — full continuation gate: val_abupt ≤ 6.0% AND val_vol_p ≤ 4.2%

**No action needed** — just keep the comments flowing. With your real steps/epoch ≈ 10,864 and current step 14,781, you should land EP2 in the next ~30 min, EP3 around step 32,600 (~5h from now).

morganmcg1:
STUDENT edward:
SENPAI-RESULT: {"terminal":false,"status":"partial_timeout","pending_arms":true,"wandb_run_ids":["n0m8mewo"],"primary_metric":{"name":"val_primary/abupt_axis_mean_rel_l2_pct","value":7.0590},"test_metric":{"name":"test_primary/abupt_axis_mean_rel_l2_pct","value":6.7181}}

## Results — Arm A (tangent frame + standard tau weights)

### Headline

The 13-epoch training plan **timed out at 4.5h training (epoch 3 best)**, so this is a 4-epoch screening result rather than a 13-epoch confirmation. The PR-specified curriculum `0:16384:3:32768:6:49152:9:65536` holds the loader at 16k vol points for 3 epochs, which inflates per-case view count and made each early epoch take ~96 min on 8 GPUs.

**The tangent-frame features did not improve WSS at the 4-epoch screening horizon. On test metrics they trail the closest internal reference (PR #1084 curvature features, same training stack minus SDF) by ~0.2–0.4pp across all WSS axes.**

### Comparison against closest no-SDF internal baseline (PR #1084 curvature, run `s3oi53o2`)

Both runs: same model (5L/512/4h/128sl), same optimiser (lion 9e-5, wd 5e-4), same loss weights, same surf→vol xattn, no SDF importance sampling. Curvature run used a 4-epoch matched schedule; tangent-frame ran a 13-epoch schedule but only got 4 epochs done.

| Metric | Tangent frame (Arm A) | Curvature ref (#1084) | Δ | Direction |
|--------|---------------------:|----------------------:|---:|:----------|
| best_epoch                | 3 (of 13 planned) | 4 (of 4)              | —     | ours timed out |
| total_train_minutes       | 271 | 196 | +75 | slower curriculum kept us at 16k vol pts |
| **val_abupt**             | **7.0590%** | **7.0901%** | **−0.031pp** | essentially identical |
| val_SP                    | 4.6924%   | 4.6292%   | +0.063pp | slightly worse |
| val_WSS                   | 7.9485%   | 7.9812%   | −0.033pp | same |
| val_WSS_x                 | 6.9026%   | 7.0096%   | −0.107pp | slightly better |
| val_WSS_y                 | 8.9385%   | 8.8053%   | +0.133pp | slightly worse |
| val_WSS_z (target axis)   | 10.5543%  | 10.5559%  | −0.002pp | identical |
| val_vol_p                 | 4.2072%   | 4.4502%   | −0.243pp | better |
| **test_abupt**            | **6.7181%** | **6.5419%** | **+0.176pp** | worse |
| test_SP                   | 4.2800%   | 4.1417%   | +0.138pp | worse |
| **test_WSS (target)**     | **7.7258%** | **7.4645%** | **+0.261pp** | **worse** |
| test_WSS_x                | 6.8134%   | 6.6645%   | +0.149pp | worse |
| test_WSS_y                | 8.6719%   | 8.2268%   | +0.445pp | worse |
| **test_WSS_z (target)**   | **9.8146%** | **9.4369%** | **+0.378pp** | **worse** |
| test_vol_p                | 4.0106%   | 4.2400%   | −0.229pp | better |

### What happened

- Pipeline correctness: `SURFACE_X_DIM=13`, `model.surface_input_dim=13`, unit-tested e1/e2 orthonormality (max dot < 1e-7). Smoke test trained one step on the corrected dataset with `surface_x.shape[-1] == 13` end-to-end.
- Training health: 0 nonfinite losses, 0 nonfinite grads, 0 step_skipped, 0 clipped grads at end. `grad/global_norm_pre_clip = 0.127` at end-of-training. Run is clean.
- Convergence trajectory was healthy: ep1 val_abupt=28.07%, ep2=7.79%, ep3=7.06%. Val WSS slope still **negative** at timeout (`val/slope/val_primary/wall_shear_rel_l2_pct/per_1k_steps = −0.093`), so the model was still improving.
- **Reading the comparison**: at 4 epochs of screening, tangent-frame features are **virtually identical to curvature features on val** but **slightly worse on test across all WSS axes**, which is the opposite of what the hypothesis predicted. The test_WSS regression (+0.26pp) is the most damning signal — `test_WSS_z` (the supposed primary beneficiary) is +0.38pp worse.
- Caveats:
  - LR schedule was tuned for 13 epochs (cosine `t_max=13`) but only ran 4 epochs, so the model's LR at end-of-training was still high (8e-5). The curvature reference had `t_max=4` so its LR had decayed properly. Comparing val/test numbers across mismatched schedules is imperfect.
  - We can't compare to the PR-cited PR #972 SOTA (val_abupt=6.126%, test_WSS=6.727%) because that run used SDF-stratified volume importance sampling, and the `--use-sdf-importance-sampling`/`--sdf-importance-alpha` flags **do not exist on `tay`**. PR #972 was closed, not merged. (Flagged earlier in this PR thread.)

### Why I think it failed

Even with the LR caveat, the per-axis WSS test deltas are systematic (+0.15 to +0.45pp on every WSS axis on test). This isn't noise — the augmented feature space appears to be either:
1. **Overfitting to val views**: val numbers are tied with curvature but test numbers regress, suggesting the extra 6 channels give the model a way to exploit val-set artifacts without learning genuine WSS structure. The hypothesis assumes a tangent frame is useful inductive bias, but a Transolver attention backbone already has access to xyz+normals which together fully determine the tangent plane (up to a 1D rotation). The new features may be redundant rather than informative.
2. **z-hat fallback discontinuity** (research note in earlier comment, FrameNet/ICCV 2019): the cross-product-with-ẑ construction has a hard discontinuity on near-vertical normals — exactly where DrivAerML roofs/hoods/underbodies sit. Without a smoothing augmentation, the model may learn to exploit this artifact rather than the actual tangent structure.

### Suggested follow-ups (in order of likely value)

1. **In-plane rotation augmentation of {e1, e2}**: per-batch random angle θ ∈ [0, 2π), rotate e1/e2 in the tangent plane. Destroys the fallback-boundary artifact. Standard fix from the frame-field literature. Cheap to add (one extra einsum in the loader).
2. **Predict WSS in the tangent frame**: decompose target into (τ·e1, τ·e2, τ·n) at the data layer, train the model to output those 3 channels, project back to xyz at the metric stage. This is more invasive but actually uses the frame as a coordinate basis rather than as redundant features. The hypothesis predicts a benefit *only* if the frame is used as a basis — appending it as features may not be enough.
3. **Fix SDF importance sampling first**: the corrected-split SOTA baseline (PR #972) requires SDF-stratified vol sampling. The current `tay` branch lacks this code path. A separate bug-fix/restoration PR adding `--use-sdf-importance-sampling` and `--sdf-importance-alpha` would let future runs reproduce baseline absolute numbers and would unblock fair comparison for all input-augmentation experiments.

### Arm B decision

Given Arm A shows no benefit (and likely small harm), **Arm B (higher tau-z weight × tangent frame) is unlikely to provide useful signal — the hypothesis was that higher tau-z would *amplify* a benefit, but there's no benefit to amplify**. Each arm costs ~5h wall clock. I'd rather hold off on Arm B and let you decide whether to (a) drop it, (b) pivot to one of the follow-ups above, or (c) re-run Arm A with a 4-epoch matched schedule for a clean apples-to-apples vs the curvature reference before committing more compute.

### Exact command used

```bash
CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 torchrun --nproc_per_node=8 --master_port=29501 train.py \
  --batch-size 4 --validation-every 2 \
  --wandb-name "edward/tangent-frame-armA" \
  --wandb-group "edward-wss-tangent-frame" \
  --kill-thresholds "10864:val_primary/abupt_axis_mean_rel_l2_pct<=30,32600:val_primary/abupt_axis_mean_rel_l2_pct<=8"
```

- **Peak memory:** 45.3GB / 95GB per GPU
- **W&B run:** `n0m8mewo` (group `edward-wss-tangent-frame`)
- **Training time:** 271.1 min (hit `train_timeout_minutes=270.0`)
- **Wall clock:** ~5.0h

morganmcg1:
ADVISOR: Thank you for the thorough analysis and clean terminal result. Closing the tangent frame hypothesis NEGATIVE.

**Decision: Do NOT launch Arm B. Close PR.**

**Assessment:**

The local tangent frame hypothesis is falsified. Your analysis of the two root causes is correct:

1. **Feature redundancy**: Transolver already receives xyz + normals. The tangent frame vectors (e1, e2) are deterministic functions of the normal, so they add no new geometric information — just redundant signal and 6 more features for the model to ignore or exploit.

2. **z-hat discontinuity**: The `cross(n, ẑ)` fallback creates a hard discontinuity at near-vertical normals (car roof, hood, underbody) — exactly where τ_z prediction matters most. This is the FrameNet/ICCV 2019 problem. The model likely learned to exploit the discontinuity artifact rather than genuine WSS structure, which explains why test_WSS_z (+0.378pp) regressed harder than test_WSS (+0.261pp).

The LR-mismatch caveat is noted — the 13-epoch cosine schedule ran 4 epochs at elevated LR ~8e-5 vs the curvature reference's matched 4-epoch schedule at lower LR. However, the systematic regression across ALL WSS axes (including test_abupt +0.176pp) and specifically the monotonic ordering (τ_z worst, τ_x least) is a directional signal that exceeds what LR mismatch alone explains. A clean 4-epoch matched rerun is not worth the 5h wall cost given the theoretical argument is already broken.

**Results summary:**

| Metric | Tangent frame (n0m8mewo) | Curvature ref (#1084) | Δ |
|--------|----------------------:|----------------------:|---:|
| val_abupt | 7.0590% | 7.0901% | −0.031pp |
| test_abupt | 6.7181% | 6.5419% | **+0.176pp WORSE** |
| test_WSS | 7.7258% | 7.4645% | **+0.261pp WORSE** |
| test_WSS_z | 9.8146% | 9.4369% | **+0.378pp WORSE** |
| test_vol_p | 4.0106% | 4.2400% | −0.229pp better |

**Follow-up direction (low priority, assigned to future student):**

The most theoretically sound successor to this experiment is **in-plane rotation augmentation** — randomly rotating the coordinate frame around the surface normal during training. This is Sim3-equivariance and is complementary to the normal direction being fixed. It does not suffer from the z-hat discontinuity and forces the model to learn τ predictions that are invariant to arbitrary in-plane orientation. Cost: ~5% training overhead, no architecture change, no new features.

However, this is a future hypothesis. Your slot is free.

**Action: Please close this PR.** Hypothesis is closed NEGATIVE. Thank you for the careful work.

morganmcg1:
STUDENT edward: Closing per advisor decision (hypothesis NEGATIVE). Tangent-frame input augmentation showed no benefit and a systematic test-side regression (test_WSS +0.261pp, test_WSS_z +0.378pp vs the closest no-SDF reference). Two root causes identified in the thread: feature redundancy (e1/e2 are deterministic functions of normals the model already receives) and the z-hat fallback discontinuity on near-vertical normals. Arm B skipped — no benefit to amplify. Follow-up direction (in-plane rotation augmentation / Sim3-equivariance around the surface normal) noted for a future student.

PR #1110, opened 2026-05-14
OHEM surface top-20% hard mining for WSS (Wave 28)

morganmcg1:
## Hypothesis

**Online Hard Example Mining (OHEM) applied to surface point-level WSS loss will force the model to focus gradient budget on spatially-hard regions (flow separation zones, A-pillars, wheel arches, rear wake), reducing τ_y and τ_z error.**

Current surface loss averages L2 error uniformly across all ~65k surface points. However, the hardest WSS points — the ones driving our 6.727% test_WSS — are concentrated in ~10–20% of the surface. Uniform averaging gives them disproportionately weak gradient signal. OHEM selects the top-K% hardest points each step and adds a supplementary loss term to amplify gradient from those regions.

**Formulation (supplementary, not replacement):**
```
L_total = L_all + 0.5 * L_hard_top20pct
```
All points still receive gradient from `L_all`. The `L_hard_top20pct` term adds extra pressure on the 20% hardest surface points. λ=0.5 keeps the supplementary term from dominating.

**Key design choices vs PR #509 (bengio-era OHEM, closed without merge):**
- K=20% (not 50%) — more aggressive focus on truly hard regions
- Supplementary formulation (not replacement) — avoids training instability from dropping 50% of points
- 2-epoch warmup (not 0) — model needs to calibrate its error distribution before OHEM is meaningful
- Tay 13-epoch schedule, 5L/512d/4H, STRING-sep, Lion optimizer (completely different from PR #509 bengio recipe)
- OHEM is NOT in the current codebase — you must implement from scratch

**Why OHEM now?** Analysis of PR #1102 ensemble shows test_tau_z=8.26% is the dominant WSS axis failure. Per-point error is spatially concentrated. Uniform averaging means these hard points are overwhelmed by ~80% of easy laminar-flow surface points that are already well-fit.

---

### Step 1: Add CLI flags to `train.py`

Add these three arguments in the argparse section (near the existing `--surface-loss-weight` flag):

```python
parser.add_argument("--ohem-surface-ratio", type=float, default=1.0,
    help="Top-K fraction of hardest surface points for supplementary OHEM loss. "
         "1.0 = disabled (default). 0.20 = top-20%% hardest points.")
parser.add_argument("--ohem-warmup-epochs", type=int, default=0,
    help="Number of epochs before OHEM activates. During warmup OHEM contribution = 0.")
parser.add_argument("--ohem-min-k", type=int, default=100,
    help="Minimum number of hard points to select per sample (floor). "
         "Prevents degenerate selection at small batch sizes.")
```

### Step 2: Add OHEM loss function to `trainer_runtime.py`

Add the following function after the existing `weighted_channel_mse` function:

```python
def ohem_supplementary_loss(
    pred: torch.Tensor,
    target: torch.Tensor,
    mask: torch.Tensor,
    ohem_ratio: float,
    min_k: int = 100,
    channel_weights: torch.Tensor | None = None,
) -> torch.Tensor:
    """Compute supplementary OHEM surface loss over top-K% hardest points.

    Args:
        pred: [B, N, C] surface predictions
        target: [B, N, C] surface targets
        mask: [B, N] boolean mask (True = valid point)
        ohem_ratio: fraction of points to select as hard (e.g. 0.20)
        min_k: minimum hard points per sample regardless of ratio
        channel_weights: [C] per-channel weights (optional)

    Returns:
        Scalar supplementary loss over hard points only.
    """
    sq_err = (pred - target).square()  # [B, N, C]

    # Per-point scalar loss for ranking — use fp32 for numerical stability
    per_point_loss = sq_err.mean(dim=-1).float()  # [B, N]

    # Mask out invalid points so they cannot be selected
    mask_dev = mask.to(device=pred.device)
    per_point_loss = per_point_loss.masked_fill(~mask_dev, float('-inf'))

    # Compute k: ratio of total points, floored at min_k, capped at min valid count
    valid_counts = mask_dev.float().sum(dim=1)  # [B]
    k = max(min_k, int(ohem_ratio * per_point_loss.shape[1]))
    k = min(k, int(valid_counts.min().item()))
    k = max(k, 1)

    # Select top-k hardest indices per sample
    _, hard_indices = per_point_loss.topk(k, dim=1, largest=True, sorted=False)
    # hard_indices: [B, k]

    # Gather hard sq_err
    idx_expanded = hard_indices.unsqueeze(-1).expand(-1, -1, pred.shape[-1])
    hard_sq_err = sq_err.gather(1, idx_expanded.to(dtype=torch.long))  # [B, k, C]

    # Apply channel weights if provided
    if channel_weights is not None:
        weights_dev = channel_weights.to(device=pred.device, dtype=pred.dtype)
        hard_sq_err = hard_sq_err * weights_dev.unsqueeze(0).unsqueeze(0)

    return hard_sq_err.mean()
```

### Step 3: Modify `train_loss()` in `train.py`

**3a. Update the function signature** to accept ohem parameters:

```python
def train_loss(
    model: nn.Module,
    batch,
    transform: TargetTransform,
    device: torch.device,
    amp_mode: str,
    *,
    surface_loss_weight: float = 1.0,
    volume_loss_weight: float = 1.0,
    surface_channel_weights: torch.Tensor | None = None,
    ohem_surface_ratio: float = 1.0,        # NEW
    ohem_warmup_active: bool = False,        # NEW: True = still in warmup, OHEM disabled
    ohem_min_k: int = 100,                   # NEW
    ohem_lambda: float = 0.5,               # NEW: weight on supplementary term
```

**3b. After the existing surface/volume loss, add the OHEM supplementary term.** Find the block where `loss = weighted_surface_loss + weighted_volume_loss` is computed and extend it:

```python
            # ... existing surface_loss and volume_loss computation unchanged ...

            # OHEM supplementary term
            ohem_active = (ohem_surface_ratio < 1.0) and (not ohem_warmup_active)
            if ohem_active:
                ohem_loss = ohem_supplementary_loss(
                    pred=out["surface_preds"],
                    target=surface_target,
                    mask=batch.surface_mask,
                    ohem_ratio=ohem_surface_ratio,
                    min_k=ohem_min_k,
                    channel_weights=surface_channel_weights,
                )
                ohem_contribution = ohem_lambda * ohem_loss
            else:
                ohem_loss = torch.zeros(1, device=device)
                ohem_contribution = torch.zeros(1, device=device)

            weighted_surface_loss = surface_loss_weight * surface_loss
            weighted_volume_loss = volume_loss_weight * volume_loss
            loss = weighted_surface_loss + weighted_volume_loss + ohem_contribution
            base_mse_loss = surface_loss + volume_loss
```

**3c. Add OHEM diagnostics** to the metrics dict before the return:

```python
    metrics["train/ohem/active"] = float(ohem_active)
    metrics["train/ohem/in_warmup"] = float(ohem_warmup_active)
    metrics["train/ohem/hard_loss"] = ohem_loss.item()
    metrics["train/ohem/lambda_contribution"] = ohem_contribution.item()
```

### Step 4: Thread ohem args through training loop call sites

In the main training loop where `train_loss` is called, compute the warmup flag from epoch count and pass all args:

```python
ohem_warmup_active = (args.ohem_warmup_epochs > 0) and (epoch < args.ohem_warmup_epochs)

loss_val, step_metrics = train_loss(
    model, batch, transform, device, amp_mode,
    surface_loss_weight=args.surface_loss_weight,
    volume_loss_weight=args.volume_loss_weight,
    surface_channel_weights=surface_channel_weights,  # existing
    ohem_surface_ratio=args.ohem_surface_ratio,
    ohem_warmup_active=ohem_warmup_active,
    ohem_min_k=args.ohem_min_k,
    ohem_lambda=0.5,
)
```

If there is a separate GradNorm training path (look for `use_gradnorm`), apply the same `ohem_warmup_active` logic there too.

---

## Run Command

```bash
cd target/ && torchrun --standalone --nproc-per-node=8 train.py \
  --model-layers 5 \
  --model-hidden-dim 512 \
  --model-heads 4 \
  --model-slices 128 \
  --optimizer lion \
  --lr 9e-5 \
  --weight-decay 5e-4 \
  --tau-y-loss-weight 1.5 \
  --tau-z-loss-weight 2.0 \
  --surface-loss-weight 2.0 \
  --use-ema \
  --ema-decay 0.999 \
  --grad-clip-norm 0.5 \
  --lr-warmup-epochs 1 \
  --use-qk-norm \
  --rff-num-features 16 \
  --lr-cosine-t-max 13 \
  --epochs 13 \
  --no-compile-model \
  --batch-size 4 \
  --validation-every 1 \
  --eval-surface-points 65536 \
  --train-volume-points 65536 \
  --eval-volume-points 65536 \
  --use-surf-to-vol-xattn \
  --ohem-surface-ratio 0.20 \
  --ohem-warmup-epochs 2 \
  --ohem-min-k 100 \
  --wandb-group tay-wave28-ohem \
  --wandb-name askeladd/ohem-surface-top20pct
```

**IMPORTANT:** Do NOT add `--sdf-importance-sampling` or `--sdf-alpha` — those flags do not exist on the tay branch and will crash argparse.

---

## EP3 Gate (kill threshold)

Check at end of epoch 3:

| Condition | Decision |
|-----------|----------|
| val_abupt ≤ 7.2% AND val_vol_p ≤ 4.5% | **PASS** — continue to EP10 |
| val_abupt ≤ 7.6% AND val_vol_p ≤ 5.0% | **MARGINAL** — continue but monitor closely |
| val_abupt > 7.6% OR val_vol_p > 5.0% | **KILL** — stop and report immediately |

**EP10 gate:** val_abupt ≤ 6.5% → PASS; ≤ 6.7% → MARGINAL; > 6.7% → KILL.

Also monitor `train/ohem/hard_loss` from EP3 onwards — if it diverges (>10× initial value) at the same time as `train/loss` spikes, kill and report. Do not wait for EP3 if training loss is clearly diverging.

---

## Win Criteria

A result qualifies as a new pool member if:
- test_WSS ≤ 6.50% AND
- test_vol_p ≤ 3.643% AND
- test_SP ≤ 3.577% AND
- val_abupt ≤ 6.20%

---

## Baseline Metrics (target to beat)

| Metric | Current SOTA (PR #972) |
|--------|------------------------|
| val_abupt | ~6.10% |
| test_abupt | 5.844% |
| test_vol_p | **3.643%** (must not exceed) |
| test_SP | **3.577%** (must not exceed) |
| test_WSS | **6.727%** ← primary target |
| test_tau_z | ~8.26% (worst axis, main bottleneck) |

---

## Reporting

After each epoch, report the full val breakdown:
- val_abupt, val_vol_p, val_SP, val_tau_y, val_tau_z
- train/ohem/hard_loss, train/ohem/active, train/ohem/in_warmup (report from EP1)

At run completion, include the terminal SENPAI-RESULT marker:
```
SENPAI-RESULT: {"terminal":true,"status":"complete","pending_arms":false,"wandb_run_ids":["<run-id>"],"primary_metric":{"name":"val_abupt_pct","value":<number>},"test_metric":{"name":"test_WSS_pct","value":<number>}}
```

---

## Background & Prior Art

- **PR #509 (bengio branch, closed 2026-05-13, never merged):** Norman's OHEM on a 5L/256d FourierPE recipe. Key lessons: use fp32 for topk ranking, add `--ohem-min-k` floor, log diagnostics from step 1. That PR used K=50% replacement formulation — we are using K=20% supplementary instead.
- **Wave 27 lesson:** Loss function changes that create conflicting gradient objectives destabilize Transolver rapidly. The supplementary formulation preserves stable gradient signal from all points while adding focus on hard ones.
- **τ_z bottleneck:** test_tau_z=8.26% in the best ensemble (PR #1102). OHEM targets the spatially concentrated hard regions that drive this metric.

### DIFF
--- a/train.py
+++ b/train.py
@@ -56,6 +56,7 @@ from trainer_runtime import (
     make_loaders,
     masked_mse,
     metric_namespace,
+    ohem_supplementary_loss,
     weighted_channel_mse,
     parse_kill_thresholds,
     primary_metric_log,
@@ -138,6 +139,10 @@ class Config:
     gradnorm_log_clip: float = 4.0
     gradnorm_ema_beta: float = 0.9
     gradnorm_min_weight: float = 0.0
+    ohem_surface_ratio: float = 1.0
+    ohem_warmup_epochs: int = 0
+    ohem_min_k: int = 100
+    ohem_lambda: float = 0.5
     debug: bool = False
 
 
@@ -219,6 +224,30 @@ def parse_args(argv: Iterable[str] | None = None) -> Config:
             "tasks (tau_y/tau_z) to climb. Default 0.0 disables the floor "
             "(matches Run 1 behavior). Recommended 0.7 for soft "
             "redistribution. Unused in mode='full'."
+        ),
+        "ohem_surface_ratio": (
+            "Top-K fraction of hardest surface points for the supplementary "
+            "OHEM (online hard example mining) loss term. 1.0 = disabled "
+            "(default). 0.20 = top-20%% hardest points per sample. The OHEM "
+            "term adds 'ohem_lambda * MSE(top-K hardest)' on top of the "
+            "standard surface_loss so all points still receive gradient."
+        ),
+        "ohem_warmup_epochs": (
+            "Number of epochs before OHEM activates. During warmup the OHEM "
+            "supplementary term is zero so the model can calibrate its error "
+            "distribution before being asked to focus on hard examples. "
+            "Default 0 = no warmup."
+        ),
+        "ohem_min_k": (
+            "Minimum number of hard points to select per sample regardless "
+            "of ohem_surface_ratio. Prevents degenerate top-K selection at "
+            "very small surface views. Default 100."
+        ),
+        "ohem_lambda": (
+            "Weight multiplier on the OHEM supplementary loss term. Final "
+            "loss = surface_loss_weight*surface_loss + volume_loss_weight*"
+            "volume_loss + ohem_lambda*ohem_hard_loss. Default 0.5 keeps "
+            "the supplementary term from dominating."
         ),
         "use_surf_to_vol_xattn": (
             "Enable surface->volume cross-attention (PR #823). After the "
@@ -321,6 +350,10 @@ def train_loss(
     surface_loss_weight: float = 1.0,
     volume_loss_weight: float = 1.0,
     surface_channel_weights: torch.Tensor | None = None,
+    ohem_surface_ratio: float = 1.0,
+    ohem_warmup_active: bool = False,
+    ohem_min_k: int = 100,
+    ohem_lambda: float = 0.5,
 ) -> tuple[torch.Tensor, dict[str, float]]:
     batch = batch.to(device)
     surface_target = transform.apply_surface(batch.surface_y)
@@ -342,9 +375,25 @@ def train_loss(
         else:
             surface_loss = masked_mse(out["surface_preds"], surface_target, batch.surface_mask)
         volume_loss = masked_mse(out["volume_preds"], volume_target, batch.volume_mask)
+
+        ohem_active = (ohem_surface_ratio < 1.0) and (not ohem_warmup_active)
+        if ohem_active:
+            ohem_loss = ohem_supplementary_loss(
+                pred=out["surface_preds"],
+                target=surface_target,
+                mask=batch.surface_mask,
+                ohem_ratio=ohem_surface_ratio,
+                min_k=ohem_min_k,
+                channel_weights=surface_channel_weights,
+            )
+            ohem_contribution = ohem_lambda * ohem_loss
+        else:
+            ohem_loss = torch.zeros((), device=device)
+            ohem_contribution = torch.zeros((), device=device)
+
         weighted_surface_loss = surface_loss_weight * surface_loss
         weighted_volume_loss = volume_loss_weight * volume_loss
-        loss = weighted_surface_loss + weighted_volume_loss
+        loss = weighted_surface_loss + weighted_volume_loss + ohem_contribution
         base_mse_loss = surface_loss + volume_loss
     return loss, {
         "base_mse_loss": float(base_mse_loss.detach().cpu().item()),
@@ -352,6 +401,10 @@ def train_loss(
         "volume_loss": float(volume_loss.detach().cpu().item()),
         "surface_loss_weighted": float(weighted_surface_loss.detach().cpu().item()),
         "volume_loss_weighted": float(weighted_volume_loss.detach().cpu().item()),
+        "ohem/active": float(ohem_active),
+        "ohem/in_warmup": float(ohem_warmup_active),
+        "ohem/hard_loss": float(ohem_loss.detach().cpu().item()),
+        "ohem/lambda_contribution": float(ohem_contribution.detach().cpu().item()),
     }
 
 
@@ -913,6 +966,11 @@ def main(argv: Iterable[str] | None = None) -> None:
                     current_train_vol_points = desired_vol_points
             if isinstance(train_loader.sampler, DistributedSampler):
                 train_loader.sampler.set_epoch(epoch)
+            ohem_warmup_active = (
+                config.ohem_surface_ratio < 1.0
+                and config.ohem_warmup_epochs > 0
+                and epoch < config.ohem_warmup_epochs
+            )
             timeout_hit = distributed_any(
                 state,
                 (time.time() - train_start) / 60.0 >= timeout_minutes,
@@ -1030,6 +1088,10 @@ def main(argv: Iterable[str] | None = None) -> None:
                         surface_loss_weight=config.surface_loss_weight,
                         volume_loss_weight=config.volume_loss_weight,
                         surface_channel_weights=surface_channel_weights if use_channel_weights else None,
+                        ohem_surface_ratio=config.ohem_surface_ratio,
+                        ohem_warmup_active=ohem_warmup_active,
+                        ohem_min_k=config.ohem_min_k,
+                        ohem_lambda=config.ohem_lambda,
                     )
                 optimizer.zero_grad(set_to_none=True)
                 global_step += 1
@@ -1070,6 +1132,18 @@ def main(argv: Iterable[str] | None = None) -> None:
                             "train/tau_z_loss_weight": config.tau_z_loss_weight,
                         }
                     )
+                    if "ohem/active" in batch_loss_metrics:
+                        train_log.update(
+                            {
+                                "train/ohem/active": batch_loss_metrics["ohem/active"],
+                                "train/ohem/in_warmup": batch_loss_metrics["ohem/in_warmup"],
+                                "train/ohem/hard_loss": batch_loss_metrics["ohem/hard_loss"],
+                                "train/ohem/lambda_contribution": batch_loss_metrics["ohem/lambda_contribution"],
+                                "train/ohem/ratio": config.ohem_surface_ratio,
+                                "train/ohem/lambda": config.ohem_lambda,
+                                "train/ohem/min_k": float(config.ohem_min_k),
+                            }
+                        )
                 if gradnorm_metrics:
                     train_log.update(gradnorm_metrics)
 
--- a/trainer_runtime.py
+++ b/trainer_runtime.py
@@ -870,6 +870,47 @@ def weighted_channel_mse(
     return pred.sum() * 0.0
 
 
+def ohem_supplementary_loss(
+    pred: torch.Tensor,
+    target: torch.Tensor,
+    mask: torch.Tensor,
+    ohem_ratio: float,
+    min_k: int = 100,
+    channel_weights: torch.Tensor | None = None,
+) -> torch.Tensor:
+    """Compute supplementary OHEM surface loss over top-K% hardest points.
+
+    pred / target: [B, N, C]; mask: [B, N] boolean (True = valid point).
+    ``ohem_ratio`` is the fraction of points to select as hard (e.g. 0.20).
+    ``min_k`` is the per-sample floor on the number of hard points selected.
+    Returns a scalar mean squared error computed only over the top-K hardest
+    points per sample, optionally per-channel weighted to match the main loss.
+    """
+    if pred.numel() == 0:
+        return pred.sum() * 0.0
+
+    sq_err = (pred - target).square()  # [B, N, C]
+    per_point_loss = sq_err.mean(dim=-1).float()  # [B, N], fp32 for stable ranking
+
+    mask_dev = mask.to(device=pred.device).bool()
+    per_point_loss = per_point_loss.masked_fill(~mask_dev, float("-inf"))
+
+    valid_counts = mask_dev.float().sum(dim=1)  # [B]
+    k = max(min_k, int(ohem_ratio * per_point_loss.shape[1]))
+    k = min(k, int(valid_counts.min().item()))
+    k = max(k, 1)
+
+    _, hard_indices = per_point_loss.topk(k, dim=1, largest=True, sorted=False)
+    idx_expanded = hard_indices.unsqueeze(-1).expand(-1, -1, pred.shape[-1])
+    hard_sq_err = sq_err.gather(1, idx_expanded.to(dtype=torch.long))  # [B, k, C]
+
+    if channel_weights is not None:
+        weights_dev = channel_weights.to(device=pred.device, dtype=hard_sq_err.dtype)
+        hard_sq_err = hard_sq_err * weights_dev.unsqueeze(0).unsqueeze(0)
+
+    return hard_sq_err.mean()
+
+
 def squared_relative_l2_loss(
     pred: torch.Tensor,
     target: torch.Tensor,


morganmcg1:
STUDENT askeladd:
SENPAI-RESULT: {"terminal":true,"status":"complete","pending_arms":false,"wandb_run_ids":["o8pt5ybd"],"primary_metric":{"name":"val_abupt_pct","value":30.7769},"test_metric":{"name":"val_WSS_pct","value":34.7992}}

## Results

**KILLED at EP3 per kill gate. OHEM caused catastrophic divergence the moment it activated.**

### Validation metrics by epoch

| Epoch | OHEM state | val_abupt | val_vol_p | val_SP | val_WSS | val_tau_x | val_tau_y | val_tau_z |
|------:|:-----------|----------:|----------:|-------:|--------:|----------:|----------:|----------:|
| 1 | warmup     | 26.36% | 16.79% | 19.81% | 28.84% | 25.08% | 34.41% | 35.69% |
| 2 | warmup     | **7.68%** | **4.52%** | **5.12%** | 8.67% | 7.52% | 9.93% | 11.32% |
| 3 | **active** | **30.78%** | **10.51%** | **25.07%** | **34.80%** | **28.17%** | **43.73%** | **46.41%** |

### EP3 gate decision

| Criterion | Threshold | Actual | Result |
|-----------|-----------|--------|--------|
| val_abupt | ≤ 7.6% (KILL above) | 30.78% | **KILL** (4× threshold) |
| val_vol_p | ≤ 5.0% (KILL above) | 10.51% | **KILL** (2× threshold) |

### Root cause: OHEM contribution magnitude is 100–1000× the base loss

OHEM's `hard_sq_err.mean()` over the top-20% hardest points lives on a wildly different scale than `surface_loss` (the all-points mean MSE):

| Metric over EP3 active steps | Value |
|---|---|
| `train/surface_loss` (all-points MSE) | typically 0.05–0.6 |
| `train/ohem/hard_loss` (top-20% MSE) | **median 5.6, p95 = 7.8, max = 4411** |
| `train/ohem/lambda_contribution` (= 0.5 × hard_loss) | p95 = 7.8, p99 = 13.4, **max = 2205.7** |
| Total loss spikes (`train/loss`) | up to **2207** at step 29023 |

The instructions specified `λ=0.5` to "keep the supplementary term from dominating", but in practice the top-20% MSE is ~30–50× the all-points MSE under normal conditions and spikes to 4000× during pathological batches (likely flow-separation regions or wheel-arch corners with large residuals). With Lion (sign-of-momentum updates), even a few catastrophic spikes destabilize the entire weight manifold; once the model degrades, OHEM finds even more hard points, creating a positive feedback loop.

The base surface/volume losses themselves remained well-behaved (0.05–0.6), so this is purely an OHEM-induced collapse — not a data or compute issue.

### Why EP2 looked OK but EP3 blew up

EP2 finished cleanly because warmup covered EP1+EP2 (ohem_warmup_epochs=2). The model at end-of-EP2 had val_abupt=7.68%, slightly above the eventual baseline. The instant OHEM activated at step 21730 (start of EP3), the first batch's hard_loss=25.06 → loss=12.57 dwarfed the well-behaved base loss (~0.02), and Lion took a destructive step. By step 23163 (~7% into EP3) the loss had spiked to 815, and the model never recovered.

### Train command used

```bash
cd target/ && torchrun --standalone --nproc-per-node=8 train.py \
  --use-ema --ema-decay 0.999 --grad-clip-norm 0.5 \
  --lr-warmup-epochs 1 --pos-encoding-mode string_separable --use-qk-norm \
  --use-surf-to-vol-xattn \
  --ohem-surface-ratio 0.20 --ohem-warmup-epochs 2 --ohem-min-k 100 --ohem-lambda 0.5 \
  --wandb-group tay-wave28-ohem --wandb-name askeladd/ohem-surface-top20pct
```

### Peak memory

Run did not OOM; this was a loss-scale / optimization failure. (Peak per-GPU memory ~40 GB; well within the 80 GB H100 budget.)

### W&B

- Run group: `tay-wave28-ohem`
- Rank-0 run id: `o8pt5ybd` (https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/o8pt5ybd)

**The hypothesis was reasonable but the formulation is unbalanced.** Focusing gradient on the hardest 20% of surface points is sensible, but `hard_loss = mean(top-K sq_err)` is not on a comparable scale to `surface_loss = mean(all sq_err)`. The ratio is ~30× under normal conditions and explodes 1000–4000× under batches with localized high-residual outlier points. λ=0.5 was 25–50× too large for these magnitudes, and there is no spike-clipping safeguard.

This is the same family of failure mode flagged in the PR background ("Wave 27 lesson: loss function changes that create conflicting gradient objectives destabilize Transolver rapidly"), but the diagnosis there ("conflicting objectives") doesn't quite fit — the OHEM term *agrees* with the surface objective, it just amplifies it by an enormous and unpredictable factor each step.

The supplementary-not-replacement design (chosen to fix PR #509's stability issues) actually makes things worse here, because we now have two unbounded loss terms summed together with no scale normalization.

### Suggested follow-ups

If you want to push OHEM further, three options ranked by likelihood of success:

1. **Self-normalized OHEM** (highest priority): scale by detached base loss to make λ scale-invariant:
   ```python
   ohem_contribution = ohem_lambda * (ohem_loss / surface_loss.detach().clamp_min(1e-8))
   ```
   This makes `λ=0.5` mean "OHEM contributes 50% extra gradient pressure per step" regardless of the absolute MSE scale. Recommend trying `λ ∈ {0.2, 0.5}` with this normalization.

2. **Spike-clipped raw OHEM**: keep the current formulation but clip per-step:
   ```python
   ohem_contribution = ohem_lambda * ohem_loss.clamp_max(10.0 * surface_loss.detach())
   ```
   Prevents the catastrophic spikes (>50× base) while keeping typical contributions intact. λ=0.5 might still be too high — recommend trying λ=0.1.

3. **Much smaller λ with raw OHEM**: just turn down λ. p95 hard_loss is ~8, so λ=0.05 gives ~0.4 contribution which is comparable to surface_loss (0.6). Still vulnerable to the 4000× spikes though.

I would also recommend **logging `(hard_loss / surface_loss)` as a real-time diagnostic** so any future OHEM run kills early if the ratio exceeds, say, 100×.

Independently: the 4411× spike at step 29023 likely corresponds to one or two specific samples in the dataset with unusual surface geometry. Worth identifying which case it is via per-sample logging — these might be the hard cases that OHEM was supposed to target, but they need preprocessing rather than gradient amplification.

morganmcg1:
ADVISOR: KILL/CLOSE confirmed. Your diagnostic is the deliverable — exceptional root-cause analysis.

## On the result

EP3 kill is unambiguous (val_abupt 30.78% vs gate 7.6%, val_vol_p 10.51% vs gate 5.0%, val_WSS 34.80%). Your trajectory data nails the failure mechanism precisely:

- **EP1+EP2 (warmup): clean** — model lands at val_abupt=7.68% at EP2 end, within normal recipe variance
- **EP3 step 21730 (OHEM activates): instant divergence** — first active batch produces `hard_loss=25.06 → loss=12.57` vs base ~0.02, Lion sign-of-momentum makes it catastrophic
- **Positive feedback loop**: degraded model → more "hard" points → larger hard_loss → larger updates → further degradation

This is a textbook scale-collapse failure, not a "conflicting objectives" failure. Your analysis correctly distinguishes the two: OHEM gradients *agree* in direction with surface gradients, but they amplify by an unpredictable 30-4000× factor each step.

## What we learned (for the experiment log)

1. **Top-K mean MSE is NOT on the same scale as all-points mean MSE** on this benchmark. Median ratio ~18-30×, p95 ~25-30×, **max 88,000×** (your worst-case spike). The distribution of point-wise sq_err is heavy-tailed enough that raw top-K means are unbounded loss-units.
2. **Lion is uniquely sensitive to loss-scale spikes** because sign-of-momentum doesn't scale with gradient magnitude. A single 4000× spike → max-size update on every parameter → manifold collapse. Adam/AdamW would have been less catastrophic because gradient magnitude would have been clipped relatively.
3. **The supplementary-not-replacement design (a Wave 27 lesson) does not by itself prevent scale failures.** It prevents direction conflicts. Scale conflicts are an independent failure mode.
4. **Implicit takeaway for future loss-augmentation work:** any auxiliary loss term derived from a subset of points (top-K, hard sample, focal, etc.) needs *scale-invariant formulation by construction*, not by hyperparameter tuning. λ tuning cannot recover from a scale that varies by 4 orders of magnitude across steps.

I'll write this up in EXPERIMENTS_LOG.md as the canonical reference for "why naive top-K OHEM fails on Transolver+Lion+Wave 28 stack."

## What's next for you

I'm closing this PR and assigning you **OHEM v2: spike-clipped with reduced λ** (a clean implementation of your Approach #2 + #3, the two follow-ups most likely to succeed). The design:

- Clip OHEM contribution at `MAX_CLIP × surface_loss.detach()` per step — guarantees bounded contribution regardless of raw magnitude
- Use λ=0.1 (5× reduction from v1) — typical contribution ~0.1× base loss, clearly supplementary
- Add `train/ohem/clip_active_pct` diagnostic to log how often the clip fires (high % means we're hitting the catastrophe regime)
- Same warmup (2 epochs), same 20% ratio, same gate

This addresses both failure modes you identified:
- Spike clipping bounds the per-step contribution
- Smaller λ bounds the typical contribution

If clip_active_pct stays near 100% throughout EP3+, that's signal the scale problem is intrinsic to this dataset's residual distribution and we should pivot away from top-K OHEM entirely. If clip_active_pct drops to <20% after a few epochs, the model is converging to a regime where top-K MSE is on a reasonable scale and OHEM might actually help.

PR coming next.

Closing #1110.

morganmcg1:
ADVISOR: Closing PR #1110 because EP3 KILL: val_abupt 30.78% vs gate 7.6% (4× over), val_vol_p 10.51% vs gate 5.0% (2× over). OHEM hard_loss spikes 1000-4000× the base surface_loss, causing Lion+sign-of-momentum to drive catastrophic divergence the moment OHEM activates at EP3. Root cause is scale-collapse (heavy-tailed sq_err distribution), not direction conflict. Detailed diagnostic + design rule logged in advisor close comment. Reassigning askeladd to OHEM v2 (spike-clipped + reduced λ)..

PR #1101, opened 2026-05-14
LR=9e-5 isolated control on corrected split (vs SOTA 1e-4)

morganmcg1:
## Hypothesis

**LR=9e-5 isolated control on corrected split** — Lower learning rate (9e-5 vs SOTA 1e-4) may produce a smoother loss landscape that generalizes better to OOD test geometries, particularly for vol_p. This is a single-variable A/B against SOTA: only `--lr` changes (1e-4 → 9e-5), everything else matches `56bcqp3m`.

**Mechanism:** Lion optimizer with lr=1e-4 converges quickly but may overshoot fine minima in the later cosine tail, especially on volume pressure which depends on long-range spatial correlations. A 10% LR reduction (9e-5) extends the effective learning window without changing the optimizer dynamics or cosine schedule shape. Prior experiments on the old dataset showed `9mm3sz7x` (lr=9e-5) achieved test_abupt=8.123% which was the wave-best at the time — however those results are void due to the dataset artifact. This is the first clean isolated test on the corrected split.

**Single variable changed:** `--lr 9e-5` (SOTA uses `--lr 1e-4`)  
**No SDF sampling, no GradNorm** — clean LR isolation.

## Baseline (current SOTA — beat this)

| Metric | SOTA value | Run | PR |
|--------|------------|-----|----|
| test_abupt (`test_primary/abupt_axis_mean_rel_l2_pct`) | **5.844%** | `56bcqp3m` (eval `zxnhtagj`) | #972 |
| test_vol_p (`test_primary/volume_pressure_rel_l2_pct`) | **3.643%** | `56bcqp3m` | #972 |
| test_surf_p (`test_primary/surface_pressure_rel_l2_pct`) | **3.577%** | `56bcqp3m` | #972 |
| test_wss (`test_primary/wall_shear_rel_l2_pct`) | **6.727%** | `56bcqp3m` | #972 |

Primary metric: `test_primary/abupt_axis_mean_rel_l2_pct` — lower is better. Must beat **5.844%** to merge.

## EMA warm-start fix

Branch `drivaerml-long-20260504` includes the EMA warm-start fix (PR #1087, commit `860d08f`): shadow is now correctly initialized from live weights at `ema-start-step` (not from random weights). Verify the fix is active before launching the long run.

### Step 1: Smoke run (required before long run)

Run a 3-epoch DDP8 smoke test to confirm config and healthy convergence:

```bash
torchrun --standalone --nproc_per_node=8 train.py \
[...]

PR #1086, opened 2026-05-13
EMA(0.999) on productive SDF α=0.25 stack — corrected-split re-test

morganmcg1:
## Hypothesis

EMA (Exponential Moving Average) with decay=0.999 gives a ~1,000-step shadow model that serves as a smoothed checkpoint for evaluation, reducing sensitivity to training noise. The wave SOTA (PR #972, run `56bcqp3m`, test_abupt=5.844%) **already uses EMA decay=0.999** — but none of the currently running SDF α-sweep arms (fern #1063 α=0.25, nezuko #1072 α=0.5, frieren #1077 α=1.0) include EMA.

PR #954's previous "EMA dead-end" verdict was from the **pre-20260511 split** (the case-split bug that caused a spurious +7-8pp val→test vol_p gap). Per Issue #1053 and `CURRENT_RESEARCH_STATE.md`, EMA decay=0.999 is **explicitly RETRACTED from dead-ends** — it needs a clean re-test on the corrected dataset.

This experiment is a direct **A/B vs fern's PR #1063** (same stack, same SDF α=0.25, same optimization config), with EMA layered on top. If EMA helps on the corrected split, this is potentially the single lever that gets us under 5.844% test_abupt without any code work.

## Current Baseline (wave SOTA — corrected split rawcanon_20260511)

| Metric | Value |
|--------|-------|
| **test_abupt** | **5.844%** |
| test_surf_p | 3.577% |
| test_vol_p | **3.643%** |
| test_wss | 6.727% |

**Source:** PR #972, run `56bcqp3m`, eval `zxnhtagj`  
**Dataset:** `/mnt/new-pvc/Processed/drivaerml_processed_rawcanon_20260511`

**A/B reference (fern, no EMA):** PR #1063 run `xfykblf9`, val_abupt best = 6.2647% at EP11, still running.

### Step 0: Verify existing SDF monkey-patch code
The SDF stratified-sampling code path is already implemented in `train.py` via a monkey-patch (from PR #1054+). Copy the exact block from fern's PR #1063 or tanjiro's prior PR #1076. Use `__class__` reassignment (NOT `types.MethodType`) with α=0.25 (matching fern exactly for clean A/B). The inverse formula MUST be `weight = 1.0 / (1.0 + α * |sdf|)` — near-surface gets HIGH weight, far-field LOW weight. Look for the `_SDFStratifiedDataset` class.

The SDF column is at `volume_x[:, 3]` (column index 3).

[...]

PR #1129, opened 2026-05-15
[dl24-nezuko] WSS H3: near-wall volume cross-attn into surface decoder

morganmcg1:
## Hypothesis

**τ_z is stuck because local surface features can't capture boundary-layer velocity gradients.**

Every experiment on this wave — including H2 (curvature features, PR #1117 in progress) — shows τ_z ≈ 9.5% regardless of what surface features we add. τ_z (vertical shear, σ=8.747% in SOTA `56bcqp3m`) is the dominant residual WSS error. The reason: τ_z is primarily driven by the velocity gradient ∂u/∂z in the boundary layer, which is a _volume_ quantity, not a surface-only quantity. Surface features like curvature, normals, and wind-exposure cannot tell the model about the velocity profile near the wall.

**H3 hypothesis:** If near-wall volume tokens (SDF < 0.05 m) attend to surface tokens during decoding, the model can incorporate boundary-layer velocity-gradient signal into WSS prediction — specifically addressing τ_z.

## Background and supporting evidence

- **H2 EP10 (tanjiro #1117, in progress):** τ_y is closing toward baseline (+0.16pp gap at EP10 vs +0.48pp at EP3). τ_z is STUCK at +0.77pp above baseline despite 10 epochs. Curvature features help lateral shear but cannot reach vertical wake physics.
- **Per-axis insight from #1086 terminal:** τ_x=5.971%, τ_y=7.362%, τ_z=8.747% — τ_z is the dominant component and ~2.4× the AB-UPT target (3.63%).
- **Tay-track finding:** local geometric features have ρ=-0.11 with |WSS| and don't help τ_z in sampling experiments. Architecture (not features) is the right lever now.
- **SDF data is available:** the training dataset `rawcanon_20260511` includes pre-computed SDF values as a volume input channel. The model already has access to SDF for each volume point, making near-wall filtering straightforward.

**Target WSS:** test_wss < 5.85% (Transolver-3 SOTA). Current SOTA test_wss = 6.727%. Gap = 0.877pp = 13% relative reduction.

## Baseline to beat (SOTA on corrected split — all test metrics)

| Metric | SOTA PR #972 `56bcqp3m` / eval `zxnhtagj` |
|---|---:|
| **test_abupt (PRIMARY)** | **5.844%** |
| test_vol_p (FLOOR ≤ 3.643%) | 3.643% |
| test_surf_p (FLOOR ≤ 3.577%) | 3.577% |
| **test_wss (TARGET: < 5.85%)** | 6.727% |
| test_wss_x | ~5.97% |
| test_wss_y | ~7.36% |
| **test_wss_z** | **8.747%** |

Stack: Lion + 6L STRING 5-sigma + GradNorm α=0.5 + EMA(0.999) + Y-sym(0.5) + bs=1 + 65k surf/vol + SDF(α=2.0) + corrected dataset.
Dataset: `/mnt/new-pvc/Processed/drivaerml_processed_rawcanon_20260511`

### Step 1: locate the surface decoder

Find the class/function that produces the final surface predictions (WSS + surface pressure) from surface tokens. In `trainer_runtime.py` (or a model file it imports), this should be the block that:
- Takes surface point features (from the Transolver encoder blocks)
- Outputs per-point predictions for τ_x, τ_y, τ_z, p_s

The typical pattern in Transolver-style code is something like `surface_head = nn.Linear(hidden_dim, n_surface_outputs)` applied to `surface_feats[-1]`.

### Step 2: add a near-wall volume cross-attention module

Add a new module before the surface head. Example implementation:

```python
import torch
import torch.nn as nn

class NearWallCrossAttention(nn.Module):
    """
    Cross-attention: surface tokens query near-wall volume tokens.
    Near-wall = |SDF| < sdf_threshold_m.
    """
    def __init__(self, hidden_dim: int, num_heads: int = 4, sdf_threshold_m: float = 0.05):
        super().__init__()
        self.sdf_threshold_m = sdf_threshold_m
            embed_dim=hidden_dim, num_heads=num_heads,
            batch_first=True, dropout=0.0
        )
        # Zero-init output proj so the module starts as a pass-through
        nn.init.zeros_(self.attn.out_proj.weight)
        nn.init.zeros_(self.attn.out_proj.bias)

    def forward(self,
                surface_feats: torch.Tensor,    # (B, N_surf, D)
                volume_feats: torch.Tensor,     # (B, N_vol, D)
                volume_sdf: torch.Tensor,       # (B, N_vol) — SDF values
               ) -> torch.Tensor:
        # Filter to near-wall volume points
        near_wall_mask = volume_sdf.abs() < self.sdf_threshold_m  # (B, N_vol)
        
        # If no near-wall points (edge case), skip
        if not near_wall_mask.any():
            return surface_feats
        
        # Gather near-wall features — variable count per batch item
        # Simplest approach: apply mask globally (pad with zeros if needed)
        # For DDP-safety, use max near-wall count across batch and pad
        max_nw = near_wall_mask.sum(dim=1).max().item()  # int
        B, D = surface_feats.shape[0], surface_feats.shape[-1]
        
        nw_feats = torch.zeros(B, max_nw, D, device=surface_feats.device,
                               dtype=surface_feats.dtype)
        nw_key_padding_mask = torch.ones(B, max_nw, device=surface_feats.device,
                                         dtype=torch.bool)  # True=ignore
        for b in range(B):
            idx = near_wall_mask[b].nonzero(as_tuple=True)[0]
            n = idx.shape[0]
            nw_feats[b, :n] = volume_feats[b, idx]
            nw_key_padding_mask[b, :n] = False  # valid tokens
        
        # Cross-attention: surface queries near-wall volume
        attn_out, _ = self.attn(
            query=surface_feats,      # (B, N_surf, D)
            key=nw_feats,             # (B, max_nw, D)
            value=nw_feats,
            key_padding_mask=nw_key_padding_mask,  # (B, max_nw) True=ignore
        )
        # Pre-norm residual
        return self.norm(surface_feats + attn_out)
```

### Step 3: wire it into the forward pass

In the model's `forward()` method, after the encoder has produced both `surface_feats` and `volume_feats`, and before the surface output head:

```python
# Instantiate in __init__ (add to the model class):
self.near_wall_cross_attn = NearWallCrossAttention(
    hidden_dim=self.hidden_dim, num_heads=4, sdf_threshold_m=0.05
)

# In forward(), just before the surface head:
# volume_sdf should be the SDF values of volume points — shape (B, N_vol)
# This is typically already available as part of volume_x[:, :, sdf_channel_idx]
surface_feats = self.near_wall_cross_attn(surface_feats, volume_feats, volume_sdf)
surface_out = self.surface_head(surface_feats)
```

You will need to identify:
1. The correct variable names for `surface_feats`, `volume_feats`, and `volume_sdf` in the existing forward pass
2. The SDF channel index in the volume input tensor (check the dataset loader or training config)
3. Where to insert the cross-attention call (just before the WSS/surface prediction head)

### Step 4: add CLI flag

Add `--use-near-wall-cross-attn` boolean flag (default False) to `train.py`. This lets you ablate it cleanly.

## Run command

```bash
  --output-dir outputs/h3_near_wall_cross_attn \
  --epochs 30 --batch-size 1 \
  --lr-warmup-epochs 1 --lr-cosine-t-max 30 \
  --use-ema --ema-decay 0.999 --ema-start-step 500 \
  --no-compile-model \
  --no-use-gradnorm \
  --use-near-wall-cross-attn \
  --wandb-group h3-near-wall-cross-attention \
  --wandb-name dl24-nezuko/h3-near-wall-cross-attn \
  --agent dl24-nezuko
```

Use lr=1e-4 (SOTA LR, not 9e-5 — we learned from #1101 that 9e-5 hurts WSS and surf_p).

## Gates

**Smoke run (before the 30-ep long run):**
Run 3 epochs. Gate: val_abupt ≤ 8.5%, val_wss ≤ 8.0%. If it diverges (>15% abupt at EP1), halt — the zero-init may not be propagating gradients correctly.

**EP3:** val_abupt ≤ 7.5%, val_wss ≤ 7.8%, val_vol_p ≤ 5.5%
**EP6:** val_wss ≤ 7.4% (H3 should start showing τ_z improvement by here)
**EP10:** val_wss ≤ 7.1%, τ_z ≤ 9.3% ← KEY GATE. If τ_z hasn't improved vs SOTA's 8.747%, H3 isn't working.
**EP15:** val_abupt ≤ 6.50%, val_wss ≤ 6.9%
**EP20:** val_abupt ≤ 6.30%, val_wss ≤ 6.85%

Post per-axis val_wss (τ_x, τ_y, τ_z) at EP3, EP6, EP10, EP15. τ_z is the key signal — if τ_z ≤ 9.0% at EP10 we're on track for a genuine WSS win.

## What to watch for

**Healthy signs:** τ_z improving faster than τ_x/τ_y — this would confirm near-wall volume info is specifically helping vertical shear.

**Warning signs:**
- Any NaN/OOM at EP1-3 (zero-init should prevent, but check gradients)
- τ_z stuck at 9.5%+ through EP6 — means the cross-attention isn't firing (check that volume_sdf is correctly wired)
- val_surf_p drift upward — the extra parameters could overfit if the near-wall filter is too broad. Reduce sdf_threshold_m to 0.03m if this happens.

## What this does NOT change

- Loss function: unchanged MSE + GradNorm α=0.5
- Learning rate: 1e-4 (SOTA)
- Training stack: full PR #972 config
- Only architectural addition: one NearWallCrossAttention module in the surface decoder

### DIFF
--- a/model.py
+++ b/model.py
@@ -295,6 +295,130 @@ class Transformer(nn.Module):
         return x
 
 
+class NearWallCrossAttention(nn.Module):
+    """Cross-attention: surface tokens query near-wall volume tokens (|SDF|<thr).
+
+    H3 hypothesis: τ_z (vertical wall-shear) is driven by boundary-layer
+    velocity gradients which are a volume quantity. By letting surface tokens
+    cross-attend to volume tokens with |SDF| < ``sdf_threshold_m`` after the
+    shared backbone, the surface head can incorporate near-wall flow info.
+
+    Inputs are expected to already be normalized (in this codebase that is
+    enforced by ``SurfaceTransolver.norm`` before the surface/volume split).
+    Implementation uses ``F.scaled_dot_product_attention`` (matching the
+    existing ``TransolverAttention`` pattern).
+
+    The output projection uses a small (std=0.01) init rather than exact
+    zero-init: a pure zero-init would zero-out gradients into Q/K/V on the
+    first iter (dL/d(act) = upstream_grad @ W_out.T == 0). A small-but-non-
+    zero init keeps the cross-attention near-identity at start while letting
+    Q/K/V receive non-trivial gradients from step 1.
+
+    Note on DDP: with this module enabled, the data loader produces view-
+    asymmetric batches (empty surface or empty volume tensors on trailing
+    views of cases where the two modalities have different point counts).
+    On those steps, Q/K/V/out projections receive no gradient, so DDP must
+    be constructed with ``find_unused_parameters=True`` to handle the per-
+    step variation in the used-parameter set across ranks.
+    """
+
+    def __init__(
+        self,
+        hidden_dim: int,
+        num_heads: int = 4,
+        sdf_threshold_m: float = 0.05,
+        max_keys: int = 1024,
+        out_proj_init_std: float = 0.01,
+    ):
+        super().__init__()
+        if hidden_dim % num_heads != 0:
+            raise ValueError(
+                f"hidden_dim ({hidden_dim}) must be divisible by num_heads ({num_heads})"
+            )
+        self.hidden_dim = hidden_dim
+        self.num_heads = num_heads
+        self.head_dim = hidden_dim // num_heads
+        self.sdf_threshold_m = sdf_threshold_m
+        self.max_keys = max_keys
+
+        self.q_proj = nn.Linear(hidden_dim, hidden_dim)
+        self.k_proj = nn.Linear(hidden_dim, hidden_dim)
+        self.v_proj = nn.Linear(hidden_dim, hidden_dim)
+        self.out_proj = nn.Linear(hidden_dim, hidden_dim)
+
+        nn.init.trunc_normal_(self.q_proj.weight, std=0.02)
+        nn.init.zeros_(self.q_proj.bias)
+        nn.init.trunc_normal_(self.k_proj.weight, std=0.02)
+        nn.init.zeros_(self.k_proj.bias)
+        nn.init.trunc_normal_(self.v_proj.weight, std=0.02)
+        nn.init.zeros_(self.v_proj.bias)
+        # Small-init for out_proj keeps the cross-attention near-identity at
+        # init (residual change is O(0.01) * normal_activations) while keeping
+        # gradients to Q/K/V populated for DDP.
+        nn.init.normal_(self.out_proj.weight, std=out_proj_init_std)
+        nn.init.zeros_(self.out_proj.bias)
+
+        self.norm = nn.LayerNorm(hidden_dim, eps=1e-6)
+
+    def forward(
+        self,
+        surface_feats: torch.Tensor,
+        volume_feats: torch.Tensor,
+        volume_sdf: torch.Tensor,
+        surface_mask: torch.Tensor | None = None,
+        volume_mask: torch.Tensor | None = None,
+    ) -> torch.Tensor:
+        # Fixed top-K: pick the K volume tokens with smallest |SDF| per sample.
+        # This gives a fixed key-tensor shape (B, K, D) across all batches and
+        # ranks — no CUDA syncs, no per-sample Python loop, no kernel
+        # recompilation. Tokens beyond ``sdf_threshold_m`` are masked out via
+        # an additive attention mask so they don't influence attention.
+        B, n_surf, D = surface_feats.shape
+        _, n_vol, _ = volume_feats.shape
+        k = min(self.max_keys, n_vol)
+        H = self.num_heads
+        d = self.head_dim
+
+        sdf_abs = volume_sdf.abs()
+        if volume_mask is not None:
+            valid_volume = volume_mask.to(dtype=torch.bool)
+            sdf_abs = torch.where(
+                valid_volume, sdf_abs, torch.full_like(sdf_abs, float("inf"))
+            )
+
+        topk_values, topk_indices = torch.topk(sdf_abs, k, dim=1, largest=False)
+        idx_expand = topk_indices.unsqueeze(-1).expand(-1, -1, D)
+        nw_feats = torch.gather(volume_feats, dim=1, index=idx_expand)
+        # True = mask out; unconditionally allow the closest key per sample to
+        # guard against the (rare) case where every nearest volume point is
+        # still beyond ``sdf_threshold_m`` (would otherwise softmax a fully
+        # -inf row and emit NaN).
+        pad_mask = topk_values >= self.sdf_threshold_m
+        pad_mask[:, 0] = False
+
+        # Build additive attention mask: 0 for valid keys, -inf for masked.
+        # Shape (B, 1, 1, k) broadcasts across (H, N_surf). Construct in fp32 so
+        # the -inf mask value survives bf16 autocast precision; SDPA upcasts the
+        # mask internally as needed.
+        attn_mask = torch.zeros(
+            B, 1, 1, k, device=surface_feats.device, dtype=torch.float32
+        )
+        attn_mask = attn_mask.masked_fill(
+            pad_mask.unsqueeze(1).unsqueeze(1), float("-inf")
+        )
+
+        q = self.q_proj(surface_feats).view(B, n_surf, H, d).transpose(1, 2)
+        k_t = self.k_proj(nw_feats).view(B, k, H, d).transpose(1, 2)
+        v_t = self.v_proj(nw_feats).view(B, k, H, d).transpose(1, 2)
+
+        out = F.scaled_dot_product_attention(q, k_t, v_t, attn_mask=attn_mask)
+        out = out.transpose(1, 2).contiguous().view(B, n_surf, D)
+        out = self.out_proj(out)
+
+        result = self.norm(surface_feats + out)
+        return _apply_token_mask(result, surface_mask)
+
+
 class SurfaceTransolver(nn.Module):
     """Grouped Transolver for surface pressure, wall shear, and volume pressure."""
 
@@ -315,6 +439,10 @@ class SurfaceTransolver(nn.Module):
         pe_kind: str = "sincos",
         pe_num_features: int = 16,
         pe_init_sigmas: list[float] | None = None,
+        use_near_wall_cross_attn: bool = False,
+        near_wall_num_heads: int = 4,
+        near_wall_sdf_threshold_m: float = 0.05,
+        near_wall_max_keys: int = 1024,
     ):
         super().__init__()
         self.space_dim = space_dim
@@ -362,6 +490,19 @@ class SurfaceTransolver(nn.Module):
             dropout=dropout,
         )
         self.norm = nn.LayerNorm(n_hidden, eps=1e-6)
+        self.use_near_wall_cross_attn = use_near_wall_cross_attn
+        self.near_wall_num_heads = near_wall_num_heads
+        self.near_wall_sdf_threshold_m = near_wall_sdf_threshold_m
+        self.near_wall_max_keys = near_wall_max_keys
+        if use_near_wall_cross_attn:
+            self.near_wall_cross_attn = NearWallCrossAttention(
+                hidden_dim=n_hidden,
+                num_heads=near_wall_num_heads,
+                sdf_threshold_m=near_wall_sdf_threshold_m,
+                max_keys=near_wall_max_keys,
+            )
+        else:
+            self.near_wall_cross_attn = None
         self.surface_out = LinearProjection(n_hidden, self.surface_output_dim)
         self.volume_out = LinearProjection(n_hidden, self.volume_output_dim)
 
@@ -433,6 +574,21 @@ class SurfaceTransolver(nn.Module):
         surface_hidden = hidden_norm[:, cursor : cursor + surface_tokens]
         cursor += surface_tokens
         volume_hidden = hidden_norm[:, cursor : cursor + volume_tokens]
+
+        if (
+            self.near_wall_cross_attn is not None
+            and surface_tokens > 0
+            and volume_tokens > 0
+        ):
+            # volume_x channel layout is [x, y, z, sdf]; SDF is channel index 3
+            volume_sdf = volume_x[:, :, 3]
+            surface_hidden = self.near_wall_cross_attn(
+                surface_feats=surface_hidden,
+                volume_feats=volume_hidden,
+                volume_sdf=volume_sdf,
+                surface_mask=surface_mask,
+                volume_mask=volume_mask,
+            )
 
         if surface_x is not None:
             surface_preds = self.surface_out(surface_hidden) * surface_mask.unsqueeze(-1)
--- a/train.py
+++ b/train.py
@@ -132,6 +132,10 @@ class Config:
     use_gradnorm: bool = False
     gradnorm_alpha: float = 1.0
     gradnorm_lr: float = 1e-3
+    use_near_wall_cross_attn: bool = False
+    near_wall_num_heads: int = 4
+    near_wall_sdf_threshold_m: float = 0.05
+    near_wall_max_keys: int = 1024
 
 
 def parse_args(argv: Iterable[str] | None = None) -> Config:
@@ -235,6 +239,10 @@ def build_model(config: Config) -> SurfaceTransolver:
         pe_kind=config.model_pe,
         pe_num_features=config.pe_num_features,
         pe_init_sigmas=parse_pe_init_sigmas(config.pe_init_sigmas),
+        use_near_wall_cross_attn=config.use_near_wall_cross_attn,
+        near_wall_num_heads=config.near_wall_num_heads,
+        near_wall_sdf_threshold_m=config.near_wall_sdf_threshold_m,
+        near_wall_max_keys=config.near_wall_max_keys,
     )
 
 
@@ -371,9 +379,9 @@ def run_eval_only(config: Config, state) -> None:
     model: nn.Module = build_model(config).to(device)
     n_params = sum(param.numel() for param in model.parameters())
     if state.enabled:
-        ddp_kwargs = {}
+        ddp_kwargs = {"find_unused_parameters": config.use_near_wall_cross_attn}
         if device.type == "cuda":
-            ddp_kwargs = {"device_ids": [state.local_rank], "output_device": state.local_rank}
+            ddp_kwargs.update({"device_ids": [state.local_rank], "output_device": state.local_rank})
         model = DistributedDataParallel(model, **ddp_kwargs)
     base_model = unwrap_model(model)
 
@@ -465,9 +473,22 @@ def main(argv: Iterable[str] | None = None) -> None:
         if config.compile_model:
             model = torch.compile(model)
         if state.enabled:
-            ddp_kwargs = {}
+            # find_unused_parameters=True: the data loader emits view-asymmetric
+            # batches (a case with surface_views < volume_views yields empty
+            # surface tensors on the trailing views, and vice versa). When
+            # surface is empty, the near-wall cross-attention produces a
+            # 0-shape output, so q_proj / k_proj / v_proj / out_proj receive no
+            # gradient on that step. Across ranks, different views land on
+            # different ranks each step, so the set of "used" parameters
+            # differs per-rank-per-step. Without find_unused_parameters,
+            # DDP's gradient AllReduce desyncs at first iteration
+            # (manifesting as the SeqNum=6 split-collective hang). Bitmap
+            # AllReduce at forward start adds a small overhead but keeps
+            # everything consistent. Required whenever
+            # use_near_wall_cross_attn is set; harmless otherwise.
+            ddp_kwargs = {"find_unused_parameters": config.use_near_wall_cross_attn}
             if device.type == "cuda":
-                ddp_kwargs = {"device_ids": [state.local_rank], "output_device": state.local_rank}
+                ddp_kwargs.update({"device_ids": [state.local_rank], "output_device": state.local_rank})
             model = DistributedDataParallel(model, **ddp_kwargs)
         base_model = unwrap_model(model)
         if state.is_main:


morganmcg1:
ADVISOR: Run `h75p7dt9` is healthy from my W&B side-check at ~EP3-4 (step ~40,720):

| Metric | EP3 | Gate | Status |
|---|---:|---:|---|
| val_abupt | 6.661% | ≤7.5% | ✅ PASS by 0.84pp |
| val_wss | 7.609% | ≤7.5% | ⚠️ Marginal +0.109pp (trending down at −0.075pp/1k steps, should clear by EP4) |
| val_vol_p | **3.654%** | (informational) | 🎯 0.011pp from SOTA test floor 3.643% |
| val_surf_p | 4.329% | — | OK |

**Continue — do not kill.** The marginal WSS miss is bounded and the descent rate suggests EP4 will clear. The val_vol_p=3.654% is intriguing — H3's near-wall cross-attention should affect WSS more than vol_p, so this might be backbone-shared-representation effect (similar to fern's H4 surprise).

Please post your official EP3 gate report when you can — include:
1. Per-axis val_wss breakdown (τ_x / τ_y / τ_z) — H3's near-wall cross-attn is hypothesized to most help τ_z (largest dominant error axis)
2. Implementation summary — how did you wire the SDF<0.05m volume tokens into the surface decoder cross-attention?
3. Crash recovery details (your previous attempts at 05:33Z / 05:51Z) — what was the bug?

Continue to EP6 gate (~13:00Z, val_wss ≤7.2%).

morganmcg1:
STUDENT dl24-nezuko:
SENPAI-RESULT: {"terminal":true,"status":"complete","pending_arms":false,"wandb_run_ids":["h75p7dt9"],"primary_metric":{"name":"val_primary/abupt_axis_mean_rel_l2_pct","value":6.22124},"test_metric":{"name":"test_primary/abupt_axis_mean_rel_l2_pct","value":6.02042}}

## Results — H3 Near-Wall Cross-Attention: **HYPOTHESIS FALSIFIED**

### TL;DR

The PR's own KEY GATE was **EP10 τ_z ≤ 9.3% — "If τ_z hasn't improved vs SOTA's 8.747%, H3 isn't working."**

Result: **val τ_z bottomed at 9.638% (EP16)** and **test τ_z=9.102%** — both **WORSE** than SOTA's 8.747%. Near-wall volume cross-attention did not improve τ_z; if anything it slightly degraded every WSS axis.

### Test metrics vs SOTA `56bcqp3m`

| Metric | SOTA `56bcqp3m` | H3 `h75p7dt9` | Δ |
|---|---:|---:|---:|
| **test_abupt (PRIMARY)** | **5.844%** | 6.020% | **+0.176pp WORSE** |
| **test_wss** (target <5.85%) | 6.727% | **7.012%** | +0.285pp WORSE (target MISSED by 1.16pp) |
| test_wss_x | ~5.97% | 6.234% | +0.26pp WORSE |
| test_wss_y | ~7.36% | 7.573% | +0.21pp WORSE |
| **test_wss_z** (H3 target axis) | **8.747%** | **9.102%** | **+0.355pp WORSE** |
| test_surf_p (floor ≤ 3.577%) | 3.577% | 3.761% | +0.184pp WORSE |
| test_vol_p (floor ≤ 3.643%) | 3.643% | **3.432%** | -0.211pp BETTER |
| wall_shear_mae | — | 0.05748 | — |

### Per-epoch val trajectory (full 30-epoch run, source: wandb `h75p7dt9`)

| EP | abupt% | wss% | τx% | τy% | τz% | surf_p% | vol_p% | Gate verdict |
|----|---:|---:|---:|---:|---:|---:|---:|---|
|  1 | 20.862 | 22.749 | 19.787 | 25.914 | 29.864 | 15.229 | 13.515 | (warmup) |
|  2 | 7.510 | 8.430 | 7.194 | 9.778 | 11.216 | 4.833 | 4.527 | |
|  3 | **6.661** | **7.609** | 6.579 | 8.563 | 10.177 | 4.329 | **3.654** | EP3 (abupt≤7.5✓, wss≤7.8✓, vol_p≤5.5✓) → **PASS** |
|  4 | 6.465 | 7.394 | 6.423 | 8.220 | 9.920 | 4.208 | 3.554 | |
|  5 | 6.380 | 7.291 | 6.332 | 8.069 | 9.834 | 4.154 | 3.512 | |
|  6 | 6.356 | **7.271** | 6.330 | 7.998 | 9.814 | 4.135 | 3.504 | EP6 (wss≤7.4✓) → **PASS by 0.13pp** |
|  7 | 6.292 | 7.207 | 6.286 | 7.907 | 9.716 | 4.089 | 3.461 | |
|  8 | 6.302 | 7.217 | 6.306 | 7.898 | 9.721 | 4.094 | 3.491 | |
|  9 | 6.261 | 7.171 | 6.263 | 7.831 | 9.679 | 4.074 | 3.458 | |
| **10** | 6.286 | **7.201** | 6.294 | 7.837 | **9.738** | 4.085 | 3.478 | **EP10 KEY GATE (wss≤7.1, τz≤9.3) → FAIL** (wss +0.10pp, τz +0.44pp) |
| 11 | 6.263 | 7.177 | 6.283 | 7.800 | 9.694 | 4.071 | 3.469 | |
| 12 | 6.256 | 7.175 | 6.286 | 7.786 | 9.690 | 4.057 | 3.460 | |
| 13 | 6.256 | 7.178 | 6.293 | 7.778 | 9.690 | 4.064 | 3.455 | |
| 14 | 6.229 | 7.149 | 6.274 | 7.743 | 9.639 | 4.044 | 3.445 | |
| **15** | 6.242 | 7.164 | 6.281 | 7.754 | 9.677 | 4.052 | 3.446 | EP15 (abupt≤6.50✓, wss≤6.9 ✗ by 0.26pp) → MIXED |
| **16** | **6.221** ⭐ | 7.155 | 6.291 | 7.723 | **9.638** | 4.028 | 3.426 | **RUN BEST** |
| 17 | 6.263 | 7.203 | 6.326 | 7.780 | 9.705 | 4.061 | 3.446 | |
| 18 | 6.274 | 7.215 | 6.337 | 7.786 | 9.733 | 4.070 | 3.446 | |
| 19 | 6.265 | 7.198 | 6.310 | 7.786 | 9.720 | 4.063 | 3.443 | |
| 20 | **6.261** | 7.202 | 6.329 | 7.774 | 9.700 | 4.054 | 3.450 | EP20 (abupt≤6.30✓, wss≤6.85 ✗ by 0.35pp) → MIXED |
| 21 | 6.282 | 7.222 | 6.333 | 7.804 | 9.757 | 4.067 | 3.451 | |
| 22 | 6.296 | 7.240 | 6.350 | 7.822 | 9.779 | 4.077 | 3.452 | |
| 23 | 6.279 | 7.219 | 6.331 | 7.801 | 9.745 | 4.065 | 3.451 | |
| 24 | 6.301 | 7.248 | 6.359 | 7.825 | 9.788 | 4.079 | 3.456 | |
| 25 | 6.310 | 7.258 | 6.369 | 7.828 | 9.803 | 4.089 | 3.460 | |
| 26 | 6.362 | 7.323 | 6.439 | 7.874 | 9.883 | 4.148 | 3.466 | |
| 27 | 6.321 | 7.272 | 6.380 | 7.840 | 9.832 | 4.093 | 3.459 | |
| 28 | 6.319 | 7.272 | 6.382 | 7.838 | 9.826 | 4.090 | 3.459 | |
| 29 | 6.324 | 7.276 | 6.386 | 7.840 | 9.831 | 4.091 | 3.472 | |
| 30 | 6.323 | 7.277 | 6.386 | 7.839 | 9.836 | 4.092 | 3.460 | terminal |

### Full-val (best EMA EP16) and test (best checkpoint)

```
full_val       loss=0.00490 abupt_axis_rel_l2_pct=6.2212 surface_p_mae=0.01043 volume_p_mae=6.50604 wall_shear_mae=0.05920 cases=34
test_surface   loss=0.00451 abupt_axis_rel_l2_pct=6.0204 surface_p_mae=0.01048 volume_p_mae=6.65246 wall_shear_mae=0.05748 cases=50
```

### Command used (exact)

```bash
  --output-dir outputs/h3_near_wall_cross_attn \
  --epochs 30 --batch-size 1 \
  --lr-warmup-epochs 1 --lr-cosine-t-max 30 \
  --use-ema --ema-decay 0.999 --ema-start-step 500 \
  --no-compile-model \
  --no-use-gradnorm \
  --use-near-wall-cross-attn \
  --wandb-group h3-near-wall-cross-attention \
  --wandb-name dl24-nezuko/h3-near-wall-cross-attn \
  --agent dl24-nezuko
```

### Run details

- **W&B group:** `h3-near-wall-cross-attention` ([all 8 ranks](https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8))
- **W&B run IDs (rank0…rank7):** `h75p7dt9`, `bj69y2id`, `swpx6rjz`, `wycghdne`, `h7in32xw`, `2o947g8q`, `uqoiuyli`, `cxf7lfu1`
- **Wall-clock:** 1320.8 minutes (~22 h 1 min) for 30 epochs on 8× H100 96 GB (8 GPUs, batch_size=1)
- **Model params:** 19.89 M (+ ~1.05 M from NearWallCrossAttention vs baseline)
- **Total steps:** 329 250 global (10 975 / epoch / rank × 30 epochs)
- **Best epoch:** 16 (selected on `val_primary/abupt_axis_mean_rel_l2_pct`, EMA shadow)
- **Peak VRAM:** not directly logged by this stack, but the run sat well within 96 GB on H100 (no OOM with same `--train-volume-points 65000 --train-surface-points 65000` as PR #972). Mask-pad fallback path through SDPA does not allocate `(B, N_surf, K)` attention matrices in fp32 thanks to `F.scaled_dot_product_attention` flash-attn backend.

### Implementation summary (in answer to the EP3 advisor comment)

- **Module** `NearWallCrossAttention` (in `model.py`): SDPA-based cross-attention, surface tokens are queries; keys/values are the **top-K** volume tokens with smallest |SDF| per sample (K=1024). Tokens beyond `sdf_threshold_m=0.05` are masked out via an additive attention mask; the single nearest volume token is always unmasked to prevent `softmax(-inf row)` → NaN.
- **Wiring (`SurfaceTransolver.forward`):** after the shared Transolver backbone produces `hidden_norm` and it is split into `surface_hidden` / `volume_hidden`, the cross-attention is applied to `surface_hidden` only, *before* the surface output projection. Volume decoder path is untouched. SDF channel is read from `volume_x[:, :, 3]` (i.e. the 4th input channel, where the dataset stores pre-computed SDF for `rawcanon_20260511`).
- **Init:** Q/K/V projections trunc_normal(std=0.02), out_proj normal(std=0.01) (not exact zero — see DDP note below).
- **CLI flag:** `--use-near-wall-cross-attn` (default False) plus `--near-wall-num-heads 4`, `--near-wall-sdf-threshold-m 0.05`, `--near-wall-max-keys 1024`.

### Crash recovery (in answer to the 05:33Z / 05:51Z question)

The two earlier 8-GPU launches hung at NCCL `SeqNum=6` (first iter) — every worker timed out on `c10d:NCCL:WatchDog`. Root cause was non-trivial:

1. **Initial misdiagnosis:** I first hypothesised a numerical issue (bf16 in the attn mask, or pure-zero out_proj killing Q/K/V gradients). Rewrote `NearWallCrossAttention` to use SDPA + `out_proj_init_std=0.01` and built the attn-mask in fp32. **Re-launched → still hung at SeqNum=6.**
2. **Inspection of NCCL traces:** ranks split between a 1-element collective (`distributed_any` in `trainer_runtime.py:131`) and the full 19.89M-element gradient AllReduce. Last-completed work=5, last-enqueued=6 on every rank — i.e. ranks disagreed on which collective came 6th.
3. **Tried `static_graph=True`** — DDP immediately erred with *"Your training graph has changed in this iteration, e.g., one parameter is unused in first iteration, but then got used in the second iteration. this is not compatible with static_graph set to True."* That error is the smoking gun: the graph really is dynamic.
4. **Real root cause:** the dataset emits *view-asymmetric* batches. `DrivAerMLSurfaceDataset._indices()` returns `torch.empty(0)` for any view whose `view_index >= group_view_count`, so a case with `surface_views < volume_views` (or vice-versa) produces trailing batches that have an **empty surface tensor** (or empty volume tensor). When surface is empty, `NearWallCrossAttention` returns a 0-shape output and its q/k/v/out projections receive no gradient that step. With `DistributedSampler` slicing those view-asymmetric batches across 8 ranks, *which* ranks get the empty modality differs per step → DDP's per-parameter gradient hooks fire in different orders on different ranks → SeqNum desync at iter 1.
5. **Fix:** swap to `find_unused_parameters=True` when `use_near_wall_cross_attn=True` (both DDP wraps in `train.py`: eval_only path and main path). Cost is one bitmap AllReduce at forward start, negligible against the 65k-point Transolver forward.
6. **Verification:** 1-epoch DDP smoke (`runs/h3_fix3_smoke`, 850 steps, 76.8 s, all 8 ranks) completed cleanly. Full 30-epoch run launched on top of the same fix completed without a single DDP error in 22 hours.

### What happened — did it work? why not?

**No — the hypothesis is falsified.** The PR's mechanism (near-wall volume cross-attention into the surface decoder) was meant to lower τ_z by exposing the surface head to boundary-layer velocity-gradient information. The result is the opposite on every WSS axis:

- **τ_z degraded by +0.36pp on test** (the explicit target axis). Across the full run, val τ_z never reached the EP10 gate of 9.3% — bottomed at 9.638% at EP16.
- **Vector τ degraded by +0.29pp** (test 7.012% vs SOTA 6.727%). Per-axis: every component got worse (τ_x +0.26pp, τ_y +0.21pp, τ_z +0.36pp).
- **surf_p degraded by +0.18pp** (4.03% val / 3.76% test vs SOTA's 3.58%), consistent with the warning in the PR ("the extra parameters could overfit if the near-wall filter is too broad"). I did not try lowering `sdf_threshold_m` to 0.03 because the WSS axes were degraded as well, not just surf_p.
- **vol_p IMPROVED by -0.21pp on test** (3.43% vs SOTA's 3.64%) — interesting, but a side-effect, not the target. The cross-attention only modifies the surface decoder; the only path for vol_p to benefit is *backward gradient* flowing from the surface attention residual into volume features through the shared backbone. So adding cross-attention essentially gave the backbone an extra auxiliary loss-shaped signal that helped volume features generalise, without translating into surface-side gains.

**Most likely explanation:** the existing 6L STRING-PE Transolver backbone is already exposing boundary-layer information to the surface tokens implicitly via shared self-attention over the concatenated (surface || volume) sequence. Adding an explicit cross-attention layer with additional parameters at fixed K=1024 selectors:
- duplicates a signal the model already had,
- moves capacity away from the surface output head,
- and (with `find_unused_parameters=True`) costs a non-trivial gradient-hook re-ordering.

The plateau in val_abupt from EP14–EP30 (6.221 ± 0.07pp) and the monotonic τ_z floor of ~9.64% (val) suggest the model is **already at the architectural limit for τ_z on this stack** — and pure cross-attention is not the unlock. The val→test gap is small and consistent (val 6.22 → test 6.02), so the result is robust, not a fluke.

### Suggested follow-ups (for advisor — not implemented)

1. **Volume-side near-wall auxiliary head, NOT surface-side cross-attention.** If τ_z is bottlenecked by *features*, the fastest signal to test is a near-wall auxiliary loss on a small MLP over the same near-wall tokens predicting τ at the closest surface point. This forces the volume features to be τ-informative without disturbing the surface decoder. Distinct from H3 because the regression head lives in the volume branch and gets back-prop directly from a wall-shear target.
2. **Surface-side τ_z-only output scaling** (no architectural change). Per-axis output gain of 2× on τ_z, with GradNorm re-balancing. Since τ_z is the dominant residual and our model is τ_z-floor-limited, simply asking the loss to put more pressure on τ_z is a low-cost falsification.
3. **K-NN-based near-wall keys, indexed by surface-point xyz.** The current top-K filter is *global per case* (the K closest volume tokens to the wall, irrespective of which surface point is querying). A τ_z-targeted variant would query, for each surface point, the K volume tokens nearest in 3D, not just nearest to the wall. This is more expensive but is a genuinely new spatial-locality signal that the shared backbone doesn't already encode.
4. **Detached-K/V cross-attention** to test whether the small vol_p win we saw was actually causal (i.e. shared-backbone gradient flow) or an attention-pattern artefact. If detached-K/V loses the vol_p win, that confirms the mechanism is backward-gradient-flow, not forward-attention, and rules out a whole family of forward-only attention tweaks.
5. **A bug-fix PR (separate from H3) for the data loader's view-asymmetric batches.** The current `_indices() -> torch.empty(0)` behaviour silently disables any cross-modality module (cross-attn, joint losses) on a non-trivial fraction of steps and requires every such module to be wrapped in `find_unused_parameters=True`. A fix would either (a) skip empty-modality batches in the sampler, or (b) emit a sentinel that downstream code can branch on cleanly. Happy to file this as a separate PR if useful.

morganmcg1:
ADVISOR: Closing PR #1129 because ADVISOR: Closing — H3 hypothesis falsified. Excellent post-mortem.

### Decisive close criteria triggered

- **Floor breach** on test_surf_p = 3.761% vs Issue #1056 floor 3.577% (+0.184pp over). Hard fail.
- **Primary regression**: test_abupt = 6.020% vs SOTA 5.844% (+0.176pp worse).
- **WSS hypothesis target axis falsified**: test_τ_z = 9.102% vs SOTA 8.747% (+0.355pp WORSE). The hypothesis's own EP10 gate (τ_z must improve vs 8.747%) was never hit.
- All three τ components degraded on test. The mechanism delivered the opposite of its design intent on every WSS axis.

### Your mechanistic read is the right one

Your analysis — that the existing 6L STRING-PE Transolver backbone already exposes boundary-layer information to surface tokens implicitly via shared self-attention over concatenated (surface || volume) sequences — is exactly right. Adding explicit cross-attention with K=1024 selectors:
- duplicated a signal the model already had,
- moved 1.05M params of capacity away from output heads,
- and forced find_unused_parameters=True (cost) without a corresponding benefit.

The persistent τ_z floor of ~9.64% (val) across the plateau EP14-EP30 is the smoking gun: the model is at the architectural limit for τ_z on this stack. Cross-attention does not unlock it.

### vol_p win via backward gradient flow — consistent with the pattern

test_vol_p −0.211pp under floor is a side-effect, not the target. Your read is correct: cross-attention only modifies the surface decoder, so vol_p improvement can only come from backward gradient flow through the shared backbone. This is the FOURTH consecutive WSS experiment (H1, H2, H4 each had vol_p improvements without WSS gains; H3 makes it four) to show this pattern.

### Crash recovery debugging was excellent work

The view-asymmetric batch root-cause analysis (DDP per-parameter gradient hooks firing in different orders on different ranks because some ranks see empty surface/volume modalities) is a textbook DDP debug story. The find_unused_parameters=True fix is the right tactical solution; your suggested follow-up #5 (proper sampler fix that emits a sentinel or skips empty-modality batches) is also a valuable infrastructure improvement we should consider as a standalone PR.

### Strategic implications: hitting the Plateau Protocol

Across the wave:
- **H1 (frieren wind-exposure input channels)** — CLOSED, floor breach
- **H2 (tanjiro curvature input channels)** — CLOSED, floor breach
- **H3 (nezuko near-wall cross-attn)** — CLOSED 02:33Z, floor breach + WSS regression on target axis
- **H4 (fern per-axis WSS weights)** — CLOSED 02:33Z, floor breach + WSS regression
- **H5 (tanjiro curvature attention bias)** — running, EP26 plateau, vol_p drifting up (likely floor breach at terminal)
- **H6 (frieren wind-exposure attention bias)** — running, EP15 plateau, vol_p drifting up (likely floor breach at terminal)
- **H7 (fern surface_loss_weight=1.5 isolation)** — running since 04:25Z

Four failures + two probable failures = the Plateau Protocol is fully active. The current SOTA stack appears to have an architectural floor on τ_z around 9.5-9.7% (val) that none of these surface-side or cross-modality interventions can break. **The unlock will come from a different strategy tier**, not from another variation of geometric-signal injection.

### Your follow-ups are excellent — picking #1 for the next experiment

Your suggestion #1 (volume-side near-wall auxiliary head predicting τ at closest surface point) is the most surgical of your five suggestions, but it's still in the cross-modality-coupling tier and may hit the same architectural-floor pattern.

I'm going to take a tier-change swing for your next assignment: **optimizer regime change** (Lion → AdamW). The reasoning: H4 showed Lion's sign-based update is contaminated by axis-noise; the plateau pattern across H5/H6/H7 may be partly due to Lion's update geometry being suboptimal for the WSS gradient landscape. AdamW with a higher LR is the cleanest single-variable bold swing.

New PR (H8) incoming. Closing #1129..

PR #1146, opened 2026-05-16
Wave 30 H9': curvature-aware surface feature (port of dl24 #1132)

morganmcg1:
## Hypothesis

**Curvature-aware surface feature (H9').** Compute per-vertex local surface curvature κ from the existing surface normals via a k-NN-of-normals statistic, append it as an 8th surface channel, and let the model learn to use it. Provides geometric prior signal in regions where the τ_z gradient is largest (A-pillar, side mirror, wing tips, rocker panel transitions) — exactly where the τ_z bottleneck is sharpest.

**Why this attacks the τ_z bottleneck:**

The τ_z structural ratio is now confirmed across **NINE distinct mechanisms** all landing in the [1.44, 1.57] band. The 4-of-4 Wave 30 model-side widening pattern (H1/H2/H4/H7 — closed) tells us the bottleneck is **not at the model architecture layer**. Tanjiro's H6 (just closed, paper-worthy) confirms the τ_z bottleneck IS at the output head level but cannot be solved by hard architectural enforcement (throws away 5–8% real GT normal-component signal).

This leaves **data-level signal injection** as the remaining attack surface. We have ONE strong external signal: an advisor working on the parallel `drivaerml-long-20260504` branch (cross-pollination thread on Issue #1056) tested *curvature attention bias* and reported `test_WSS=6.609%` (−0.118pp under SOTA) and `test_τ_z=8.592%` (−0.155pp on dominant axis). That's a real WSS gain AND a real τ_z gain, on the same underlying dataset. We have not tested any curvature-aware signal on tay.

**Mechanism intuition:**
1. **Curvature correlates with τ_z magnitude.** Vehicle geometry has sharp curvature concentrated on the A-pillar, mirror housings, wing tips, and rocker panel transitions — these are the same regions where τ_z is large (separation, vortex shedding) and where the model's τ_z prediction is worst.
2. **Existing normals don't encode curvature directly.** The model sees `(nx, ny, nz)` per vertex but has to infer curvature from the spatial variation of normals across attention windows. A scalar curvature channel hands this for free.
3. **Orthogonal to all 8 in-flight Wave 30 axes.** Operates at the input feature layer, not architecture, loss, sampling, output, or input distribution. Stacks cleanly with H8 mirror-aug (frieren #1143) if both win.

**Theoretical basis:**
- Stokes-friction WSS scales with the wall-normal velocity gradient, which in turbulent boundary layers couples to local surface curvature via the favorable/adverse pressure gradient.
[...]

PR #1143, opened 2026-05-16
Wave 30 H8: Mirror-Symmetry Data Augmentation (x-z plane y-flip)

morganmcg1:
## Hypothesis

**Mirror-Symmetry Data Augmentation (H8).** Apply random left-right (y-axis) mirror reflection to training samples with 50% probability. DrivAerML is generated at **zero yaw and zero pitch**, so every case has *exact* mirror symmetry across the x-z plane. The model has never been told this, and is wasting capacity learning the symmetry from finite data.

**Why this attacks the τ_z bottleneck:**

The τ_z structural ratio is now confirmed across **9 distinct mechanisms** (loss weighting, output decoupling, EMA, sampling, depth, curriculum, mag decomposition, SDF FAR-field, per-channel heads) all landing in the [1.44, 1.57] band. Every mechanism so far has attacked the architecture/loss/sampling at training time. **None has attacked the input distribution.** Mirror augmentation:

1. **Doubles the effective dataset size** (400 → 800 unique-orientation views per epoch) — directly addresses the OOD test-set generalization gap.
2. **Forces τ_y representations to be sign-flip-equivariant** — the model can no longer co-adapt τ_z predictions to τ_y sign patterns, breaking any spurious τ_y/τ_z correlation that biases the τz/τx structural ratio.
3. **Is exactly equivariant** for this dataset — unlike rotation augmentation (which would break the y-vs-z geometry distinction), the mirror is documented in the DrivAerML data paper as an exact geometric symmetry.

**This is the EIGHTH Wave 30 attack axis** — orthogonal to ALL 7 architectural axes in flight:

| Wave 30 axis | What it touches | Mirror-aug touches |
|---|---|---|
| H1 cylindrical coords (#1139) | Input coord frame | Cartesian unchanged |
| H2 normal Fourier (#1136) | Input features | StringSep unchanged |
| H3 soft slice routing (#1138) | Attention logits | Attention unchanged |
| H4 hard MoE routing (#1141) | Attention mask | Attention unchanged |
| H5 Y-architecture (#1137) | Backbone split | Backbone unchanged |
| H6 local-frame WSS (#1134) | Output head | Output head unchanged |
| H7 normal aux head (#1140) | Aux gradient | Aux gradient unchanged |
| **H8 mirror aug (this PR)** | **Input distribution** | — |

Mirror aug operates entirely at the data collation layer, upstream of every model component. It is provably orthogonal to all seven active axes and can be stacked on any future winner.

[...]

PR #1153, opened 2026-05-16
Wave 30 H14: asymmetric LR for surface output head (5× backbone)

morganmcg1:
## Hypothesis (H14 — Asymmetric Learning Rate for Surface Output Head)

The Wave 30 fleet has now closed **6 model-side / architecture-layer attacks** (H1 cylindrical, H2 spectral, H3 soft-routing, H4 hard-routing, H5 Y-arch, H7 normal-aux). The architecture-layer attack surface is **DEFINITIVELY exhausted** for the τ_z structural bottleneck — all 6 closures show the same engaged-but-neutral signature (val widening 1.50–1.55 → test collapse to 1.44–1.47 baseline band).

H6 (tanjiro #1134) broke that pattern via hard τ·n=0 projection → test τz/τx=1.281 (mechanism PASS, absolute FAIL): definitive proof that **the bottleneck is at the output-head layer**, but hard projection throws away ~5-8% of real GT normal-component signal.

The current Wave 30 in-flight fleet covers:
- **Loss layer (3-strong)**: H6' soft τ·n penalty (#1147), H12 magnitude-weighted MSE (#1151), H13 tangent/normal anisotropic (#1152)
- **Data-input layer (3-strong)**: H8 mirror-symmetry aug (#1143), H9' curvature input feature (#1146), H11 multi-scale kNN context (#1150)
- **Output-head layer (1-strong)**: H10 vector-decoupled head (#1148)
- **Optimization layer (0-strong)**: ← THIS IS THE COMPLETELY UNDER-ATTACKED SURFACE

**H14 fills the optimization-layer gap**: split optimizer parameter groups so the `surface_out.*` MLP receives a **5× higher learning rate** than the backbone. Two motivations:

1. **The H6 mechanism PASS confirmed the output head IS the bottleneck location**. Pushing more gradient signal into those specific parameters is the most direct attack consistent with that diagnostic.
2. **Standard Kaggle / ImageNet precedent**: classifier heads consistently benefit from higher LR than feature backbones during fine-tuning (e.g., DeiT/timm `head_init_scale`, MAE linear probing protocol, ULMFit slanted triangular). The intuition: the output projection sees a fixed feature distribution and needs to rapidly adapt to the task, while the backbone needs gentler updates to preserve geometry-aware structure.

**Mechanism if H14 wins**:
- `surface_out` parameters update with effective LR ~2.5e-3 (5× the 5e-4 base) under cosine schedule
- The 2-layer surface MLP (n_hidden=512 → 512 → 4) has ~263k params, ~0.5% of total
- Heightened LR on these 263k params drives faster adaptation of the τ-direction-and-magnitude projection without disturbing the slice-attention representation
- If test τz/τx drops below baseline 1.45 (without floor breach), it confirms output-head capacity-budget was the real constraint

**Mechanism if H14 loses (likely outcomes)**:
- Same engaged-but-neutral signature → output-head LR budget is NOT the issue; problem is upstream (the backbone representation doesn't encode τ_z well enough at any output projection)
- Training instability (NaN, divergence) → asymmetric LR ratio too aggressive; falsifies the simple "more LR on head" hypothesis but motivates a follow-up sweep at 2-3x

## Instructions

**File to modify:** `target/train.py`, specifically `build_optimizer` at line 246 onward.

### Implementation outline

Add a new Config field near `lr`:

```python
@dataclass
class Config:
    ...
    lr: float = 5e-4
    weight_decay: float = 5e-4
    # H14 asymmetric LR for surface output head
    surface_out_lr_multiplier: float = 1.0  # 1.0 = baseline; 5.0 = H14 active arm
```

Modify `build_optimizer` to split parameters into two groups:

```python
def build_optimizer(model: nn.Module, config: Config) -> torch.optim.Optimizer:
    optimizer_name = config.optimizer.lower()

    # Split named_parameters into surface_out head vs. everything else.
    head_params = []
    backbone_params = []
    for name, p in model.named_parameters():
        if not p.requires_grad:
            continue
        # Match `surface_out.*` regardless of DDP wrapping (module.surface_out.* or surface_out.*)
        if "surface_out" in name.split("."):
            head_params.append(p)
        else:
            backbone_params.append(p)

    base_lr = config.lr
    head_lr = base_lr * float(config.surface_out_lr_multiplier)

    # Log split for sanity
    backbone_count = sum(p.numel() for p in backbone_params)
    head_count = sum(p.numel() for p in head_params)
    print(
        f"[build_optimizer] backbone params: {backbone_count:,} @ lr={base_lr:.2e}; "
        f"head (surface_out): {head_count:,} @ lr={head_lr:.2e} "
        f"(multiplier={config.surface_out_lr_multiplier:.2f})"
    )

    param_groups = [
        {"params": backbone_params, "lr": base_lr, "name": "backbone"},
        {"params": head_params,     "lr": head_lr,  "name": "surface_out_head"},
    ]

    if optimizer_name == "adamw":
        return torch.optim.AdamW(param_groups, weight_decay=config.weight_decay)
    if optimizer_name == "lion":
        from lion_pytorch import Lion
        return Lion(
            param_groups,
            weight_decay=config.weight_decay,
            betas=(config.lion_beta1, config.lion_beta2),
            use_triton=False,
        )
    raise ValueError(f"Unknown optimizer '{config.optimizer}'. Supported: adamw, lion.")
```

### LR scheduler compatibility check

The cosine LR scheduler (`get_lr_scheduler` or similar — find it in `trainer_runtime.py` or the main training loop near `scheduler = ...`) typically scales each param_group's LR by the same cosine factor. PyTorch's `CosineAnnealingLR` and friends preserve per-group ratios automatically: at any step, `group['lr'] = group['initial_lr'] * cosine_factor`. **Verify** that the scheduler in use does this. If a custom scheduler is hard-coded to a single `lr` value, modify it to iterate per-group.

Specifically check `target/trainer_runtime.py` for the LR schedule implementation and confirm both param groups will be scaled. If they're not, you'll need to add per-group scaling.

### Log per-group LR to W&B every epoch

Add to the W&B `log_metrics` dict in the train loop:
```python
metrics["lr/backbone"] = optimizer.param_groups[0]["lr"]
metrics["lr/surface_out_head"] = optimizer.param_groups[1]["lr"]
metrics["lr/head_to_backbone_ratio"] = (
    optimizer.param_groups[1]["lr"] / max(optimizer.param_groups[0]["lr"], 1e-12)
)
```

Should hold steady at 5.0 throughout training (modulo cosine scaling).

### Sanity-check: matching baseline at multiplier=1.0

Before launching the H14 arm at 5×, **run a 1-epoch viability check at multiplier=1.0** to confirm:
- The split optimizer produces bit-identical (or near-identical, within numerical noise) results vs the original `model.parameters()` path
- Both param groups have nonzero param counts
- The W&B `lr/head_to_backbone_ratio` logs as exactly 1.0
- No NaN in either group's gradients

This protects against subtle bugs (e.g., missing some params via the `"surface_out" in name.split(".")` rule).

### Reproduce command (H14 active arm, multiplier=5.0)

```bash
cd target/
  --batch-size 4 \
  --epochs 13 \
  --lr 5e-4 \
  --lr-warmup-epochs 1 \
  --optimizer lion \
  --weight-decay 5e-4 \
  --eval-surface-points 65536 \
  --train-volume-points 49152 \
  --eval-volume-points 65536 \
  --tau-z-loss-weight 2.0 \
  --surface-loss-weight 2.0 \
  --model-layers 5 \
  --model-hidden-dim 512 \
  --model-heads 4 \
  --model-slices 128 \
  --use-qk-norm \
  --surface-out-lr-multiplier 5.0 \
  --wandb-group wave30_asymmetric_lr_h14 \
  --wandb-name alphonse/h14-headlr5x-rank{rank}
```

(If your CLI parser auto-generates `--surface-out-lr-multiplier` from the dataclass via reflection — and it does at line ~233-241 of `train.py` — no extra parse_args change is needed.)

### Expected EP gate guidance

- **EP1 sanity**: val_abupt ≤ 33% (warmup floor); `lr/head_to_backbone_ratio = 5.0` confirmed in W&B logs
- **EP3 PASS**: val_abupt ≤ 6.5%
- **EP6 PASS**: val_abupt ≤ 6.3%
- **EP10 PASS**: val_abupt ≤ 6.15%
- **EP13 terminal**: full SENPAI-RESULT with val + test on abupt, WSS, SP, vol_p, τ_x, τ_y, τ_z, plus the per-group LR traces

### Falsification panel

Report at terminal:
1. val + test on all 7 metrics
2. val τz/τx and test τz/τx trajectory (EP3 / EP6 / EP10 / EP13)
3. Per-group LR traces (should be smooth cosine for both, ratio fixed at 5.0)
4. Surface_out grad norm vs backbone grad norm trajectory if logged
5. Comparison vs the 6 closed model-side attacks (test τz/τx 1.44–1.47 band): does H14 break this band?

## Notes

- No ensembles — single-model only.
- Post per-epoch val readout from EP3 onward including the new LR diagnostics.
- Single arm `surface_out_lr_multiplier=5.0`; if mechanism engages but doesn't beat baseline, follow-up sweep over {2.0, 3.0, 10.0} is queued.
- Sanity-check at multiplier=1.0 for a single epoch BEFORE the full 18h run, to catch any param-split bugs.

### DIFF
--- a/train.py
+++ b/train.py
@@ -105,6 +105,7 @@ class Config:
     use_surf_to_vol_xattn: bool = False
     tau_y_loss_weight: float = 1.0
     tau_z_loss_weight: float = 1.0
+    surface_out_lr_multiplier: float = 1.0
     amp_mode: str = "bf16"
     num_workers: int = -1
     pin_memory: bool = True
@@ -245,18 +246,48 @@ def parse_args(argv: Iterable[str] | None = None) -> Config:
 
 def build_optimizer(model: nn.Module, config: Config) -> torch.optim.Optimizer:
     optimizer_name = config.optimizer.lower()
-    if optimizer_name == "adamw":
-        return torch.optim.AdamW(
-            model.parameters(),
-            lr=config.lr,
-            weight_decay=config.weight_decay,
+
+    head_params: list[nn.Parameter] = []
+    backbone_params: list[nn.Parameter] = []
+    head_param_names: list[str] = []
+    for name, p in model.named_parameters():
+        if not p.requires_grad:
+            continue
+        if "surface_out" in name.split("."):
+            head_params.append(p)
+            head_param_names.append(name)
+        else:
+            backbone_params.append(p)
+
+    base_lr = float(config.lr)
+    multiplier = float(config.surface_out_lr_multiplier)
+    head_lr = base_lr * multiplier
+
+    backbone_count = sum(p.numel() for p in backbone_params)
+    head_count = sum(p.numel() for p in head_params)
+    print(
+        f"[build_optimizer] backbone params: {backbone_count:,} @ lr={base_lr:.2e}; "
+        f"head (surface_out): {head_count:,} @ lr={head_lr:.2e} "
+        f"(multiplier={multiplier:.4f}); head tensors: {head_param_names}"
+    )
+    if head_count == 0:
+        raise ValueError(
+            "[build_optimizer] surface_out head group has 0 params; check naming "
+            "(expected to match 'surface_out' as a dot-separated component)."
         )
+
+    param_groups = [
+        {"params": backbone_params, "lr": base_lr, "name": "backbone"},
+        {"params": head_params, "lr": head_lr, "name": "surface_out_head"},
+    ]
+
+    if optimizer_name == "adamw":
+        return torch.optim.AdamW(param_groups, weight_decay=config.weight_decay)
     if optimizer_name == "lion":
         from lion_pytorch import Lion
 
         return Lion(
-            model.parameters(),
-            lr=config.lr,
+            param_groups,
             weight_decay=config.weight_decay,
             betas=(config.lion_beta1, config.lion_beta2),
             use_triton=False,
@@ -1049,10 +1080,19 @@ def main(argv: Iterable[str] | None = None) -> None:
                     and config.weight_log_every > 0
                     and global_step % config.weight_log_every == 0
                 )
+                last_lrs = scheduler.get_last_lr()
+                backbone_lr = float(last_lrs[0])
+                head_lr_value = float(last_lrs[1]) if len(last_lrs) > 1 else backbone_lr
+                head_to_backbone_ratio = (
+                    head_lr_value / backbone_lr if backbone_lr > 1e-12 else 0.0
+                )
                 train_log: dict[str, object] = {
                     "global_step": global_step,
                     "train/lr": current_lr,
                     "lr": current_lr,
+                    "lr/backbone": backbone_lr,
+                    "lr/surface_out_head": head_lr_value,
+                    "lr/head_to_backbone_ratio": head_to_backbone_ratio,
                     "train/vol_points": float(current_train_vol_points),
                     "train/step_skipped": 1.0 if skip_step else 0.0,
                     "train/nonfinite_loss": 1.0 if loss_is_nonfinite else 0.0,
@@ -1175,10 +1215,21 @@ def main(argv: Iterable[str] | None = None) -> None:
                 or (timeout_hit and n_batches > 0)
             )
 
+            epoch_last_lrs = scheduler.get_last_lr()
+            epoch_backbone_lr = float(epoch_last_lrs[0])
+            epoch_head_lr = (
+                float(epoch_last_lrs[1]) if len(epoch_last_lrs) > 1 else epoch_backbone_lr
+            )
+            epoch_head_to_backbone_ratio = (
+                epoch_head_lr / epoch_backbone_lr if epoch_backbone_lr > 1e-12 else 0.0
+            )
             log_metrics: dict[str, object] = {
                 "train/epoch_loss": epoch_train_loss,
-                "train/lr": scheduler.get_last_lr()[0],
-                "lr": scheduler.get_last_lr()[0],
+                "train/lr": epoch_backbone_lr,
+                "lr": epoch_backbone_lr,
+                "lr/backbone": epoch_backbone_lr,
+                "lr/surface_out_head": epoch_head_lr,
+                "lr/head_to_backbone_ratio": epoch_head_to_backbone_ratio,
                 "epoch_time_s": dt,
                 "global_step": global_step,
             }


morganmcg1:
STUDENT alphonse:
SENPAI-RESULT: {"terminal":true,"status":"diverged","pending_arms":false,"wandb_run_ids":["ci9ipu1x"],"primary_metric":{"name":"val_primary/abupt_axis_mean_rel_l2_pct","value":34.6500},"test_metric":{"name":"test_primary/abupt_axis_mean_rel_l2_pct","value":null}}

## Results — H14 Asymmetric Head LR (multiplier=5.0)

**Hypothesis FALSIFIED via training instability.** Lion-optimized head at lr=2.5e-3 (5× backbone 5e-4) diverged in EP2 with non-finite gradients on 100% of subsequent optimizer steps. Mechanism engaged cleanly (ratio=5.0 verified throughout) — the failure mode matches the PR's predicted *"asymmetric LR ratio too aggressive"* outcome and motivates the queued 2-3× follow-up sweep.

### Param-split sanity (mult=1.0, run `7xq2kpa8`) — PASS

| Check | Result |
|---|---|
| head tensors found | `surface_out.{0.weight, 0.bias, 2.weight, 2.bias}` (4 tensors) |
| head params | **264,708** (vs PR estimate ~263k) |
| backbone params | 15,998,869 |
| `lr/head_to_backbone_ratio` | **1.0** exactly |
| EP1 train_epoch_loss | 0.9613 (matches baseline trajectory) |
| EP1 final step loss | 0.279 (smooth descent, grad norms 4-15 throughout) |
| NaN/Inf in gradients | **none** |

The split logic (`"surface_out" in name.split(".")`) is correct and complete. No subtle bugs.

### H14 main arm (mult=5.0, runs `ci9ipu1x` rank0 + 7 sibling DDP ranks)

| Epoch | val_abupt | val_SP | val_VP | val_WSS | τ_x | τ_y | τ_z | τz/τx | train/epoch_loss |
| 1 | 34.6500% | 25.327% | 24.873% | 37.032% | 31.793% | 44.628% | 46.630% | 1.467 | 0.972 |
| 2 | 279.7212% | 80.458% | 285.237% | 383.505% | 93.040% | 818.331% | 121.541% | 1.307 | **96798.50** |

Baseline (PR #972): val_abupt **6.126%**, test_abupt **5.844%** — H14 EP1 was **5.66× worse** even before divergence. Best-checkpoint never updated past EP1 (`best_checkpoint/updated=0`).

### Step-level divergence trajectory (rank 0)

| Step | epoch | train/loss | grad_norm_pre_clip | clipped? | nonfinite_grad? |
| 0 | 1 warmup | 3.626 | 4.50 | 1 | 0 |
| 1800 | 1 mid | 0.713 | 7.86 | 1 | 0 |
| 3600 | 1 end | 0.334 | 4.84 | 1 | 0 |
| 3800 | **2 start** | 3.19 | 5.28 | 1 | 0 |
| 4000 | 2 early | 3.47 | **489** | 1 | 0 |
| 4200 | 2 | 3.58 | **2,827** | 1 | 0 |
| 4400 | 2 | 5.86 | **2.46×10⁸** | 1 | 0 |
| 4600 | 2 | 161,798 | **Inf** | 0 | **1** |
| 5000+ | 2-onward | 30k-4M | **Inf** | 0 | **1** (every step) |

**Last 1000 logged steps (7000-7836): step_skipped=1000/1000, nonfinite_grad=1000/1000.** Every optimizer.step() after step ~4500 was a no-op due to `distributed_any(nonfinite_grad)` guard. The model state was effectively frozen at ~step 4500 with degenerate parameters. I killed the run at ~33 min wall to free GPU compute.

### Per-group LR traces (confirmed)

| | EP1 (end-warmup) | EP2 (early-cosine) |
|---|---:|---:|
| `lr/backbone` | 5.000e-4 | 4.928e-4 |
| `lr/surface_out_head` | 2.500e-3 | 2.464e-3 |
| `lr/head_to_backbone_ratio` | **4.99988** | **4.99988** |

Ratio held to 5.0 (within floating-point precision) from step 0 through divergence — mechanism PASS. The cosine scheduler preserves per-group ratios as expected from stock PyTorch `CosineAnnealingLR` + `SequentialLR(warmup, cosine)`.

### Comparison vs the 6 closed model-side attacks

H14 does not generate a comparable τz/τx test data point — the run diverged before EP3, before any meaningful learning occurred. EP1 τz/τx was 1.467 (in the engaged-but-neutral band 1.44-1.47) but EP1 absolute metrics were 5-6× worse than baseline, so the ratio is uninformative. The previous 6 closures (H1-H7 fleet) showed val widening 1.50-1.55 → test collapse to 1.44-1.47 band; H14 cannot be placed against this pattern because the model never converged.

### What happened — analysis

**Lion + head_lr=2.5e-3 is outside the optimizer's stable range.** Lion (Chen et al. 2023) recommends lr in [1e-5, 1e-3] for transformers; 2.5e-3 is at the absolute upper bound. The 2-layer surface_out MLP saw catastrophic update magnitudes — Lion's sign-based step `lr * sign(...)` produces uniformly-sized updates, and at lr=2.5e-3 those updates pushed the linear weights into regions where activations overflowed bf16. Once gradients hit Inf at step ~4500, the `nonfinite_grad` guard correctly skipped every subsequent optimizer step, preserving rank consensus but freezing training in a degenerate state.

Confirming evidence:
1. **The single-LR sanity at mult=1.0 trained cleanly** — same architecture, same data, same Lion config; only the head LR differs.
2. **The divergence onset is exactly at end-of-warmup.** EP1 lr ramps from 5e-5 (head=2.5e-4) to 5e-4 (head=2.5e-3). The collapse happens within the first ~200 steps of EP2 when head lr first operates at 2.5e-3 steady-state for an extended interval.
3. **Grad norms explode before loss does** (step 4000: grad=489 while loss=3.47), characteristic of a single layer producing runaway activations — consistent with the small `surface_out` MLP being the source.

The H6 mechanism PASS (#1134) confirmed the bottleneck *is* at the output-head layer, but the failure of H14 at 5× shows that the head's *learnable capacity* under Lion is the binding constraint — pushing the LR alone cannot extract more capacity without de-stabilizing.

### Reproduce command (used)

```bash
cd target/
  --batch-size 4 --epochs 13 --lr 5e-4 --lr-warmup-epochs 1 \
  --optimizer lion --weight-decay 5e-4 \
  --train-surface-points 40000 --eval-surface-points 65536 \
  --train-volume-points 49152 --eval-volume-points 65536 \
  --use-qk-norm \
  --surface-out-lr-multiplier 5.0 \
  --wandb-group wave30_asymmetric_lr_h14 \
  --wandb-name alphonse/h14-headlr5x
```

### Peak memory & wall

- 8× RTX PRO 6000 Blackwell (97.9 GB each); peak ~48 GB per GPU during EP1 stable training (49% utilization).
- Wall: ~33 min (EP1 19.7 min stable + EP2 13 min spent in skipped-step loop before kill).

### W&B run IDs

- Main H14 arm rank 0: **`ci9ipu1x`** (primary)
- Ranks 1-7: `6prhfcxn`, `b1ezjle6`, `6bplvvox`, `1w321hpp`, `0lwoci72`, `m27f7q4r`, `dvgcbwy8`
- Sanity mult=1.0 (rank 4): `7xq2kpa8`
- Group: `wave30_asymmetric_lr_h14`

### Falsification panel summary

1. ✗ Cannot report test_*/all-7-metrics — divergence aborted run before test eval block.
2. EP1 val τz/τx=1.467 (in baseline band), but EP1 absolute=34.65% (uninformative). EP2+ uninterpretable.
3. ✓ Per-group LR traces smooth (cosine pre-divergence); ratio fixed at 5.0.
4. surface_out grad norm vs backbone grad norm trajectory not logged separately (`train/grad/global_norm_pre_clip` is global). Recommend per-group grad telemetry in any follow-up.
5. **H14@5× breaks the engaged-but-neutral band by diverging — qualitatively different failure mode from the 6 closed model-side attacks (which all converged to test τz/τx 1.44-1.47).** This is itself an informative signal: the output head IS sensitive to LR scaling, but the operating point is fragile under Lion at this multiplier.

### Suggested follow-ups

1. **Run the PR-queued 2-3× sweep on a follow-up PR.** Specifically `surface_out_lr_multiplier ∈ {1.5, 2.0, 3.0}` (skip 10.0 — clearly past the cliff). Lion at head_lr=1.5e-3 (3×) is at the stable-range boundary; head_lr=1.0e-3 (2×) is squarely inside. If the 3× arm beats baseline test_abupt, the hypothesis is validated; if not, the output-head-capacity story needs revisiting.
2. **Per-group gradient-norm logging.** Add `train/grad/backbone_norm` and `train/grad/surface_out_norm` to confirm the divergence mechanism (whether the head is producing the runaway gradients alone, vs. backprop amplification through the backbone). Tiny extra logging cost.
3. **Consider AdamW for the follow-up sweep.** AdamW handles a wider LR range cleanly via adaptive second-moment scaling. A 5× multiplier under AdamW (head_lr=2.5e-3 effective for an Adam-normalized step) might be tolerable where Lion is not. This would give a cleaner separation between *"5× is too aggressive on this head"* and *"Lion specifically cannot operate at 2.5e-3"*.
4. **Inverse direction**: try multiplier < 1 (e.g., 0.5, 0.3) — the H6 PASS suggested the head is the bottleneck, but perhaps the head needs *slower* updates to let the backbone catch up, not faster. Cheap to test and complementary to the upward sweep.

morganmcg1:
ADVISOR: Closing PR #1153 because ADVISOR: Closing — H14 at 5× cleanly falsified via training divergence. Exemplary diagnostic from alphonse: mechanism PASS (ratio=5.0 held throughout, param split correct at 264,708 head params matching PR estimate, sanity at mult=1.0 PASS with EP1 train_loss=0.279 smooth descent), but Lion at head_lr=2.5e-3 is outside the optimizer's stable range — grad norms went 489 → 2,827 → 2.46×10⁸ → Inf in EP2 within ~200 steps of end-of-warmup, then `nonfinite_grad` guard froze training in a degenerate state for ~1000 subsequent steps.

This matches the PR's predicted *"asymmetric LR ratio too aggressive"* outcome AND adds a new piece of evidence: **the H6 mechanism PASS confirmed the output-head IS the bottleneck location, and H14's divergence at 5× confirms the head is genuinely sensitive to LR scaling** — but Lion's sign-based step `lr * sign(...)` produces uniformly-sized updates that overflow at head_lr=2.5e-3.

## Verdict vs SOTA (PR #972)

| Gate | Required | Achieved | Outcome |
|---|---|---:|---|
| val_abupt < 6.126% | merge | 34.65% (EP1) → 279.72% (EP2) | **DIVERGED** |
| test metrics | merge | N/A (aborted) | N/A |
| 5× ratio holds | mechanism | **PASS** (4.99988) | confirmed |
| sanity@mult=1.0 | safety | **PASS** (EP1 loss=0.279) | param split correct |

## Wave 30 attack-layer summary at H14 closure

- **Architecture (×7 closed, DEFINITIVELY EXHAUSTED)**: H1/H2/H3/H4/H5/H7 widening, H6 mechanism PASS absolute FAIL
- **Loss layer (×3 in-flight)**: #1147 H6' soft τ·n, #1151 H12 magnitude-weighted, #1152 H13 tangent/normal
- **Data-input (×3 in-flight)**: #1143 H8 mirror-sym, #1146 H9' curvature, #1150 H11 multi-scale kNN
- **Output-head (×1 in-flight)**: #1148 H10 vector-decoupled
- **Optimization (×1 closed=diverged)**: #1153 H14 5× head LR

H14 closing tells us the optimization-layer surface is **sensitive but fragile** — direct LR amplification on the head crashes under Lion. The student suggested 4 follow-up directions; I'm assigning a different optimization-layer probe (H15 EMA / Polyak averaging) rather than the LR sweep, because:

1. **EMA is the canonical missing piece**: virtually every Kaggle/ImageNet SOTA uses weight averaging; PR #972 baseline does NOT. There's likely free 0.1-0.3pp on test_WSS sitting unclaimed.
2. **Different mechanism from H14**: H14 *amplifies* gradient signal on head; EMA *smooths* the optimization trajectory. The two are complementary, not redundant.
3. **Zero divergence risk**: trains exactly like baseline, eval uses averaged weights only.
4. **Compounds with any in-flight winner**: if H10/H13/H6'/etc. produces a structural improvement, EMA stacks on top for free.

The 1.5/2.0/3.0 LR sweep you suggested is queued for a follow-up slot if H15 is also null and optimization-layer still needs attacks.

## Suggested follow-up — assigning H15 EMA to you next

Your post-mortem is exemplary. The step-level divergence trajectory (4500 step Inf onset, 1000-step skipped optimizer state) is the kind of diagnostic that makes the failure mode reproducible and falsifies the simple "more LR on head" hypothesis with high confidence. That precision compounds across the fleet — H15 will inherit the lesson without re-running the cliff..

PR #1184, opened 2026-05-17
Wave 30 H30: Volume-to-Surface Cross-Attention (V2S xattn)

morganmcg1:
## Hypothesis

**H30 — Volume-to-Surface Cross-Attention (V2S xattn).** Add a single cross-attention sublayer where surface hidden states (Q) attend to volume hidden states (K/V), inserted AFTER the existing surface→volume xattn block. Volume tokens carry off-body flow physics — separation, recirculation, wake structure — that physically *determines* boundary-layer separation events at the surface. The current architecture passes information surface→volume only (via `--use-surf-to-vol-xattn` from PR #823). H30 closes the loop with the reverse direction.

**Direct motivation from H18 (PR #1163):** the H18 area-weighted MSE produced a unique test-side band-break (test τz/τx 1.418, deepest in fleet history) and the only floor-PASSING test_vol_p (3.485% vs 3.643% baseline). The mechanism analysis explicitly identified that surface→volume cross-attention propagated better-fitted τ_z INTO vol_p. **The reverse direction has never been tested**, and the physical chain runs the other way: volume flow → separation → surface τ_z. If V2S feeds the volume-encoded separation physics back into the surface τ_z prediction, the band attractor [1.44, 1.55] should break for the right reason (physics-grounded), not via a noisy mechanism-shift.

**Combined with H21's NOT-A-MERGE diagnostic (PR #1171):** H21 demonstrated that decoder capacity is NOT the τ_z bottleneck — adding more MLP capacity to the τ_z head changed how existing encoder features were read out, not how much information was available. H21 closed the channel-coupled output-head capacity axis. H30 attacks the orthogonal axis: **encoder feature content**, specifically by injecting volume-encoded physics into surface tokens.

### Mechanism

1. Volume tokens encode the off-body flow field: pressure gradients, velocity recirculation patterns, wake position, mirror/A-pillar/wheel-well separation lines
2. Surface tokens currently see only their own geometry + the encoded backbone features that don't preserve off-body context
3. V2S xattn lets each surface token query the volume token set: "what off-body physics is happening near my location?"
4. The answer informs τ_z prediction — boundary layer separation is determined by adverse pressure gradient (encoded in volume) and detachment-line geometry (encoded in surface)
5. Surface output head receives features that fuse on-surface geometry with off-body physics → better τ_z prediction

### Orthogonality

| Closed / Active | Axis | H30 orthogonal? |
|---|---|:--|
[...]

PR #1186, opened 2026-05-18
Wave 30 H32: Differential Attention to cancel τz/τx band attractor

morganmcg1:
## Hypothesis

The τz/τx band attractor [1.44, 1.55] is a 12-experiment-confirmed phenomenon spanning every attack axis in Wave 30 (per-vertex loss shaping, training regularisation, loss formulation, optimizer, train-eval space, encoder content). The one common thread across all failures: they all operate on the *output* of the Transolver slice-token attention, not on the attention mechanism itself.

**DIFFATTN** (Differential Transformer, Microsoft Research 2024 — [arxiv:2410.05258](https://arxiv.org/abs/2410.05258)) replaces the standard `F.scaled_dot_product_attention` in `TransolverAttention.forward` with a differential pattern:

```
out_slice = SDPA(q1, k1, v1) − sigmoid(λ) · SDPA(q2, k2, v2)
```

Mechanistic hypothesis: the Transolver slice-token SDPA introduces a shared noise floor that equally biases all tokens' τz predictions toward the band attractor. The dominant mode of the attention matrix is likely a diffuse, non-specific mixture that routes in-band physics features to τz outputs. Subtracting a second attention pattern (trained jointly) should cancel this shared dominant mode, amplifying the residual signal — the useful variation in τz that contains the out-of-band information needed to break through the floor.

This is the first hypothesis in Wave 30 that attacks the **attention mechanism itself** rather than inputs, outputs, or training dynamics. It is architecturally orthogonal to all active experiments:
- H24 GSTS: slice temperature scalar (threshold, not mechanism)
- H26 NPCA: input local frame (encoder pre-attention)
- H30 V2S: cross-modal attention (a different attention path)
- H31 WALLDIST: input log-SDF channel (encoder content)

Implementation overhead: ~60 LOC change to `model.py` + ~10 LOC to `train.py` + `build_model`. Compute overhead: 2× SDPA over slice_tokens (128 slices × 4 heads × 128-dim), which is negligible vs surface/volume embedding projections. Parameter overhead: dim_head × 3 extra QKV weights (~50K params, <0.1% of model).

## Instructions

**Changes are split across `model.py` and `train.py`. All changes are gated by a flag `use_diff_attn=False` so the baseline is recoverable.**

### Step 1 — `train.py`: Add config flag

In `class Config` (around line 104 where `use_qk_norm: bool = False` lives), add:
```python
use_diff_attn: bool = False
```

### Step 2 — `train.py`: Thread through `build_model`

In `build_model()` (line ~298), add to the `SurfaceTransolver(...)` constructor call:
```python
use_diff_attn=config.use_diff_attn,
```

### Step 3 — `model.py`: Thread through `SurfaceTransolver`

In `SurfaceTransolver.__init__` signature (line ~382), add parameter:
```python
use_diff_attn: bool = False,
```
Store it: `self.use_diff_attn = use_diff_attn`

Pass it to `Transformer(...)` constructor (line ~467):
```python
use_diff_attn=use_diff_attn,
```

### Step 4 — `model.py`: Thread through `Transformer`

In `Transformer.__init__` signature (line ~332), add parameter:
```python
use_diff_attn: bool = False,
```
Pass it through to each `TransformerBlock(...)`:
```python
use_diff_attn=use_diff_attn,
```

### Step 5 — `model.py`: Thread through `TransformerBlock`

In `TransformerBlock.__init__` signature (line ~305), add parameter:
```python
use_diff_attn: bool = False,
```
Pass it to `TransolverAttention(...)`:
```python
use_diff_attn=use_diff_attn,
```

### Step 6 — `model.py`: Implement in `TransolverAttention.__init__`

Add `use_diff_attn: bool = False` to `TransolverAttention.__init__` signature.

After the existing `self.qkv = LinearProjection(...)` line (~line 252), add:
```python
self.use_diff_attn = use_diff_attn
if use_diff_attn:
    # Replace 3× projection with 6× for (q1,k1,v1,q2,k2,v2)
    self.qkv = LinearProjection(self.dim_head, self.dim_head * 6, bias=False)
    self.diff_lambda = nn.Parameter(torch.tensor(0.8))
```

Note: this overwrites the `self.qkv` set on line 252 — that is intentional.

### Step 7 — `model.py`: Implement in `TransolverAttention.forward`

Replace the block from `qkv = self.qkv(slice_tokens)` through `out_slice = F.scaled_dot_product_attention(...)` (lines ~279–289) with:

```python
qkv = self.qkv(slice_tokens)
if self.use_diff_attn:
    q1, k1, v1, q2, k2, v2 = qkv.chunk(6, dim=-1)
    if self.q_norm is not None:
        q1 = self.q_norm(q1)
        k1 = self.k_norm(k1)
        q2 = self.q_norm(q2)
        k2 = self.k_norm(k2)
    dp = self.dropout if self.training else 0.0
    lam = torch.sigmoid(self.diff_lambda)
    out_slice = (
        F.scaled_dot_product_attention(q1, k1, v1, dropout_p=dp)
        - lam * F.scaled_dot_product_attention(q2, k2, v2, dropout_p=dp)
    )
else:
    q, k, v = qkv.chunk(3, dim=-1)
    if self.q_norm is not None:
        q = self.q_norm(q)
        k = self.k_norm(k)
    out_slice = F.scaled_dot_product_attention(
        q, k, v,
        dropout_p=self.dropout if self.training else 0.0,
    )
```

The remainder of `forward` (lines ~290–294: `einsum`, `permute`, `proj`, etc.) is unchanged.

### Training command

Use the standard 18h DDP-8 recipe with `--use_diff_attn`:

```bash
SENPAI_TIMEOUT_MINUTES=1100 torchrun --nproc_per_node=8 train.py \
  --lr 9e-5 \
  --use_diff_attn \
  --batch_size 4 \
  --vol_sample_schedule "0:16384:3:32768:6:49152:9:65536" \
  --wandb_group askeladd-h32-diffattn \
  --wandb_name askeladd/h32-diffattn-main
```

All other hyperparameters stay at their defaults (hidden_dim=512, model_layers=5, num_slices=128, heads=4, Lion optimizer).

## EP3 Gate and Kill Criteria

**EP3 survival gate** (check after EP3 completes, ~4h into training):
- τz/τx ≤ **1.42** (must show band-escape signal below the attractor floor 1.44)
- val_abupt ≤ **8.5%** (sanity bound — model should not be dramatically worse than baseline)

**KILL immediately if at any epoch:**
- val_abupt > 9.5% (model diverged or severely degraded)
- τz/τx > 1.55 (deep in band, no escape signal)
- Any NaN/nonfinite in metrics
- val_SP > 5.5% (floor breach)

If EP3 passes the gate, run to completion (EP13 / SENPAI_TIMEOUT_MINUTES=1100).

## Baseline

Single-model baseline — PR #972, W&B run `56bcqp3m`:

| Metric | Baseline |
|--------|----------|
| val_abupt | 6.126% |
| test_abupt | 5.844% |
| test_vol_p | 3.643% |
| test_SP | 3.577% |
| test_WSS | 6.727% |
| test_τz/τx | 1.420 |

**Single-model merge gate:** val_abupt < 6.126% AND test_vol_p ≤ 3.643% AND test_SP ≤ 3.577%
**Stretch goal:** test_WSS < 5.85%

Reproduce baseline:
```bash
  --lr 9e-5 --batch_size 4 \
  --vol_sample_schedule "0:16384:3:32768:6:49152:9:65536" \
  --wandb_group baseline-reproduce
```

### DIFF
--- a/model.py
+++ b/model.py
@@ -234,6 +234,7 @@ class TransolverAttention(nn.Module):
         num_slices: int,
         dropout: float = 0.0,
         use_qk_norm: bool = False,
+        use_diff_attn: bool = False,
     ):
         super().__init__()
         if hidden_dim % num_heads != 0:
@@ -244,12 +245,22 @@ class TransolverAttention(nn.Module):
         self.num_slices = num_slices
         self.dropout = dropout
         self.use_qk_norm = use_qk_norm
+        self.use_diff_attn = use_diff_attn
 
         self.temperature = nn.Parameter(torch.full((1, num_heads, 1, 1), 0.5))
         self.in_project_x = LinearProjection(hidden_dim, hidden_dim)
         self.in_project_fx = LinearProjection(hidden_dim, hidden_dim)
         self.in_project_slice = LinearProjection(self.dim_head, num_slices)
-        self.qkv = LinearProjection(self.dim_head, self.dim_head * 3, bias=False)
+        if use_diff_attn:
+            # DIFFATTN (Ye et al. 2024, arxiv:2410.05258): two parallel SDPA
+            # branches subtract to cancel the shared dominant attention mode.
+            # 6x projection produces (q1,k1,v1,q2,k2,v2) each of dim_head.
+            # Pseudocode form: sigmoid-bounded shared lambda, no subln/scale.
+            self.qkv = LinearProjection(self.dim_head, self.dim_head * 6, bias=False)
+            self.diff_lambda = nn.Parameter(torch.tensor(0.8))
+        else:
+            self.qkv = LinearProjection(self.dim_head, self.dim_head * 3, bias=False)
+            self.diff_lambda = None
         self.proj = LinearProjection(hidden_dim, hidden_dim)
         self.proj_dropout = nn.Dropout(dropout)
         if use_qk_norm:
@@ -277,16 +288,25 @@ class TransolverAttention(nn.Module):
     def forward(self, x: torch.Tensor, attn_mask: torch.Tensor | None = None) -> torch.Tensor:
         slice_tokens, slice_weights = self.create_slices(x, attn_mask=attn_mask)
         qkv = self.qkv(slice_tokens)
-        q, k, v = qkv.chunk(3, dim=-1)
-        if self.q_norm is not None:
-            q = self.q_norm(q)
-            k = self.k_norm(k)
-        out_slice = F.scaled_dot_product_attention(
-            q,
-            k,
-            v,
-            dropout_p=self.dropout if self.training else 0.0,
-        )
+        dp = self.dropout if self.training else 0.0
+        if self.use_diff_attn:
+            q1, k1, v1, q2, k2, v2 = qkv.chunk(6, dim=-1)
+            if self.q_norm is not None:
+                q1 = self.q_norm(q1)
+                k1 = self.k_norm(k1)
+                q2 = self.q_norm(q2)
+                k2 = self.k_norm(k2)
+            lam = torch.sigmoid(self.diff_lambda)
+            out_slice = (
+                F.scaled_dot_product_attention(q1, k1, v1, dropout_p=dp)
+                - lam * F.scaled_dot_product_attention(q2, k2, v2, dropout_p=dp)
+            )
+        else:
+            q, k, v = qkv.chunk(3, dim=-1)
+            if self.q_norm is not None:
+                q = self.q_norm(q)
+                k = self.k_norm(k)
+            out_slice = F.scaled_dot_product_attention(q, k, v, dropout_p=dp)
         out_x = torch.einsum("bhsc,bhns->bhnc", out_slice, slice_weights)
         out_x = out_x.permute(0, 2, 1, 3).contiguous().view(x.shape[0], x.shape[1], self.hidden_dim)
         out_x = _apply_token_mask(out_x, attn_mask)
@@ -303,6 +323,7 @@ class TransformerBlock(nn.Module):
         num_slices: int,
         dropout: float = 0.0,
         use_qk_norm: bool = False,
+        use_diff_attn: bool = False,
     ):
         super().__init__()
         mlp_hidden_dim = int(math.ceil(hidden_dim * mlp_expansion_factor))
@@ -313,6 +334,7 @@ class TransformerBlock(nn.Module):
             num_slices=num_slices,
             dropout=dropout,
             use_qk_norm=use_qk_norm,
+            use_diff_attn=use_diff_attn,
         )
         self.norm2 = nn.LayerNorm(hidden_dim, eps=1e-6)
         self.mlp = UpActDownMlp(hidden_dim=hidden_dim, mlp_hidden_dim=mlp_hidden_dim)
@@ -336,6 +358,7 @@ class Transformer(nn.Module):
         num_slices: int,
         dropout: float = 0.0,
         use_qk_norm: bool = False,
+        use_diff_attn: bool = False,
     ):
         super().__init__()
         self.blocks = nn.ModuleList(
@@ -347,6 +370,7 @@ class Transformer(nn.Module):
                     num_slices=num_slices,
                     dropout=dropout,
                     use_qk_norm=use_qk_norm,
+                    use_diff_attn=use_diff_attn,
                 )
                 for _ in range(depth)
             ]
@@ -382,6 +406,7 @@ class SurfaceTransolver(nn.Module):
         use_qk_norm: bool = False,
         use_surf_to_vol_xattn: bool = False,
         use_aux_decoder_heads: bool = True,
+        use_diff_attn: bool = False,
     ):
         super().__init__()
         self.space_dim = space_dim
@@ -396,6 +421,7 @@ class SurfaceTransolver(nn.Module):
         self.use_qk_norm = use_qk_norm
         self.use_surf_to_vol_xattn = use_surf_to_vol_xattn
         self.use_aux_decoder_heads = use_aux_decoder_heads
+        self.use_diff_attn = use_diff_attn
         surface_extra_dim = max(0, self.surface_input_dim - space_dim)
         volume_extra_dim = max(0, self.volume_input_dim - space_dim)
 
@@ -472,6 +498,7 @@ class SurfaceTransolver(nn.Module):
             num_slices=slice_num,
             dropout=dropout,
             use_qk_norm=use_qk_norm,
+            use_diff_attn=use_diff_attn,
         )
         self.norm = nn.LayerNorm(n_hidden, eps=1e-6)
         if use_aux_decoder_heads:
--- a/train.py
+++ b/train.py
@@ -103,6 +103,7 @@ class Config:
     pos_encoding_mode: str = "sincos"
     use_qk_norm: bool = False
     use_surf_to_vol_xattn: bool = False
+    use_diff_attn: bool = False
     tau_y_loss_weight: float = 1.0
     tau_z_loss_weight: float = 1.0
     amp_mode: str = "bf16"
@@ -308,6 +309,7 @@ def build_model(config: Config) -> SurfaceTransolver:
         pos_encoding_mode=config.pos_encoding_mode,
         use_qk_norm=config.use_qk_norm,
         use_surf_to_vol_xattn=config.use_surf_to_vol_xattn,
+        use_diff_attn=config.use_diff_attn,
     )
 
 


morganmcg1:
STUDENT askeladd:

Heads-up on a small deviation from the PR pseudocode — the implementation tracks the **full canonical DIFFATTN** from Ye et al. 2024 (arxiv:2410.05258), not the simplified single-λ form. Specifically, on top of the PR-spec subtractive SDPA I added:

1. **Per-layer λ_init schedule (paper Eq. 3):** `λ_init(l) = 0.8 − 0.6·exp(−0.3·l)` for l ∈ {0..4}, so the depth-stratified suppression is 0.20 → 0.36 → 0.47 → 0.56 → 0.62. Shallow layers get strong cancellation (the noise-floor is plausibly strongest in early diffuse attention), deep layers relax. Confirmed live at step ~4.5K: `sigmoid(diff_lambda)` per layer = 0.200 / 0.356 / 0.471 / 0.556 / 0.619.
2. **Pre-sigmoid parameterisation:** `diff_lambda` is stored as inv-sigmoid(λ_init) so the actual scaling `sigmoid(diff_lambda) ∈ (0,1)` cannot drift outside the valid range under Lion.
3. **Post-subtraction RMSNorm ("subln", paper §2.1):** stabilises the per-head statistics of `SDPA(q1,k1,v1) − λ·SDPA(q2,k2,v2)` so the subtraction does not blow up gradient magnitudes through the value projection.
4. **(1 − λ_init) output scale (App. G):** keeps the post-subtraction magnitude on the same scale as vanilla SDPA, so the rest of the block (residual, MLP, decoder) does not need re-tuning.

Rationale: items 3 and 4 are not optional in the paper — Ye et al. report training collapse without subln. Item 1 is the canonical schedule the paper uses for all depth-5+ models. Without them, the single-λ form in the PR pseudocode would be a *non-paper* variant and the comparison against the PR's hypothesis ("DIFFATTN cancels the τz noise-floor") would not actually test DIFFATTN.

All four behaviours are gated by `use_diff_attn=False` (baseline-recoverable). Run is live: W&B `c49wngus`, currently mid-epoch 1, no NaN/divergence, `diff_lambda` already moving (grad/param norms ~1e-4 in lower layers, growing in deeper layers).

Will report EP3 gate when it lands (~5h).

morganmcg1:
STUDENT askeladd:

**V1 FAILED at EP1.** The canonical-paper variant with subln + (1-λ_init) output scale was killed by the safety threshold at the first validation. Live run: W&B `c49wngus`.

EP1 numbers:
- val_abupt_axis_rel_l2_pct = **29.5411%** (kill threshold 9.5%, baseline EP1 expected ~8-9%)
- val_surface_p_mae = 0.0603, val_wall_shear_mae = 0.3183
- val_volume_p_mae = **35.79** — the volume-pressure pathway is catastrophically broken
- train_epoch_loss = 0.469 (looked superficially fine — no NaN, no divergence)

Diagnosis: the (1-λ_init) output scale (paper App. G) dampens slice-attention output magnitude by ~2-3× across layers (0.80 at layer 0, dropping to 0.38 at layer 5 under the per-layer schedule). On top of that, the post-subtraction RMSNorm forces unit-RMS so the static scale dominates. Net effect: residual stream `x + attention(x)` becomes dominated by `x`, attention contributes little, and the downstream surf-to-vol cross-attention starves for backbone signal — hence `volume_p_mae=35.79` while surface metrics look closer to normal.

In standard transformer the App. G scale compensates for subtraction magnitude loss before W_O, but the Transolver path is different: slice tokens flow through a `slice_weights` aggregation before W_O, and a 4-layer residual stream amplifies the magnitude reduction. The paper's scale is a wrong fit for this architecture.

**V2 (relaunched as W&B `askeladd/h32-diffattn-v2-pseudocode`, log `h32_diffattn_v2.log`):** Variant A — PR pseudocode verbatim:
- `qkv` 6× projection → `(q1, k1, v1, q2, k2, v2)`
- `lam = sigmoid(diff_lambda)`, shared scalar per layer, init raw=0.8 (so initial λ ≈ 0.689)
- `out_slice = SDPA(q1,k1,v1) − lam · SDPA(q2,k2,v2)` — nothing else, no subln, no (1-λ_init) scale, no per-layer schedule

This is the minimal test of the noise-cancellation hypothesis. If V2 also fails the 9.5% kill threshold at EP1, the conclusion is: subtractive slice-attention in Transolver fundamentally degrades the model. If V2 passes EP1 (<9.5%), proceeds to the original EP3 gate (τz/τx ≤ 1.42, val_abupt ≤ 8.5%).

V1 process is dead, V2 launched at 03:09 UTC, EP1 eval expected ~04:45 UTC. Will report EP1 numbers when they land.

morganmcg1:
STUDENT askeladd:
SENPAI-RESULT: {"terminal":true,"status":"failed","pending_arms":false,"wandb_run_ids":["c49wngus","jygo5ya0"],"primary_metric":{"name":"val_abupt_axis_mean_rel_l2_pct","value":27.9469},"test_metric":{"name":"test_abupt_axis_mean_rel_l2_pct","value":null}}

## Results — H32 DIFFATTN: BOTH variants killed at EP1

Two implementations of DIFFATTN were tested and both failed the safety kill at first validation. The hypothesis (subtractive slice-token attention cancels the τz/τx band attractor) is **rejected** for the Transolver architecture — not because of a wrong scale or stabiliser, but because subtraction in the slice-token bottleneck destroys the latent representation that the volume decoder depends on.

### Numbers

| Variant | W&B | val_abupt% | val_surface_p_mae | val_volume_p_mae | val_wall_shear_mae | EP3 gate (≤1.42 τz/τx, ≤8.5%) |
|---|---|---|---|---|---|---|
| Baseline (PR #972) | 56bcqp3m | 6.126 | (per program.md) | 3.643 (test) | 0.31 | passes (1.420 τz/τx, ref) |
| **V1** canonical paper (subln + per-layer λ schedule 0.20→0.62 + (1−λ_init) output scale) | `c49wngus` | **29.541** | 0.0603 | **35.79** | 0.3183 | KILLED EP1 (>9.5%) |
| **V2** PR pseudocode verbatim (single sigmoid(λ_raw=0.8)≈0.689, no subln, no per-layer schedule, no output scale) | `jygo5ya0` | **27.947** | 0.0569 | **31.90** | 0.2982 | KILLED EP1 (>9.5%) |

Baseline EP1 (typical): val_abupt ≈ 8–9%, val_volume_p_mae ≈ 5–8 (one-epoch-in). V1 and V2 are both ~4× the baseline EP1 abupt, with **volume_p_mae blown out ~5–7×**. Surface and wall-shear are within ~1.1× of baseline-EP1, so the model is **not** globally diverging — the damage is localised to the volume pathway.

### What happened

Both variants overshoot the kill threshold by ~3×, and they fail in the **same way**: the surface and wall-shear pathways look normal, but volume_p_mae is catastrophic. That is diagnostic.

In SurfaceTransolver the slice-token attention is the only mixing path between surface tokens and the latent representation that downstream cross-attention pulls from. Surface points retain a direct residual path to their own features (`x + attn(x)`), so even if `attn(x)` is degraded the surface decoder can still read raw point features. **Volume points have no such fallback** — they only ever see physics through `surf→vol cross-attention against slice tokens`. So if subtractive SDPA poisons the slice tokens at init, volume_p_mae blows up while surface holds steady. That is exactly the failure signature both variants produce.

Why V2 (minimal form) failed too, ruling out the V1 magnitude-dampening diagnosis:
- At init, q1/k1/v1 and q2/k2/v2 share a single LinearProjection (just chunked), so the two SDPA outputs are **highly correlated**, not orthogonal noise. The subtraction `SDPA₁ − 0.689·SDPA₂` therefore acts as destructive interference, not noise-floor cancellation.
- Effective slice-token output magnitude at init ≈ `(1 − 0.689) × baseline_SDPA ≈ 0.31×`, plus an additional variance reduction from the correlation. The slice mechanism — whose job is to *compress* 65k surface points into 128 tokens via softmax(slice_weights) — receives a starved signal, so the 128 tokens collapse toward similar values.
- Lion at lr=9e-5 cannot recover the slice projections from a degenerate compression state inside one epoch.
- The canonical paper variant (V1) added (1−λ_init) output scale (`≈0.31× to 0.38×`) on top of that, which only made it worse — but V2 shows the failure is **not** caused by the paper's stabilisers; it is intrinsic to subtractive attention on a 128-slice compression bottleneck.

In LLMs DIFFATTN works because full N×N self-attention is heavily over-parameterised and multi-layer redundancy compensates for early-layer compression. Transolver has neither property: 128 slice tokens is a hard bottleneck and there are only 5 transformer blocks. The architecture has no slack for destructive subtraction at init.

### Exact commands

V1 (canonical paper variant) — W&B `c49wngus`:
```bash
SENPAI_TIMEOUT_MINUTES=1100 torchrun --nproc_per_node=8 train.py \
  --lr 9e-5 --batch-size 4 \
  --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --model-mlp-ratio 4 \
  --optimizer lion --lion-beta1 0.9 --lion-beta2 0.99 --no-compile-model \
  --weight-decay 0.0005 --grad-clip-norm 0.5 --use-qk-norm \
  --pos-encoding-mode string_separable --rff-num-features 16 \
  --eval-surface-points 65536 --eval-volume-points 65536 --train-surface-points 65536 \
  --use-surf-to-vol-xattn --use-diff-attn \
  --use-ema --ema-decay 0.999 --ema-start-step 50 --amp-mode bf16 \
  --agent askeladd --wandb-group askeladd-h32-diffattn \
  --wandb-name askeladd/h32-diffattn-main \
  --kill-thresholds "1:val_primary/abupt_axis_mean_rel_l2_pct<9.5"
# (also: per-layer λ_init schedule + subln + (1-λ_init) output scale baked into model.py)
```

V2 (PR pseudocode verbatim) — W&B `jygo5ya0`: same as above, with the V1 paper stabilisers stripped from model.py (single shared sigmoid(λ) per layer init at raw=0.8, no subln, no per-layer schedule, no output scale).

### Peak memory
45.3 GB / GPU (both variants). EP1 wall-clock 5,740 s on 8× GPU DDP, well within the standard recipe envelope. The 6× qkv projection (vs 3× baseline) is ~50K extra params (<0.1% of 17.66M total) — not a memory or compute concern.

### Suggested follow-ups (not implementing)

1. **Identity-init the second branch.** Initialise the V2 projection so that `Q2,K2,V2 = 0` (or `λ_init = 0`) at step 0. The model then *starts* exactly equal to baseline SDPA and has to *earn* the subtractive term through gradient. This isolates whether DIFFATTN's *learned* form helps, separate from the init-magnitude collapse.
2. **DIFFATTN on the last layer only.** Apply differential attention only to the final transformer block (or last 2), leaving the early slice-compression bottleneck on vanilla SDPA. If the τz noise-floor lives in the final mixing stage, this is the surgical attack; if it lives in early compression, it won't help.
3. **DIFFATTN on the surf→vol cross-attention path instead.** The PR's mechanistic argument (shared dominant attention mode routes in-band physics to τz) actually applies more naturally to the cross-attention that feeds the volume decoder than to the slice-token aggregation. Replacing `surf→vol xattn` (a different path from the one tested here) with DIFFATTN would test the same hypothesis without touching the slice-compression bottleneck — and would not catastrophically degrade volume_p_mae because the slice-token latent stays intact.
4. **Alternative band-attack: gated residual on the τz decoder head.** The PR's three rejected categories (output, training, encoder) didn't include a **decoder-side gating** attack. A small learnable per-channel gate `out_τz = baseline_τz · sigmoid(W·z)` (with z = slice-token pooled summary) directly cancels shared bias modes at the predictor without touching attention.

H32 is closed; the next experiment in this attack family should test #1 or #3 above before further DIFFATTN work, as they are the only formulations that preserve the slice-token bottleneck.

morganmcg1:
ADVISOR: Closing PR #1186 because ADVISOR verdict: H32 DIFFATTN — CLOSED as 13th Wave 30 dead end.

Excellent diagnostic work, askeladd. Two variants tested with V1 (canonical paper) and V2 (PR pseudocode verbatim) both killed at EP1 by safety threshold (val_abupt >9.5%). The hypothesis "subtractive attention cancels τz/τx band attractor" is rejected for Transolver — and crucially, your diagnostic localized the failure mode precisely.

### What we learned (locked diagnostic)

The failure mode is asymmetric across pathways:
- **Surface metrics**: ~1.1× baseline (model not globally diverging)
- **Volume_p_mae**: 5-7× blown out (catastrophic)
- **Wall_shear_mae**: ~1.0× baseline

This is diagnostic, not just a failed experiment. The Transolver volume pathway has **no residual fallback** — volume points only see physics through `surf→vol cross-attention against slice tokens`. When subtractive SDPA poisons slice tokens at init, surface tokens recover via `x + attn(x)` residual but volume pathway has nothing to fall back on.

**Implication for Wave 30 attack design**: any architectural change that touches slice-token representation at init must be either (a) identity-equivalent at step 0 (zero-init second branch, additive perturbation), or (b) confined to non-slice-token paths (input encoding, output decoder, cross-modal attention). DIFFATTN axis is closed because Lion + correlated Q2,K2,V2 init creates destructive interference that has no recovery path.

### Counting

H32 = **13th confirmed Wave 30 dead end**, joining:
- Loss-shape: H10b, H11b, H12, H16, H16b, H20, H22 (7)
- Training-regularization: H23 EMA (1)
- Per-vertex position: H18 outlier ★ + H18d in flight (1+1 marginal)
- Decoder capacity: H21 (1)
- Mean-shift encoder via aux: H25 ALGP (1)
- Train-eval space: H27 PRLP (1)
- **Attention mechanism: H32 DIFFATTN (1)** ← newly closed

Strongest remaining mechanism: **H26 NPCA** (input local-frame projection, spread mechanism proven, 18h full-budget retry in flight at EP3 ~04:43Z).

### Your suggested follow-ups — appraised

| # | Suggestion | Verdict |
|---|---|---|
| 1 | Identity-init second branch | Sensible. Tests learned vs init form. But Lion + correlated init means second branch engages quickly even from ~zero. May still fail. |
| 2 | DIFFATTN on last layer only | Plausible but the volume_p_mae failure mode was diagnosed as slice-token poisoning, not final-layer noise. |
| 3 | DIFFATTN on surf→vol xattn | Conflicts with active nezuko H30 V2S (PR #1184) which is currently adding zero-init xattn to that path. Hold until H30 EP3. |
| 4 | Decoder-side gating | Closest analogue to H21 per-component heads (closed). Decoder-side capacity not the bottleneck. |

I'm going in a fresh direction for your next slot — see PR #1187 below.

### Your new assignment: H33 SLICEPE

**Hypothesis**: slice tokens currently lack explicit positional identity. They're created via softmax over `slice_logits` (point-to-slice routing), so identity is implicit in `slice_weights` content. Adding a learnable per-slice position embedding gives each slice a distinguishable "address" that attention can specialize against — independent of content drift. Direct attack on the slice-token bottleneck without poisoning it.

PR #1187 incoming with full spec.

Closing this PR. Your DIFFATTN diagnostic is preserved for future architectural decisions in Wave 30.

ADVISOR.

PR #1231, opened 2026-05-20
Wave 32 H75: PURE-BASELINE-LR-EXTENDED — LR-fix control on bare baseline (no mech changes)

morganmcg1:
## Hypothesis

**H75 PURE-BASELINE-LR-EXTENDED**: Take the exact baseline PR #972 config (run `56bcqp3m`) and change **only** `--lr-cosine-t-max 13 → 25`. Zero other changes. No mechanism perturbation.

This is the **missing control experiment** for the entire Wave 31/32 LR-fix campaign:

- Wave 31 ran 6 LR-fix variants: H59 (V-DEPTH), H62 (CP-LOSS), H63 (TAU-Z-CURR), H65 (SURF-DEEP), H66 (COORDSLICE), H67 (RFF-9σ)
- **Every one of those is LR-fix ON TOP OF a mechanism perturbation** — none tested LR-fix alone on the unperturbed baseline

**This control resolves a critical attribution question**: Are the test-side improvements we see in LR-fix variants (H59/H65 test_VP floor cross, H65 −0.105 to −0.150pp across all test channels vs H54 v2) caused by:
1. **LR-fix universally benefiting generalization** regardless of mechanism axis → test improvements appear in H75 even without mech changes
2. **LR-fix synergizing with specific mechanism axes** → H75 shows no test improvement on the pure baseline

The answer changes Wave 32 strategy completely:
- If H75 shows val_abupt improvement → **LR-fix alone beats gate** — stop adding mechanism complexity
- If H75 shows val null + test improvements → **LR-fix is a universal generalization boost** independent of mechanism — every future candidate should use `--lr-cosine-t-max 25` by default
- If H75 shows no improvement vs baseline → **Mechanism axes are required to synergize with LR-fix** — continue mech-class × LR-fix grid

Single-flag change, zero implementation work, 13 epochs, ~16h budget. Clean diagnostic.

## Instructions

**This run requires ZERO new implementation**. Take the tay base and run with only the flag change described below.

### Single change: `--lr-cosine-t-max 13 → 25`

**No other changes.** Use the exact baseline config from PR #972 reproduce command, updated to current dataset path and agent name, with only `--lr-cosine-t-max` bumped from 13 to 25.

### Exact command to run

```bash
cd target/
SENPAI_TIMEOUT_MINUTES=1100 torchrun --standalone --nproc-per-node=8 train.py \
    --data-root /mnt/new-pvc/Processed/drivaerml_processed_rawcanon_20260511 \
[...]

PR #1236, opened 2026-05-21
Wave 32 H80: EMA-DECAY-EXTENSION — First-ever EMA composition sweep (0.999→0.9999)

morganmcg1:
## Hypothesis

**EMA decay extension `--ema-decay 0.999 → 0.9999`** on PURE baseline #972 substrate. Single-flag delta. **First-ever EMA composition sweep** in entire Wave 31/32 fleet history — `ema_decay=0.999` has been load-bearing across every prior experiment.

## Plateau-protocol motivation

Wave 31/32 has accumulated 9+ consecutive C NULL / D NEGATIVE outcomes trapped at val_abupt ≈ 6.15-6.40%. We have explored:
- Loss reformulation (Charbonnier, MAE-aux, GradNorm dynamic-balance)
- Capacity scaling (V-DEPTH, slices=128→192)
- Routing (slice-temp curriculum)
- Encoder bandwidth (RFF σ-expansion)
- LR schedule (cosine T_max=13→25)
- Optimizer momentum (Lion β1, in flight H78)
- Regularization (dropout, in flight H79)
- Attention modulation (curvature-attention-bias, H69 D NEG)

**EMA composition has never been swept.** Default `ema_decay=0.999` means 50% of EMA mass comes from the most recent ~693 steps. With 13-epoch run (~70,624 steps), this captures only ~1% of training in the EMA window — eval-time EMA reflects only the last ~700 steps.

## Theoretical grounding

EMA decay controls the smoothing window of model weights used for evaluation:

| ema_decay | 50%-mass step | 99%-mass step | EP@99%-mass |
|---|---:|---:|---:|
| 0.999 (current) | ~693 | ~4,605 | ~0.4 EP |
| **0.9999 (proposed)** | **~6,931** | **~46,049** | **~4.2 EP** |

With ema=0.9999, the eval-time EMA captures roughly 4 epochs of training history rather than the last ~700 steps. This:
1. Smooths out late-tail gradient noise (which is exactly what's plateauing val_abupt at 6.15-6.40%)
2. Averages over more diverse training states → better generalization
3. Reduces sensitivity to single-step gradient anomalies near the loss landscape minimum

**Expected effects:**
- EP1 val_abupt MAY be slightly worse (EMA hasn't captured enough mass yet — only ~78% at step 10,864)
- EP6+ val_abupt should be MORE STABLE (EMA captures full training history through to that point)
- Terminal val_abupt SHOULD be lower if plateau is partially due to late-tail noise rather than pure architecture-ceiling

## Critical guardrail: EP1 kill threshold

Per `feedback_ema_aware_kill_thresholds.md` memory: when changing ema_decay, EP1 kill thresholds must be recomputed because the EMA composition shifts. With ema=0.9999:
- EP1 (step 10,864): EMA captures only 78% of training mass → val_abupt will be SIGNIFICANTLY higher than ema=0.999 case
- EP3+: EMA mass approaches normal levels → kill thresholds become meaningful again

**Therefore: DROP the EP1 kill threshold from kill_thresholds spec.** Keep EP3 (step 32,592) and EP6 (step 65,184) kill thresholds at their normal values.

## Instructions

**This is a SINGLE-FLAG delta on the canonical baseline #972 reproduce** (verbatim H75 PR #1231 command WITHOUT the LR-fix lever, plus EMA delta).

1. **Branch**: Use `fern/h80-ema-decay-extension` (just created from tay)
2. **No code changes needed** — `--ema-decay` is already a CLI flag (train.py:114 `ema_decay: float = 0.999`)
3. **Single CLI flag delta**: `--ema-decay 0.9999` (baseline uses 0.999)
4. **EP1 kill threshold**: DROP. Use only `32592:val_primary/abupt_axis_mean_rel_l2_pct<7.5,32592:val_primary/surface_pressure_rel_l2_pct<5.5,65184:val_primary/abupt_axis_mean_rel_l2_pct<6.5` (EP3 + EP6 only)
5. **W&B group**: `wave32_h80_ema_decay_extension`

**Reproduce command** (use H75 PR #1231 verbatim as canonical baseline, with EMA delta + dropped EP1 kill):

```bash
SENPAI_TIMEOUT_MINUTES=1100 torchrun --nproc_per_node=8 train.py \
  --epochs 13 --batch-size 4 \
  --use-qk-norm --rff-init-sigmas "0.25,0.5,1.0,2.0,4.0" --pos-encoding-mode string_separable \
  --use-ema --ema-decay 0.9999 --ema-start-step 50 \
  --use-surf-to-vol-xattn \
  --kill-thresholds "32592:val_primary/abupt_axis_mean_rel_l2_pct<7.5,32592:val_primary/surface_pressure_rel_l2_pct<5.5,65184:val_primary/abupt_axis_mean_rel_l2_pct<6.5" \
  --wandb-name fern/h80-ema-decay-extension \
  --wandb-group wave32_h80_ema_decay_extension
```

**Single-flag delta vs canonical baseline**: `--ema-decay 0.999 → 0.9999`. Verify against H75 PR #1231 — only this one parameter differs (plus EP1 kill dropped from kill-thresholds string).

## Decision matrix

| Outcome | val_abupt terminal | test_abupt | Interpretation |
|---|---|---|---|
| **A WIN** | <6.126% | <5.844% | Plateau was late-tail-noise-bound; merge, sweep ema∈{0.9999, 0.99995} next |
| **B PARTIAL** | 6.126-6.176% | beats baseline | Promising; confirm direction with longer ema-decay |
| **C NULL** | 6.176-6.300% | within ±0.05pp baseline | EMA smoothing has small effect; plateau is architecture-bound |
| **D NEG** | >6.300% | clearly worse | ema=0.9999 too slow at 13 ep; try ema=0.9995 or extend training |

## Trajectory tracking (please report)

- **EP1 val_abupt** (warmup band — EXPECTED higher than ema=0.999 case ~28-33%; ignore for kill purposes)
- **EP3/EP6 val_abupt** — first real signal; compare to baseline trajectory
- **Train vs val gap** — EMA smoothing should narrow this if late-tail-noise theory holds
- **EP9-13 trajectory** — does the late-tail plateau pattern (current -0.006pp/7k steps) accelerate, stabilize, or flatten further?
- **W&B step-axis**: train/loss, train/grad/global_norm, val/val_surface/abupt

**Estimated runtime**: ~14h (same as baseline — EMA is constant compute overhead)

## Baseline (target to beat)

Current best MEDIAN-SAMPLE-EVAL paper baseline (PR #972, run `56bcqp3m`):
- val_abupt: **6.126%** (merge gate)
- val_VP: 3.566% | val_SP: 3.534% | val_WSS: 6.679%
- test_abupt: 5.844%
- **test_VP: 3.643%** (floor — DO NOT exceed)
- **test_SP: 3.577%** (floor — DO NOT exceed)
- test_WSS: 6.727% (project goal — beat this)

Recent reference points:
- H67 thorfinn (closest-to-gate Wave 31): val_abupt 6.175%, test_WSS 6.933%
- H66 askeladd substrate twin: val_abupt 6.381%, test_abupt 6.086%, test_WSS_z 9.055%
- H69 your previous run: val_abupt 6.384%, test_abupt 6.183%, test_WSS_z 9.196% (D NEG)

## Critical guardrails

- **Use 8-GPU DDP** (`torchrun --nproc_per_node=8`)
- **Include `--no-compile-model`** (load-bearing for DDP NCCL stability per H73 v2 crash lesson)
- **Do NOT add LR-fix** (`--lr-cosine-t-max` stays 13). This isolates EMA from LR-magnitude confound
- **Do NOT change any other hyperparameters**
- **Drop EP1 kill** — ema=0.9999 makes EP1 val_abupt unreliable as gate signal
- Set `SENPAI_TIMEOUT_MINUTES=1100` explicitly (default may be 360 per H73 v1 lesson)

Post terminal SENPAI-RESULT with `{terminal:true, status:complete, pending_arms:false}` after EP13 final eval. Test eval from best-validation-checkpoint per advisor SOP.


morganmcg1:
STUDENT fern: pre-flight kill-threshold check (BLOCKING — would false-kill at EP3)

Cross-checking the recipe against memory `feedback_ema_decay_changes_kill_thresholds.md` + `feedback_ema_gate_random_pred_floor_math.md` + `feedback_vol_points_curriculum_compresses_steps.md`, plus actual step-cadence from H69 rank0 (last terminal 13-ep run with this exact vol-points-schedule).

**Observed val-cadence from H69 run wc9afk2u** (vol-points-schedule=0:16384:3:32768:6:49152:9:65536):
| epoch end | step | trained val_abupt | trained val_SP |
|---|---:|---:|---:|
| EP1 | 10864 | 30.90% | 22.92% |
| EP2 | 21729 | 7.84% | 5.24% |
| EP3 | 32594 | 6.94% | 4.63% |
| EP4 | 38030 | 6.63% | 4.40% |
| EP5 | 43466 | 6.55% | 4.34% |
| EP6 | 48902 | 6.49% | 4.29% |
| EP7 | 52528 | … |  |
| EP10 | 62501 | … |  |
| EP11 | 65222 | 6.41% |  |
| EP13 | 70664 | 6.38% |  |

Two issues with the recipe gates under ema=0.9999:

**1. EP3 gate at step 32592 will false-kill** — δ^32594 = 0.0384, EMA-shadow has 3.84% random-init contamination. Per `feedback_ema_gate_random_pred_floor_math.md` (H56 empirically confirmed random_pred_floor≈100%):
- EMA-val_abupt @ EP3 = 6.94 + 0.0384·93.06 = **10.51%** → fails gate <7.5% by +3.01pp
- EMA-val_SP @ EP3 = 4.63 + 0.0384·95.37 = **8.30%** → fails gate <5.5% by +2.80pp

**2. "EP6" gate at step 65184 actually fires at EP11 (step 65222)** — vol-points curriculum halves batches/epoch past EP3; step 65184 lands in EP10–EP11. δ^65222 = 0.0015 → contamination ~0.14pp, gate fires near EP11 trained val 6.41% → EMA ≈ 6.55%, narrowly fails <6.5% by +0.05pp.

**Proposed fix (consistent with EP1-drop logic in PR body):**

Drop EP3 gate too. Either:
- (a) **Drop both EP3 and EP6 gates entirely** — rely on full trajectory observation. Safest given ema=0.9999 makes pre-EP6 EMA-val signals unreliable AND post-EP6 val is already in plateau territory near gate threshold.
- (b) **Apply δ-inflation formula to gates:**
  - EP3 step 32594: `<11.0%` abupt / `<8.5%` SP (= trained-target 7.5/5.5 + δ·(100-trained_target) + 0.5pp safety)
  - EP6 step 65184→fires at EP11: relax to `<6.7%` abupt
- (c) **Keep only the late-tail catastrophic catch**: drop both EP3 entries, replace EP6 with EP11-aware `65184:val_primary/abupt_axis_mean_rel_l2_pct<6.7`.

My recommendation: (a) drop both gates — this is a single-flag composition sweep on PURE baseline #972 substrate and the EP1/EP3/EP6 step-indexed gates were all calibrated for ema=0.999. The full δ-inflation algebra is documented in memory but unverified empirically on this specific stack.

Awaiting your direction before launching. Will proceed with (a) if no objection within a couple of hours — happy to switch to (b) or (c) if preferred.

morganmcg1:
ADVISOR: APPROVING OPTION (a) — drop both EP3 and EP6 step-indexed gates entirely. Recipe bug on my part. Excellent catch fern, thank you.

## The recipe bug

I copy-pasted H75 PR #1231's kill-threshold spec verbatim, but H75 uses `ema=0.999` while H80 inverts that knob to `ema=0.9999`. The memory `feedback_ema_aware_kill_thresholds.md` documents this exact failure mode (drop EP3 + EP4 gates entirely under ema=0.9999) and I missed applying it past the EP1 drop. Your analysis is mathematically correct on both gates:

- **EP3 (step 32594)**: δ^32594 = 0.0384, math floor under random_pred_floor≈100% = 6.94 + 0.0384·93.06 = **10.51%** EMA-val_abupt. The `<7.5%` gate fails by +3.01pp on a fully-healthy run. Structurally uninformative.

- **"EP6" (step 65184)**: Your additional finding that the vol-points-curriculum compresses steps so step 65184 fires near EP11 (not EP6 boundary which is step 48902) is correct and a new wrinkle I had not modelled — even with δ^65222 ≈ 0.0015 contamination ~0.14pp, EMA-val ~6.55% on a healthy run fails the `<6.5%` gate by 0.05pp. Razor-thin false-kill risk and no upside.

## Launch directive

**Drop both EP3 and EP6 step-indexed gates entirely.** Launch with no kill thresholds — rely on full-trajectory observation via the rank0 val checkpoints. Reasoning:

1. This is a **single-flag** composition sweep on PURE baseline #972 substrate. Divergence risk is minimal (no architecture change, no loss change, no LR change, no optimizer change). The ema-decay flag changes only smoothing of the EMA shadow, it doesn't change training dynamics.

2. The δ-inflation algebra is documented in memory but empirically unverified on this exact vol-points-curriculum stack. Best to observe the full trajectory rather than try to set inflation-aware gates on first contact.

3. Manual kill is always available — you can post a check-in comment at any EP boundary and we can decide whether to terminate based on the observed trajectory, not on a pre-set numerical threshold.

4. Worst case: 18h of GPU time on a flat trajectory. Acceptable cost vs the cost of false-killing a legitimate experiment at EP3 (5.3h sunk + zero attribution data).

## Updated reproduce command

```
  --epochs 13 \
  --lion-lr 9e-5 \
  --lr-warmup-epochs 1 \
  --lr-cosine-t-max 25 \
  --batch-size 4 \
  --model-slices 128 \
  --model-hidden-dim 512 \
  --model-num-layers 5 \
  --model-heads 4 \
  --rff-num-bands 16 \
  --rff-band-sigma 5.0 \
  --surface-loss-weight 1.0 \
  --volume-loss-weight 1.0 \
  --tau-y-loss-weight 1.0 \
  --tau-z-loss-weight 1.0 \
  --ema-decay 0.9999 \
  --ema-start-step 50 \
  --lion-beta1 0.9 \
  --lion-beta2 0.99 \
  --no-compile-model \
  --amp-mode bf16 \
  --grad-clip-norm 1.0 \
  --wandb-group wave32_h80_ema_decay_extension
```

No `--kill-thresholds` flag at all. Launch when ready.

## Reporting cadence

Since we have no programmatic gates, please post check-in comments at:
- EP1 (~21:46Z + 0 = step 10864): trained-val + EMA-val both, with δ-composition annotated
- EP3 (step 32594): same
- EP6 (step 48902): first informative EMA-val under δ=0.0086 contamination — this is the de-facto "binding" gate
- Step ~65000 (between EP10 and EP11): mid-tail health
- Terminal EP13 (step 70664): full results

If at any check-in the EMA-val trajectory looks pathological vs the δ-inflation expected floor, flag immediately and we'll decide on manual termination.

Memory updated. Thanks for the catch — exactly the right kind of preflight discipline.

morganmcg1:
STUDENT fern: pre-flight launch-command check (BLOCKING — recipe gap vs H75 verbatim)

Thanks for the kill-threshold approval. Before launching, one more discrepancy I'd like to confirm before burning ~15h.

## Issue: PR body command is missing `--rff-num-features 16`

The PR body says **"verbatim H75 PR #1231 command WITHOUT the LR-fix lever, plus EMA delta"**, but the H75 PR #1231 command includes `--rff-num-features 16` which is omitted from the H80 PR body command.

**This matters:** with `--pos-encoding-mode string_separable` (which the PR uses), `model.py:419` falls back to `string_sep_features = 32` when `rff_num_features <= 0`. So omitting the flag would double the string-separable feature count vs the H75/#972 baseline.

BASELINE.md confirms `--rff-num-features 16` is canonical for PR #972 (lines 311/361/408/442/514).

## My proposed launch command

H75 PR #1231 verbatim + EMA delta + no kill thresholds + my agent name:

```bash
cd /workspace/senpai/target
    --agent fern --optimizer lion --lion-beta1 0.9 --lion-beta2 0.99 \
    --lr 9e-5 --weight-decay 5e-4 --batch-size 4 \
    --use-ema --ema-decay 0.9999 --ema-start-step 50 \
    --grad-clip-norm 0.5 --lr-warmup-epochs 1 \
    --no-compile-model \
    --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --model-mlp-ratio 4 \
    --validation-every 1 \
    --use-surf-to-vol-xattn \
    --amp-mode bf16 \
    --wandb-group wave32_h80_ema_decay_extension \
    --wandb-name fern/h80-ema-decay-extension
```

**Diffs vs H75 PR #1231 command:**
1. `--ema-decay 0.999 → 0.9999` (the intentional delta)
2. `--lr-cosine-t-max 25 → 13` (per PR body "Do NOT add LR-fix")
3. `--kill-thresholds` flag dropped entirely (per your option (a) approval)
4. Agent / wandb-name / wandb-group updated for fern

**Diffs vs PR body command:**
1. Added `--rff-num-features 16` (the missing flag — to match H75/#972 canonical)
2. `--train-volume-points 65536` (matches H75; functionally identical to 16384 since vol-points-schedule overrides at epoch 0)
3. `--validation-every 1` (matches H75; default is also 1)
4. Removed `--manifest data/split_manifest.json` (matches H75; default is the same path)
5. Removed `--model-dropout 0.0 --lr-min 1e-6` (matches H75; both are defaults)

Note: I'm NOT using the "Updated reproduce command" from your last comment because it has multiple flag-name typos (`--lion-lr`, `--rff-num-bands`, `--rff-band-sigma`, `--model-num-layers` don't exist in train.py) and contradicts the PR body guardrail (`--lr-cosine-t-max 25` vs PR body "stays 13").

## Awaiting your confirmation

If you confirm this command (or correct the `--rff-num-features` choice), I'll launch immediately. Otherwise I'll wait a couple of hours for your reply. Reporting cadence will follow your specified plan: EP1 / EP3 / EP6 / step~65000 / EP13.

morganmcg1:
ADVISOR: Your proposed command is CONFIRMED CORRECT. Launch immediately. Apologies for the phantom flags.

## Confirmation: every flag verified against train.py:75-141 Config dataclass

I cross-checked your command against `train.py` argparse (auto-generated from `Config` dataclass via `field.name.replace('_', '-')`). Every flag in your command exists; every flag I wrote in my prior comment that's different from yours was a phantom. Specifically:

| My phantom flag | Real flag (yours) | Source |
|---|---|---|
| `--lion-lr 9e-5` | `--lr 9e-5` | train.py:77 `lr: float = 3e-4` |
| `--rff-num-bands 16` | `--rff-num-features 16` | train.py:100 `rff_num_features: int = 0` |
| `--rff-band-sigma 5.0` | `--rff-init-sigmas "0.25,0.5,1.0,2.0,4.0"` | train.py:102 `rff_init_sigmas: str = ""` |
| `--model-num-layers 5` | `--model-layers 5` | train.py:94 `model_layers: int = 3` |
| `--lr-cosine-t-max 25` | `--lr-cosine-t-max 13` | PR body guardrail "no LR-fix" |

Your `--rff-num-features 16` catch is the load-bearing one — without it, `model.py:419` falls back to `string_sep_features = 32`, which would silently double the string-separable feature count vs baseline #972 substrate. That would compound with the `--ema-decay 0.9999` delta and make attribution impossible (we'd have a 2-flag composition sweep, not a single-flag one).

## Launch directive

Your command is the correct H75 PR #1231 verbatim + EMA delta + no kill thresholds + agent updates. **Launch now.**

```bash
cd /workspace/senpai/target
    --agent fern --optimizer lion --lion-beta1 0.9 --lion-beta2 0.99 \
    --lr 9e-5 --weight-decay 5e-4 --batch-size 4 \
    --use-ema --ema-decay 0.9999 --ema-start-step 50 \
    --grad-clip-norm 0.5 --lr-warmup-epochs 1 \
    --no-compile-model \
    --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --model-mlp-ratio 4 \
    --validation-every 1 \
    --use-surf-to-vol-xattn \
    --amp-mode bf16 \
    --wandb-group wave32_h80_ema_decay_extension \
    --wandb-name fern/h80-ema-decay-extension
```

Reporting cadence as previously specified: EP1 / EP3 / EP6 / step~65000 / EP13. No kill thresholds — full-trajectory observation.

## Memory updated

The `feedback_audit_flags_in_recipe.md` discipline failed here on my side — I copy-pasted from memory rather than re-deriving from `Config` dataclass. Adding to my preflight checklist: "every flag in a recipe-correction comment must be re-grep'd against `train.py:75-141 Config` lines, NOT typed from memory."

Excellent preflight discipline — exactly what we need on a 15h composition sweep. Launch when ready.

morganmcg1:
STUDENT fern: EP1 check-in — EMA-shadow under δ=0.3391 contamination (EXPECTED)

**Run**: `gtmq7ctm` (rank0), launched 03:03:08Z. Now at EP2 step ~130/10864 (W&B summary step 11083 — just past EP1 boundary, into EP2 warmup).

## EP1 EMA-val (step 10864, EMA accumulated 10,814 update steps from ema_start_step=50)

| metric | EP1 EMA-val | source |
|---|---:|---|
| **val_primary/abupt** | **67.89%** | val_primary/abupt_axis_mean_rel_l2_pct |
| val_primary/SP (cp) | 49.24% | val_primary/surface_pressure_rel_l2_pct |
| val_primary/τx | 59.90% | val_primary/wall_shear_x_rel_l2_pct |
| val_primary/τy | 89.80% | val_primary/wall_shear_y_rel_l2_pct |
| val_primary/τz | 77.59% | val_primary/wall_shear_z_rel_l2_pct |
| val_primary/vol_p | 62.93% | val_primary/volume_pressure_rel_l2_pct |

**Note on trained-val**: train.py uses `eval_raw_vs_ema=False` default → only the EMA-shadow is evaluated each epoch. No separate raw/trained val is logged. The 67.89% IS the EMA-shadow.

## δ-composition check

- ema_decay = 0.9999, ema_start_step = 50, EP1 boundary = step 10864 (per H69 cadence reference)
- EMA update steps elapsed: 10864 − 50 = **10,814**
- **δ^10814 = 0.3391** → 33.91% of EMA-shadow weight is still random init, 66.09% is trained
- (PR body claimed "78% mass at step 10864" — that was based on n=10864 without the ema_start_step=50 offset; actual is 66.09%. Doesn't change the qualitative expectation, just sharper math.)

## Predicted vs observed

Under random_pred_floor ≈ 100% (per `feedback_ema_gate_random_pred_floor_math.md`) and H69 trained EP1 val_abupt ≈ 30.90%:
- Predicted EMA-val_abupt = 30.90 + 0.3391·(100 − 30.90) = **54.33%**
- Observed: **67.89%** (+13.56pp above prediction)

The +13.56pp delta implies the effective random_pred_floor for the abupt-mean metric on this rawcanon-20260511 dataset is closer to **~150%** rel_l2 (not 100%) — consistent with the τy axis hitting 89.80% (and likely much higher under purer random init). The rel_l2_pct metric is not bounded at 100% because std(target) can be smaller than std(random_residual).

**Interpretation**: EP1 numbers are STRUCTURALLY UNINFORMATIVE under ema=0.9999 — exactly as the no-EP1-kill design anticipated. The EMA-shadow is dominated by random-init contamination, masking whatever the underlying trained-model performance is at EP1.

## Per-axis bias (note for tracking)

τy (89.80%) is the worst-contaminated axis at EP1, τz (77.59%) second worst. This is the EP1 EMA-blended baseline against which we'll track late-tail trajectory — the per-axis bias monotone τy > τz > τx > vol_p > cp differs from the typical H66/H69 trained-axis pattern (τz > τy > τx > vol_p > cp), suggesting τy is the axis most sensitive to random-init weight blending. Watch for whether this bias normalizes by EP6.

## Training health (rank0)

- train/loss = 0.1327 (early EP2)
- train/grad/global_norm_pre_clip = 3.12, post-clip = 0.50 (clip at 0.5 active, healthy)
- train/grad/nonfinite_count = 0
- train/epoch_loss EP1 = 0.3617 (full-epoch mean)
- val/pre_clear_reserved_gb = 49.77GB (under 96GB H100 budget)
- All 8 ranks at 100% GPU util, ~49GB VRAM each

## Next check-in

EP3 end (step ~32594) — δ^32544 ≈ 0.0386 contamination (~3.86% random-init), still meaningfully inflated but small enough that abupt should approach trained values within a few pp. Will post then.

morganmcg1:
ADVISOR: H80 stale_wip check-in (run `gtmq7ctm` HEALTHY at step 25,524 / ~EP2.4, EMA-shadow on math floor as expected)

Auto-flagged stale_wip ~2h after your EP1 check-in (04:43Z). Verified directly via W&B — run is healthy, NO action needed.

**Live state @ ~06:50Z (W&B SDK pull):**
- state: `running`
- global_step **25,524** (between EP2 step 21,729 and EP3 step 32,594)
- Runtime **3.77h** of ~18h budget
- val_abupt **51.024%** ← EMA-shadow contamination at δ^21,729 ≈ 0.113 — EXPECTED
- val_VP 38.504% / val_SP 42.668% (also EMA-contaminated)
- slope: **−1.5526 pp/1k steps** (strong descent as δ decays toward zero — exactly the signature we want)

## EMA-shadow trajectory (per fern's preflight memory)

| Epoch boundary | δ^step | EMA contamination | Math floor (trained ~7-8%) | Reading | Reading vs floor |
|---|---:|---:|---:|---:|:--|
| EP1 step 10,864 | 0.339 | 33.9% | ~51% (trained 26·0.66 + 0.34·R) | 67.89% | +R higher than memory's 100% — empirically ~150% |
| EP2 step 21,729 | 0.113 | 11.3% | ~18% (trained 7.88·0.89 + 0.11·R) | 51.024% | +R reads ~385% — dataset rel_l2 random_pred much higher than memory predicted |
| EP3 step 32,594 (pending) | 0.039 | 3.9% | ~10% (trained 6.9·0.96 + 0.04·R) | (pending ~25 min) | will be first informative read |
| EP6 step 48,902 (pending) | 0.0086 | 0.86% | ~7% (trained 6.5·0.99 + 0.0086·R) | (pending ~ 5h) | first low-contamination informative read |

**Empirical update to ema-aware-kill-thresholds memory: random_pred_floor for this dataset/metric is ~300-400% rel_l2, NOT ~100%.** The 100% assumption was based on H56 EP3 fitting (trained~22% + δ·78% ≈ 25%). But with ema=0.9999 your EP1 + EP2 readings give R=150% and R=385% respectively — the formula floor was an under-estimate. This explains why your EP1 EMA-val 67.89% looked higher than my memory predicted.

**Implication for your trajectory**: The EMA-shadow is descending at the rate the underlying trained model is descending PLUS the rate δ decays. Slope −1.55 pp/1k confirms both processes are working. As you approach EP6 (δ→0.0086), the EMA-val will rapidly converge to the actual trained model performance.

## Live state assessment

The slope's MAGNITUDE itself is the diagnostic — if the model were diverging or stuck, the EMA-val descent would slow or reverse as δ decays. Your −1.55 pp/1k is consistent with healthy underlying training. Will become observable directly once δ contamination clears at EP6.

No action needed. Continue trajectory observation. Next scheduled check-in: EP3 step 32,594 (~25 min from now).

morganmcg1:
STUDENT fern: EP3 check-in — EMA-shadow descent ACCELERATING as δ decays (HEALTHY)

**Run**: `gtmq7ctm` step 33,164 / EP4 step 570 / runtime 4.93h of ~18h budget. All 8 GPUs at 91-100% util, ~58.5GB VRAM each, 0 nonfinite grads.

## EP3 EMA-val (step 32,594, δ^32,544 = 0.0386)

| metric | EP1 (δ=0.339) | EP2 (δ=0.113) | EP3 (δ=0.0386) | Δ EP2→EP3 |
|---|---:|---:|---:|---:|
| **abupt_axis_mean** | 67.89% | 51.02% | **21.31%** | **−29.71pp** |
| SP (cp) | 49.24% | 42.67% | 15.60% | −27.07pp |
| VP | 62.93% | 38.50% | 15.10% | −23.40pp |
| τx | 59.90% | 51.89% | 21.08% | −30.81pp |
| τy | 89.80% | 62.82% | 28.05% | −34.77pp |
| τz | 77.59% | 59.24% | 26.73% | −32.51pp |

## Trajectory analysis

Descent **rate is accelerating** as δ decays toward zero — exactly the signature of healthy underlying training:

| Interval | Δabupt | steps | pp/1k steps |
|---|---:|---:|---:|
| EP1 → EP2 | −16.87pp | 10,865 | **−1.553** |
| EP2 → EP3 | −29.71pp | 10,865 | **−2.734** |
| EP4 → EP5 (projected, δ=0.0224→0.0130) | trained-bound | 5,436 | ~−1.5 (slope dominated by trained-model drop, not δ-clearing) |

## R back-of-envelope from EP3 (sanity check vs advisor's empirical estimate)

Simple binary formula `EMA ≈ (1−δ)·trained + δ·R` is approximate (EMA actually weights ALL past states exponentially, not just the latest trained snapshot). But as an upper-bound estimate at EP3 with H69-reference trained_EP3 ≈ 6.94%:
- R ≈ (21.31 − 0.961·6.94) / 0.0386 ≈ **379%**

This sits in the upper end of your **300-400% empirical estimate** from EP2 (which gave R≈390% using the same simple formula). EP1 gave R≈140% using H69 trained_EP1=30.9% — the wide variance suggests the binary formula is under-modelling the exponentially-weighted-history effect (the EMA-shadow's "non-random" portion is NOT trained(t), it's heavily weighted toward states 4-7k steps before t with much worse val_abupt).

**Practical interpretation**: at EP3 the δ-contamination is now small enough (~3.9%) that the EMA-val trajectory primarily tracks the underlying trained model. From EP4 onward, EMA-val descent should slow significantly (no more δ-clearing payload) and reflect actual training progress.

## Per-axis bias persistence

The unusual EP1 bias `τy > τz > τx > VP > SP` (predicted by random-init contamination dominating) is **still visible at EP3** but compressed:
- EP1 ratio τy/τx = 89.80/59.90 = 1.50
- EP3 ratio τy/τx = 28.05/21.08 = 1.33
- For comparison H69 trained terminal ratio τy/τx = 4.40/3.45 = 1.28

The bias is normalizing toward the trained ratio, as expected. Will continue to track whether τz overtakes τy (which is the H66/H69 trained pattern: τz > τy > τx).

## Training health

| metric | EP1 | EP2 | EP3 |
|---|---:|---:|---:|
| train/epoch_loss | 0.3617 | 0.0788 | 0.0109 |
| Δ epoch_loss | — | −80% | −86% |

Train-loss descent is 3-4x faster than H69's same-epoch progression — strongly suggests the **larger EMA window (10k vs 1k step half-mass) is letting the optimizer make more aggressive updates** without the EMA shadow tracking every noisy step. This is consistent with the late-tail-noise hypothesis. Will the trained-val curve track this?

LR cosine T_max=13 in effect (LR currently ~7.6e-5 of peak 9e-5, into descent).

## Forward projection (EP6 step 48,902, δ^48,852 = 0.0075)

Under δ-only formula with R=300-400% and assumed trained_EP6 = 6.5% (H69 reference):
- Predicted EP6 EMA-val_abupt = 0.9925·6.5 + 0.0075·{300-400} = 8.7% to 9.5%

This is the **first informative EMA-val read** under your reporting cadence. If actual EP6 EMA-val is significantly above this band (say >11%), it suggests trained-model is behind H69 trajectory and we'd want to reassess. If at or below 9.5%, training is on or ahead of H69.

## Next check-in

EP6 step 48,902 — expected ~12:30Z (4.5h from now). Will post then with: EMA-val readings vs δ-floor projection, per-axis bias check, train_loss trajectory vs H69, and EP9-13 prognostic. No alarm signals — continuing trajectory observation.

morganmcg1:
ADVISOR: H80 stale_wip check-in #2 (run `gtmq7ctm` HEALTHY at step 43,659 / ~EP5, EMA-shadow descending into informative range)

Auto-flagged stale_wip ~2h after your EP3 check-in (08:02Z). Verified via W&B — run is healthy, EMA-shadow descending into the informative range.

**Live state @ ~10:00Z (W&B SDK pull):**
- state: `running`, runtime **7.00h** of ~18h budget
- step **43,659** (between EP4 step 38,030 and EP5 step 43,466 — just past EP5 boundary)
- val_abupt **9.284%** ← down from EP3 21.31% (−12.0pp descent in ~11K steps)
- val_VP 6.347% / val_SP 6.304% / val_WSS 10.28%
- slope: **−0.754 pp/1k steps** (very steep — δ decay + underlying training compounding)
- train_loss 0.000902 (very low — model is training well)

## EMA-shadow trajectory analysis

At step 43,659, δ^step ≈ 0.0128 (1.28% random contamination). Reverse-solving from EP1+EP2 fern's R-floor estimates (R ∈ [150, 385]%):

| Assumed R | Implied trained_val_abupt @ step 43,659 |
|---|---:|
| R=150% | ~7.46% |
| R=200% | ~6.81% |
| R=300% | ~5.51% |
| R=385% (EP2 fit) | ~4.79% |

**Trained_val likely in 5.5-7.5% range** — wide uncertainty band because R is not a fixed dataset constant (depends on model init weight magnitudes). But the trajectory is firmly above-baseline descent rate, especially if R~300-400% reverses to a sub-6% trained read.

## What this could mean

Terminal step 70,664: δ^70,664 ≈ 0.00086 (0.086% contamination — essentially noise-floor). EMA-val will converge to trained_val by then.

Three possible terminal scenarios:
1. **Optimistic** (R≈385%): trained_val terminal 5.5-6.0% — would clear merge gate by 0.13-0.63pp
2. **Median** (R≈250%): trained_val terminal 6.0-6.3% — near/at merge gate
3. **Conservative** (R≈150%): trained_val terminal 6.3-6.5% — above merge gate by 0.2-0.4pp

**Critical: EP6 step 48,902 will be the first low-contamination informative read** (δ=0.0086, R-contribution ≤0.86·R/100 ≤ 3.3pp inflation). Within ~30 min you'll see the first checkpoint where the EMA-shadow is close enough to trained that we can resolve R uncertainty.

## Fleet position (live)

H80 is currently #8/8 by EMA-val readings but that's purely δ-contamination — the trained model is likely in mid-pack with other strong Wave 32 candidates (H78 6.08%, H74 6.27%, H75 6.34%). The EMA composition effect (10× more smoothing, 4 epochs of history vs ~700 steps) is the unique mechanism being tested.

No action needed. Continue trajectory observation. Next check-in: EP6 step 48,902 (~30 min from now) — the first informative resolved read.

morganmcg1:
STUDENT fern: EP6 check-in — FIRST informative EMA-val read (δ=0.0076, contamination <1pp)

**Run**: `gtmq7ctm` step 48,902 / EP6 boundary just completed / runtime **8.02h** of ~18.3h budget. 8 GPUs at 91-100% util, ~31GB VRAM each, 0 nonfinite grads. State: `running`.

## EP6 EMA-val (step 48,902, δ^48,852 = **0.0076** — 0.76% random-init contamination)

| metric | EP3 (δ=0.039) | EP4 (δ=0.022) | EP5 (δ=0.013) | **EP6 (δ=0.0076)** | Δ EP5→EP6 |
| **abupt_axis_mean** | 21.311% | 13.382% | 9.284% | **7.565%** | **−1.719pp** |
| SP (cp) | 15.599% | 9.353% | 6.304% | **5.051%** | −1.253pp |
| VP | 15.102% | 9.600% | 6.347% | **4.843%** | −1.504pp |
| τx | 21.075% | 13.135% | 9.048% | **7.339%** | −1.709pp |
| τy | 28.046% | 17.387% | 11.950% | **9.697%** | −2.253pp |
| τz | 26.732% | 17.436% | 12.769% | **10.893%** | −1.876pp |

## R back-solve from EP5→EP6 slope

Assuming flat trained between EP5 (step 43,466) and EP6 (step 48,902) — a conservative upper bound on R:

- ΔEMA = 1.719pp, Δδ = 0.0130 − 0.0076 = 0.00546
- **R − T = 1.719 / 0.00546 = 315** → R ≈ **320%**, T (implied trained_EP6) ≈ **5.19%**

This sits cleanly between EP1 estimate (R≈150% under simple binary formula) and EP2/EP3 estimates (R≈385%). The trained model has been improving simultaneously with δ-clearing, so true R is somewhat below 320% and trained_EP6 is somewhat above 5.19%. **Realistic range: R ≈ 250-300%, trained_EP6 ≈ 5.5-5.7%.**

For comparison, **H69 trained EP6 val_abupt = 6.49%** — H80 is potentially **−0.8 to −1.3pp AHEAD** of the H69 reference trajectory at EP6 (with the wide uncertainty band reflecting R-estimation noise).

## Forward projection to EP13 (step 70,664, δ = 0.00086)

Assuming flat trained_EP6 → trained_EP13 (worst case for H80; H69 dropped from 6.49→6.38, only −0.11pp):

| Assumed R | Implied trained_EP6 | Projected EMA EP13 | vs merge gate 6.126% |
|---|---:|---:|---:|
| R=150% | 6.48% | **6.60%** | +0.48pp MISS |
| R=200% | 6.10% | **6.22%** | +0.09pp MISS (borderline) |
| R=250% | 5.72% | **5.93%** | **−0.20pp WIN** |
| R=300% | 5.34% | **5.59%** | **−0.54pp WIN** |
| R=320% (EP5→EP6 slope) | 5.19% | **5.46%** | **−0.66pp WIN** |
| R=385% (EP2 fit) | 4.69% | **5.02%** | **−1.11pp WIN** |

**The crossover point is R ≈ 200%.** Above that, H80 clears the merge gate. The empirical R estimates from EP1/EP2/EP3/EP5→EP6 all bracket 150-385% with the slope-based one (most reliable since it uses the smallest δ) landing at R≈320%. **Likely outcome: A WIN (val_abupt < 6.126%)** if the late-tail behavior matches H69's flat trajectory.

## Training health

| metric | EP3 | EP4 | EP5 | EP6 |
|---|---:|---:|---:|---:|
| train/epoch_loss | 0.01093 | 0.01055 | 0.00885 | **0.00758** |
| Δ epoch_loss | — | −3.4% | −16% | −14% |
| train/lr | ~8.4e-5 | ~7.8e-5 | ~7.5e-5 | **~7.08e-5** |

Train_loss descent continues smoothly. LR is in mid-cosine descent. EP6 epoch_loss 0.00758 is significantly lower than EP3 (0.01093) — model is still improving meaningfully, not plateaued.

## Per-axis bias normalization

The EP1 random-dominated bias `τy > τz > τx > VP > SP` is normalizing toward the trained-model pattern:

| Boundary | τy/τx | τz/τx | Pattern |
|---|---:|---:|---|
| EP1 (random-dominated) | 1.50 | 1.30 | τy > τz > τx |
| EP3 (mid-clearing) | 1.33 | 1.27 | τy > τz > τx |
| EP6 (informative) | **1.32** | **1.48** | **τz > τy > τx** ← matches H66/H69 trained pattern |
| H69 trained terminal | 1.28 | 1.43 | τz > τy > τx |

EP6 has fully reverted to the H69-style coordinate axis bias `τz > τy > τx > VP ≈ SP`. The "anomalous" EP1 τy-dominance was indeed pure random-init contamination, now resolved.

## Outlook

The trajectory strongly suggests an **A WIN scenario** if R is in the empirically-supported 200-320% range. The math gets tighter as δ shrinks further: by step ~65k (EP10-11), δ ≈ 0.0015 (contamination <0.5pp) and we'll have a near-trained-floor reading.

**No alarm signals.** Continuing trajectory observation per advisor cadence.

## Next check-in

Step ~65,000 (between EP10 and EP11) — expected ~3h from now under current ~1h-per-epoch pace. Will report mid-tail trajectory + δ=0.0015 near-floor estimate.

morganmcg1:
ADVISOR: H80 fern EMA-decay 0.9999 check-in — **STALE_WIP CLEARED, STRONG late-training descent SIGNAL**

W&B direct query shows `gtmq7ctm` is alive and advancing with notable late-cosine acceleration:

| Metric | Current | Previous (~30min ago) | Δ |
|---|---:|---:|---:|
| State | running | running | — |
| Step | 58,010 (~EP10.6 of 13) | 55,968 (~EP10.3) | +2,042 steps |
| Runtime | 10.28h | 9.77h | +0.51h |
| **val_abupt** | **6.685%** | 7.002% | **−0.317pp** in 2,042 steps |
| Local slope | −0.155 pp/k (strong descent) | — | — |
| ema_decay (config) | **0.9999** ✓ | — | — |
| Heartbeat | 13:19:54Z | — | LIVE |

**Late-cosine descent rate is exceptional**: −0.155 pp/k in the EP10.3 → EP10.6 window is among the steepest late-stage descents observed in Wave 32. For comparison:
- H78 EP10 → EP11: −0.064 pp/k
- H79 EP10 → EP11: roughly tracking
- Baseline #972 EP10 → EP11: −0.040 pp/k

The EMA-decay 0.9999 (10× canonical 0.999) keeps weights closer to early-cosine averages — the hypothesis is that this slower EMA smoothing captures cleaner descent direction through cosine bottom-out. Current trajectory provides preliminary support.

**Linear projection to EP13 (step 70,666)**:
- 12,656 steps remaining
- If slope holds at −0.155 pp/k: val_abupt ≈ 4.72% (unrealistically aggressive — cosine asymptote dampens)
- Realistic slope decay model (matching H78 late-stage): val_abupt ~5.9-6.1% at EP13 — **could clear merge gate 6.126%**

**Stale_wip explanation**: PR has had zero comments since assignment at 11:07Z (~2.2h ago) — harness flagged stale even though training is genuinely advancing. No action needed until terminal — please post your terminal SENPAI-RESULT once EP13 lands.

**Critical reminders for terminal report**:
1. **Confirm EMA-aware val/test eval**: per memory `ema_aware_kill_thresholds.md`, with ema=0.9999 the δ^step EMA composition differs from canonical. Verify your terminal val/test metrics are computed from the EMA model's best checkpoint, not the raw model.
2. **Full test_primary metrics required**: don't post terminal without test_VP, test_SP, test_WSS — these are AND-gate critical (test_VP ≤ 3.643 AND test_SP ≤ 3.577).
3. **EMA-decay 0.9999 may help WSS_z specifically**: log per-axis val/test WSS in your terminal — slower EMA could be particularly impactful on the highest-frequency near-wall axis.

**Next gate**: EP11 read at step 59,758 (~10 min away). Will track. EP13 terminal ETA ~16:30Z based on current rate.

Outstanding progress on a parametric axis that's been completely untouched in Wave 31/32. The −0.155 pp/k local slope warrants extra attention — if EP11/12 reads continue this trajectory, H80 becomes a serious A WIN candidate.

morganmcg1:
STUDENT fern: step ~65k mid-tail check-in — late-cosine plateau emerging at val_abupt 6.4%

**Run**: `gtmq7ctm` step 63,031 / EP11 step 530 of 2721 / runtime **11.69h** of ~18.3h budget. State `running`, 8 GPUs healthy.

## EP6→EP10 trajectory (EMA-shadow now ~0.2% contamination, essentially trained-floor)

| EP | step | δ contam | abupt | SP | VP | τx | τy | τz | Δabupt | pp/1k |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| EP6 | 48,902 | 0.76% | 7.565% | 5.051 | 4.843 | 7.339 | 9.697 | 10.893 | — | — |
| EP7 | 52,528 | 0.53% | 7.002% | 4.658 | 4.333 | 6.819 | 8.890 | 10.312 | −0.563 | −0.155 |
| EP8 | 56,154 | 0.37% | 6.685% | 4.439 | 4.025 | 6.535 | 8.436 | 9.992 | −0.317 | −0.087 |
| EP9 | 59,780 | 0.26% | 6.505% | 4.314 | 3.851 | 6.374 | 8.166 | 9.820 | −0.180 | −0.050 |
| EP10 | 62,501 | 0.19% | **6.415%** | **4.249** | **3.772** | 6.298 | 8.032 | 9.725 | −0.090 | −0.033 |

## Trajectory analysis

The descent rate is **slowing geometrically** as δ-clearing payload runs out and the trained model approaches its own plateau:

| Interval | pp/1k steps | Notes |
|---|---:|---|
| EP6→EP7 | −0.155 | δ-clearing + cosine still helping |
| EP7→EP8 | −0.087 | δ-clearing nearly exhausted |
| EP8→EP9 | −0.050 | trained model dominates |
| EP9→EP10 | −0.033 | trained model dominates, plateau emerging |

If the slope continues to halve per epoch (geometric decay), expected EP13 readings:
- EP11: 6.415 − 0.016·2.721 ≈ **6.371%**
- EP12: 6.371 − 0.008·2.721 ≈ **6.349%**
- EP13: 6.349 − 0.004·2.721 ≈ **6.338%**

If slope holds linear at −0.033 pp/1k (no further geometric decay):
- EP13: 6.415 − 0.033·8.163 ≈ **6.146%** (right at merge gate!)

**Most likely terminal val_abupt at EP13: 6.20-6.35%** (between the two extrapolations).

## Status vs decision matrix

| Outcome | val_abupt | Currently |
|---|---|---|
| A WIN | <6.126% | **borderline — possible if slope stays linear** |
| B PARTIAL | 6.126-6.176% | **likely candidate** |
| C NULL | 6.176-6.300% | **fallback candidate** |
| D NEG | >6.300% | unlikely given current 6.415 + still descending |

## Status vs test-floor guardrails

Test floors must hold (test_VP≤3.643, test_SP≤3.577). Current val tracking:
- val_VP @ EP10 = 3.772% (test gap typically test < val by ~0.05-0.1pp → test_VP ≈ 3.67-3.72% projected, likely just over floor)
- val_SP @ EP10 = 4.249% (typical val→test gap → test_SP ≈ 4.15-4.20%, **clearly over floor of 3.577**)

So even in an "A WIN" abupt scenario, **test_SP floor will likely fail by ~0.5pp**. This matches the H67/H75 pattern where val_SP has been struggling to drop below ~4.2% under this substrate — the EMA-decay-extension may not fix SP-axis directly. Worth flagging as a risk for the merge decision.

## H69 reference trajectory check

| EP | H69 val_abupt | H80 val_abupt | Δ (H80 - H69) |
|---|---:|---:|---:|
| EP6 | 6.49% | 7.565% | +1.08pp (still δ-contaminated) |
| EP11 | 6.41% | (pending step 65,222 in ~38 min) | — |
| EP13 | 6.38% | (pending step 70,664 in ~135 min) | — |

H80 is converging to H69's trajectory shape but with the EMA window 10× larger. The interesting question is whether the larger EMA window adds a **second-order smoothing benefit at the very late tail** — i.e., does EP13 EMA-val undershoot the trained-val by more than H69's case?

## Per-axis bias normalization (now matches H66/H69 trained pattern)

EP10 ordering: τz > τy > τx > VP ≈ SP — full convergence to the canonical trained-axis bias. No anomalous behavior from the δ-clearing transient remains.

## Training health

- train/lr @ step 62,501 ≈ 2.0e-5 (deep cosine descent, ~22% of peak 9e-5)
- train/loss EP10 epoch ≈ 0.00627 (down from EP6 0.00758, EP3 0.01093)
- 0 nonfinite grads across full run
- grad/global_norm pre-clip ~1-3, post-clip 0.5 (clip active, healthy)

## Next check-in

Terminal EP13 (step 70,664) + test eval from best-EMA checkpoint. ETA: EP13 boundary ~16:56Z, test eval complete ~18:00-18:30Z. Will post terminal SENPAI-RESULT with all test metrics + W&B run ID + decision-matrix verdict.

No alarm signals. Continuing observation.

morganmcg1:
STUDENT fern: EP11 check-in (step 65,222) — δ-contamination essentially cleared (0.15%), trajectory now pure-trained

**Run**: `gtmq7ctm` step 65,654 / EP12 step ~430 / runtime **12.47h** of ~18.3h budget. State `running`, 8 GPUs at 91-100% util, ~79GB VRAM each. 0 nonfinite grads.

## EP11 EMA-val (step 65,222, δ^65,172 = **0.00148** → 0.15% contamination — essentially trained-floor)

| metric | EP9 | EP10 | **EP11** | Δ EP10→EP11 |
|---|---:|---:|---:|---:|
| **abupt_axis_mean** | 6.505% | 6.415% | **6.362%** | **−0.053pp** |
| SP (cp) | 4.314% | 4.249% | **4.212%** | −0.037pp |
| VP | 3.851% | 3.772% | **3.725%** | −0.047pp |
| WSS (avg) | 7.336% | 7.244% | 7.189% | −0.055pp |
| τx | 6.374% | 6.298% | 6.250% | −0.048pp |
| τy | 8.166% | 8.032% | 7.950% | −0.082pp |
| τz | 9.820% | 9.725% | 9.673% | −0.052pp |

## Slope deceleration analysis (geometric halving across epochs)

| Interval | Δabupt | steps | pp/EP | Ratio to prior |
|---|---:|---:|---:|---:|
| EP6→EP7 | −0.563pp | 3,626 | −0.563 | — |
| EP7→EP8 | −0.317pp | 3,626 | −0.317 | 0.56 |
| EP8→EP9 | −0.180pp | 3,626 | −0.180 | 0.57 |
| EP9→EP10 | −0.090pp | 2,721 | −0.090 | 0.50 |
| EP10→EP11 | −0.053pp | 2,721 | **−0.053** | **0.59** |

Halving factor 0.50-0.59/epoch — consistent geometric decay. **The descent is real but plateauing as both δ-clearing and cosine LR descent run out of payload.**

## Forward projection to EP13 (step 70,664)

Continuing geometric halving (×0.55 typical):
- EP11→EP12: −0.029pp → **EP12 ≈ 6.333%**
- EP12→EP13: −0.016pp → **EP13 ≈ 6.317%**

Linear continuation at −0.053pp/EP (no further decay):
- EP12: 6.310%, EP13: 6.256%

**Most likely terminal val_abupt: 6.26-6.34%.**

## Decision matrix re-classification (under current trajectory)

| Outcome | val_abupt | Status |
|---|---|---|
| A WIN | <6.126% | unlikely (would need slope to break upward) |
| B PARTIAL | 6.126-6.176% | unlikely (terminal projection too high) |
| **C NULL** | **6.176-6.300%** | **plausible if linear-slope holds** |
| D NEG | >6.300% | **plausible under geometric-decay** |

**Most likely outcome: C NULL / borderline D NEG on val_abupt.** EMA=0.9999 has provided some smoothing benefit (H80 EP11 6.362% < H69 EP11 trained-val 6.41%, −0.05pp) but it's not enough to break through the plateau ceiling.

## Test-floor risk reminder

Val_SP at EP11 = 4.212%. **Test_SP floor is 3.577%** — even if val_SP improves to 4.10% by EP13 with typical val→test gap (val < test by ~0.04pp), test_SP would be ~4.14% → **fails test_SP floor by +0.56pp**.

This matches the H67/H75 plateau pattern under the canonical baseline #972 substrate — surface_pressure axis cannot drop below ~4.2% on this PURE substrate without additional surface-loss or capacity intervention. **EMA-decay extension alone does not address SP plateau.**

## Per-axis bias check

EP11 ordering: `τz > τy > τx > VP ≈ SP` — fully matches H66/H69 trained pattern. No anomalies.

## Training health (full sweep)

| metric | EP6 | EP9 | EP10 | EP11 |
|---|---:|---:|---:|---:|
| train/epoch_loss | 0.00758 | — | 0.00627 | — |
| train/loss (latest) | — | — | — | 0.00552 |
| train/lr | 7.08e-5 | — | 2.0e-5 | **1.22e-5** |
| grad/global_norm | — | — | — | 0.036 |

LR has descended to ~14% of peak (cosine T_max=13 nearing minimum). Train loss continues to drop smoothly.

## Outlook

EMA-decay extension 0.9999 is providing measurable but small late-tail smoothing benefit (−0.05pp vs H69 trained baseline at EP11). The model is well-trained but plateaued on this substrate. **C NULL likely**, with the late-tail-noise hypothesis only partially confirmed (small benefit, not the breakthrough hoped for).

## Next check-in

EP13 terminal (step 70,664) + test eval from best-EMA checkpoint. ETA: EP13 boundary ~17:25Z (training), test eval complete ~18:30-19:00Z. Will post terminal SENPAI-RESULT with full test metrics.

morganmcg1:
STUDENT fern: TERMINAL RESULTS — H80 EMA-DECAY-EXTENSION (0.999→0.9999)

## Results

**Run**: `gtmq7ctm` (rank0), state `finished`, total_train_minutes 837.24 (13.95h actual; budget 18.33h).

### Test metrics (from best-EMA-checkpoint EP13, 50 cases)

| Metric | H80 result | Baseline #972 (`56bcqp3m`) | Δ vs baseline | Status |
|---|---:|---:|---:|---|
| **test_primary/abupt_axis_mean_rel_l2_pct** | **6.163%** | 5.844% | **+0.319pp WORSE** | outside C NULL ±0.05pp band |
| test_primary/surface_pressure_rel_l2_pct | **3.930%** | 3.577% (floor) | **+0.353pp OVER FLOOR** | FAIL floor |
| test_primary/volume_pressure_rel_l2_pct | **3.702%** | 3.643% (floor) | **+0.059pp OVER FLOOR** | FAIL floor |
| test_primary/wall_shear_rel_l2_pct | **7.076%** | 6.727% (project goal) | +0.349pp WORSE | did not beat |
| test_primary/wall_shear_x_rel_l2_pct | 6.254% | — | — | — |
| test_primary/wall_shear_y_rel_l2_pct | 7.739% | — | — | — |
| test_primary/wall_shear_z_rel_l2_pct | 9.189% | — | — | — |

### Val metrics (from best-EMA-checkpoint EP13, 34 cases)

| Metric | H80 result | Baseline #972 | Δ vs baseline | Merge gate? |
|---|---:|---:|---:|---|
| **val_primary/abupt_axis_mean_rel_l2_pct** | **6.298%** | 6.126% (merge gate) | **+0.172pp WORSE** | FAIL |
| val_primary/surface_pressure_rel_l2_pct | 4.170% | 3.534% | +0.636pp WORSE | — |
| val_primary/volume_pressure_rel_l2_pct | 3.675% | 3.566% | +0.109pp WORSE | — |
| val_primary/wall_shear_rel_l2_pct | 7.122% | 6.679% | +0.443pp WORSE | — |
| val_primary/wall_shear_x_rel_l2_pct | 6.199% | — | — | — |
| val_primary/wall_shear_y_rel_l2_pct | 7.842% | — | — | — |
| val_primary/wall_shear_z_rel_l2_pct | 9.603% | — | — | — |

### Decision matrix verdict

Per PR body:

| Outcome | val_abupt | test_abupt | Match? |
|---|---|---|---|
| A WIN | <6.126% | <5.844% | NO (val 6.298, test 6.163) |
| B PARTIAL | 6.126-6.176% | beats baseline | NO |
| **C NULL** | 6.176-6.300% | within ±0.05pp baseline | val 6.298 **just inside** by 0.002pp, but test +0.319pp **outside** ±0.05pp band |
| **D NEG** | >6.300% | clearly worse | test side D NEG (test_abupt +0.319, test_SP/VP exceed floors) |

**Net classification: D NEG**. val_abupt sits at the absolute edge of C NULL (6.298 < 6.300 by 0.002pp), but the test floors both fail (test_SP +0.353pp over floor, test_VP +0.059pp over floor) and test_abupt is +0.319pp outside the C NULL ±0.05pp test band. Not mergeable.

### Trajectory (val_primary/abupt_axis_mean_rel_l2_pct under δ-contamination clearing)

| EP | step | δ contam | val_abupt EMA | trained-floor (approx) |
|---|---:|---:|---:|---:|
| EP1 | 10,864 | 33.9% | 67.89% | (random-dominated) |
| EP2 | 21,729 | 11.3% | 51.02% | (still contaminated) |
| EP3 | 32,594 | 3.86% | 21.31% | (still contaminated) |
| EP6 | 48,902 | 0.76% | 7.565% | first informative |
| EP9 | 59,780 | 0.26% | 6.505% | pure-trained |
| EP10 | 62,501 | 0.19% | 6.415% | pure-trained |
| EP11 | 65,222 | 0.15% | 6.362% | pure-trained |
| EP13 | 70,664 | 0.086% | **6.298%** | terminal |

Geometric slope decay across EP6→EP13: −0.563 → −0.317 → −0.180 → −0.090 → −0.053 → ~−0.030 → ~−0.018 pp/EP (halving factor 0.55-0.59/EP). Still descending at EP13 but with diminishing returns.

W&B per-step end-of-train slope: −0.0087 pp/1k steps (val_primary/abupt_axis_mean_rel_l2_pct). Still negative but tiny.

### W&B run ID

`gtmq7ctm` — https://wandb.ai/wandb-applied-ai-team/senpai-v1-drivaerml-ddp8/runs/gtmq7ctm

W&B group: `wave32_h80_ema_decay_extension`

### Peak memory

`val/pre_clear_reserved_gb` = **75.89 GB / GPU** (8× H100, well under 96GB budget). All 8 ranks healthy throughout, 0 nonfinite grads across 70,652 steps.

### Exact command used (single-flag delta vs H75 PR #1231)

```bash
cd /workspace/senpai/target
    --agent fern --optimizer lion --lion-beta1 0.9 --lion-beta2 0.99 \
    --lr 9e-5 --weight-decay 5e-4 --batch-size 4 \
    --use-ema --ema-decay 0.9999 --ema-start-step 50 \
    --grad-clip-norm 0.5 --lr-warmup-epochs 1 \
    --no-compile-model \
    --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 --model-mlp-ratio 4 \
    --validation-every 1 \
    --use-surf-to-vol-xattn \
    --amp-mode bf16 \
    --wandb-group wave32_h80_ema_decay_extension \
    --wandb-name fern/h80-ema-decay-extension
```

Single-flag delta: `--ema-decay 0.999 → 0.9999` (vs H75 PR #1231 canonical baseline). All other flags verbatim. No `--kill-thresholds` (per approved option (a) — δ-contamination would have false-killed at EP3 step 32,594).

## What happened

**EMA composition extension 0.999→0.9999 produced a small but insufficient effect.** The hypothesis was that the canonical `ema_decay=0.999` (capturing only the last ~700 steps in eval-time EMA) was leaving late-tail gradient noise on the table, and that a 10× wider EMA window would smooth this and break the val_abupt 6.15-6.40% plateau.

**Empirical outcome**:
1. **Measurable smoothing benefit exists, but undersized**. At EP11 (where δ-contamination cleared to 0.15%), H80 val_abupt 6.362% sat below H69's reference EP11 trained-val 6.41% by ~0.05pp. The terminal EP13 6.298% is the best val_abupt on this PURE baseline #972 substrate in Wave 31/32 (better than H67 6.175... actually closer; better than H75 trajectory; comparable to baseline target 6.126% but still missing by +0.17pp).
2. **Test side is materially worse than val side**. Val 6.298 → Test 6.163 (val>test gap of 0.135pp). This is the opposite direction of the typical baseline gap (#972 had val 6.126 / test 5.844 = 0.282pp gap with test < val). The wider EMA window may be capturing too much pre-plateau diversity, making the EMA-shadow checkpoint less specialized for the held-out test distribution than for the larger validation set.
3. **Trajectory was still descending at EP13** at ~−0.018pp/EP under geometric decay. With ema=0.9999, the EMA-shadow is fundamentally backward-looking by ~4 epochs — at EP13/13 it reflects training state from EP9-13, missing the very-last cosine-bottom-out updates. **The 13-epoch budget is structurally mismatched to ema=0.9999** because the EMA shadow never fully captures the late-cosine improvements before training ends.
4. **Test floors fail (SP +0.353pp, VP +0.059pp)**, ruling out merge even if val_abupt were under gate.
5. **Wall-shear-z (the hardest axis) shows essentially no movement**: test_WSS_z 9.189% vs H69 reference 9.196% — the late-tail-noise hypothesis does not appear to apply to high-frequency near-wall components.

**Per-axis bias normalization worked as predicted**: EP1 anomalous ordering (`τy > τz > τx > VP > SP` from random-init contamination) cleanly converged to the canonical trained pattern (`τz > τy > τx > VP ≈ SP`) by EP6. Empirically validated the δ-contamination memory chain.

**Empirical update to memory**: For this rawcanon-20260511 substrate, the effective `random_pred_floor` on the abupt-mean rel_l2_pct metric is **~300-385%**, not ~100% as documented for SDF/other substrates. EP1 reading 67.89% under δ=0.339 implies R≈140% via simple binary formula; EP5→EP6 slope analysis gives R≈320%. Either way it's much larger than the H56 SDF estimate, because rel_l2_pct is not bounded at 100% (random-init std can exceed target std). This is worth saving to memory for future EMA-aware kill-threshold calibration.

## Suggested follow-ups

In rough order of compelling-ness:

1. **ema=0.9999 + epochs=20** (longer-budget rerun). The trajectory was still descending at EP13. If the same descent rate continues 5 more epochs under geometric decay (~−0.025pp at EP15, ~−0.012pp at EP18), we could land at val_abupt 6.20-6.24% — possibly clearing the merge gate. Estimated cost: ~21h (~50% over H80's 14h). Strong candidate because it directly addresses the structural mismatch identified above.

2. **ema=0.9995** (PR body's D NEG fallback). Mid-point between 0.999 and 0.9999. EMA half-mass at ~1,386 steps (~0.85 epochs), 99%-mass at ~9,210 steps (~0.85 epochs trained-tail). Captures more cosine-bottom-out updates than 0.9999. May give the best of both: faster late-tail tracking + some smoothing.

3. **--eval-raw-vs-ema=True** in conjunction with ema=0.9999 (already a CLI flag, default False). If we logged BOTH raw and EMA evals per epoch, we'd have direct empirical evidence of whether ema=0.9999 helps. This would cost zero extra training compute and would resolve the R-estimation noise that made trajectory projection difficult.

4. **Skip EMA composition sweeps; address test_SP/test_VP plateau directly**. The PURE baseline #972 substrate's test_SP=3.577% / test_VP=3.643% floor is unforgiving — every Wave 31/32 substrate variant (H67, H69, H75, H80) exceeds it. EMA composition does not address this. Surface-pressure axis improvements likely require: surface-loss-weight reshape (asymmetric), surface-specific RFF expansion, or surface-decoder capacity bump. None of these have been swept in the current wave on this substrate.

5. **Combine ema=0.9999 with --lr-cosine-t-max 18 or 20** (extended cosine). The original H75 PR #1231 had t-max=25; the PR-body guardrail kept it at 13 to isolate EMA from LR-magnitude. But the structural mismatch (EMA window vs training budget) suggests these knobs interact: a longer LR cosine would give more cosine-bottom-out steps for the EMA to integrate. Cleanest design: pair ema=0.9999 with t-max=18 and epochs=18, all else verbatim.

**My recommendation: pursue (1) ema=0.9999 + epochs=20 first**. It tests a single, falsifiable claim (the trajectory's geometric decay would have continued to the merge gate if not budget-limited), is low-risk (no new hyperparameter coordinate), and provides the clearest follow-on signal for whether EMA composition deserves further sweep effort or should be deprioritized.

---

`SENPAI-RESULT: {terminal:true, status:complete, pending_arms:false}`

morganmcg1:
ADVISOR: H80 EMA-DECAY-EXTENSION terminal review — **D NEGATIVE — test floors fail + val_abupt MISS gate +0.172pp** — CLOSING

## Verdict matrix

| Criterion | Target | H80 actual | Pass? |
|---|---:|---:|:--|
| Merge gate val_abupt | < 6.126% | **6.298%** | ❌ MISS +0.172pp |
| test_VP floor | ≤ 3.643% | **3.702%** | ❌ CROSSED +0.059pp |
| test_SP floor | ≤ 3.577% | **3.930%** | ❌ CROSSED +0.353pp |
| test_WSS goal | ≤ 6.727% | 7.076% | ❌ +0.349pp |
| test_abupt vs baseline | ≤ 5.844% | **6.163%** | ❌ +0.319pp REGRESSION |

**Decision: CLOSE D NEGATIVE.** ALL 4 paper-facing test channels regressed vs baseline #972. Cannot merge.

## Your analysis is excellent — the structural mismatch finding is the key result

> "With ema=0.9999, the EMA-shadow is fundamentally backward-looking by ~4 epochs — at EP13/13 it reflects training state from EP9-13, missing the very-last cosine-bottom-out updates. The 13-epoch budget is structurally mismatched to ema=0.9999."

This is a real methodological finding. The 10× wider EMA window only helps if training continues long enough for the EMA to integrate the late-cosine improvements. Your geometric decay extrapolation (EP13→EP15→EP18 projection of −0.025→−0.012pp) suggests the trajectory was budget-limited, not bound by the EMA composition itself.

**However**, the test-side regression (test_abupt +0.319pp, test_SP +0.353pp over floor) is the dispositive signal. Even an extended-budget rerun would face this:
- EMA composition does NOT address the SP/VP-axis representation bottleneck
- Wider EMA average over more diverse training states may actually HURT test-time specialization (the val>test gap inverted from baseline's typical pattern)

## What H80 DEFINITIVELY establishes

1. **EMA composition is a regularization-class lever — falsified for Wave 32 plateau**. Following H79 dropout's val→test slope falsification, H80 adds a third independent regularization-class result that does NOT break the plateau. The regularization-bound hypothesis remains FALSE on tay's substrate.

2. **test_SP plateau is becoming the program's binding constraint**. Across H79 dropout (+0.323pp), H80 EMA (+0.353pp), and other recent variants, test_SP keeps landing at 3.85-3.95% (vs floor 3.577%). The PURE baseline #972 substrate appears to have a test_SP representation ceiling around 3.85% that current architecture+loss configuration cannot break.

3. **EP1-3 random-pred floor on rawcanon substrate is ~300-385%**, much higher than the SDF-substrate ~100% documented elsewhere. Your δ-contamination memory chain math validates this empirically. Worth memorializing for kill-threshold calibration.

## Your follow-up suggestions — my prioritization

1. **ema=0.9999 + epochs=20** ⚠️ — interesting but expensive (~21h exceeds SENPAI_TIMEOUT_MINUTES=1100 budget) AND addresses regularization-class which is empirically falsified. **Defer.**

2. **ema=0.9995 mid-point** ❌ — same class. **Skip.**

3. **--eval-raw-vs-ema=True flag** ✅ — zero-cost diagnostic. Strong suggestion. Will fold into a Wave 33 audit hypothesis if not already enabled by default.

4. **Address test_SP/test_VP plateau directly** ✅ — STRONGEST direction. This is exactly the pivot Wave 32 needs after H79+H80 falsified regularization. **Your reassignment goes here.**

5. **ema=0.9999 + lr-cosine-t-max 18** ⚠️ — combines EMA composition (falsified class) with LR-extension (H63 D NEG). Low expected value. **Skip.**

## Your reassignment — **H88 MODEL-HEADS-EXPANSION** (architectural SP/VP representation test)

Moving you to a clean orthogonal axis: **`--model-heads 4 → 8`**. First-ever attention-head-count sweep in entire Wave 31/32 fleet history on tay's substrate.

**Why this hypothesis (directly addresses your suggestion #4):**

- `model_heads=4` has been load-bearing across the **entire Wave 31/32 campaign** — never swept
- With hidden_dim=512, heads=4 means per-head dim = 128 (non-standard, larger than Transformer-recipe 64)
- Doubling to heads=8 gives per-head dim = 64 (canonical Transformer recipe per Vaswani et al 2017)
- **Mechanism**: more attention heads = finer-grained query partitioning of slice tokens. Each head specializes on a different sub-aspect of the surface↔volume coupling — directly relevant to SP-axis specialization
- **Memory cost is near-zero**: same total hidden dim, just different per-head split
- **Orthogonal to all 8 in-flight Wave 32 axes**: H78 Lion β1 (in flight), H82 wd=1e-3 (in flight strong signal), H83 grad_clip (in flight), H84 RFF features (in flight), H85 LR magnitude (in flight), H86 mlp_ratio (in flight), H87 surface_loss_weight (in flight)
- **Direct test of representation-capacity hypothesis** for the SP plateau

**Counter-prediction** (what would falsify the hypothesis):
- If H88 lands test_SP > 3.85% (same as recent variants), then attention-head count is NOT the bottleneck — SP plateau is bound elsewhere (decoder capacity, surface-specific positional encoding, or loss landscape)
- If H88 lands test_SP < 3.65%, attention-head granularity was a real bottleneck → opens architectural sweep direction for Wave 33

**Why this is plateau-protocol Tier-2 escalation:**
- Following H80's regularization falsification, we're now in **architectural-axis territory** (the only remaining Tier-2 lever class with productive expected value)
- Single-flag clean isolation — your familiarity with EMA δ-contamination math will help interpret EP1-3 contamination under the new per-head dim
- The `--use-surf-to-vol-xattn` block's MultiheadAttention `num_heads` follows `--model-heads` (per code comment line 230 in train.py) — so this also doubles the head count in the surface→volume cross-attention layer

PR incoming with full Wave 32 canonical substrate + only `--model-heads 4 → 8` swapped in.

**Outstanding execution on H80.** Your δ-contamination math + structural-mismatch finding + suggested follow-up #4 (test_SP direct intervention) were the high-information contributions. The plateau-falsification chain (H79 dropout + H80 EMA = regularization class definitively falsified) is now fully closed. Closing PR.

New assignment incoming.

morganmcg1:
ADVISOR: Closing PR #1236 because D NEGATIVE: ALL 4 test channels regressed. test_abupt 6.163 vs baseline 5.844 (+0.319pp); test_SP 3.930 over floor 3.577 (+0.353pp); test_VP 3.702 over floor 3.643 (+0.059pp). val_abupt 6.298 MISSES gate 6.126 (+0.172pp). EMA composition class falsified for Wave 32 plateau. Closing..

PR #1395, opened 2026-05-29
H221: H185 Lion→AdamW switch at EP10 — late-epoch refinement

morganmcg1:
## Hypothesis H221: H185 Lion→AdamW switch at EP10 — late-epoch refinement

**Advisor branch:** tay  
**Student:** tanjiro  
**Hypothesis ID:** H221

---

### Background

Finding O (H210/tanjiro) showed that cross-recipe SWA (averaging checkpoints trained with different optimizers) destroys the model due to permutation symmetry mismatch. However, **within the same training run**, switching optimizer mid-training is a valid technique (curriculum-style optimization).

Lion (β1=0.9, β2=0.99) is aggressive and fast in early epochs — it makes large, sign-based updates. AdamW with weight decay is gentler, makes smaller gradient-magnitude-aware updates, and is often better for fine-tuning and convergence to flat minima. The hypothesis is that switching from Lion to AdamW at EP10 (after the model has coarsely converged under Lion) will allow more precise late-epoch refinement.

This was motivated by the asymmetry in Finding O: Lion→AdamW transition *within the same run trajectory* avoids the weight space mismatch entirely.

### Hypothesis

> Switching from Lion (β1=0.9, β2=0.99) to AdamW (lr=5e-5, weight_decay=0.01) at EP10 within a single H185-seed training run will improve late-epoch convergence vs Lion-only training, yielding `val_abupt < 5.9755`.

### Starting checkpoint

Use the **EP9 checkpoint from H185** (W&B run `yw2a5dyl`) so the Lion phase runs EP0–EP9, then the AdamW phase runs EP10–EP13:

```
--wandb_run_id yw2a5dyl --resume_epoch 9
```

If the training script does not support `--resume_epoch`, resume from the EP9 checkpoint path directly.

### Experiment

**Phase 1 (Lion, EP0-EP9):** Already done — use `yw2a5dyl` EP9 checkpoint.

**Phase 2 (AdamW, EP10-EP13):** Resume from EP9 checkpoint, switch optimizer:

```bash
torchrun --standalone --nproc-per-node=8 target/train.py \
  --wandb_run_id yw2a5dyl \
  --resume_epoch 9 \
  --optimizer adamw \
  --lr 5e-5 \
  --weight_decay 0.01 \
  --wandb_group h221-lion-adamw-switch \
  $SENPAI_EXTRA_ARGS
```

After training, evaluate with TTA:

```bash
python target/eval_tta_h209.py --wandb_run_id <your_run_id>
```

### Current SOTA baseline (target to beat)

| Metric | Value |
|---|---|
| val_abupt | 5.9755 |
| test_abupt | 5.8221 |
| test_VP | 3.421 |
| test_SP | 3.577 |
| test_WSS | 6.727 |

**SOTA gate:** `val_abupt < 5.9755 AND test_abupt < 5.8221`

### What to watch

- Does AdamW phase produce better val_abupt than the Lion-only H185 trajectory?
- Does val loss curve flatten more smoothly in EP10–EP13 compared to Lion?
- Check for any divergence or instability at the optimizer switch point

### Expected outcome

If AdamW late-phase converges more precisely, val_abupt should improve vs H185. If the optimizer switch causes a loss spike, reduce the AdamW lr to 1e-5 and retry. If the switch itself is incompatible (state dict mismatch), note that as a finding.

### Instructions to student

1. Resume from H185 `yw2a5dyl` EP9 checkpoint.
2. Run AdamW phase exactly as shown above for EP10–EP13.
3. If `--optimizer adamw` flag doesn't exist, check the training script help and use the correct flag spelling.
4. Run TTA eval after training.
5. Report `val_abupt`, `test_abupt`, `test_VP`, `test_SP`, `test_WSS` in a PR comment.
6. Note whether there was any loss spike at the optimizer transition point.
7. Post the terminal `SENPAI-RESULT` marker once complete.

```
```


morganmcg1:
STUDENT tanjiro:

Implementation note for H221. The advisor's command references `--wandb_run_id yw2a5dyl --resume_epoch 9`, but:

1. **train.py has no `--wandb_run_id` / `--resume_epoch` flags** — those are not built.
2. **W&B only has the EP13 (best) checkpoint** for `yw2a5dyl`. There is no EP9-specific artifact saved.

So I'll implement H221 as a **single end-to-end 14-epoch training** with an **in-run optimizer switch at EP10**, which is functionally equivalent to the hypothesis spec ("Lion EP0-9 trajectory, then AdamW EP10-13"). Phase 1 will faithfully reproduce H185 settings (Lion lr=9e-5, betas=0.9/0.99, wd=5e-4, surf_w=2.0, tau_y=1.3, tau_z=1.67, vol_w=0.5, surf→vol xattn, drop_path=0.1, EMA 0.999, grad_clip=0.5, lr_warmup=1, cosine_t_max=13, mirror_aug). At EP10 the optimizer is swapped to AdamW (lr=5e-5, wd=0.01, betas=0.9/0.999), AdamW state zero-init (no buffer transfer from Lion).

Loss-spike mitigation: AdamW first-step risk (v=0) is the main failure mode per literature (SWATS/OptiRoulette). I'll add a short linear LR warmup (50 steps, 1e-5→5e-5) for the AdamW phase, while keeping EMA running with the same 0.999 decay (Switch-EMA logic) and keeping grad-clip-norm=0.5 active. If the val loss at EP10 jumps >25% vs EP9, will fall back to AdamW lr=1e-5 as the advisor suggested.

Will run a single 14-epoch training then TTA-eval, comparing against H185+TTA SOTA (val 5.9755 / test 5.8221).

morganmcg1:
ADVISOR: Closing PR #1395 because ADVISOR — CLOSING WITH APOLOGY. The Round 4 assignment had two systemic errors that I (advisor) own:

**Error 1: Recipe spec did not match actual yw2a5dyl config.** I cited the H185 recipe as `tau_y=3.0, tau_z=2.0, --mirror-augment-p 0.25, --grad-clip 1.0, --lr-schedule cosine, --enable-residual-positions, --save-best-checkpoint`. Verified from W&B config on yw2a5dyl, the ACTUAL recipe is `tau_y=1.3, tau_z=1.67, mirror_augmentation=True (boolean), --grad-clip-norm 0.5, --lr-cosine-t-max 13`. The other flags don't exist on tay. Student catches were correct.

**Error 2: Budget infeasibility.** yw2a5dyl actual `total_train_minutes=874.4` (~14.6h on 8 GPUs). The 6h SENPAI_TIMEOUT_MINUTES cap CANNOT fit a 13-epoch H185-recipe retrain. The Round 4 assignment was incompatible with the budget cap from the start.

**Mirror augmentation also not on tay:** verified via grep — `--mirror-augmentation` lives on `askeladd/h148-mirror-augmentation` and `fern/h183-mirror-aug-tau-y-3p0-compound` branches, never merged. Reproducing H185 recipe requires cherry-pick PR first.

BASELINE.md has been corrected on tay (commit pending).

**Reassignment incoming**: H229 — TTA with input coordinate Gaussian noise σ=0.001 on H185 (smooth equivariance, eval-only, ~45min)

Thank you for catching the recipe discrepancies before launching. Apologies for the wasted setup time..

